PR #3, opened 2015-07-08
Added a stream block size parameter to change the size of InputStreams' buffers

asandryh:
This patch resolves ORC-22.


omalley:
My concern with this is that this is really a characteristic of the InputStream rather than the user facing side. Let me take a stab at what I think this should look like.

asandryh:
Ok, now I understand your point of view. In my head, I saw it as a parameter specified by an [educated] user.

PR #5, opened 2015-07-26
ORC-9. Create a vector type for timestamp columns

omalley:
Create a vector type for timestamp columns that separates out the seconds from the nanoseconds.

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -854,6 +854,8 @@ namespace orc {
 
   class ReaderImpl : public Reader {
   private:
+    const int64_t epochOffset;
+
     // inputs
     std::unique_ptr<InputStream> stream;
     ReaderOptions options;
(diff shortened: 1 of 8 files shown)


c++/src/Reader.cc
```diff
@@ -854,6 +854,8 @@ namespace orc {
 
   class ReaderImpl : public Reader {
   private:
+    const int64_t epochOffset;
+
```
asandryh: ReaderImpl::epochOffset is not used (we seem to call getEpochOffset() everywhere).

c++/src/Reader.cc
```diff
@@ -993,12 +995,25 @@ namespace orc {
...
+    epoch.tm_sec = 0;
+    epoch.tm_min = 0;
+    epoch.tm_hour = 0;
+    epoch.tm_mday = 1;
+    epoch.tm_mon = 0;
+    epoch.tm_year = 2015 - 1900;
+    epoch.tm_isdst = 0;
+    return static_cast<int64_t>(mktime(&epoch));
+  }
+
```
asandryh: I'm a bit confused: Why is the offset 115 years and 1 day, instead of 45 years (2015 - 1970)?

c++/src/Reader.cc
omalley: The ColumnReader uses StripeStreamsImpl::getEpochOffset() and the implementation of that method uses ReaderImpl::epochOffset. The initializer for ReaderImpl::epochOffset uses the static method getEpochOffset, but it should only be called once per a reader.

c++/src/Reader.cc
omalley: That is the representation of 1 Jan 2015 00:00:00, so it is converting that date into a time_t. Look at the man page for mktime.

omalley: I guess I should add a comment saying that :).

asandryh:
Ah, yes, the manual says tm_year is # of years since 1900, and tm_day is the day of the month starting from 1, not 0. I should've RTFM.
A comment here will be helpful.

PR #6, opened 2015-08-05
ORC-24. C++ reader for direct string encodings occasionally skips bytes.

omalley:
ORC-24. C++ reader for direct string encodings occasionally skips bytes.

### DIFF
--- a/c++/test/TestColumnReader.cc
+++ b/c++/test/TestColumnReader.cc
@@ -1029,6 +1029,13 @@ TEST(TestColumnReader, testStringDirectShortBufferWithNulls) {
   }
 }
 
+/**
+ * Tests ORC-24.
+ * Requires:
+ *   * direct string encoding
+ *   * a null value where the unused length crosses the streaming block
+ *     and the actual value doesn't
+ */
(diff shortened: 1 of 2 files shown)


c++/test/TestColumnReader.cc
```diff
@@ -1029,6 +1029,65 @@ TEST(TestColumnReader, testStringDirectShortBufferWithNulls) {
...
+  std::unique_ptr<Type> rowType = createStructType();
+  rowType->addStructField(createPrimitiveType(STRING), "col0");
+  rowType->assignIds(0);
+
+  std::unique_ptr<ColumnReader> reader = buildReader(*rowType, streams);
+
+  StructVectorBatch batch(25, *getDefaultPool());
+  StringVectorBatch *strings = new StringVectorBatch(25, *getDefaultPool());
+  batch.fields.push_back(strings);
+  strings->length[0] = 5;
```
asandryh: Probably worth adding a comment here that the first element in this batch is null, so the length shouldn't matter (because this is exactly what this unit tests confirms)

PR #7, opened 2015-08-25
ORC-25. Clean up Coverity warnings.

omalley:
These fixes mostly make sure things are defined. I don't believe any of the coverity warnings were real problems, but I wanted to get the list cleared out.

### DIFF
--- a/c++/src/MemoryPool.cc
+++ b/c++/src/MemoryPool.cc
@@ -88,7 +88,7 @@ namespace orc {
 
   template <class T>
   void DataBuffer<T>::reserve(uint64_t newCapacity){
-    if (newCapacity > currentCapacity) {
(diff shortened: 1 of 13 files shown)


c++/src/MemoryPool.cc
```diff
@@ -88,7 +88,7 @@ namespace orc {
 
   template <class T>
   void DataBuffer<T>::reserve(uint64_t newCapacity){
-    if (newCapacity > currentCapacity) {
+    if (newCapacity > currentCapacity || !buf) {
```
asandryh: This is a redundant condition that will never be used. Does Coverity complain about this?

omalley: No, coverity was complaining about needing null checks on buf in other methods. With this change, even if newCapacity is 0, buf is always non-null.

PR #9, opened 2015-10-12
ORC-29. Enable ColumnPrinter to print only selected columns.

asandryh:


c++/include/orc/ColumnPrinter.hh
```diff
@@ -47,6 +47,6 @@ namespace orc {
   };
 
   ORC_UNIQUE_PTR<ColumnPrinter> createColumnPrinter(std::string&,
-						    const Type& type);
+						    const Type& type, const std::vector<bool>* selectedColumns = nullptr);
```
omalley: In the public API, you need to use ORC_NULLPTR to compile on non-C++ 0x11.

Please line break your addition to keep it within 80 chars.

tools/src/FileContents.cc
```diff
@@ -48,11 +49,31 @@ void printContents(const char* filename, const orc::ReaderOptions opts) {
...
+        << "[--columns=column1,column2,...]\n" ;
     return 1;
   }
   try {
+    const std::string COLUMNS_PREFIX = "--columns=";
+    std::list<int64_t> cols;
+
+    // Read command-line options
+    char *param, *value;
+    for (int i = 2; i < argc; i++) {
```
omalley: You should also accept the more standard ordering of "file-contents --columns=1,2 filename.orc"

c++/include/orc/ColumnPrinter.hh
asandryh: Thanks for proofreading! Sorry I missed the line break. Will fix and resubmit the patch shortly.

c++/src/ColumnPrinter.cc
```diff
@@ -252,7 +256,7 @@ namespace orc {
       break;
 
     case STRUCT:
-      result = new StructColumnPrinter(buffer, type);
+      result = new StructColumnPrinter(buffer, type, selectedColumns);
```
omalley: You also need to push the selected columns down through the list, map, and union types. Otherwise, you won't be able to select columns below them.

tools/src/FileContents.cc
```diff
@@ -48,11 +49,31 @@ void printContents(const char* filename, const orc::ReaderOptions opts) {
...
+      if ( (param = std::strstr(argv[i], COLUMNS_PREFIX.c_str())) ) {
```
omalley: What are the semantics? Are the fields above the selected ones automatically included? What about the types below the selected one?

c++/src/ColumnPrinter.cc
asandryh: Actually, this is the intended implementation: only specify top-level (logical) columns. Otherwise, there is no way to distinguish between logical and physical columns.

asandryh: For example, if an ORC file contains columns INT, STRUCT<STRING, BOOLEAN>, running
./file-contents --columns=2 file.orc
will select the STRUCT column. If we allowed selection of subcolumns, then it is unclear which column the above command will select: STRUCT or STRING.

tools/src/FileContents.cc
omalley: It would be much more user friendly to select by column name rather than column id, which given complex types are hard to know. At that point, you might start with only selecting top level columns with something like "--columns=field1,field12", which would mean all of the types under those types.

Eventually, it would be nice to support virtual column names like "length" and "value" for lists, and "length, "key", and "value" for maps. Nested structures would look like "outer12.inner3" or "outer12.key".

asandryh: Yes, originally I wanted to use column names, too. Unfortunately, they are optional. If column names are missing, how would a user select specific columns?

omalley:
Sorry for the long pause on this one.

I wanted to make some changes to Type that it made sense to roll together with this capability. In particular, rather than making ColumnPrinter understand the selected vector, I wanted to add a method to Reader to getSelectedType, which returns the schema filtered by the selected vector. Note that the column ids in the selected type are the same as the original file schema, so getting statistics works.

You can see my take in https://github.com/omalley/orc/tree/orc-29

Other changes:
* I removed the Type.assignIds(0), which was always annoying. Now column ids are assigned automatically when the first one is requested.
* I removed the uncalled kind2String.
* I added Type.getMaximumColumnId, which returns the largest column id under the given type. That makes it easy to sweep through the selected vector to hit the right ids.
* I added a ReaderOptions.include(std::vector<std::string>) that selects column names. Nothing uses or tests it yet, but it will let us extend the tools to use name-based column selection.
* I moved the createRowBatch from the Reader to the Type class. 
* I moved the Type class to its own Type.hh out of Vector.hh.
* I simplified the internal implementation of TypeImpl.
* I tried to be more consistent to use uint64_t for columnIds.
* Added TestType for testing the TypeImpl.
* Fix the Reader.seekToRow implementation to use ColumnReader->skip rather than next.

asandryh:
This is a significantly bigger change than my original pull request. :) I agree that some of the subtype logic and handling of selected columns needed to be moved.

A few comments:
1) Type.getMaximumColumnId and Type.getColumnId share a lot of code (they are identical but the last line). Should this common code be moved to a separate function? Alternatively, the following can also work, but it's a bit ugly:
uint64_t TypeImpl::getMaximumColumnId() const {
    getColumnId();
    return static_cast<uint64_t>(maximumColumnId);
}    

2) Methods ReaderOptions::include(std::vector<std::string>) and  ReaderOptions::include(std::list<uint64_t>) reset each other's settings. It looks like they are not meant to be used together . Is this an intended design?

3) Since the logic it that type->getSubtype(i).getMaximumColumnId() + 1 = type->getSubtype(i+1).getColumnId(), can we add this to TestType?

omalley:
Ok, I pulled out the tree climbing code into TypeImpl::ensureIdAssigned(). I updated the comments on the ReaderOptions::include methods to clarify that calling either one resets the currently selected columns. I also added a check in TestType that ensures that ids are assigned correctly.

omalley:
Here's the proposed update: https://github.com/omalley/orc/commit/108ddbfc629f0305bc7fed6f054cc40432b4078a

asandryh:
+1. Thanks!

PR #10, opened 2015-10-13
Fixed ORC-21: Add functionality to estimate dynamic memory requirements

asandryh:
An upper bound on memory requirements is provided by two components:
- Reader::memoryUse() returns an upper bound on its memory needs. It depends on the file and columns read.
- ColumnBatch::memoryUse() returns an upper bound on its memory needs. It depends on the file, columns, and number of rows read.

The new utility FileMemory.cc compares estimated and actual memory usage.


omalley:
Sorry, it has taken me so long to get to this patch. You've been very patient.  I tried to rebase & squash this down to a single commit and it has conflicts. Can you squash this down to a single commit that applies cleanly to apache/master?

PR #12, opened 2015-11-06
Fixed ORC-21: Add dynamic memory usage estimation.

asandryh:
An upper bound on memory requirements is provided by two components:
* Reader::memoryUse() returns an upper bound on its memory needs. It depends on the file and columns read.
* ColumnBatch::memoryUse() returns an upper bound on its memory needs. It depends on the file, columns, and number of rows read.

The new utility FileMemory.cc compares estimated and actual memory usage.

### DIFF
--- a/c++/include/orc/Reader.hh
+++ b/c++/include/orc/Reader.hh
@@ -718,7 +718,7 @@ namespace orc {
     * all stripes are considered).
      * @return upper bound on memory use
      */
-    virtual uint64_t memoryUse(int stripeIx=-1) = 0;
+    virtual uint64_t getMemoryUse(int stripeIx=-1) = 0;
   };
 }
 
--- a/c++/include/orc/Reader.hh
+++ b/c++/include/orc/Reader.hh
@@ -709,6 +709,16 @@ namespace orc {
      * @return a string of bytes with the file tail
      */
     virtual std::string getSerializedFileTail() const = 0;
+
+    /**
+     * Estimate an upper bound on heap memory allocation by the Reader
+     * based on the information in the file footer.
+     * The bound is less tight if only few columns are read or compression is used.
+     * @param stripeIx index of the stripe to be read (if not specified,
+    * all stripes are considered).
+     * @return upper bound on memory use
+     */
+    virtual uint64_t memoryUse(int stripeIx=-1) = 0;
   };
 }
 
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -143,7 +143,12 @@ namespace orc {
     /**
      * Heap memory used by the batch.
      */
-    virtual int64_t memoryUse();
+    virtual uint64_t getMemoryUsage();
+
+    /**
+     * Check whether the batch length varies depending on data.
+     */
+    virtual bool hasVariableLength();
 
   private:
     ColumnVectorBatch(const ColumnVectorBatch&);
@@ -157,7 +162,7 @@ namespace orc {
     DataBuffer<int64_t> data;
     std::string toString() const;
     void resize(uint64_t capacity);
-    int64_t memoryUse();
+    uint64_t getMemoryUsage();
   };
 
   struct DoubleVectorBatch: public ColumnVectorBatch {
@@ -165,7 +170,7 @@ namespace orc {
     virtual ~DoubleVectorBatch();
     std::string toString() const;
     void resize(uint64_t capacity);
-    int64_t memoryUse();
+    uint64_t getMemoryUsage();
 
     DataBuffer<double> data;
   };
@@ -175,7 +180,7 @@ namespace orc {
     virtual ~StringVectorBatch();
     std::string toString() const;
     void resize(uint64_t capacity);
-    int64_t memoryUse();
+    uint64_t getMemoryUsage();
 
     // pointers to the start of each string
     DataBuffer<char*> data;
@@ -188,7 +193,8 @@ namespace orc {
     virtual ~StructVectorBatch();
     std::string toString() const;
     void resize(uint64_t capacity);
-    int64_t memoryUse();
+    uint64_t getMemoryUsage();
+    bool hasVariableLength();
 
     std::vector<ColumnVectorBatch*> fields;
   };
@@ -198,7 +204,8 @@ namespace orc {
     virtual ~ListVectorBatch();
     std::string toString() const;
     void resize(uint64_t capacity);
-    int64_t memoryUse();
+    uint64_t getMemoryUsage();
+    bool hasVariableLength();
 
     /**
      * The offset of the first element of each list.
@@ -215,7 +222,8 @@ namespace orc {
     virtual ~MapVectorBatch();
     std::string toString() const;
     void resize(uint64_t capacity);
-    int64_t memoryUse();
+    uint64_t getMemoryUsage();
+    bool hasVariableLength();
 
     /**
      * The offset of the first element of each list.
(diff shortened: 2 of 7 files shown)


c++/include/orc/Vector.hh
```diff
@@ -140,6 +140,11 @@ namespace orc {
      */
     virtual void resize(uint64_t capacity);
 
+    /**
+     * Heap memory used by the batch.
+     */
+    virtual int64_t memoryUse();
```
omalley: I think the use of -1 as a special value is problematic. I think you should have it return the current usage for all types and add a method hasVariableLength() that returns if it contains a list or map.
Please use uint64_t.
Please rename to getMemoryUsage().

c++/include/orc/Reader.hh
```diff
@@ -709,6 +709,16 @@ namespace orc {
...
+
+    /**
+     * Estimate an upper bound on heap memory allocation by the Reader
+     * based on the information in the file footer.
+     * The bound is less tight if only few columns are read or compression is used.
+     * @param stripeIx index of the stripe to be read (if not specified,
+    * all stripes are considered).
+     * @return upper bound on memory use
+     */
+    virtual uint64_t memoryUse(int stripeIx=-1) = 0;
```
omalley: Please use getMemoryUse().

c++/src/Reader.cc
```diff
@@ -1061,7 +1063,6 @@ namespace orc {
 
     schema = convertType(footer->types(0), *footer);
     schema->assignIds(0);
-    previousRow = (std::numeric_limits<uint64_t>::max)();
```
omalley: This is a bad change.

c++/src/Reader.cc
```diff
@@ -1081,8 +1081,8 @@ namespace orc {
   }
 
   void ReaderImpl::selectType(const Type& type) {
-    if (!selectedColumns[static_cast<size_t>(type.getColumnId())]) {
-      selectedColumns[static_cast<size_t>(type.getColumnId())] = true;
+    if (!selectedColumns[type.getColumnId()]) {
```
omalley: Removing these casts causes clang breakage.

c++/src/Reader.cc
```diff
@@ -1070,9 +1071,8 @@ namespace orc {
         columnId != included.end(); ++columnId) {
       if (*columnId == 0) {
         selectType(*(schema.get()));
-      } else if (*columnId <=
-                 static_cast<int64_t>(schema->getSubtypeCount())) {
-        selectType(schema->getSubtype(static_cast<uint64_t>(*columnId-1)));
+      } else if (*columnId <= static_cast<int64_t>(schema->getSubtypeCount())) {
+        selectType(schema->getSubtype(*columnId-1));
```
omalley: This is also wrong.

c++/src/Reader.cc
```diff
@@ -1364,6 +1364,111 @@ namespace orc {
...
+      }
+    }
+
+    bool hasStringColumn = false;
+    uint64_t nSelectedStreams = 0;
+    for (int i=0; !hasStringColumn && i < footer->types_size(); i++) {
+      if (selectedColumns[i]) {
+        const proto::Type& type = footer->types(i);
+        nSelectedStreams += maxStreamsForType(type) ;
+        switch (type.kind()) {
```
omalley: you need to static cast the kind to int in order to avoid a clang warning.

c++/src/Reader.cc
```diff
@@ -1433,6 +1538,7 @@ namespace orc {
   }
 
   void ReaderImpl::startNextStripe() {
+    reader.reset(); // ColumnReaders use lots of memory; free old memory first
```
omalley: I'm not sure, but this likely causes a performance regression.

c++/include/orc/Vector.hh
asandryh: Sure, using a separate method is cleaner than returning -1.
Just to confirm: getMemoryUsage() or getMemoryUse()?

c++/src/Reader.cc
asandryh: Why? In the code immediately above (around lines 1055-1060) we already set previousRow. 
This line is a left-over from a commit I made months ago when I implemented ReaderImpl::seekToRow() method. I forgot to remove it. I discovered it recently while testing because it conflicts with the logic of startNextStripe() and seekToRow() methods.

c++/src/Reader.cc
asandryh: Sorry, that's my mistake. Will restore the cast.

c++/src/Reader.cc
asandryh: Will fix this, too.

c++/src/Reader.cc
asandryh: Yes, will add casting here.

c++/src/Reader.cc
asandryh: Why? On line 1550, we assign reader a new value. It destroys the previous value anyway. What calling reset() explicitly does for us is it frees up all the dynamic memory used in the existing instance of ColumnReader BEFORE creating a new ColumnReader and allocating all the buffers. Without calling reset(), the deallocation happens AFTER a new reader is created, so for a brief moment we use 2x memory than we really need. It messes up our memory estimates, too.

c++/include/orc/Vector.hh
omalley: I'd prefer getMemoryUsage, but either is ok.

c++/src/Reader.cc
omalley: sorry, I missed that.

c++/src/Reader.cc
omalley: Sorry, I missed that it was the reset on the unique_ptr rather than the ColumnReader. 

In my profiling, one of the major costs for files with little stripe sizes is rebuilding the ColumnReader and the associated buffer. At some point, we need to clean it up so that it reuses the current ColumnReader and buffers.

c++/src/Reader.cc
```diff
@@ -1364,6 +1365,111 @@ namespace orc {
...
+        if (maxDataLength < stripe) {
+          maxDataLength = stripe;
+        }
+      }
+    }
+
+    bool hasStringColumn = false;
+    uint64_t nSelectedStreams = 0;
+    for (int i=0; !hasStringColumn && i < footer->types_size(); i++) {
+      if (selectedColumns[i]) {
```
omalley: clang needs:
if (selectedColumns[static_cast<size_t>(i)]) {

to avoid a message about using signed int where an unsigned int is needed.

c++/src/Reader.cc
```diff
@@ -1364,6 +1365,111 @@ namespace orc {
...
+    }
+
+    // Account for firstRowOfStripe.
+    memory += firstRowOfStripe.capacity() * sizeof(uint64_t);
+
+    // Decompressors need buffers for each stream
+    uint64_t decompressorMemory = 0;
+    if (compression != CompressionKind_NONE) {
+      for (int i=0; i < footer->types_size(); i++) {
+        if (selectedColumns[i]) {
```
omalley: you need the static cast to size_t here too.

c++/src/Vector.cc
```diff
@@ -196,6 +246,17 @@ namespace orc {
     }
   }
 
+  uint64_t MapVectorBatch::getMemoryUsage() {
+    return ColumnVectorBatch::getMemoryUsage()
+           + static_cast<uint64_t>(offsets.capacity() * sizeof(int64_t))
+           + keys->getMemoryUsage();
```
omalley: You have a spurious ';' here that is making your results and tests wrong.

omalley:
I pushed my fixes for this patch to omalley/orc-21.

c++/src/Vector.cc
asandryh: Argggh, this is a silly one. Thank you for catching this bug! I didn't even get a compiler warning on this one.

PR #14, opened 2016-01-08
ORC-36. Create a test program that writes ORC files from future versions.

omalley:


c++/src/TypeImpl.cc
```diff
@@ -416,7 +416,7 @@ namespace orc {
   std::unique_ptr<Type> buildSelectedType(const Type *fileType,
                                           const std::vector<bool>& selected) {
     if (fileType == nullptr || !selected[fileType->getColumnId()]) {
-      return nullptr;
+      return std::unique_ptr<Type>();
     }
```
asandryh: I have already pulled in this fix.

asandryh:
+1

PR #17, opened 2016-01-21
ORC-39. Create docker scripts for building and testing in various

omalley:
Linux distributions.


omalley:
I updated the patch with a new run-all.sh script that will test all 8 variants of linux.

.. Owen

asandryh:
+1. All distros work fine.

PR #18, opened 2016-02-23
ORC-10. Correct bug when moving ORC files between timezones.

omalley:
This patch:
  * Create the new Timezone class that reads the tzfiles and interprets them.
  * Changes the timestamp reader to use the writer's timezone if the writer included it.
  * Changes the timestamp column vector representation to UTC so that we don't need to do a second lookup and translation into the local timezone.
  * ColumnPrinter is changed to assume the column vector is in UTC.
  * Makes the test for >2038 support explicitly set the timezone.
  * Adds variables for the run-all.sh docker script that lets you change the repository and branch to test.
  * Removes a spurious -O0 from the tools' cmake file.

I've tested this patch on MacOS 10.11 and all of the docker scripts.


c++/src/Timezone.cc
```diff
@@ -0,0 +1,964 @@
...
+                          LOCAL_TIMEZONE + ": " +
+                          strerror(errno));
+    }
+    std::vector<char> buffer(static_cast<size_t>(linkStatus.st_size + 1));
+    ssize_t len = readlink(LOCAL_TIMEZONE, &buffer[0], buffer.size());
+    if (len == -1 || static_cast<size_t>(len) >= buffer.size()) {
+      throw TimezoneError(std::string("Can't read local timezone link ") +
+                          LOCAL_TIMEZONE + ": " +
+                          strerror(errno));
+    }
```
asandryh: This will throw. By default, /etc/localtime is a file, not a symlink, so `readlink()` will return -1.

omalley: Ok, I've fixed that in the pull request.

asandryh:
+1

c++/src/ColumnReader.cc
```diff
@@ -327,7 +329,9 @@ namespace orc {
             nanoBuffer[i] *= 10;
           }
         }
-        secsBuffer[i] += epochOffset;
+        int64_t writerTime = secsBuffer[i] + epochOffset;
+        secsBuffer[i] = writerTime +
```
wgtmac: @omalley I think here introduces an inconsistency between reader and writer. The secsBuffer[i] in the TimestampVectorBatch of reader side will not get same value on the java writer/reader side.

PR #19, opened 2016-02-23
ORC-43. FileMemory.cc throws uncaught exception.

omalley:


tools/src/FileMemory.cc
```diff
@@ -95,37 +127,11 @@ int main(int argc, char* argv[]) {
...
+  try {
+    processFile(filename, cols, batchSize);
+    return 0;
+  } catch (std::exception& ex) {
+    std::cerr << "Caught exception: " << ex.what() << "\n";
+    return 1;
   }
-  std::cout << "\nActual max memory used: " << actualMemory << "\n";
-
-  return 0;
```
asandryh: Don't you get a compiler error if you remove this `return 0;` statement? It's unreachable, but some compilers can still complain.

asandryh:
+1 LGTM.

tools/src/FileMemory.cc
omalley: gcc and clang are both fine with it at -Wall.

PR #20, opened 2016-04-06
ORC-46. Reserve CompressionKind values for LZ4 and ZSTD.

electrum:


omalley:
This seems reasonable. Are you working on implementing the compression? I'd love to get the integration in so that others could experiment with them.

electrum:
We will implement them in Presto using this compression code (zstd coming soon): https://github.com/airlift/aircompressor

omalley:
A couple of thoughts:
* So Presto has a writer now? If so, we should extend the WriterVersion with a new id for it so that we can handle bugs that are specific to that implementation.
* I assume you'll follow the same compression block structure that we used with zlib and snappy. (Three bytes of header followed by the untagged compressed bytes.) I'm very concerned that we don't break compatibility with the C++ and Java readers.
* That is *really* cool that there is a pure Java LZO, Snappy, and LZ4 implementation that is Apache licensed. Does it have a release that is pushed to Maven central?

martint:
> So Presto has a writer now? If so, we should extend the WriterVersion with a new id for it so that we can handle bugs that are specific to that implementation.

We're going to start working on a writer soon.

> I assume you'll follow the same compression block structure that we used with zlib and snappy. (Three bytes of header followed by the untagged compressed bytes.) I'm very concerned that we don't break compatibility with the C++ and Java readers.

Yeah, most likely.

> That is really cool that there is a pure Java LZO, Snappy, and LZ4 implementation that is Apache licensed. Does it have a release that is pushed to Maven central?

http://search.maven.org/#artifactdetails%7Cio.airlift%7Caircompressor%7C0.3%7Cjar

PR #27, opened 2016-05-24
ORC-52. Create ORC InputFormat and OutputFormat implementations for org.apache.hadoop.mapreduce API

omalley:


java/mapreduce/src/java/org.apache.orc.mapreduce/OrcInputFormat.java
```diff
@@ -0,0 +1,95 @@
...
+    Configuration conf = taskAttemptContext.getConfiguration();
+    Reader file = OrcFile.createReader(split.getPath(),
+        OrcFile.readerOptions(conf)
+            .maxLength(OrcConf.MAX_FILE_LENGTH.getLong(conf)));
+    TypeDescription schema =
+        TypeDescription.fromString(OrcConf.SCHEMA.getString(conf));
+    Reader.Options options = new Reader.Options()
+        .range(split.getStart(), split.getLength())
+        .useZeroCopy(OrcConf.USE_ZEROCOPY.getBoolean(conf))
+        .skipCorruptRecords(OrcConf.SKIP_CORRUPT_DATA.getBoolean(conf));
```
wagnermarkd: These two lines seem to duplicate logic in RecordReaderImpl: https://github.com/apache/orc/blob/7a4fe255e84bea097ca478d2008d257e07f8808a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java#L184. As far as I can see that's the only reason conf is passed at all. It could be cleaner if Reader.Options's constructor pulled the requisite values from the conf and only stored/exposed those.

java/mapreduce/src/java/org.apache.orc.mapreduce/OrcOutputFormat.java
```diff
@@ -0,0 +1,83 @@
...
+      "orc.mapreduce.output.skip-temporary-directory";
+
+  @Override
+  public RecordWriter<NullWritable, V>
+       getRecordWriter(TaskAttemptContext taskAttemptContext
+                       ) throws IOException {
+    Configuration conf = taskAttemptContext.getConfiguration();
+    Path filename = getDefaultWorkFile(taskAttemptContext, EXTENSION);
+    Writer writer = OrcFile.createWriter(filename,
+        OrcFile.writerOptions(conf)
```
wagnermarkd: Same question as reader options: Can we pull all this boiler plate into the writerOptions method?

java/mapreduce/src/test/org/apache/orc/mapreduce/TestMrUnit.java
```diff
@@ -0,0 +1,202 @@
...
+        context.write(nada, output);
+      }
+    }
+  }
+
+  /**
+   * This class is intended to support MRUnit's object copying for input and
+   * output objects.
+   *
+   * Real mapreduce contexts should NEVER use this class.
```
wagnermarkd: Isn't there the same challenge for true MR contexts? The reducer needs to be aware of the schema for the values which it is serializing. Orc

omalley: My goal with OrcKey and OrcValue was to encode the respective types once in the JobConf rather than encode it per a value. They let you control the key and value types of the shuffle with knobs that don't conflict with either the input or output.

So if you are shuffling with two OrcStructs, you would define your JobConf like (filling in the "..." with the appropriate fields):

orc.mapred.key.type=struct<...>
orc.mapred.value.type=struct<vals:array<..>
mapreduce.map.output.key.class=OrcKey
mapreduce.map.output.value.class=OrcValue

I guess that I should actually improve that to:

orc.mapred.input.type=struct<...>
orc.mapred.output.type=struct<...>
orc.mapred.map.output.key.type=struct<...>
orc.mapred.map.output.value.type=struct<...>
mapreduce.map.output.key.class=OrcKey
mapreduce.map.output.value.class=OrcValue

The orc.mapred.input.type setting is only necessary if your application wants to use the schema evolution to convert to a specific type. The orc.mapred.output.type would control the schema of the output format. 

Does that make sense?

java/mapreduce/src/java/org.apache.orc.mapreduce/OrcInputFormat.java
omalley: Yeah, let me figure out how to merge the common code.

wagnermarkd:
LGTM. The new names are helpful.

PR #29, opened 2016-05-26
ORC-53. Make the complex types Comparable so they can be put as keys in OrcMap

omalley:


java/mapreduce/src/java/org/apache/orc/mapred/OrcList.java
```diff
@@ -71,4 +71,30 @@ public void readFields(DataInput input) throws IOException {
...
+      Writable ours = get(e);
+      Writable theirs = other.get(e);
+      if (ours == null) {
+        if (theirs != null) {
+          return 1;
+        }
+      } else if (theirs == null) {
+        return -1;
+      } else {
+        int val = ((Comparable<Writable>) ours).compareTo(theirs);
```
wagnermarkd: Comparing two complex types with different nested types will cause trouble here. To avoid this, we need to first compare the schemas, then the values. This applies to all the complex types.

omalley:
Ok, this needed a relatively large change. To compare the schemas, I wanted to make TypeDescription implement Comparable, but the equals and hashCode were set up to match against the id, which would have largely screwed things up. So, I made SchemaEvolution use a HashMap<Integer, TypeDescription> and used the id as the key. That is more straightforward and let me change the semantics of equals and hashCode to be more reasonable. While I was in TypeDescription, I also made it Serializable since FindBugs was complaining about the OrcList and OrcMap having non-serializable fields.

I also added a log4j.properties file into the test resources so that the log messages come out to the console during unit tests.

I also changed the 4 complex ORC types to implement WritableComparable, which ORC-53 had done to better match with MapReduce's requirements for key types.

java/mapreduce/src/test/org/apache/orc/mapred/TestOrcStruct.java
```diff
@@ -123,4 +123,18 @@ public void testCompare() {
...
+
+  @Test
+  public void testSchemaInCompare() {
+    TypeDescription leftType = TypeDescription.fromString("struct<s:string,i:int>");
+    TypeDescription rightType = TypeDescription.fromString("struct<s:string,j:bigint>");
+    OrcStruct left = new OrcStruct(leftType);
+    OrcStruct right = new OrcStruct(rightType);
+    assertEquals(-1, left.compareTo(right));
+    assertEquals(1, right.compareTo(left));
+    left.setAllFields(new Text("123"), new IntWritable(123));
```
wagnermarkd: I checked out your branch and got a compilation error for setAllFields, which was introduced for ORC-52.

java/mapreduce/src/java/org/apache/orc/mapred/OrcMap.java
```diff
@@ -83,4 +83,52 @@ public void readFields(DataInput input) throws IOException {
...
+      return result;
+    }
+    Iterator<Map.Entry<K,V>> ourItr = entrySet().iterator();
+    Iterator<Map.Entry<K,V>> theirItr = other.entrySet().iterator();
+    while (ourItr.hasNext() && theirItr.hasNext()) {
+      Map.Entry<K,V> ourItem = ourItr.next();
+      Map.Entry<K,V> theirItem = theirItr.next();
+      K ourKey = ourItem.getKey();
+      K theirKey = theirItem.getKey();
+      int val = ((Comparable<K>) ourKey).compareTo(theirKey);
```
wagnermarkd: Changing the type of K to extend WritableComparable would remove the need for this cast

omalley:
The compilation was working for me, but you had a good point that in the current version that everything should implement WritableComparable since obviously compareTo needs to recurse all the way down.

Can you try the current version?

wagnermarkd:
Everything works on the orc-52 branch, which is based on orc-53, but with orc-53 I still see the previous error for TestOrcStruct and this new one for OrcRecordReader:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project orc-mapreduce: Compilation failure: Compilation failure:
[ERROR] /home/mwagner/git/orc/java/mapreduce/src/java/org/apache/orc/mapred/OrcRecordReader.java:[94,15] no suitable method found for setFieldValue(int,org.apache.hadoop.io.Writable)
[ERROR] method org.apache.orc.mapred.OrcStruct.setFieldValue(int,org.apache.hadoop.io.WritableComparable) is not applicable
[ERROR] (argument mismatch; org.apache.hadoop.io.Writable cannot be converted to org.apache.hadoop.io.WritableComparable)
```

PR #35, opened 2016-06-06
ORC-67 Fix for reading GMT timezone info on MacOS.

omalley:


spothineni:
Very clean fix. I tested this commit. working fine. Thanks for the help.

PR #38, opened 2016-06-12
ORC-69. Add batch option support in orc-contents and orc-scan tools.

xunzhang:
cc @omalley , plz review.

### DIFF
--- a/tools/src/FileContents.cc
+++ b/tools/src/FileContents.cc
@@ -26,11 +26,13 @@
 #include <iostream>
 #include <string>
 
-void printContents(const char* filename, const orc::ReaderOptions opts) {
+void printContents(const char* filename,
+                   const orc::ReaderOptions opts,
+                   uint64_t batchSize) {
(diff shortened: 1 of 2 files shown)


xunzhang:
I have tested the option `--batch` with `orc-contents` and `orc-scan` in my environment.

omalley:
I buy that we may want to test the batch size for performance tests, but orc-scan is the important tool there since you don't really want to benchmark the conversion into JSON.

With that in mind, I've made a variant of this patch that:
* reverts the change to orc-contents
* switches orc-scan to use getopt_long
* adds some test infrastructure so that we can test the executables
* adds tests of the nominal and off-nominal invocations of orc-scan.

Please see my changes on https://github.com/omalley/orc/tree/orc-69

xunzhang:
LGTM.
I like the getopt solution. Other tools may also use this way in the future.
And, feel free to commit without my name. Thanks!

PR #40, opened 2016-06-16
ORC-54: Evolve schemas based on field name rather than index

wagnermarkd:
Change evolution to compare column names when building mapping.

Misc notes:
- I've added a setting orc.tolerate.missing.schema to allow working with data before HIVE-4243 was fixed. This is important for anyone who has older data, though it limits the evolutions that can be supported
- A lot of changes in RecordReaderImpl to make the reader column vs writer column difference explicit
- SARG tests are passing, but there's no test coverage for SARG + evolution. Will work on that.
- Removed some unneeded checked exceptions.

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -82,6 +82,12 @@ public enum OrcConf {
       "If ORC reader encounters corrupt data, this value will be used to\n" +
           "determine whether to skip the corrupt data or throw exception.\n" +
           "The default behavior is to throw exception."),
+  TOLERATE_MISSING_SCHEMA("orc.tolerate.missing.schema",
+      "hive.exec.orc.tolerate.missing.schema",
+      true,
+      "Writers earlier than HIVE-4243 may have inaccurate schema metadata.\n"
+          + "This setting will enable best effort schema evolution rather\n"
+          + "than rejecting mismatched schemas"),
   MEMORY_POOL("orc.memory.pool", "hive.exec.orc.memory.pool", 0.5,
       "Maximum fraction of heap that can be used by ORC file writers"),
   DICTIONARY_KEY_SIZE_THRESHOLD("orc.dictionary.key.threshold",
--- a/java/core/src/java/org/apache/orc/impl/ReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ReaderImpl.java
@@ -572,7 +572,14 @@ public class ReaderImpl implements Reader {
     boolean[] include = options.getInclude();
     // if included columns is null, then include all columns
     if (include == null) {
-      include = new boolean[types.size()];
+      int size;
+      TypeDescription readSchema = options.getSchema();
+      if (readSchema != null){
+        size = readSchema.getMaximumId() + 1;
+      } else {
+        size = types.size();
+      }
+      include = new boolean[size];
       Arrays.fill(include, true);
       options.include(include);
     }
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -27,21 +27,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import org.apache.orc.BooleanColumnStatistics;
-import org.apache.orc.Reader;
-import org.apache.orc.RecordReader;
-import org.apache.orc.TypeDescription;
-import org.apache.orc.ColumnStatistics;
-import org.apache.orc.CompressionCodec;
-import org.apache.orc.DataReader;
-import org.apache.orc.DateColumnStatistics;
-import org.apache.orc.DecimalColumnStatistics;
-import org.apache.orc.DoubleColumnStatistics;
-import org.apache.orc.IntegerColumnStatistics;
-import org.apache.orc.OrcConf;
-import org.apache.orc.StringColumnStatistics;
-import org.apache.orc.StripeInformation;
-import org.apache.orc.TimestampColumnStatistics;
+import org.apache.orc.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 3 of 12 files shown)


java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -27,21 +27,7 @@
...
-import org.apache.orc.DataReader;
-import org.apache.orc.DateColumnStatistics;
-import org.apache.orc.DecimalColumnStatistics;
-import org.apache.orc.DoubleColumnStatistics;
-import org.apache.orc.IntegerColumnStatistics;
-import org.apache.orc.OrcConf;
-import org.apache.orc.StringColumnStatistics;
-import org.apache.orc.StripeInformation;
-import org.apache.orc.TimestampColumnStatistics;
+import org.apache.orc.*;
```
omalley: Please configure your IDE to not replace imports with wildcards.

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
```diff
@@ -572,7 +572,14 @@ public RecordReader rows(Options options) throws IOException {
     boolean[] include = options.getInclude();
     // if included columns is null, then include all columns
     if (include == null) {
-      include = new boolean[types.size()];
+      int size;
+      TypeDescription readSchema = options.getSchema();
+      if (readSchema != null){
+        size = readSchema.getMaximumId() + 1;
```
omalley: I'd suggest doing
```
if (readSchema == null) {
  readSchema = schema;
}
```

java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
```diff
@@ -18,25 +18,33 @@
...
+    public IllegalEvolutionException(String msg) {
+      super(msg);
+    }
+  }
+
   private final Map<Integer, TypeDescription> readerToFile;
   private final boolean[] included;
   private final TypeDescription readerSchema;
   private static final Logger LOG = LoggerFactory.getLogger(SchemaEvolution.class);
+  private static final Pattern missingMetadataPattern = Pattern.compile("_col\\d+");
```
omalley: You should check for WriterVersion < HIVE_4243 before bothering to check the column names. The vast majority of files from before HIVE_4243 have the synthetic column names.

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -82,6 +82,12 @@
       "If ORC reader encounters corrupt data, this value will be used to\n" +
           "determine whether to skip the corrupt data or throw exception.\n" +
           "The default behavior is to throw exception."),
+  TOLERATE_MISSING_SCHEMA("orc.tolerate.missing.schema",
+      "hive.exec.orc.tolerate.missing.schema",
+      true,
+      "Writers earlier than HIVE-4243 may have inaccurate schema metadata.\n"
+          + "This setting will enable best effort schema evolution rather\n"
```
omalley: I'd suggest that the comment include something about using position in the top level structure as the fallback and that it support adding new columns to the end and changing types, but not deleting or reordering the top level columns.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -933,7 +926,7 @@ void createStreams(List<OrcProto.Stream> streamDescriptions,
     for (OrcProto.Stream streamDesc : streamDescriptions) {
       int column = streamDesc.getColumn();
       if ((includeColumn != null &&
-          (column < included.length && !includeColumn[column])) ||
+          (column < includeColumn.length && !includeColumn[column])) ||
```
omalley: should this be writerIncluded?

java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
```diff
@@ -45,16 +53,65 @@ public SchemaEvolution(TypeDescription readerSchema, boolean[] included) {
...
       this.readerSchema = readerSchema;
     }
-    buildMapping(fileSchema, this.readerSchema);
+
+    boolean useFieldNames;
+    if (!hasColumnNames(fileSchema)){
+      if (!allowMissingMetadata && !fileSchema.equals(readerSchema)){
+        throw new RuntimeException("Found that schema metadata is missing from file. This is likely caused by a writer earlier than HIVE-4243. Will not try to reconcile schemas");
+      } else {
+        LOG.warn("Found that schema metadata is missing from file. This is likely caused by a writer earlier than HIVE-4243. Will try to reconcile schemas based on index");
```
omalley: How about "Found that column names are missing from file, which is caused by ORC writers before HIVE-4243."

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -201,6 +201,10 @@ public void nextVector(ColumnVector previous,
     public BitFieldReader getPresent() {
       return present;
     }
+
```
omalley: What do you think about moving the building of writerIncluded into the SchemaEvolution class? I think it would be pretty natural there and building it would come naturally out of buildMapping.

java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
```diff
@@ -85,55 +142,78 @@ void buildMapping(TypeDescription fileType,
...
+        break;
+      case UNION:
+      case MAP:
+      case LIST: {
+        // these must be an exact match
+        List<TypeDescription> fileChildren = fileType.getChildren();
+        List<TypeDescription> readerChildren = readerType.getChildren();
+        if (fileChildren.size() == readerChildren.size()) {
+          for (int i = 0; i < fileChildren.size(); ++i) {
+            buildMapping(fileChildren.get(i), readerChildren.get(i), useFieldNames);
```
omalley: when you are recursing, useFieldNames should always be true. Prior to HIVE-4243 only the top level column names were lost. In recursive types the field names were correct.

java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
```diff
@@ -85,55 +142,78 @@ void buildMapping(TypeDescription fileType,
...
+            }
+
+            buildMapping(fileField, readerField, true);
+          }
+        } else {
           int jointSize = Math.min(fileChildren.size(), readerChildren.size());
-          for(int i=0; i < jointSize; ++i) {
-            buildMapping(fileChildren.get(i), readerChildren.get(i));
+          for (int i = 0; i < jointSize; ++i) {
+            buildMapping(fileChildren.get(i), readerChildren.get(i), false);
```
omalley: This should be true also to match field names down below the root.

omalley:
This is looking good. I rebased it to the current trunk and undid some of the spacing changes. My version is here:
https://github.com/omalley/orc/tree/orc-54

I also left some comments in the pull request.

PR #41, opened 2016-06-20
ORC-58: Move code for reading rows from Reader to RowReader

majetideepak:
This PR divides the current Reader class into Reader class and RowReader class
Reader class reads the footer and metadata from an ORC file.
RowReader class reads the rows from the file.


majetideepak:
@omalley, please let me know your feedback. Thanks!

tools/src/FileContents.cc
```diff
@@ -27,8 +27,10 @@
 #include <string>
 
 void printContents(const char* filename, const orc::ReaderOptions opts) {
-  std::unique_ptr<orc::Reader> reader;
-  reader = orc::createReader(orc::readLocalFile(std::string(filename)), opts);
+  std::unique_ptr<orc::Reader> fileReader;
```
swalkaus: Pendantic: use reader and rowReader consistently (notice the camelCase).

c++/include/orc/Reader.hh
```diff
@@ -645,7 +645,72 @@ namespace orc {
...
+     * @param stripeIx index of the stripe to be read (if not specified,
+     *        all stripes are considered).
+     * @return upper bound on memory use
+     */
+    virtual uint64_t getMemoryUse(int stripeIx=-1) = 0;
+ 
+  };
+
+  /**
+   * The interface for reading footer in ORC files.
```
swalkaus: "The interface for reading ORC file meta-data."

swalkaus:
I might rename "Reader" something like "FileTailReader" but, in general, the refactoring makes sense. The schema can be read and columns metadata known before column selection is decided.

c++/src/Reader.cc
```diff
@@ -1108,13 +1107,75 @@ namespace orc {
...
+    uint64_t getRowNumber() const override;
+
+    void seekToRow(uint64_t rowNumber) override;
+
+    MemoryPool* getMemoryPool() const ;
+
+  };
+
+  class ReaderImpl : public Reader {
+    friend class RowReaderImpl;
```
swalkaus: Avoid using friend.

c++/src/Reader.cc
```diff
@@ -1108,13 +1107,75 @@ namespace orc {
...
     void updateSelected(const std::list<uint64_t>& fieldIds);
     void updateSelected(const std::list<std::string>& fieldNames);
 
   public:
+   /**
+    * Constructor that lets the user specify additional options.
+    * @param filereader the object to read from
+    * @param options options for reading
+    */
+    RowReaderImpl(const ReaderImpl* filereader,
```
swalkaus: Should filereader be a "Reader" or "ReaderImpl"?

majetideepak: We will have to extend the `Reader` interface with a notion of `metadata`, `ReaderOptions` etc. to achieve this. I am not sure what the right design choice is here.

c++/include/orc/Reader.hh
```diff
@@ -645,7 +645,72 @@ namespace orc {
...
+     * struct<col0:int,col2:string>.
+     * @return the root type
+     */
+    virtual const Type& getSelectedType() const = 0;
+
+    /**
+     * Get the selected columns of the file.
+     */
+    virtual const std::vector<bool> getSelectedColumns() const = 0;
+
```
omalley: You have some trailing spaces. You can probably configure your IDE to automatically trim them.

c++/include/orc/Reader.hh
```diff
@@ -645,7 +645,72 @@ namespace orc {
...
+    /**
+     * Estimate an upper bound on heap memory allocation by the Reader
+     * based on the information in the file footer.
+     * The bound is less tight if only few columns are read or compression is
+     * used.
+     * @param stripeIx index of the stripe to be read (if not specified,
+     *        all stripes are considered).
+     * @return upper bound on memory use
+     */
+    virtual uint64_t getMemoryUse(int stripeIx=-1) = 0;
```
omalley: This should stay in the reader.

c++/include/orc/Reader.hh
```diff
@@ -782,85 +847,42 @@ namespace orc {
...
 
     /**
-     * Seek to a given row.
-     * @param rowNumber the next row the reader should return
+     * @param include update with new columns
+     * @return a RowReader to read the rows
      */
-    virtual void seekToRow(uint64_t rowNumber) = 0;
+    virtual ORC_UNIQUE_PTR<RowReader>
+    getRowReader(const std::list<uint64_t>& include) const = 0;
```
omalley: Let's go ahead and make a RowReaderOptions class and pass that in here. It is almost guaranteed that the include vector will not be the only option that we want to pass in. One of the options will be to specify the include vector.

omalley:
The ReaderOptions needs to split in half too:

* ReaderOptions
  * setTailLocation
  * setErrorStream
  * setSerializedFooter

* RowReaderOptions
  * include
  * range
  * throwOnHive11DecimalOverflow
  * forcedScaleOnHive11Decimal

c++/src/Reader.cc
```diff
@@ -1062,37 +1062,36 @@ namespace orc {
...
     const Timezone& localTimezone;
 
     // inputs
-    std::unique_ptr<InputStream> stream;
-    ReaderOptions options;
-    const uint64_t fileLength;
-    const uint64_t postscriptLength;
     std::vector<bool> selectedColumns;
-
+    std::shared_ptr<InputStream> stream;
```
omalley: I'd suggest creating a single class that has all of the state for the ReaderImpl. Then it can have the postscript, InputStream, etc. and be shared between the ReaderImpl and the set of RowReaderImpls that share the same ReaderImpl.

majetideepak:
Included review comments.
This patch is huge because I split the monolithic Reader.cc into various files
- Reader.hh/Reader.cc: ReaderImpl and RowReaderImpl
- Statistics.hh/Statistics.cc: ColumnStatisticsImpl, StatisticsImpl, Type*StatisticsImpl
- StripeStream.hh/StripeStream.cc: StripeStreamsImpl, StreamInformationImpl, StripeInformationImpl
- Options.hh: ReaderOptions, RowReaderOptions

@omalley any feedback ?

omalley:
This is looking good. To compile on MacOS, I needed to replace some of the std::move with REDUNDANT_MOVE. See omalley:pr/41

majetideepak:
Replaced with REDUNDANT_MOVE. valgrind does not show any memory leaks due to this patch.

omalley:
I think we should have FileContents be a shared ptr so that if the user deallocates the Reader, the RowReader won't reach into deallocated memory.

majetideepak:
I updated the patch to use a `shared_ptr` for `FileContents`.

omalley:
Deepak, 
   We got a notice of a new problem from Coverity. Can you look at it?


** CID 173749:  Uninitialized members  (UNINIT_CTOR)
/c++/src/Reader.cc: 259 in orc::RowReaderImpl::RowReaderImpl(std::shared_ptr<orc::FileContents>, const orc::RowReaderOptions &)()


________________________________________________________________________________________________________
*** CID 173749:  Uninitialized members  (UNINIT_CTOR)
/c++/src/Reader.cc: 259 in orc::RowReaderImpl::RowReaderImpl(std::shared_ptr<orc::FileContents>, const orc::RowReaderOptions &)()
253         } else {
254           previousRow = firstRowOfStripe[firstStripe]-1;
255         }
256
257         ColumnSelector column_selector(contents.get());
258         column_selector.updateSelected(selectedColumns, options);
>>>     CID 173749:  Uninitialized members  (UNINIT_CTOR)
>>>     Non-static class member "rowsInCurrentStripe" is not initialized in this constructor nor in any functions that it calls.
259       }
260
261       const RowReaderOptions& RowReaderImpl::getRowReaderOptions() const {
262         return options;
263       }
264

majetideepak:
@omalley 
https://issues.apache.org/jira/browse/ORC-127
https://github.com/apache/orc/pull/77
Thanks.

PR #46, opened 2016-07-06
ORC-80:C++ build breaks due to warnings with gcc-4.9

majetideepak:


tools/test/ToolTest.cc
```diff
@@ -163,7 +163,7 @@ int runProgram(const std::vector<std::string>& command,
     if (result == -1) {
       std::cerr << "Error: " << strerror(errno) << "\n";
     } else if (WIFSIGNALED(status)) {
-      std::cerr << "Fatal signal: " << WTERMSIG(signal) << "\n";
+      std::cerr << "Fatal signal: " << WTERMSIG(status) << "\n";
```
majetideepak: @xunzhang let me know if `status` is incorrect here.

PR #55, opened 2016-08-16
ORC-54: Evolve schemas based on field name rather than index

omalley:
This is an updated version of Mark's patch that fixes evolution of ACID files and rebases it to the current trunk.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -27,10 +27,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.orc.BloomFilterIO;
 import org.apache.orc.BooleanColumnStatistics;
-import org.apache.orc.Reader;
-import org.apache.orc.RecordReader;
-import org.apache.orc.TypeDescription;
 import org.apache.orc.ColumnStatistics;
 import org.apache.orc.CompressionCodec;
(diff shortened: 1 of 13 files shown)


java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
```diff
@@ -20,59 +20,132 @@
...
+            throw new RuntimeException("Found that schema metadata is missing"
+                + " from file. This is likely caused by"
+                + " a writer earlier than HIVE-4243. Will"
+                + " not try to reconcile schemas");
+          } else {
+            LOG.warn("Column names are missing from this file. This is"
+                + " caused by a writer earlier than HIVE-4243. The reader will"
+                + " reconcile schemas based on index. File type: " +
+                this.fileSchema + ", reader type: " + this.readerSchema);
+            positionalLevels = isAcid ? 2 : 1;
```
prasanthj: What does positional level mean? Is it real row level? 
Does acid file schema look like this struct<struct<[acid_cols]>,struct[real_cols]>>? If so can you leave a comment about it?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -853,6 +862,15 @@ private void readStripe() throws IOException {
     }
   }
 
+  private boolean isFullRead() {
+    for (Boolean isColumnPresent : writerIncluded){
```
prasanthj: nit: boolean

PR #56, opened 2016-08-17
ORC-93 Remove log message about seeking into an empty stream.

omalley:


prasanthj:
lgtm +1

omalley:
I added a test case for the case of having a 0 byte InStream with an empty DiskRange in it.

PR #57, opened 2016-08-20
ORC-96: Pass Context to Orc tree readers

prasanthj:
This patch passed Context object to tree readers. With that we don't have to repeatedly change signatures of tree readers.


omalley:
+1

PR #60, opened 2016-09-14
ORC-101 fix broken encoding for string and decimal bloom filters

omalley:


java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -105,6 +105,12 @@
           "dictionary or not will be retained thereafter."),
   BLOOM_FILTER_COLUMNS("orc.bloom.filter.columns", "orc.bloom.filter.columns",
       "", "List of columns to create bloom filters for when writing."),
+  BLOOM_FILTER_WRITE_VERSION("orc.bloom.filter.write.version",
+      "orc.bloom.filter.write.version", OrcFile.BloomFilterVersion.UTF8.toString(),
+      "Which version of the bloom filter should we write."),
```
prasanthj: Any reason to have this config? If we are going to default to UTF8 anyways we don't need this right?

prasanthj: I see from below to provide the option of writing both default and utf-8 we are using this config. Can you use enum of "default","utf-8" and add a description saying default will write both versions.

omalley: If orc.bloom.filter.write.version is set to original, it will put both streams so that the files are usable by old versions of orc.

java/core/src/java/org/apache/orc/util/BloomFilter.java
```diff
@@ -130,7 +125,7 @@ public void addString(String val) {
     if (val == null) {
       add(null);
     } else {
-      add(val.getBytes());
+      add(val.getBytes(Charset.defaultCharset()));
```
prasanthj: Should we make the default "UTF-8" and provide an alternate addString() that accepts Charset?

java/core/src/java/org/apache/orc/impl/WriterImpl.java
```diff
@@ -1901,7 +2014,11 @@ void writeBatch(ColumnVector vector, int offset,
           HiveDecimal value = vec.vector[0].getHiveDecimal();
           indexStatistics.updateDecimal(value);
           if (createBloomFilter) {
-            bloomFilter.addString(value.toString());
+            String str = value.toString();
+            if (bloomFilter != null) {
+              bloomFilter.addString(str);
```
prasanthj: Can you plz add a comment here for not using UTF-8 for decimals?

java/core/src/java/org/apache/orc/util/BloomFilterIO.java
```diff
@@ -0,0 +1,71 @@
...
+   * @param builder the builder to write to
+   * @param bloomFilter the bloom filter to serialize
+   */
+  public static void serialize(OrcProto.BloomFilter.Builder builder,
+                               BloomFilter bloomFilter) {
+    long[] bitset = bloomFilter.getBitSet();
+    if (builder.getBitsetCount() != bitset.length) {
+      builder.clear();
+      for(int i=0; i < bitset.length; ++i) {
+        builder.addBitset(bitset[i]);
```
prasanthj: can we make this byte[]? Protobuf deserializes this as List<Long> IIRC (Sergey and Dain reported).

java/core/src/java/org/apache/orc/OrcConf.java
omalley: The default is "utf8". Maybe we should use "both" and "utf8"?

java/core/src/java/org/apache/orc/util/BloomFilter.java
omalley: No, this one is explicitly the default charset to be compatible. BloomFilterUtf8 is the fixed one.

java/core/src/java/org/apache/orc/OrcFile.java
```diff
@@ -231,6 +232,33 @@ public static Reader createReader(Path path,
...
+      return id;
+    }
+
+    public static BloomFilterVersion fromString(String s) {
+      for (BloomFilterVersion version: values()) {
+        if (version.id.equals(s)) {
+          return version;
+        }
+      }
+      throw new IllegalArgumentException("Unknown BloomFilterVersion " + s);
```
omalley: I don't see how. To get the value, the code will always do this conversion after getting the string from OrcConf, so the error will be caught coming out.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -106,49 +198,58 @@ public OrcIndex readRowIndex(StripeInformation stripe,
...
+        }
+        int column = stream.getColumn();
+        if (stream.hasKind() && range.getOffset() <= offset) {
+          switch (stream.getKind()) {
+            case ROW_INDEX:
+              if (included == null || included[column]) {
+                ByteBuffer bb = range.getData().duplicate();
+                bb.position((int) (offset - range.getOffset()));
+                bb.limit((int) (bb.position() + stream.getLength()));
+                indexes[column] = OrcProto.RowIndex.parseFrom(InStream.create("index",
```
omalley: I just copied the code that was there, but sure.

java/core/src/java/org/apache/orc/util/BloomFilterIO.java
omalley: We'd need to change the protobuf, which won't be backwards compatible. I guess we could do that for the new bloom_filter_utf8 stream.

omalley:
Ok, the latest push has a few changes:
* bloom_filter_utf8 streams use a new encoding with bytes instead of long[]. This is much more efficient for performance and storage size.
* all column types now have bloom_filter_utf8 streams (largely to get the new representation)
* the default is to just write the new bloom_filter_utf8 streams that old readers will ignore. There is an option to write both bloom_filter and bloom_filter_utf8 streams to support old readers.
* there is an option for new readers to ignore the old bloom filters.
* files generated after hive-12055 will correctly use the utf8 encoding even for the bloom_filter stream.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -1168,8 +1193,9 @@ public OrcIndex readRowIndex(int stripeIndex, boolean[] included,
       sargColumns = sargColumns == null ?
           (sargApp == null ? null : sargApp.sargColumns) : sargColumns;
     }
-    return dataReader.readRowIndex(stripe, stripeFooter, included, indexes, sargColumns,
-        bloomFilterIndex);
+    return dataReader.readRowIndex(stripe, evolution.getFileType(0), stripeFooter,
```
prasanthj: Should the index change based on ACID file or not?

omalley: Actually RecordReaderImpl.mapSargColumnsToOrcInternalColIdx already maps column names to the physical file column ids. It handles ACID files by looking for the field names in row field instead of the root. So at this level, it should be consistently using the file id.

prasanthj:
lgtm, +1

PR #65, opened 2016-11-16
ORC-110: [C++] getTimezoneByFilename is not thread safe

majetideepak:
…multiple threads

### DIFF
--- a/c++/src/Timezone.cc
+++ b/c++/src/Timezone.cc
@@ -690,6 +690,8 @@ namespace orc {
    * Results are cached.
    */
   const Timezone& getTimezoneByFilename(const std::string& filename) {
+    // ORC-110
+    pthread_mutex_lock(&timezone_mutex);
     std::map<std::string, Timezone*>::iterator itr =
       timezoneCache.find(filename);
     if (itr != timezoneCache.end()) {
@@ -730,6 +732,7 @@ namespace orc {
     }
     Timezone* result = new TimezoneImpl(filename, buffer);
     timezoneCache[filename] = result;
+    pthread_mutex_unlock(&timezone_mutex);
     return *result;
   }
 
@@ -737,10 +740,7 @@ namespace orc {
    * Get the local timezone.
    */
   const Timezone& getLocalTimezone() {
-    pthread_mutex_lock(&timezone_mutex);
-    const Timezone& timezone = getTimezoneByFilename(LOCAL_TIMEZONE);
-    pthread_mutex_unlock(&timezone_mutex);
-    return timezone;
+    return getTimezoneByFilename(LOCAL_TIMEZONE);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


c++/src/Timezone.cc
```diff
@@ -650,6 +651,7 @@ namespace orc {
     DIAGNOSTIC_IGNORE("-Wglobal-constructors")
     DIAGNOSTIC_IGNORE("-Wexit-time-destructors")
   #endif
+  static pthread_mutex_t timezone_mutex;
   static std::map<std::string, Timezone*> timezoneCache;
```
majetideepak: `getTimezoneByFilename` updates this global cache causing a race.

ClydeZhao:
I came across this too! Please merge.

ClydeZhao:
@majetideepak I think it is better to add the mutex at line:730 `timezoneCache[filename] = result;`. 

Moreover, the c++11 `std::mutex` looks better here.

majetideepak:
I believe ORC library must be compatible without C++11 support too.
I don't think just locking the write `timezoneCache[filename] = result;` is sufficient. The reads must be protected too. A write by one thread could be mutating an element being read by another thread.

ClydeZhao:
The last commit has problems. Line 696 and throws will lead to deadlock. I think we should use [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) here.

majetideepak:
You are right. I added an unlock before each throw and return. 
I am in favor of RAII as well. However, I don't think there are any RAII locks before C++11.

ClydeZhao:
Should we write our own? It is not complicated after all. I am thinking a class like this would be good enough:
```c++
  /**
   * Lock guard for pthread_mutex_t object using RAII
   * The Lock is automatically release when exiting current scope.
   */
  class Lock {
    public:
      explicit Lock(pthread_mutex_t* mutex) : mutex_ptr_(mutex) {
        pthread_mutex_lock(mutex_ptr_);
      }
      ~Lock() { pthread_mutex_unlock(mutex_ptr_); }
    private:
       // no default constructor
       Lock();
       // prohibit copying
       Lock(const Lock&);
       Lock& operator=(const Lock&);

       pthread_mutex_t* mutex_ptr_;
  };
```

majetideepak:
This is a good idea. Patch coming up.

majetideepak:
The tests are failing because of this error
[ERROR] Plugin org.apache.maven.plugins:maven-remote-resources-plugin:1.5 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-remote-resources-plugin:jar:1.5: Could not transfer artifact org.apache.maven.plugins:maven-remote-resources-plugin:pom:1.5 from/to central (https://repo.maven.apache.org/maven2): hostname in certificate didn't match: <repo.maven.apache.org> != <repo1.maven.org> OR <repo1.maven.org> -> [Help 1]

majetideepak:
@ClydeZhao @omalley any comments ?

omalley:
This is looking great, thanks!

The CI build passed and it built and passed tests on my mac. Let me run it through the docker scripts to make sure it passes on each of the Linux platforms that we support.

ClydeZhao:
LGTM :)

PR #70, opened 2016-12-06
ORC-117 ConvertTreeReaderFactory gets an index out of bounds

omalley:
ConvertTreeReaderFactory gets an ArrayIndexOutOfBoundsException when columns are added and types changed.


prasanthj:
LGTM, +1

PR #74, opened 2016-12-16
ORC-123: Fix memory leaks in ToolTest.cc

majetideepak:


tools/test/ToolTest.cc
```diff
@@ -82,8 +83,10 @@ int runProgram(const std::vector<std::string>& command,
                std::string &err) {
 
   // create temporary filenames for stdout and stderr
-  char *stdoutName = strdup("/tmp/orc-test-stdout-XXXXXXXX");
-  char *stderrName = strdup("/tmp/orc-test-stderr-XXXXXXXX");
```
majetideepak: `strdup` does a malloc internally and must be freed.

omalley: Would it be easier to use std::array or std::vector?

majetideepak: `std::string` should work too. I am updating the patch.

PR #76, opened 2017-01-04
ORC-119. Create an API to separate out layout from the writer.

omalley:


dain:
Can you describe why you want to make this change?

omalley:
The original version is https://issues.apache.org/jira/browse/HIVE-14453

The goal is to build support for a write through cache, although Sergey is the owner of the use case.

PR #78, opened 2017-01-06
ORC-128. Add getStatistics to Writer API

omalley:
Allow user to getStatistics while writing files.

### DIFF
--- a/java/core/src/java/org/apache/orc/Writer.java
+++ b/java/core/src/java/org/apache/orc/Writer.java
@@ -111,4 +111,16 @@ public interface Writer {
    * @param userMetadata - user metadata
    */
   public void appendUserMetadata(List<OrcProto.UserMetadataItem> userMetadata);
+
+  /**
+   * Get the statistics about the columns in the file. The output of this is
+   * based on the time at which it is called. It shall use all of the currently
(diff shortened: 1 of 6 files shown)


dain:
Just curious, how do you plan on using this?

prasanthj:
@dain Hive already uses stats API (reader side and writer side) to get basic statistics like (numRows, rawDataSize, etc.) from the footer to avoid row-by-row stats gathering. This new API is to extend the same for column statistics (although ORC is missing NDV at this point).

dain:
I'm curious how the writer statistics are used since it is the "end  of the line".  Is it just to display to a user (e.g., logging), or is the engine making decisions based on the information?

prasanthj:
Hive auto-gathers statistics during write (INSERT, CTAS..). Just before closing the file, the file sink operator gets the statistics from Writer, publishes it for aggregation by the client. This just pushes the stats collection part from processOp (row-by-row or vector batch processing) to closeOp.

Similarly Reader side interface is used by ANALYZE queries to compute statistics just by reading footer.

prasanthj:
>  is the engine making decisions based on the information?

Not at this point.

PR #80, opened 2017-01-13
ORC-126. Remove our fork of storage-api.

omalley:
This patch removes our fork of Hive's storage api and replaces it with the official release.


omalley:
The CI failed because the hive storage api 2.2.0 release hasn't been synced to maven central yet.

PR #81, opened 2017-01-16
fix up release build instructions

graphaelli:


omalley:
I've committed this and deployed it to the site. Thanks!

PR #84, opened 2017-01-19
ORC-132. Implement a merge file method and fix the number of rows wri…

omalley:
…tten

during merge. (omalley)

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcFile.java
+++ b/java/core/src/java/org/apache/orc/OrcFile.java
@@ -694,7 +694,6 @@ public class OrcFile {
    * @param schema the writer schema
    * @param fileVersion the writer fileVersion
    * @param writerVersion the writer writerVersion
-   * @param compressionSize the compression buffer size
    * @param rowIndexStride the row index stride
    * @param compression the compression that was used
    * @param userMetadata the user metadata
@@ -705,7 +704,6 @@ public class OrcFile {
   static boolean readerIsCompatible(TypeDescription schema,
                                     Version fileVersion,
                                     WriterVersion writerVersion,
-                                    int compressionSize,
                                     int rowIndexStride,
                                     CompressionKind compression,
                                     Map<String, ByteBuffer> userMetadata,
@@ -720,12 +718,6 @@ public class OrcFile {
     if (reader.getCompressionKind() != compression) {
       LOG.info("Can't merge {} because of different compression {} vs {}",
           path, reader.getCompressionKind(), compression);
-      return false;
-    }
-    if (compression != CompressionKind.NONE &&
-        reader.getCompressionSize() != compressionSize) {
-      LOG.info("Can't merge {} because of different compression sizes {} vs {}",
-          path, reader.getCompressionSize(), compressionSize);
       return false;
     }
     if (reader.getFileVersion() != fileVersion) {
@@ -782,7 +774,7 @@ public class OrcFile {
    */
   public static List<Path> mergeFiles(Path outputPath,
                                       WriterOptions options,
-                                      Path... inputFiles) throws IOException {
+                                      List<Path> inputFiles) throws IOException {
     Writer output = null;
     byte[] buffer = new byte[0];
     Configuration conf = options.getConfiguration();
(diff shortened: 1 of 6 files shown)


java/core/src/java/org/apache/orc/OrcFile.java
```diff
@@ -642,4 +671,187 @@ public static Writer createWriter(Path path,
...
+   * @param outputPath the output file
+   * @param options the options for writing with although the options related
+   *                to the input files' encodings are overridden
+   * @param inputFiles the list of files to merge
+   * @return the list of files that were successfully merged
+   * @throws IOException
+   */
+  public static List<Path> mergeFiles(Path outputPath,
+                                      WriterOptions options,
+                                      Path... inputFiles) throws IOException {
```
prasanthj: List instead of varargs? Or another interface that converts List to varargs.

java/core/src/java/org/apache/orc/OrcFile.java
```diff
@@ -642,4 +671,187 @@ public static Writer createWriter(Path path,
...
+      int stripeNum = 0;
+      result.add(input);
+
+      for(StripeInformation stripe: reader.getStripes()) {
+        int length = (int) stripe.getLength();
+        if (buffer.length < length) {
+          buffer = new byte[length];
+        }
+        long offset = stripe.getOffset();
+        inputStream.readFully(offset, buffer, 0, length);
```
prasanthj: wrap this inside try .. catch? readFully throws EOFException

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -2944,4 +2948,164 @@ public void testWriterVersion() throws Exception {
...
+  static ByteBuffer fromString(String s) {
+    return ByteBuffer.wrap(s.getBytes(StandardCharsets.UTF_8));
+  }
+
+  static byte[] fromInt(int x) {
+    return Integer.toHexString(x).getBytes(StandardCharsets.UTF_8);
+  }
+
+  @Test
+  public void testGoodMerge() throws Exception {
```
prasanthj: Will be good to have some tests for bad merge (incompatibility tests)

java/core/src/java/org/apache/orc/OrcFile.java
omalley: Really? It isn't clear to me that List<Path> is that much easier than a Path[] and the Path[] requires 1 fewer objects. But Ok.

java/core/src/java/org/apache/orc/OrcFile.java
omalley: No, I'd rather have the whole method blow out than hide the exception. It can only happen in the case of a broken ORC file.

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
omalley: Ok.

omalley: The second merge in testGoodMerge, I did have two files that are dropped for being incompatible.

java/core/src/java/org/apache/orc/OrcFile.java
prasanthj: Any exception in the entire code can leave the output file open and not readable. Isn't it?

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
prasanthj: ah! ok. Missed that.

java/core/src/java/org/apache/orc/OrcFile.java
omalley: True, but the same happens if you have a Writer that throws an exception.

omalley:
The new version of the patch:
* uses the maximum buffer size of the input files rather than reject them.
* uses List rather than Path[] as the input parameter
* tests that the buffer size is updated correctly
* renames the testGoodMerge to testMerge since it tests both good and bad merges

java/core/src/java/org/apache/orc/OrcFile.java
prasanthj: Appending stripes throwing exception could be for different reasons than writer and some of them are still avoidable. For example: Merging v11 files without stripe statistics will return null for statList throwing NPE. This could be skipped or avoided without leaving corrupted output file.

omalley: The current WriterImpl.appendStripe() will fail if it is called on files that don't have stripe statistic, so that feels like we can handle it in a follow up bug.

If the merge can't complete, throwing an exception seems like the best course. I guess we can delete the partially written output file if that feels cleaner to you.

PR #85, opened 2017-01-23
Orc-100 fixing maven compile site errors and ensuring build failure in case of findbug errors.

tarunk2:
#### Changes include:
- Fixing `mvn clean compile site` errors. `tools` module should have findbugs `exclude.xml`.
- Maven build should fail if there are any findbug errors.

#### Testing:
- [x] maven build passes locally and there no findbug errors.

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -199,25 +199,29 @@
             <target>1.7</target>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>findbugs-maven-plugin</artifactId>
+          <version>3.0.3</version>
+          <configuration>
+            <effort>Max</effort>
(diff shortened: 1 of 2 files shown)


omalley:
Thanks. We also need to add the plugin to the build sections of the children poms and unfortunately fix a few items.  Does this look ok? https://github.com/omalley/orc/commits/orc-100

tarunk2:
Thanks, looks good. Alternatively, we can move `findbugs-maven-plugin` outside of `pluginManagement` and it won't need explicit plugin in children poms.

tarunk2:
@omalley should i pull your commit in this PR or you will merge changes from your fork?

omalley:
I can do the merge, thanks though!

omalley:
I did also move the plugin to the build in the root pom.

PR #87, opened 2017-01-27
ORC-135: PPD for timestamp is wrong when reader and writer timezones are different

prasanthj:
When reader and writer timezones are different, PPD evaluation does not offset the timezone when reading the min and max values. This can result is wrong PPD evaluation and hence incorrect results.

Example:
Table written in US/Eastern timezone. All values in this table are "2007-08-01 00:00:00.0".
**PPD disabled**
```
hive> set hive.optimize.index.filter=false;
hive> select ORDER_DATE from ORDER_FACT_small where ORDER_DATE='2007-08-01 00:00:00.0' limit 1;
2007-08-01 00:00:00.0
OK
```

**PPD enabled**
```
set hive.optimize.index.filter=true;
select ORDER_DATE from ORDER_FACT_small where ORDER_DATE='2007-08-01 00:00:00.0' limit 1;
OK
```
No rows are returned when PPD is enabled (reader timezone is UTC)

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
@@ -19,6 +19,9 @@ package org.apache.orc.impl;
 
 import java.sql.Date;
 import java.sql.Timestamp;
+import java.util.Calendar;
+import java.util.GregorianCalendar;
+import java.util.TimeZone;
 
 import org.apache.hadoop.hive.serde2.io.HiveDecimalWritable;
 import org.apache.hadoop.hive.common.type.HiveDecimal;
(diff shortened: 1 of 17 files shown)


java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -19,6 +19,9 @@
 
 import java.sql.Date;
 import java.sql.Timestamp;
+import java.util.Calendar;
+import java.util.GregorianCalendar;
+import java.util.TimeZone;
```
prasanthj: These imports are not required. Will remove it in the next patch.

omalley:
Ok, I took a pass at this:

https://github.com/omalley/orc/tree/orc-135

Changes:
* Block all use of PPD on timestamps from before ORC-135.
* Block PPD on timestamp with numeric constraints
* TimestampColumnStatistics.getMinimum and getMaximum return Timestamps in the local timezone.
* Loading old TimestampColumnStatistics loads the min and max assuming local timezone.
* Add utilities to SerializationUtils for conversion to and from utc.
* Make BloomFilter.encoding an int so that we can detect future values.
* Future BloomFilter encodings are not used.
* Some more test cases

PR #90, opened 2017-02-04
ORC-135. Fixing PPD for timestamps across timezones.

omalley:


omalley:
Of course we should rebase & squash before committing.

PR #96, opened 2017-02-27
ORC-151. Minimize the size of the tools uber jar.

omalley:
This patch takes the tools uber jar from 29mb to 23mb.


l1x:
Thank you!

PR #99, opened 2017-03-03
ORC-148: Move ORC PPD failure messages from WARN to DEBUG level

prasanthj:


java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -468,16 +470,15 @@ static TruthValue evaluatePredicateRange(PredicateLeaf predicate, Object min,
       }
       // in case failed conversion, return the default YES_NO_NULL truth value
     } catch (Exception e) {
-      if (LOG.isWarnEnabled()) {
+      if (LOG.isDebugEnabled()) {
```
omalley: This isn't ok. We can't take exceptions and make them invisible. We need to avoid the exception.

omalley:
Here's a more complete solution. 

https://github.com/omalley/orc/tree/orc-148

It will still get 1 log line per a stripe, but that is much better than 1 per a row group. It also downgrades the expected exceptions to info. I considered downgrading it further to debug,
but we really should even fix those via schema evolution.

prasanthj:
lgtm, +1.

omalley:
I realized that we really shouldn't be using the variable, but rather the sarg leaf as the thing we are caching the exceptional results for. (So that we avoid excessive logging)

PR #100, opened 2017-03-16
ORC-157. Fix C++ unit tests to be more timezone agnostic.

omalley:
This change:
1. Introduces a new version of the examples/orc_split_elim.orc file that was written by the current master version of ORC as examples/orc_split_elim_new.orc.
2. The old version is left as an example with broken statistics.
3. The contents test uses the new version.


amorskoy:
Checked - tests passed now

PR #102, opened 2017-03-21
ORC-166 : add codec pool to ORC; make sure end is called on underlyin…

sershe:
…g codecs (Sergey Shelukhin)


java/core/src/java/org/apache/orc/impl/HadoopShimsCurrent.java
```diff
@@ -33,31 +34,62 @@
  */
 public class HadoopShimsCurrent implements HadoopShims {
 
-  private static class DirectDecompressWrapper implements DirectDecompressor {
-    private final org.apache.hadoop.io.compress.DirectDecompressor root;
+  private static class SnappyDirectDecompressWrapper implements DirectDecompressor {
```
prasanthj: Is this change required? I don't see any change that require having separate Snappy and Zlib wrappers.

PR #103, opened 2017-03-21
ORC-167: Missing SchemaEvolution API breaks hive compatibility

prasanthj:


java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
```diff
@@ -128,6 +128,11 @@ public SchemaEvolution(TypeDescription fileSchema,
   }
 
   @Deprecated
+  public SchemaEvolution(TypeDescription fileSchema, Reader.Options options) {
+    this(fileSchema, null, options);
```
omalley: That doesn't work. We removed that one in ORC-147 precisely because we need compatibility with Hive 2.1.

prasanthj:
Closing this as this issue is fixed in HIVE-15841

PR #105, opened 2017-04-11
ORC-170. TypeDescription parent links aren't always initialized.

omalley:


java/core/src/test/org/apache/orc/TestTypeDescription.java
```diff
@@ -198,4 +198,16 @@ public void testParserExtraStuff() {
...
+  @Test
+  public void testConnectedSubtrees() {
+    TypeDescription type =
+        TypeDescription.fromString("struct<field1:array<struct<field2:int>>>");
+    TypeDescription leaf = type.getChildren().get(0)
+        .getChildren().get(0)
+        .getChildren().get(0);
+    assertEquals(3, leaf.getId());
+    assertEquals(0, type.getId());
+    assertEquals(3, leaf.getId());
```
prasanthj: can you also add a small test for map<> type?

PR #109, opened 2017-04-17
ORC-172: Fix C++ build issue with Centos6

majetideepak:
The patch for ORC-122 requires explicit `copy` and `move` of the `Timezone` class since it contains a `unique_ptr` as one of its member. The simpler fix is to make the member a shared_ptr. A similar change is made for the `TimezoneCache` map. This is reasonable since this code is not on the critical path.


majetideepak:
@omalley  You raised this issue last week. I verified this fix on docker. Thanks!

omalley:
I just committed this (squashed and stripping trailing whitespace).
Thanks, Deepak!

PR #110, opened 2017-04-19
ORC-87: [C++] Handle missing timezone conversion for timestamp statistics

majetideepak:


c++/src/Statistics.cc
```diff
@@ -279,20 +277,28 @@ namespace orc {
...
-      maximum = stats.maximum();
+      _hasMinimum = stats.has_minimumutc() || (stats.has_minimum() && statContext.writerTimezone);
+      _hasMaximum = stats.has_maximumutc() || (stats.has_maximum() && statContext.writerTimezone);
+
+      if (stats.has_minimumutc()) {
+        minimum = stats.minimumutc();
+      } else {
+        minimum = stats.minimum() + statContext.writerTimezone->getEpoch();
+      }
+      if (stats.has_minimumutc()) {
```
omalley: You mean has_maximumutc() here.

c++/src/Reader.cc
```diff
@@ -526,16 +526,25 @@ namespace orc {
...
     return std::unique_ptr<Statistics>
       (new StatisticsImpl(metadata->stripestats
-                          (static_cast<int>(stripeIndex)),
-                          hasCorrectStatistics()));
+                          (static_cast<int>(stripeIndex)), statContext));
   }
 
   std::unique_ptr<Statistics> ReaderImpl::getStatistics() const {
+    StatContext statContext;
+    statContext.correctStats = hasCorrectStatistics();
```
omalley: You need the timezone here too don't you?

c++/src/Statistics.hh
```diff
@@ -23,10 +23,17 @@
...
 
+#include "Timezone.hh"
 #include "TypeImpl.hh"
 
 namespace orc {
 
+  struct StatContext {
+    bool correctStats;
+    const Timezone* writerTimezone;
+    StatContext() : correctStats(false), writerTimezone(NULL) {}
```
omalley: Shouldn't the constructor take the correctStats and writerTimezone and make them const?

c++/src/Reader.cc
majetideepak: `getStatistics()` is returning the statistics in the file footer. As far as I understood, there is no timezone information available in the file footer. Looks like the timezone is stored for each stripe.
I am also curious about this design choice. I would imagine the file footer to have the timezone information. Is it possible for different stripes to have different timezone values?

c++/src/Statistics.cc
majetideepak: You are right. Will fix this.

c++/src/Statistics.hh
majetideepak: We can change `pointer to const` later too.
 But you are right that it is more efficient to use const for the values too. Will change this.

omalley:
I think we should probably add new methods for timestamp statistics for getLowerBound and getUpperBound that will be the same for the new version of the statistics, but be a looser bound in the case of the older files.

majetideepak:
@omalley I have added methods for timestamp statistics for getLowerBound an getUpperBound

c++/src/Statistics.cc
```diff
@@ -279,20 +277,55 @@ namespace orc {
...
+        // multiply the offset by 1000 to convert to millisecond
+        maximum = stats.maximum() + (statContext.writerTimezone->getVariant(writerTimeSec).gmtOffset) * 1000;
+        upperBound = maximum;
+      } else {
+        maximum = 0;
+        // add 1 day to handle unknown TZ
+        upperBound = stats.maximum() +  (SECONDS_PER_DAY * 1000);
+      }
+      // Add 1 millisecond to account for microsecond precision of values
+      maximum += 1;
```
omalley: I don't understand this and the unit tests pass without it.

majetideepak: The timestamp values have microsecond precision `eg: 2037-01-01 00:00:00.000999`.
However, I see that the statistics computed have millisecond precision `2037-01-01 00:00:00.000` and they are not rounded off to the ceil. This is to account for the ceil. I will update the test.
Thanks! for catching this.

omalley: Actually, it should push up the upperBound, but maximum is less confusing as it was.

c++/src/Reader.cc
```diff
@@ -526,16 +526,22 @@ namespace orc {
     if (metadata.get() == nullptr) {
       throw std::logic_error("No stripe statistics in file");
     }
+    auto currentStripeInfo = footer->stripes(static_cast<int>(stripeIndex));
```
wgtmac: Should we consider some older compliers which don't support c++11 very well? I don't see anywhere else is using keyword auto.

c++/src/Statistics.cc
```diff
@@ -279,20 +277,55 @@ namespace orc {
...
+        minimum = stats.minimumutc();
+        lowerBound = minimum;
+      } else if (statContext.writerTimezone) {
+        int64_t writerTimeSec = stats.minimum() / 1000;
+        // multiply the offset by 1000 to convert to millisecond
+        minimum = stats.minimum() + (statContext.writerTimezone->getVariant(writerTimeSec).gmtOffset) * 1000;
+        lowerBound = minimum;
+      } else {
+        minimum = 0;
+        // subtract 1 day 1 hour (25 hours) in milliseconds to handle unknown TZ and daylight savings
```
majetideepak: @omalley Note the change here to 25 hours. The additional 1 hour is to account for daylight savings.

PR #111, opened 2017-04-20
ORC-173: [C++] Expose Indexes as API

majetideepak:


majetideepak:
Rebases off PR#110. This should be committed after that.

c++/include/orc/Statistics.hh
```diff
@@ -321,6 +321,32 @@ namespace orc {
      * @return maximum value
      */
     virtual int64_t getMaximum() const = 0;
+
+    /**
+     * check whether column has a lowerBound
+     * @return true if column has a lowerBound
+     */
+    virtual bool hasLowerBound() const = 0;
```
wgtmac: I saw these lines of changes are also in ORC-87. Is it a good idea to remove them here to make this change more concise?

majetideepak: I rebase this patch off ORC-87. That is why the changes appear in the overall patch.
I just squashed the commits relevant to only this JIRA. There are two commits now. Once is the ORC-87 commit. The second commit is for this JIRA.

c++/src/Statistics.hh
```diff
@@ -605,6 +676,43 @@ namespace orc {
     }
   };
 
+  class StripeStatisticsImpl: public StripeStatistics {
+  private:
+    std::unique_ptr<StatisticsImpl> columnStats;
+    std::vector<std::vector<std::unique_ptr<const ColumnStatistics> > > rowIndexStats;
```
wgtmac: Similar to my previous comment in ORC-87. If we need to be compatible with older C++ standards, this line may not work.

tools/src/FileStatistics.cc
```diff
@@ -55,6 +55,11 @@ void printStatistics(const char *filename) {
         std::cout << "--- Column " << k << " ---" << std::endl;
         std::cout << stripeStats->getColumnStatistics(k)->toString()
                   << std::endl;
+        for(unsigned int r = 0; r < stripeStats->getNumberOfRowIndexStats(k); ++r) {
```
wgtmac: Is it better to make printing row index stats optional as it may be too verbose?

c++/src/Statistics.hh
majetideepak: If using `unique_ptr` is your concern, I believe this is handled by the ORC Adaptor header.
`./c++/src/Adaptor.hh:86:  #define unique_ptr auto_ptr`

wgtmac: auto_ptr is not copy-assignable so we can't create a container of auto_ptr. This is the problem.

majetideepak: Correct! Making it a `shared_ptr`

omalley:
* You've introduced a trailing space in FileStatistics.cc.
* It would be good if you switched to using getopt_long in FileStatistics, like we did in FileMetadata.cc.
* Since you added a new example file, you should add it to TestMatch.cc and make an expected file for it.
* Please keep your column width to 100 characters.
* You should remove min & max from string, date, and decimal statistics for files before HIVE-8732.

majetideepak:
Currently, statistics of all types before WriterVersion HIVE-8732 are ignored.
Do you want to limit the check only to `string`, `date`, and `decimal` statistics ?

PR #113, opened 2017-04-27
ORC-176: Refactor common classes for writer and reader

wgtmac:
This is mainly a refactoring change for ORC-176, including:
1. Extracted common classes and functions into Common.hh and Common.cc;
2. Put InputStream interface and its implementations from Compression.hh to Stream.hh;
3. Refactored ColumnStatistic classes to reuse code as many as possible and added setters for writer;
4. Added more functions in Int128 class for decimal operations;
5. Added buildTypeFromString() method in Type class to construct ORC type.


c++/include/orc/Statistics.hh
```diff
@@ -37,12 +37,76 @@ namespace orc {
...
-    virtual uint64_t getNumberOfValues() const = 0;
+    virtual uint64_t getNumberOfValues() const {
+      return valueCount;
+    }
+
+    /**
+     * Set the number of values in this column
+     * @param newValueCount new number of values to be set
+     */
+    virtual void setNumberOfValues(uint64_t newValueCount) {
```
majetideepak: Are these and the below methods required in the public API?
I think they belong to the Impl classes in `c++/src/Statistics.hh`.

c++/include/orc/Statistics.hh
```diff
@@ -37,12 +37,76 @@ namespace orc {
...
+     * Convert statistics to protobuf version
+     * @param pbStats output of protobuf stats
+     */
+    virtual void toProtoBuf(proto::ColumnStatistics& pbStats) const {
+      pbStats.set_hasnull(hasNullValue);
+      pbStats.set_numberofvalues(valueCount);
+    }
+
+  protected:
+    uint64_t valueCount;
```
majetideepak: The class members should belong to the Impl class in  `c++/include/orc/Statistics.hh` as well.
What is the reason behind moving them here?

c++/src/Common.cc
```diff
@@ -0,0 +1,107 @@
...
+ * limitations under the License.
+ */
+
+#include "orc/Common.hh"
+
+#include <sstream>
+
+namespace orc {
+
+  std::string compressionKindToString(CompressionKind kind) {
```
majetideepak: Put all these methods pertaining to types in `c++/src/TypeImpl.cc`

c++/src/Stream.cc
```diff
@@ -0,0 +1,222 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
```
majetideepak: Use `InputStream.cc` instead?

wgtmac:
@majetideepak Yes I think it will be much easier if I separate it into more smaller PRs. Originally we wanted to do this large PR because it may take a lot of effort for us to split them. As I saw your PRs of ORC-87 and ORC-173, I decide to split them because your code may have conflicts with ours, e.g. in TimestampColumnStatistics. I will update this PR later to make it smaller and create more JIRAs to reflect the change.

c++/src/Stream.cc
wgtmac: We will add OutputStream to this file so I think name it Stream.cc is fine.

c++/include/orc/Common.hh
```diff
@@ -0,0 +1,167 @@
...
+
+#include "orc/Vector.hh"
+#include "orc/Type.hh"
+#include "Exceptions.hh"
+#include "wrap/orc-proto-wrapper.hh"
+
+#include <string>
+
+namespace orc {
+  enum CompressionKind {
```
wgtmac: I think these are better to stay in Common.hh as Type.hh is specific for schema types. Same as below.

wgtmac:
@majetideepak Thanks for your previous review. I have replied to your comments and also have deleted some changes to make this commit more concise. Can you review it again?

c++/include/orc/Statistics.hh
wgtmac: If we put this into Impl classes, then there are many duplicate code which is not elegant. So we decided to lift them up to the base classes so that duplicate codes are reduced significantly while we were implementing the writer. What do you think?

majetideepak:
@wgtmac I will provide feedback shortly. For some reason, Github is not saving my comments.

c++/include/orc/Common.hh
majetideepak: I think all the ORC types including `Compression`, `Stream`, etc. should go to `Type.hh`. 
I would do away with `Common.hh`.
@omalley should give his view here.

c++/src/Stream.cc
majetideepak: We must clearly distinguish the different kinds of `Streams` :). We have ORC Streams and also the in-memory streams. Maybe we can create a new folder `io` and put all the file and memory streams there.

majetideepak:
@wgtmac I made some more comments. Thanks for minimizing the patches.
I learnt the following rules of thumb for C++ apache projects, which might also help us improve the ORC C++ library significantly. I feel this is the right opportunity to do this.
1. An implementation must follow the PIMPL idiom. The public headers must include only the minimal API and all the class members, implementation will go in the Impl classes.
2. We must follow a coding style. How about we pick the google C++ coding style https://google.github.io/styleguide/cppguide.html?
3. We must try to disambiguate the class names and file names as much as possible so that anyone new to the code base can quickly understand the code.

Let me know what your thoughts are. Thanks.

c++/src/Stream.cc
wgtmac: This is a good idea. I will do that.

wgtmac:
@majetideepak I would absolutely support 2 and 3 as it will definitely improve the code quality and readability. As for 1, I would say let's do it in caution. At present we have classes like ColumnStatistics and ColumnReader which have a one-to-many inheritance relation. Particularly, class ColumnStatistics has many duplicate code which takes more effort for use when we were adding functions to it for implementing writer. This can be improved first.

majetideepak:
I will fix the ColumnStatistics code. Templates should help here.

wgtmac:
@majetideepak That's cool. You may create a separate JIRA for it and I will continue [ORC-184](https://issues.apache.org/jira/browse/ORC-184) after your fix is done. Thanks!

wgtmac:
@omalley Can you help review this change? It has no new code and our main argument is on the correct place (and naming) to put these lines. Thanks!

omalley:
I'm sorry, but I committed ORC-183. Can you rebase this patch as a single commit on top of the current master.

wgtmac:
@omalley I've created a separate [PR](https://github.com/apache/orc/pull/118/) based on the current master per your request and thus close this one.

PR #114, opened 2017-04-27
ORC-181: ByteRleDecoder doesn't seek correctly

wgtmac:
Fixed by explicitly adding "ByteRleDecoderImpl::" to the front of skip() function in the function ByteRleDecoderImpl::seek(PositionProvider& location).


wgtmac:
@omalley I've added a test case to cover this case. Thanks for the suggestion!

PR #115, opened 2017-04-28
ORC-183: Add a method in Type to build type

wgtmac:
Added static Type* buildTypeFromString(const std::string& input)

### DIFF
--- a/c++/include/orc/Type.hh
+++ b/c++/include/orc/Type.hh
@@ -86,7 +86,7 @@ namespace orc {
     /**
      * Build a Type object from string text representation.
      */
-    static Type* buildTypeFromString(const std::string& input);
+    static ORC_UNIQUE_PTR<Type> buildTypeFromString(const std::string& input);
   };
 
   const int64_t DEFAULT_DECIMAL_SCALE = 18;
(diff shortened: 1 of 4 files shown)


c++/include/orc/Type.hh
```diff
@@ -82,6 +82,11 @@ namespace orc {
      * @return a reference to the union type
      */
     virtual Type* addUnionChild(ORC_UNIQUE_PTR<Type> fieldType) = 0;
+
+    /**
+     * Build a Type object from string text representation.
+     */
+    static Type* buildTypeFromString(const std::string& input);
```
omalley: Let's make it a unique_ptr<Type> to make it clear that the caller owns the pointer.

c++/src/TypeImpl.cc
```diff
@@ -491,4 +491,190 @@ namespace orc {
     return std::unique_ptr<Type>(result);
   }
 
+  Type * Type::buildTypeFromString(const std::string& input) {
```
omalley: This should probably be factored in to smaller functions. Take a look at the equivalent java https://github.com/apache/orc/blob/master/java/core/src/java/org/apache/orc/TypeDescription.java#L396 and you'll see parseName, parseCategory, parseInteger, etc.

wgtmac: Make sense. I will do the refactoring soon.

PR #116, opened 2017-05-02
ORC-185: [C++] Simplify Statististics Implementation

majetideepak:


majetideepak:
@wgtmac   how does the second commit look?

c++/src/Statistics.hh
```diff
@@ -41,49 +41,181 @@ namespace orc {
...
+    T maximum_;
+    T sum_;
+  public:
+    InternalStatisticsImpl() {
+      hasNull_ = false;
+      hasMinimum_ = false;
+      hasMaximum_ = false;
+      hasSum_ = false;
+      hasTotalLength_ = false;
+      totalLength_ = -1;
```
wgtmac: Why the default value is set to -1 not 0?

c++/src/Statistics.hh
```diff
@@ -41,49 +41,181 @@ namespace orc {
...
+      valueCount_ = -1;
```
wgtmac: Same as above. Also, we may need some functions like void increase(uint64 count) to increase valueCount_.

c++/src/Statistics.hh
```diff
@@ -41,49 +41,181 @@ namespace orc {
   };
 
 /**
+ * Internal Statistics Implementation
+ */
+
+  template <typename T>
```
wgtmac: We may need some functions like void increase(uint64 count) to increase valueCount. I can add them when needed. 
My main concern for using templates is that we need to compare, update, merge ColumnStatistics, and transform to protobuf version for implementing writers and using templates will also introduce some duplicate code. It means we still need to do template specialization for different types like Date, Timestamp, Decimal, etc. if we want to let class ColumnStatistics to handle the update (e.g. use ColumnStatistics<T>::update(T value) to update min/max for type T). Otherwise we may need to let specific ColumnWriters to be responsible for update (e.g. DecimalColumnWriter to compare min/max of decimal values and then use setMax/setMin of ColumnStatistics<Decimal> to update the values).

c++/src/Statistics.hh
```diff
@@ -41,49 +41,181 @@ namespace orc {
...
+  class InternalStatisticsImpl {
+  private:
+    bool hasNull_;
```
wgtmac: For the naming here, underscore is not consistent with the current code base. (Only current XXColumnStatistics classes use "_" in the beginning of members but I don't think they should be fixed)

wgtmac:
Overall the PIMPL looks nicer than the current implementation. My main concern for using templates is that we need to compare & update values of different types, merge ColumnStatistics, and transform to protobuf version for implementing writers and using templates may not help in this case. It means we still need to do template specialization for different types like Date, Timestamp, Decimal, etc. if we want to let class ColumnStatistics to handle the update (e.g. use ColumnStatistics<T>::update(T value) to update min/max for type T). Otherwise we may need to let specific ColumnWriters to be responsible for update (e.g. DecimalColumnWriter to compare min/max of decimal values and then use setMax/setMin of ColumnStatistics<Decimal> to update the values).

c++/src/Statistics.hh
majetideepak: 0 is a valid value. I think it is safe if we initialize to an invalid value `-1`.

c++/src/Statistics.hh
majetideepak: This is according to the google coding style we agreed upon. I am going to slowly change the entire code base to follow the google coding style. The style recommends class variables to end with an `_`
https://google.github.io/styleguide/cppguide.html#Variable_Names

majetideepak:
You have to specialize the code for non-primitive types. There is no other option. I would say specialization of methods like `compare, update` inside `InternalStatisticsImpl`  is the right and simple way to go forward. This class should be responsible for handling the statistics updates.
ColumnWriters will just have an instance of the InternalStatisticsImpl and call the corresponding methods to `update` on writing a new value.

c++/src/Statistics.hh
wgtmac: That sounds good. We also need suggestion from @omalley about the coding style.

c++/src/Statistics.hh
wgtmac: If I add a function called void update(std::string str) for StringColumnStatistics to update string stats and it will see problem. For the first string, it needs to change totalLength_ to its length. For the following strings we use addition. This works but the code is not elegant. 

Similarly, if I add a function called void increase(uint64_t count), the same thing happens. I think making default value to 0 is more cleaner in these cases.

majetideepak: Will make this change. However, we have to initialize the first value using `setMinimum(), setMinimum()` for min, max fields. We could use the update methods from the second value being written.

c++/src/Statistics.hh
```diff
@@ -64,103 +64,63 @@ namespace orc {
...
+      valueCount_ = 0;
     }
 
     ~InternalStatisticsImpl() {}
 
     // GET / SET totalLength_
-    bool hasTotalLength() const {
-      return hasTotalLength_;
-    }
+    bool hasTotalLength() const {return hasTotalLength_;}
```
wgtmac: Spaces need to be used here to separate braces from implementation. [Horizontal_Whitespace](https://google.github.io/styleguide/cppguide.html#Horizontal_Whitespace)

c++/src/Statistics.hh
```diff
@@ -64,103 +64,63 @@ namespace orc {
...
-      hasNull_ = hasNull;
-    }
+    void setHasNull(bool hasNull) {hasNull_ = hasNull;}
   };
 
   typedef InternalStatisticsImpl<char> InternalCharStatistics;
-  typedef InternalStatisticsImpl<uint64_t> InternalIntegerStatistics;
-  typedef InternalStatisticsImpl<uint32_t> InternalDateStatistics;
-  typedef InternalStatisticsImpl<int64_t> InternalTimestampStatistics;
+  typedef InternalStatisticsImpl<uint64_t> InternalBooleanIntegerStatistics;
```
wgtmac: Maybe you can separate them as follows:
typedef InternalStatisticsImpl<uint64_t> InternalIntegerStatistics;
typedef InternalStatisticsImpl<uint64_t> InternalBooleanStatistics;
The mixed name is confusing.

c++/src/Statistics.hh
wgtmac: Yep. I will add our implementation after your PR is merged.

c++/src/Statistics.hh
majetideepak: This is a typo actually. I only meant `Boolean`.

c++/src/Statistics.hh
xndai: Not sure if this is right time to enforce a new coding style. We have an existing code base roughly about 10,000 lines. It's quite a bit of effort to convert existing code into proposed style. What's the problem with current style? Or ask it differently - what is the benefit of moving over to google style?

majetideepak: The problem with the current style is that there are no guidelines and is not consistent as well. I see some variables beginning with an `_` and some don't.
In other projects, I found the google style very helpful to read code. Especially when I have to revisit the code after some months. The style allows me to easily distinguish a class member vs a local member. Now that we plan to add the writer, the code base is only going to increase significantly. I felt this is the right time to introduce a standard moving forward. A standard helps people new to the code base too.

xndai: Personally I don't have any problem with the current style. I don't even notice the inconsistency (should have pay more attentions to style, :)). But to me, missing writer is a big functionality gap that needs to be filled in quickly.

Having said that I think this is eventually a call from @omalley and other major committers. If we really want to enforce a new style, we should always update the existing code first before adding new codes with new style. Otherwise the code base is going to look even worse until we fix everything.

omalley:
Yeah, we should probably adopt some coding guidelines. I'm not wild about Google's trailing underscore names.

majetideepak:
I will update this patch to remove the trailing underscores.

PR #119, opened 2017-05-11
ORC-177 OutputStream interface and implementation

xndai:
1. Create output stream interface and implementation classes
2. Add memory input output stream for the convenince of UT
3. Add corresponding unit test

Change-Id: I2495345453c3070fa43290edb2f7a012e117069b


xndai:
There are type conversion errors which are not caught by MacOS clang. I am investigating why.

xndai:
hi @omalley , can you please take a look at this? Thx.

PR #120, opened 2017-05-11
ORC-184: Refactor ColumnStatistics classes for writer

wgtmac:
1. Added several functions to ColumnStatistic class inclduing hasNull, merge, reset, toProtoBuf, etc.
2. Added functions to scale up/down for Int128 and comparision function for Decimal types.

### DIFF
--- a/c++/include/orc/Int128.hh
+++ b/c++/include/orc/Int128.hh
@@ -332,5 +332,23 @@ namespace orc {
     int64_t highbits;
     uint64_t lowbits;
   };
+
+  /**
+   * Scales up an Int128 value
+   * @param value the Int128 value to scale
+   * @param power the scale offset. Result of a negative factor is undefined.
+   * @param overflow returns whether the result overflows or not
+   * @return the scaled value
+   */
+  Int128 scaleUpInt128ByPowerOfTen(Int128 value,
+                                   int32_t power,
+                                   bool &overflow);
+  /**
+   * Scales down an Int128 value
+   * @param value the Int128 value to scale
+   * @param power the scale offset. Result of a negative factor is undefined.
+   * @return the scaled value
+   */
+  Int128 scaleDownInt128ByPowerOfTen(Int128 value, int32_t power);
 }
 #endif
--- a/c++/include/orc/Statistics.hh
+++ b/c++/include/orc/Statistics.hh
@@ -40,6 +40,35 @@ namespace orc {
     virtual uint64_t getNumberOfValues() const = 0;
 
     /**
+     * Check whether column has null value
+     * @return true if has null value
+     */
+    virtual bool hasNull() const = 0;
+
+    /**
+     * Increases count of values
+     * @param count number of values to be increased
+     */
+    virtual void increase(uint64_t count) = 0;
+
+    /**
+     * Merges another statistics
+     * @param other statistics to be merged
+     */
+    virtual void merge(const ColumnStatistics& other) = 0;
+
+    /**
+     * Convert statistics to protobuf version
+     * @param pbStats output of protobuf stats
+     */
+    virtual void toProtoBuf(proto::ColumnStatistics& pbStats) const = 0;
+
+    /**
+     * reset column statistics to initial state
+     */
+    virtual void reset() = 0;
+
+    /**
      * print out statistics of column if any
      */
     virtual std::string toString() const = 0;
@@ -59,6 +88,12 @@ namespace orc {
     virtual bool hasTotalLength() const = 0;
 
     virtual uint64_t getTotalLength() const = 0;
+
+    /**
+     * update stats by a new value
+     * @param length of the value
+     */
+    virtual void update(size_t length) = 0;
   };
 
   /**
@@ -76,6 +111,13 @@ namespace orc {
 
     virtual uint64_t getFalseCount() const = 0;
     virtual uint64_t getTrueCount() const = 0;
+
+    /**
+     * update stats by a new value
+     * @param value new value to update
+     * @param repetitions the repetitions of the boolean value
+     */
+    virtual void update(bool value, size_t repetitions) = 0;
   };
 
   /**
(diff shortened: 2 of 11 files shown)


wgtmac:
@omalley @majetideepak @xndai
As I am waiting for [PR of ORC-185](https://github.com/apache/orc/pull/116) being resolved, I haven't added all functions that we have implemented for column statistics to avoid huge effort for refactoring later. In this PR, I just followed the style of current code base which puts function declaration in the class ColumnStatistics and implements them in every ColumnStatisticsImpl class and it contains a lot of copying and pasting duplicate code (e.g. in c++/src/Statistics.hh). Actually in our internal implementation, we put the implementation of functions like hasNull(), getNumberOfValues(), increase(), merge(), reset(), etc. into the base class **ColumnStatistics**. What are your thoughts?

majetideepak:
I am strongly for rebasing this on top of ORC-185.
1) Exposing the functions you mentioned in the public API is bad design.
2) ORC-185 adds templates, which will further reduce code duplication.

wgtmac:
On Java side, ColumnStatistics and [Type]ColumnStatistics are very simple interfaces with minimum getter functions. It is the duty of ColumnStatisticsImpl class to define/implement most common interfaces and act as the base class in most writer implementations. Other [Type]ColumnStatisticsImpls just extend ColumnStatisticsImpl and implement [Type]ColumnStatistics to add their own logic to remove duplicate code. However, on C++ side, ColumnStatisticsImpl and [Type]ColumnStatisticsImpl have no inheritance relation like Java side. We cannot simply make [Type]ColumnStatisticsImpl inherit ColumnStatisticsImpl as it may introduce virtual inheritance issue because they have a common base class. Templates do not solve this issue as there is still no base class defining all common interfaces. In this case, we can expect that we need to use dynamic_cast almost everywhere if we want to use ColumnStatistcis abstraction; otherwise, we need to bind each [Type]ColumnStatistics to [Type]ColumnWriter leading to another place with huge amount of duplicate code.

majetideepak:
You are right that we cannot make  [Type]ColumnStatisticsImpl inherit ColumnStatisticsImpl. That is why I introduced `InternalStatisticsImpl` in ORC-185.
With that, you will just use `InternalStatisticsImpl<Type>` in your templated [Type]ColumnWriterImpl. 
I guess without the design for the ColumnWriter, it is hard to discuss at this point.

Can you open a separate PR for your ColumnWrite code? Things will get more clear then.
We can converge on the necessary overall changes.

wgtmac:
We are progressively working on moving all the code here. As there is a big gap between the current code and our code, it is not possible to create a PR for ColumnWriter right now. 

This PR is not a final decision, but a piece of code to show our ideas and introduce the discussion as we actually need some sort of ColumnStatistics class to act as the base class in the ColumnWriter implementation. To use your InternalStatisticsImpl<Type>, I have following thoughts:
1) InternalStatisticsImpl<Type> is an implementation class, so the interface functions like increase, merge, reset, etc. should still be defined in a base class somewhere;
2) We can implement ColumnWriter class using templates as well to adopt InternalStatisticsImpl<Type>; but ColumnWriter and ColumnStatistics are slightly different since ColumnWriter has more types and less common code, so I doubt template is not a good choice for ColumnWriter class.
3) BTW, I really appreciate your refactoring in your PR @majetideepak. I would like to add our code based on your changes. Before that, we have to achieve a consensus on which class is the best choice of the base class in ColumnWriter implementation (like the ColumnStatisticsImpl on the java side). In our design, it is ColumnStatistics itself.

majetideepak:
To answer your concerns,
1) The functions like increase, merge, reset, etc. can be added to InternalStatisticsImpl also. I don't see the need for a base class. The implementation for non-primitive types must be specialized anyway.
2) I don't see how the Types can be different between the ColumnWriter and the ColumnStatistics. Can you give an example? If your [Type]ColumnWriter design is similar to the existing [Type]ColumnReader design, then we can easily stick in an InternalStatisticsImpl<Type> instance.

I opened ORC-185 as a pattern we could use to overcome the lack of a base class with [Type]ColumnStatisticsImpl. Ultimately, we should definitely do what is easier to get the Writer code in with minimal code changes possible. However, without the ColumnWriter design, it is hard to enumerate and discuss other options for ColumnStatistics.
However, putting methods that should be internal, in the public API is definitely not the right approach.

wgtmac:
Yes, our current ColumnWriter design follows the approach of ColumnReader. One base class ColumnWriter takes care of all common work, and specific inherited ColumnWriters like IntegerColumnWriter, StructColumnWriter, BooleanColumnWriters, etc. manage their own logic. To integrate ColumnStatistics, we have two options (correct me if I'm wrong):
1. Put a based ColumnStatistics class in the base ColumnWriter like I said previously, and in each TypeColumnWriter, use dynamic type of ColumnStatistics to do its type-specific logic. In this way, we can save many duplicate code by polymorphism in the base ColumnWriter class; (our current code is written in this fashion)
2. There is no ColumnStatistic variable in the base class ColumnWriter. Instead, each inherited TypeColumnWriter class uses one TypeColumnStatisticsImpl (not TypeColumnStatistics), e.g. IntegerColumnWriter uses IntegerColumnStatisticsImpl. This is feasible except for a little bit more duplicate stuff compared with the 1st approach.

Your InternalColumnStatisticsImpl design is more inclined to the 2nd approach as it is still not a base class to use in the base class ColumnWriter. We can discuss ColumnWriter later as it is difficult without actual code. To save both of our time and effort and avoid significant change of our current ColumnWriter implementation, I will update this PR accordingly based on the approach 2 after your PR is merged. @majetideepak

majetideepak:
I see your concern now.
My idea is to directly use InternalIntegerStatisticsImpl from ORC-185 inside the IntegerColumnWriter. If we use the existing IntegerColumnStatisticsImpl, then we need to duplicate all the new methods being introduced. 
I think if I add a base class to the InternalIntegerStatisticsImpl in ORC-185, your problem should be solved.

wgtmac:
For not making it more complex, I think your current work in ORC-185 is good enough. I also agree that it is not right place to put them into ColumnStatistics. I will fix this in the next commit after ORC-185.

majetideepak:
Sounds good then. Thanks.

omalley:
I'm sorry, I closed the wrong pull request.

PR #121, opened 2017-05-16
Check the result schema after column selection

dongjoon-hyun:
For me, it seems to fail.
```
java.lang.AssertionError: 
Expected :2
Actual   :3
```
cc @omalley .


dongjoon-hyun:
Thank you! :)

omalley:
For external reference, the confusion was that when you use column pruning in ORC, the resulting schema still has the unselected columns, but they are always null. We should probably update the documentation to make that clear.

PR #122, opened 2017-05-16
ORC-192 Implement zlib compresion stream

xndai:
Implement zlib compressor based on the output stream. Add corresponding
UTs. Also rename the existing test suite from TestCompression into
TestDecompression.

Change-Id: I6c594cc8f8a769078bbe561056adfd5d0389dbc1

### DIFF
--- a/c++/src/Compression.cc
+++ b/c++/src/Compression.cc
@@ -898,7 +898,8 @@ DIAGNOSTIC_POP
         (new BufferedOutputStream(pool, outStream, bufferCapacity, blockSize));
     }
     case CompressionKind_ZLIB: {
-      int level = (strategy == CompressionStrategy_SPEED) ? -1 : 9;
+      int level = (strategy == CompressionStrategy_SPEED) ?
+              Z_BEST_SPEED + 1 : Z_DEFAULT_COMPRESSION;
       return std::unique_ptr<BufferedOutputStream>
         (new ZlibCompressionStream(
                 outStream, level, bufferCapacity, blockSize, pool));
--- a/c++/src/Compression.cc
+++ b/c++/src/Compression.cc
@@ -33,6 +33,254 @@
 
 namespace orc {
 
+  class CompressionStreamBase: public BufferedOutputStream {
+  public:
+    CompressionStreamBase(OutputStream * outStream,
+                          int compressionLevel,
+                          uint64_t capacity,
+                          uint64_t blockSize,
+                          MemoryPool& pool);
+
+    virtual bool Next(void** data, int*size) override = 0;
+    virtual void BackUp(int count) override;
+
+    virtual std::string getName() const override = 0;
+    virtual uint64_t flush() override;
+
+    virtual bool isCompressed() const override { return true; }
+    virtual uint64_t getSize() const override;
+
+  protected:
+    void writeHeader(char * buffer, size_t compressedSize, bool original) {
+      buffer[0] = static_cast<char>((compressedSize << 1) + (original ? 1 : 0));
+      buffer[1] = static_cast<char>(compressedSize >> 7);
+      buffer[2] = static_cast<char>(compressedSize >> 15);
+    }
+
+    // Buffer to hold uncompressed data until user calls Next()
+    DataBuffer<unsigned char> rawInputBuffer;
+
+    // Compress level
+    int level;
+
+    // Compressed data output buffer
+    char * outputBuffer;
+
+    // Size for compressionBuffer
+    int bufferSize;
+
+    // Compress output position
+    int outputPosition;
+
+    // Compress output buffer size
+    int outputSize;
+  };
+
+  CompressionStreamBase::CompressionStreamBase(OutputStream * outStream,
+                                               int compressionLevel,
+                                               uint64_t capacity,
+                                               uint64_t blockSize,
+                                               MemoryPool& pool) :
+                                                BufferedOutputStream(pool,
+                                                                     outStream,
+                                                                     capacity,
+                                                                     blockSize),
+                                                rawInputBuffer(pool, blockSize),
+                                                level(compressionLevel),
+                                                outputBuffer(nullptr),
+                                                bufferSize(0),
+                                                outputPosition(0),
+                                                outputSize(0) {
+    // PASS
+  }
+
+  void CompressionStreamBase::BackUp(int count) {
+    if (count > bufferSize) {
+      throw std::logic_error("Can't backup that much!");
+    }
+    bufferSize -= count;
+  }
+
+  uint64_t CompressionStreamBase::flush() {
+    void * data;
+    int size;
+    if (!Next(&data, &size)) {
+      throw std::logic_error("Failed to flush compression buffer.");
+    }
+    BufferedOutputStream::BackUp(outputSize - outputPosition);
+    bufferSize = outputSize = outputPosition = 0;
+    return BufferedOutputStream::flush();
+  }
+
+  uint64_t CompressionStreamBase::getSize() const {
+    return BufferedOutputStream::getSize() -
+           static_cast<uint64_t>(outputSize - outputPosition);
+  }
+
+  /**
+   * Streaming compression base class
+   */
+  class CompressionStream: public CompressionStreamBase {
+  public:
+    CompressionStream(OutputStream * outStream,
+                          int compressionLevel,
+                          uint64_t capacity,
+                          uint64_t blockSize,
+                          MemoryPool& pool);
+
+    virtual bool Next(void** data, int*size) override;
+    virtual std::string getName() const override = 0;
+
+  protected:
+    // return total compressed size
+    virtual uint64_t doStreamingCompression() = 0;
+  };
+
+  CompressionStream::CompressionStream(OutputStream * outStream,
+                                       int compressionLevel,
+                                       uint64_t capacity,
+                                       uint64_t blockSize,
+                                       MemoryPool& pool) :
+                                         CompressionStreamBase(outStream,
+                                                               compressionLevel,
+                                                               capacity,
+                                                               blockSize,
+                                                               pool) {
+    // PASS
+  }
+
+  bool CompressionStream::Next(void** data, int*size) {
+    if (bufferSize != 0) {
+      // adjust 3 bytes for the compression header
+      if (outputPosition + 3 >= outputSize) {
+        int newPosition = outputPosition + 3 - outputSize;
+        if (!BufferedOutputStream::Next(
+          reinterpret_cast<void **>(&outputBuffer),
+          &outputSize)) {
+          throw std::logic_error(
+            "Failed to get next output buffer from output stream.");
+        }
+        outputPosition = newPosition;
+      } else {
+        outputPosition += 3;
+      }
+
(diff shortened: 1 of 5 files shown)


xndai:
@omalley can you please take a look at this? Thx.

c++/src/Compression.cc
```diff
@@ -33,6 +33,254 @@
...
+                                            compressionLevel,
+                                            capacity,
+                                            blockSize,
+                                            pool) {
+    init();
+  }
+
+  uint64_t ZlibCompressionStream::doStreamingCompression() {
+    if (deflateReset(&strm) != Z_OK) {
+      throw std::logic_error("Failed to reset inflate.");
```
majetideepak: The decompression logic uses `ParseError` for failed library calls.
We should follow the same here. `ParseError(ZlibCompressionStream failed to reset deflate)`

c++/src/Compression.cc
```diff
@@ -33,6 +33,254 @@
...
+
+  bool CompressionStream::Next(void** data, int*size) {
+    if (bufferSize != 0) {
+      // adjust 3 bytes for the compression header
+      if (outputPosition + 3 >= outputSize) {
+        int newPosition = outputPosition + 3 - outputSize;
+        if (!BufferedOutputStream::Next(
+          reinterpret_cast<void **>(&outputBuffer),
+          &outputSize)) {
+          throw std::logic_error(
```
majetideepak: Use `ParseError` here

c++/src/Compression.cc
```diff
@@ -33,6 +33,254 @@
...
+      throw std::logic_error("Can't backup that much!");
+    }
+    bufferSize -= count;
+  }
+
+  uint64_t CompressionStreamBase::flush() {
+    void * data;
+    int size;
+    if (!Next(&data, &size)) {
+      throw std::logic_error("Failed to flush compression buffer.");
```
majetideepak: use `ParseError` here.

c++/src/Compression.cc
```diff
@@ -33,6 +33,254 @@
...
+
+    strm.avail_in = static_cast<unsigned int>(bufferSize);
+    strm.next_in = rawInputBuffer.data();
+
+    do {
+      if (outputPosition >= outputSize) {
+        if (!BufferedOutputStream::Next(
+          reinterpret_cast<void **>(&outputBuffer),
+          &outputSize)) {
+          throw std::logic_error(
```
majetideepak: use `ParseError` here.

c++/src/Compression.cc
```diff
@@ -33,6 +33,254 @@
...
+
+      int ret = deflate(&strm, Z_FINISH);
+      outputPosition = outputSize - static_cast<int>(strm.avail_out);
+
+      if (ret == Z_STREAM_END) {
+        break;
+      } else if (ret == Z_OK) {
+        // needs more buffer so will continue the loop
+      } else {
+        throw std::logic_error("Failed to deflate input data.");
```
majetideepak: use `ParseError` here.

c++/src/Compression.cc
```diff
@@ -33,6 +33,254 @@
...
+DIAGNOSTIC_IGNORE("-Wold-style-cast")
+
+  void ZlibCompressionStream::init() {
+    strm.zalloc = Z_NULL;
+    strm.zfree = Z_NULL;
+    strm.opaque = Z_NULL;
+
+    if (deflateInit2(&strm, level, Z_DEFLATED, -15, 8, Z_DEFAULT_STRATEGY)
+        != Z_OK) {
+      throw std::logic_error("Error while calling deflateInit2() for zlib.");
```
majetideepak: use `ParseError` here.

c++/src/Compression.cc
```diff
@@ -636,6 +884,33 @@ DIAGNOSTIC_POP
...
+                      uint64_t bufferCapacity,
+                      uint64_t blockSize,
+                      MemoryPool& pool) {
+    switch (static_cast<int64_t>(kind)) {
+    case CompressionKind_NONE: {
+      return std::unique_ptr<BufferedOutputStream>
+        (new BufferedOutputStream(pool, outStream, bufferCapacity, blockSize));
+    }
+    case CompressionKind_ZLIB: {
+      int level = strategy == CompressionStrategy_SPEED ? 1 : 9;
```
majetideepak: Adding braces here improves readbility.
`(strategy == CompressionStrategy_SPEED)`

majetideepak: We should keep the default level to `Z_DEFAULT_COMPRESSION` (-1). Only if `strategy` is specified, we must set accordingly. I think the Java code does something similar.

c++/src/Compression.cc
xndai: Throwing ParseError is understandable for reader, but would be weird for writer. If we really want to distinguish the exceptions from compression, we might come up with a different exception type. But so far I don't see a strong need here.

c++/src/Compression.cc
xndai: Will do. Thx.

c++/src/Compression.cc
majetideepak: I think logic_error is also not apt for runtime failures. ParseError has been used throughout the read path.
Moving forward, you will have to create an Exception class for the write path. How about we create a `WriteError` exception class similar to `ParseError`?

majetideepak: Alternatively, a better option is to rename the `ParseError` to `ORCError`, and use it throughout the code base.

xndai: how about just using std::runtime_error? I still don't see a need to create yet another exception class.

majetideepak: ParseError is extended from std::runtime_error.
Applications that integrate with the ORC library would want to distinguish exceptions thrown by the ORC library versus thrown by other libraries.

xndai: We also throw std::logic_error in other places. If that's the goal, I'd suggest we use one single type of exception for all Orc code, e.g. OrcException. You can open a JIRA on this, and it can be done in a separate change.

majetideepak: Sounds good!

xndai: Or we keep RuntimeError and LogicError both defined in orc namespace to differentiate different types of errors.

majetideepak: Filed ORC-196

xndai: For now, I will just use std::runtime_error in compression path. We can come back and fix that later.

majetideepak: yes! use std::runtime_error for now.

xndai:
@majetideepak @omalley please take another look.

c++/src/Compression.hh
```diff
@@ -20,9 +20,15 @@
 #define ORC_COMPRESSION_HH
 
 #include "io/InputStream.hh"
+#include "io/OutputStream.hh"
 
 namespace orc {
 
+  enum CompressionStrategy {
+    CompressionStrategy_SPEED = 0,
```
majetideepak: Can you add a `CompressionStrategy_DEFAULT` here?

c++/src/Compression.cc
```diff
@@ -636,6 +884,33 @@ DIAGNOSTIC_POP
...
+                      uint64_t bufferCapacity,
+                      uint64_t blockSize,
+                      MemoryPool& pool) {
+    switch (static_cast<int64_t>(kind)) {
+    case CompressionKind_NONE: {
+      return std::unique_ptr<BufferedOutputStream>
+        (new BufferedOutputStream(pool, outStream, bufferCapacity, blockSize));
+    }
+    case CompressionKind_ZLIB: {
+      int level = (strategy == CompressionStrategy_SPEED) ? -1 : 9;
```
majetideepak: Compare the three options, `CompressionStrategy_DEFAULT`, `CompressionStrategy_SPEED`, `CompressionStrategy_COMPRESSION` and set the level accordingly.

xndai: what are the corresponding levels for these three settings? Are they documented anywhere?

majetideepak: You can see the compression levels here
http://www.zlib.net/manual.html
```
#define Z_NO_COMPRESSION         0
#define Z_BEST_SPEED             1
#define Z_BEST_COMPRESSION       9
#define Z_DEFAULT_COMPRESSION  (-1)
```

xndai: According to this - https://orc.apache.org/docs/hive-config.html, there are only two compression strategy defined: SPEED and COMPRESSION. I also checked Java implementation, SPEED maps to zlib level Z_BEST_SPEED + 1, and COMPRESSION maps to Z_DEFAULT_COMPRESSION. I will do the same for C++.

Java implementation for your reference -

WriterImpl.java

`
    CompressionCodec result = physicalWriter.getCompressionCodec();
    if (result != null) {
      switch (kind) {
        case BLOOM_FILTER:
        case DATA:
        case DICTIONARY_DATA:
        case BLOOM_FILTER_UTF8:
          if (compressionStrategy == OrcFile.CompressionStrategy.SPEED) {
            result = result.modify(EnumSet.of(CompressionCodec.Modifier.FAST,
                CompressionCodec.Modifier.TEXT));
          } else {
            result = result.modify(EnumSet.of(CompressionCodec.Modifier.DEFAULT,
                CompressionCodec.Modifier.TEXT));
          }
          break;
        case LENGTH:
        case DICTIONARY_COUNT:
        case PRESENT:
        case ROW_INDEX:
        case SECONDARY:
          // easily compressed using the fastest modes
          result = result.modify(EnumSet.of(CompressionCodec.Modifier.FASTEST,
              CompressionCodec.Modifier.BINARY));
          break;
        default:
          LOG.info("Missing ORC compression modifiers for " + kind);
          break;
      }
    }

`

ZlibCodec.java

`
public CompressionCodec modify(/* @Nullable */ EnumSet<Modifier> modifiers) {

    if (modifiers == null) {
      return this;
    }

    int l = this.level;
    int s = this.strategy;

    for (Modifier m : modifiers) {
      switch (m) {
      case BINARY:
        /* filtered == less LZ77, more huffman */
        s = Deflater.FILTERED;
        break;
      case TEXT:
        s = Deflater.DEFAULT_STRATEGY;
        break;
      case FASTEST:
        // deflate_fast looking for 8 byte patterns
        l = Deflater.BEST_SPEED;
        break;
      case FAST:
        // deflate_fast looking for 16 byte patterns
        l = Deflater.BEST_SPEED + 1;
        break;
      case DEFAULT:
        // deflate_slow looking for 128 byte patterns
        l = Deflater.DEFAULT_COMPRESSION;
        break;
      default:
        break;
      }
    }
    return new ZlibCodec(l, s);
  }
`

majetideepak: You are right!

majetideepak:
+1 LGTM

xndai:
Hi @majetideepak , what happens next after you approved? are you able to accept this pull request?

majetideepak:
@xndai I should get rights to merge by next week. @omalley  will merge them until then.

prasanthj:
I just merged this commit. Thanks @xndai for the contribution and @majetideepak  for the reviews!

PR #123, opened 2017-05-16
ORC-184: Refactor ColumnStatistics classes for writer

wgtmac:
1. Added setters for ColumnStatisticsImpl;
2. Added merge, update, increase, reset, etc. to ColumnStatisticsImpl;
3. Supported comparision for Decimal types;
4. Added UTs for Decimal and ColumnStatistics.

### DIFF
--- a/c++/include/orc/Statistics.hh
+++ b/c++/include/orc/Statistics.hh
@@ -391,8 +391,6 @@ namespace orc {
      */
     virtual uint32_t getNumberOfRowIndexStats(uint32_t columnId) const = 0;
   };
-
-
 }
 
 #endif
--- a/c++/src/Statistics.hh
+++ b/c++/src/Statistics.hh
@@ -19,6 +19,7 @@
 #ifndef ORC_STATISTICS_IMPL_HH
 #define ORC_STATISTICS_IMPL_HH
 
+#include "orc/Common.hh"
 #include "orc/Int128.hh"
 #include "orc/OrcFile.hh"
 #include "orc/Reader.hh"
@@ -70,7 +71,7 @@ namespace orc {
 
     ~InternalStatisticsImpl() {}
 
-    // GET / SET totalLength_
+    // GET / SET _totalLength
     bool hasTotalLength() const { return _hasTotalLength; }
 
     void setHasTotalLength(bool hasTotalLength) { _hasTotalLength = hasTotalLength; }
@@ -79,7 +80,7 @@ namespace orc {
 
     void setTotalLength(uint64_t totalLength) { _totalLength = totalLength; }
 
-    // GET / SET sum_
+    // GET / SET _sum
     bool hasSum() const { return _hasSum; }
 
     void setHasSum(bool hasSum) { _hasSum = hasSum; }
@@ -88,7 +89,7 @@ namespace orc {
 
     void setSum(T sum) { _sum = sum; }
 
-    // GET / SET maximum_
+    // GET / SET _maximum
     bool hasMaximum() const { return _hasMaximum; }
 
     T getMaximum() const { return _maximum; }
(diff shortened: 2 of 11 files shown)


wgtmac:
@omalley I have created a new PR here for ORC-184 based on the current master branch. Please review it, thanks!

wgtmac:
@majetideepak how does this PR look to you?

majetideepak:
@wgtmac I will take a look at it in some time tonight. Thanks.

c++/include/orc/Statistics.hh
```diff
@@ -392,7 +392,14 @@ namespace orc {
...
   };
 
-
+  /**
+   * Create ColumnStatistics for writers
+   * @param type of column
+   * @param enableStringComparison whether enable string columns comparision
+   * @return ColumnStatistics instances
+   */
+  std::unique_ptr<ColumnStatistics> createColumnStatistics(
```
wgtmac: Make sense. I will move it to src/Statistics.hh

c++/include/orc/Vector.hh
```diff
@@ -194,6 +194,14 @@ namespace orc {
     int32_t scale;
   };
 
+  /**
+   * Compares two decimals
+   * @param decimal1 1st decimal to compare
+   * @param decimal2 2nd decimal to compare
+   * @return true is decimal1 is less than decimal2; otherwise false
+   */
+  bool operator<(Decimal decimal1, Decimal decimal2);
```
wgtmac: Decimal and Int128 are in the public APIs as well.

majetideepak: Libraries usually don't define these operators in the public API.
It is also odd that we only define one operator and not the other operators.
All other non-primitive types don't have these operators defined as well. It looks odd just to add this one. Can we just use `compare(Decimal, Decimal)` api inside?

c++/src/Statistics.hh
```diff
@@ -106,15 +106,50 @@ namespace orc {
...
+      _valueCount += other._valueCount;
+
+      if (other._hasMinimum) {
+        if (!_hasMinimum) {
+          _hasMinimum = _hasMaximum = true;
+          _minimum = other._minimum;
+          _maximum = other._maximum;
+        } else {
+          // all template types should support operator<
+          if (_maximum < other._maximum) {
```
wgtmac: @majetideepak I agree with you for not putting opeator< into public API. The season to implement operator< is to simplify the logic here, otherwise we have to specialize InternalStatisticsImpl<Decimal>::merge(const InternalStatisticsImpl& other). Now there are two options:
1. add Decimal::compare() and use the template specilization for Decimal;
2. move opeator< of decimal here to simplify the merge logic.
Which do you prefer or you have other ideas? Thanks!

majetideepak: The first option of using templates seems better to me.
Can use `common.hh` and `common.cc` to put the following.
```
// Return True if val1 < val2
// Return False otherwise
template <typename T>
static inline bool Compare(const T& val1, const T& val2) {
    return val1 < val2
}
// Specialize for Decimal
template <>
static inline bool Compare(const Decimal& val1, const Decimal& val2) {
   ....
}
```

wgtmac: If we declare the template function in the Common.hh and define it in Common.cc. Then we have to declare it for every type as the template function is instantiated in Common.o while used in Statistics.o.

wgtmac: I've fixed it and created a new PR.  I have removed "static" identifier as this template function is used in several modules including UTs. Please review it again. Thanks!

majetideepak:
@wgtmac  looks like you have to resolve a merge conflict.

wgtmac:
@majetideepak Resolved. Can you retry? Thanks!

majetideepak:
@prasanthj  can you make a pass and merge this too? Thanks.

prasanthj:
I am seeing some merged conflicts with this PR. @wgtmac could you please rebase the patch after the recent commits?

wgtmac:
@prasanthj  I've rebased it just now. Please retry. Thanks!

prasanthj:
Merged with master. Thanks @wgtmac for the rebase and the contribution! Thanks @majetideepak for the reviews!

PR #124, opened 2017-05-17
ORC-193. Refactor TreeWriter API in WriterImpl.

omalley:
For this patch, I decided to just fix the API to TreeWriter. I'll make a follow up patch that just pulls the TreeWriter classes out of WriterImpl.


prasanthj:
lgtm, +1

PR #125, opened 2017-05-17
ORC-194 Move TreeWriters classes out of WriterImpl

omalley:


omalley:
I just committed this with ded204a4a10bfad1ed739fc98f612a41005640c5v .

PR #126, opened 2017-05-18
ORC-191 Implement RLE v1 encoder

xndai:
Implement integer, byte and boolean RLE encoder, and add corresponding
UTs. Only v1 is implemented at this point.

Change-Id: Ia03b0d8bf98a843e45da8a5d349efeff9fb4fa00

### DIFF
--- a/c++/src/ByteRLE.cc
+++ b/c++/src/ByteRLE.cc
@@ -27,6 +27,268 @@
 namespace orc {
 
   const size_t MINIMUM_REPEAT = 3;
+  const size_t MAXIMUM_REPEAT = 127 + MINIMUM_REPEAT;
+  const size_t MAX_LITERAL_SIZE = 128;
+
+  ByteRleEncoder::~ByteRleEncoder() {
+    // PASS
+  }
+
+  class ByteRleEncoderImpl : public ByteRleEncoder {
+  public:
+    ByteRleEncoderImpl(std::unique_ptr<BufferedOutputStream> output);
+    virtual ~ByteRleEncoderImpl();
+
+    /**
+     * Encode the next batch of values
+     * @param data to be encoded
+     * @param numValues the number of values to be encoded
+     * @param notNull If the pointer is null, all values are read. If the
+     *    pointer is not null, positions that are false are skipped.
+     */
+    virtual void add(const char* data, uint64_t numValues,
+                      const char* notNull) override;
+
+    /**
+     * Get size of buffer used so far.
+     */
+    virtual uint64_t getBufferSize() const override;
+
+    /**
+     * Flushing underlying BufferedOutputStream
+    */
+    virtual uint64_t flush() override;
+
+    virtual void recordPosition(PositionRecorder* recorder) const override;
+
+  protected:
+    std::unique_ptr<BufferedOutputStream> outputStream;
+    char * literals;
+    int numLiterals;
+    bool repeat;
+    int tailRunLength;
+    int bufferPosition;
+    int bufferLength;
+    char * buffer;
+
+    void writeByte(char c);
+    void writeValues();
+    void write(char c);
+  };
+
+  ByteRleEncoderImpl::ByteRleEncoderImpl(
+                                std::unique_ptr<BufferedOutputStream> output)
+                                  : outputStream(std::move(output)) {
+    literals = new char[MAX_LITERAL_SIZE];
+    numLiterals = 0;
+    tailRunLength = 0;
+    repeat = false;
+    bufferPosition = 0;
+    bufferLength = 0;
+    buffer = nullptr;
+  }
+
+  ByteRleEncoderImpl::~ByteRleEncoderImpl() {
+    // PASS
+  }
+
+  void ByteRleEncoderImpl::writeByte(char c) {
+    if (bufferPosition == bufferLength) {
+      int addedSize = 0;
+      if (!outputStream->Next(reinterpret_cast<void **>(&buffer), &addedSize)) {
+        throw std::bad_alloc();
+      }
+      bufferPosition = 0;
+      bufferLength = addedSize;
+    }
+    buffer[bufferPosition++] = c;
+  }
+
+  void ByteRleEncoderImpl::add(
+                               const char* data,
+                               uint64_t numValues,
+                               const char* notNull) {
+    for (uint64_t i = 0; i < numValues; ++i) {
+      if (!notNull || notNull[i]) {
+        write(data[i]);
+      }
+    }
+  }
+
+  void ByteRleEncoderImpl::writeValues() {
+    if (numLiterals != 0) {
+      if (repeat) {
+        writeByte(static_cast<char>(numLiterals - static_cast<int>(MINIMUM_REPEAT)));
+        writeByte(literals[0]);
+     } else {
+        writeByte(static_cast<char>(-numLiterals));
+        for (int i = 0; i < numLiterals; ++i) {
+          writeByte(literals[i]);
+        }
+      }
+      repeat = false;
+      tailRunLength = 0;
+      numLiterals = 0;
+    }
+  }
+
+  uint64_t ByteRleEncoderImpl::flush() {
+    writeValues();
+    outputStream->BackUp(bufferLength - bufferPosition);
+    uint64_t dataSize = outputStream->flush();
+    bufferLength = bufferPosition = 0;
+    return dataSize;
+  }
+
+  void ByteRleEncoderImpl::write(char value) {
+    if (numLiterals == 0) {
+      literals[numLiterals++] = value;
+      tailRunLength = 1;
+    } else if (repeat) {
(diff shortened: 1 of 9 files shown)


xndai:
@omalley @majetideepak Can you also take a look at this? Thanks.

majetideepak:
@xndai I will take a look at this sometime tomorrow. Sorry for the delay.

c++/src/ByteRLE.cc
```diff
@@ -27,6 +27,272 @@
...
+
+  ByteRleEncoderImpl::~ByteRleEncoderImpl() {
+    // PASS
+  }
+
+  void ByteRleEncoderImpl::writeByte(char c) {
+    if (bufferPosition == bufferLength) {
+      int addedSize = 0;
+      if (!outputStream->Next(reinterpret_cast<void **>(&buffer), &addedSize)) {
+        throw std::bad_alloc();
```
majetideepak: When does `BufferedOutputStream::Next` return `false`?

c++/src/RLEv1.cc
```diff
@@ -26,8 +26,173 @@
...
+const int MAX_DELTA = 127;
+const int MIN_DELTA = -128;
+const int MAX_LITERAL_SIZE = 128;
+
+RleEncoderV1::RleEncoderV1(
+                          std::unique_ptr<BufferedOutputStream> outStream,
+                          bool hasSigned):
+                          outputStream(std::move(outStream)) {
+  isSigned = hasSigned;
+  literals = new int64_t[MAX_LITERAL_SIZE];
```
majetideepak: use `std::vector<int64_t>` literals_buf as a buffer instead.

c++/src/RLEv1.hh
```diff
@@ -26,6 +26,59 @@
...
+    std::unique_ptr<BufferedOutputStream> outputStream;
+    bool isSigned;
+    int64_t* literals;
+    int numLiterals;
+    int64_t delta;
+    bool repeat;
+    int tailRunLength;
+    int bufferPosition;
+    int bufferLength;
+    char * buffer;
```
majetideepak: `char*` (remove space)

c++/test/TestByteRLEEncoder.cc
```diff
@@ -0,0 +1,231 @@
...
+                       uint64_t numValues,
+                       char* notNull) {
+
+    std::unique_ptr<SeekableInputStream> inStream(
+      new SeekableArrayInputStream(memStream.getData(), memStream.getLength()));
+
+    std::unique_ptr<ByteRleDecoder> decoder =
+      createByteRleDecoder(std::move(inStream));
+
+    char* decodedData = new char[numValues];
```
majetideepak: use `std::vector<char> data_buf` as a buffer.

c++/test/TestByteRLEEncoder.cc
```diff
@@ -0,0 +1,231 @@
...
+        }
+      }
+    }
+  }
+
+  char * generateData(uint64_t numValues,
+                      uint64_t numNulls = 0,
+                      char ** notNull = nullptr) {
+    generateNotNull(numValues, numNulls, notNull);
+    char * res = new char[numValues];
```
majetideepak: pass `res` as an argument. Let's allocate and deallocate at the same place.

c++/test/TestByteRLEEncoder.cc
```diff
@@ -0,0 +1,231 @@
...
+
+namespace orc {
+
+  const int DEFAULT_MEM_STREAM_SIZE = 1024 * 1024; // 1M
+
+  void generateNotNull(uint64_t numValues,
+                       uint64_t numNulls,
+                       char ** notNull) {
+    if (numNulls != 0 && notNull != nullptr) {
+      *notNull = new char[numValues];
```
majetideepak: pass `notNull` as an argument. Let's allocate and deallocate at the same place.

c++/test/TestByteRLEEncoder.cc
```diff
@@ -0,0 +1,231 @@
...
+        res[i] = static_cast<char>(std::rand() % 256);
+    }
+    return res;
+  }
+
+  char * generateBoolData(uint64_t numValues,
+                          uint64_t numNulls = 0,
+                          char ** notNull = nullptr) {
+    generateNotNull(numValues, numNulls, notNull);
+    char * res = new char[numValues];
```
majetideepak: pass `res` as an argument.

c++/test/TestByteRLEEncoder.cc
```diff
@@ -0,0 +1,231 @@
...
+    uint64_t block = 1024;
+    BufferedOutputStream bufStream(*pool, &memStream, capacity, block);
+
+    std::unique_ptr<BufferedOutputStream> outStream(
+        new BufferedOutputStream(*pool, &memStream, capacity, block));
+
+    std::unique_ptr<ByteRleEncoder> encoder =
+      createByteRleEncoder(std::move(outStream));
+
+    char* data = generateData(102400);
```
majetideepak: pass a `std::vector<char>` buffer.

c++/test/TestByteRLEEncoder.cc
```diff
@@ -0,0 +1,231 @@
...
+    BufferedOutputStream bufStream(*pool, &memStream, capacity, block);
+
+    std::unique_ptr<BufferedOutputStream> outStream(
+        new BufferedOutputStream(*pool, &memStream, capacity, block));
+
+    std::unique_ptr<ByteRleEncoder> encoder =
+      createByteRleEncoder(std::move(outStream));
+
+    char* notNull = nullptr;
+    char* data = generateData(102400, 377, &notNull);
```
majetideepak: allocate via `std::vector<char>`

c++/src/ByteRLE.cc
```diff
@@ -27,6 +27,272 @@
...
+
+    void writeByte(char c);
+    void writeValues();
+    void write(char c);
+  };
+
+  ByteRleEncoderImpl::ByteRleEncoderImpl(
+                                std::unique_ptr<BufferedOutputStream> output)
+                                  : outputStream(std::move(output)) {
+    literals = new char[MAX_LITERAL_SIZE];
```
xndai: I use unique_ptr here to indicate ByteRleEncoderImpl has full ownership of this buffer. Thanks for catching that.

c++/src/ByteRLE.cc
xndai: Next() is a method we inherit from protobuf ZeroCopyOutputStream. Current implementation of BufferedOutputStream doesn't return false, but we still want this check to comply with the function signature.

c++/src/ByteRLE.cc
```diff
@@ -27,6 +27,272 @@
...
+    /**
+     * Flushing underlying BufferedOutputStream
+    */
+    virtual uint64_t flush() override;
+
+    virtual void recordPosition(PositionRecorder* recorder) const override;
+
+  protected:
+    std::unique_ptr<BufferedOutputStream> outputStream;
+    char * literals;
```
xndai: done.

c++/src/ByteRLE.cc
```diff
@@ -27,6 +27,272 @@
...
+    int numLiterals;
+    bool repeat;
+    int tailRunLength;
+    int bufferPosition;
+    int bufferLength;
+    char * buffer;
```
xndai: done.

c++/src/ByteRLE.cc
xndai: thanks for catching that. I add a free() call. Regarding to RAII, I think it's a bit overkilled here. The array is guaranteed to be allocated in constructor and freed in destructor. The logic is quite clear.

c++/src/RLEv1.cc
xndai: Same as above.

c++/src/RLEv1.hh
xndai: done

c++/test/TestByteRLEEncoder.cc
xndai: I have fixed these as well as the same issues in TestRLEv1Encoder.cc. See my next iteration.

c++/test/TestByteRLEEncoder.cc
xndai: I don't see a huge difference here - we are manipulating raw buffer anyway. I would just leave them there unless you feel strongly about it.

majetideepak: Since this is only in the tests, let's leave it as is for now. Thanks.

xndai:
Thanks @majetideepak  for reviewing. @prasanthj @omalley please help review and/or accept the PR. Thx!

prasanthj:
Merged with master. Thanks @xndai  for the patch and @majetideepak  for the reviews!

PR #128, opened 2017-05-24
ORC-178 Implement Basic C++ Writer and Writer Option

xndai:
1. Add basic Writer and WriterOption
2. Add StructColumnWriter and IntegerColumnWriter. With them, we will be
able to write a complete ORC file that contains only int columns. To
limit the scope of this change, we will add more column writers later.
3. Add a base class for column statistics impl classes. This
is to be used by the base class of ColumnWriter so we don't have to
duplicate a bunch of logics everywhere.
4. Right now the UTs are pretty primative. We will add more UTs
(especially for stats and index) as we are adding more column writers.
At this moment, it's really hard to extract more UTs from our code base
without intorducing additional column writers.

Change-Id: I694dda600136e5d285e70a8124aa1cc334f4ef14


majetideepak:
@xndai  I will complete my review by Tuesday.

xndai:
@majetideepak have you got a chance to take a look? Sorry for the big diff.

majetideepak:
@xndai I am half way through. This is an important patch and I just want to make sure we get it right. Sorry for the delay!

xndai:
@majetideepak totally understand. take your time!

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,294 @@
...
+     */
+    virtual ORC_UNIQUE_PTR<ColumnVectorBatch> createRowBatch(uint64_t size
+                                                             ) const = 0;
+
+    /**
+     * Add a row batch into current writer.
+     * @param rowsToAdd the row batch data to write.
+     */
+    virtual void add(ColumnVectorBatch& rowsToAdd) = 0;
+
```
majetideepak: Will be helpful to have `virtual bool canAdd(ColumnVectorBatch& rowsToAdd) = 0;` return true if there is enough buffer available to add the ColumnVectorBatch, false otherwise.

c++/src/ColumnWriter.cc
```diff
@@ -0,0 +1,507 @@
...
+
+  void StructColumnWriter::resetIndex() {
+    ColumnWriter::resetIndex();
+
+    for (uint32_t i = 0; i < children.size(); ++i) {
+      children[i]->resetIndex();
+    }
+  }
+
+  class IntegerColumnWriter : public ColumnWriter {
```
majetideepak: are you planning to template this class for other types?

c++/include/orc/OrcFile.hh
```diff
@@ -119,6 +120,17 @@ namespace orc {
...
+
+  /**
+   * Create a writer to write the ORC file.
+   * @param type the type of data to be written
+   * @param stream the stream to write to
+   * @param options the options for writing the file
+   */
+  ORC_UNIQUE_PTR<Writer> createWriter(
+                                      const Type& type,
+                                      OutputStream * stream,
```
xndai: I will fix all occurrences of the space issue.

c++/src/ColumnWriter.cc
```diff
@@ -0,0 +1,507 @@
...
+
+    virtual void createRowIndexEntry() override;
+
+    virtual void writeIndex(
+      std::vector<proto::Stream> &streams) const override;
+
+    virtual void resetIndex() override;
+
+  private:
+    std::vector<ColumnWriter *> children;
```
xndai: std::unique_ptr can be redefined as auto_ptr when unique_ptr is not supported by the current version of STL. See Adaptor.hh.in -

`#ifndef ORC_CXX_HAS_UNIQUE_PTR
  #define unique_ptr auto_ptr
#endif
`
And auto_ptr is not supposed to be used in STL container.

c++/src/ColumnWriter.cc
xndai: Not necessarily. There are just no much common logics between different types. We can re-examine this in next PR which includes other type writers.

c++/src/Writer.cc
```diff
@@ -0,0 +1,659 @@
...
+        }
+      }
+    } else {
+      stripeRows += rowsToAdd.numElements;
+      columnWriter->add(rowsToAdd, 0, rowsToAdd.numElements);
+    }
+
+    if (columnWriter->getEstimatedSize() >= options.getStripeSize()) {
+      WriteStripe();
+    }
```
xndai: We won't throw exception here. If estimated size is smaller than stripe size, it just means the current data added are not big enough to form a stripe. We continue keeping data in memory until a subsequent add() that satisfies this check, or a call to Writer::close() which forces us to flush any remaining data as the last stripe of file.

c++/include/orc/Writer.hh
xndai: We won't need canAdd() method. We always keep a fixed size buffer (determined by stripe size) and flush data to file as a stripe when the buffer is full.

majetideepak: Thanks! I misunderstood the API.

c++/src/Writer.cc
```diff
@@ -0,0 +1,659 @@
...
+  }
+
+  class WriterImpl : public Writer {
+  private:
+    std::unique_ptr<ColumnWriter> columnWriter;
+    std::unique_ptr<BufferedOutputStream> compressionStream;
+    std::unique_ptr<BufferedOutputStream> bufferedStream;
+    std::unique_ptr<StreamsFactory> streamsFactory;
+    OutputStream* outStream;
+    WriterOptions options;
```
xndai: We better keep a copy of WriterOptions within Writer class. Image below use case -

{
    WriterOption option;
    // set writer option
    // ...
    return createWriter(.. , option);
}

The writer object will be holding a destructed WriterOptions upon function returns.

majetideepak: The current implementation is inconsistent with the `Reader()` API.  https://github.com/apache/orc/blob/master/c%2B%2B/src/Reader.hh#L181
I would prefer both to be similar.
Also, why not move the class definitions to a header `writer.hh`?

xndai: I would prefer updating Reader instead to *not* use a const reference as member variable.  It's never been a good idea to do so. unique_ptr or shared_ptr should be used to enforce the ownership/lifecycle of an object. But in this case, a simple copy would be just fine.

And the reason we don't define WriterImpl in header file is because it's an implementation that needn't to be exposed.

xndai: I fixed ReaderOptions in the last commit. Please take a look.

xndai:
hi, @omalley do you want to take a look as well? Thanks!

xndai:
hi @majetideepak @omalley , if you don't have any more feedback, can you please help accept this PR? Thanks!

majetideepak:
@xndai Unfortunately, I don't have access to merge yet.

xndai:
@majetideepak No problem. I just talked to @omalley and looks like he's going to review this change soon.

xndai:
I add the writer id to file footer. Please take a look when you have a chance. Thx. @omalley

omalley:
Sorry for not getting to this sooner. I'm looking at it now.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * Get whether or not to have block padding.
+     * @return if not set, return default value which is false.
+     */
+    bool getBlockPadding() const;
+
+    /**
+     * Set Run length encoding version
+     */
+    WriterOptions& setRleVersion(RleVersion version);
```
omalley: Should be FileVersion.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,294 @@
...
+
+    /**
+     * Get the dictionary key size threshold.
+     */
+    double getDictionaryKeySizeThreshold() const;
+
+    /**
+     * Set whether or not to have block padding.
+     */
+    WriterOptions& setBlockPadding(bool padding);
```
omalley: Let's just use setPaddingTolerance(0) to turn off block padding.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * Get the compression strategy.
+     * @return if not set, return default value which is speed.
+     */
+    CompressionStrategy getCompressionStrategy() const;
+
+    /**
+     * Set the writer version.
+     */
+    WriterOptions& setWriterVersion(WriterVersion version);
```
omalley: The user doesn't get to choose the WriterVersion. It is based on the version of the software that writes the file.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,294 @@
...
+     * Get whether or not to write statistics (file statistics,
+     * stripe statistics, etc.)
+     * @return if not set, the default is true
+     */
+     bool getEnableStats() const;
+
+    /**
+     * Set whether or not to update string min/max stats
+     */
+    WriterOptions& setEnableStrStatsCmp(bool enable);
```
omalley: I don't understand this. If they get statistics, they will get the min/max.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * Get whether or not to update string min/max stats
+     * @return if not set, the default is false
+     */
+    bool getEnableStrStatsCmp() const;
+
+    /**
+     * Set whether or not to write row group index
+     */
+    WriterOptions& setEnableIndex(bool enable);
```
omalley: We don't want lots of variants for users to write the files with. Most users take the default and the others can either choose to completely disable them or increase the index stride. Both of which save space, but reduce the effectiveness of predicate push down.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,294 @@
...
+     * Get writer timezone
+     * @return writer timezone
+     */
+    const Timezone* getTimezone() const;
+
+    /**
+     * Set writer timezone
+     * @param zone writer timezone name
+     */
+    WriterOptions& setTimezone(const std::string& zone);
```
omalley: Let's not give this option. In the long term, it shouldn't matter and I'd rather not add it to the public api.

c++/src/ColumnWriter.cc
omalley: I think we've dropped the last OS version that failed ORC_CXX_HAS_UNIQUE_PTR, so we should undo the test and remove that support.

c++/src/ColumnWriter.cc
```diff
@@ -0,0 +1,507 @@
...
+  }
+
+  void IntegerColumnWriter::recordPosition() const {
+    ColumnWriter::recordPosition();
+    rleEncoder->recordPosition(rowIndexPosition.get());
+  }
+
+  std::unique_ptr<ColumnWriter> buildWriter(
+                                            const Type& type,
+                                            StreamsFactory& factory,
```
omalley: The factory object should be const&.

c++/src/ColumnWriter.cc
xndai: We could do that. Unfortunately our own code base doesn't support unique_ptr, removing this will create difficulties for us to integrate Orc code.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,294 @@
...
+
+  enum CompressionStrategy {
+    CompressionStrategy_SPEED = 0,
+    CompressionStrategy_COMPRESSION
+  };
+
+  enum RleVersion {
+    RleVersion_1,
+    RleVersion_2
+  };
```
xndai: I will add a FileVersion enum. I will still keep RleVersion enum for creating different version of encoder but will remove it from public header.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * Get the strip size.
+     * @return if not set, return default value.
+     */
+    uint64_t getStripeSize() const;
+
+    /**
+     * Set the block size.
+     */
+    WriterOptions& setBlockSize(uint64_t size);
```
xndai: The block size is essentially the size of compression block. It's used to initialize compression stream and output stream. Our test shows that block size does affect compression efficiency. That's why I leave it as an option.

c++/include/orc/Writer.hh
xndai: Sure. Will do.

c++/include/orc/Writer.hh
xndai: Will do.

c++/include/orc/Writer.hh
xndai: Computing min/max for string column is expensive due to memcpy() and strcmp(). So I leave it optional.

c++/include/orc/Writer.hh
xndai: Do you mean we don't need this option? Java interface does have it.

c++/include/orc/Writer.hh
xndai: Again Java interface has this option, but I can remove it.

c++/src/ColumnWriter.cc
xndai: sure.

xndai:
Hi @omalley , please see my replies and the new commit. Let me know if you have further questions.

c++/include/orc/Writer.hh
omalley: Ok, after tracking through the code I understand what they are being used for now. The source of confusion is that C++ and Java are using the names very differently.

C++ | Java | Use
-----| -----| ------
setBlockSize  | bufferSize | set the compression chunk size
setBufferSize | n/a             | set space pre-allocated for each stream
n/a                  | blockSize  | set the HDFS block size

C++'s setBuffer size is an artifact of preallocating the stream in a single chunk. I think we should use the compression chunk size for now, remove setBufferSize, either rename or document really well C++'s setBlockSize, and improve the memory management later.

c++/include/orc/Writer.hh
omalley: The memcpy isn't that expensive since it only happens when there is a new minimum or maximum. This would require a change to the ORC specification and would likely break current readers. We can discuss making such a change after we have performance numbers, but I really don't expect computing string statistics to be a noticeable cost.

c++/include/orc/Writer.hh
omalley: Well, the Java uses rowIndexStride(0) to mean disable. We can leave this one.

c++/include/orc/Writer.hh
omalley: The Java doesn't have it. Please remove it.

c++/include/orc/Writer.hh
xndai: It's not just memcpy() but also strcmp(), which is called for every string row. Based on our test results (TPC-H 1G data), string min, max stats calculation adds about 10% overhead to the writer. 

For now, I will just remove this option so it's consistent with spec.

c++/include/orc/Writer.hh
xndai: Sorry for the wording, I will rename setBlockSize() into setCompressionBlockSize(), and remove setBufferSize(). Also we don't want the initial stream capacity to be too small otherwise there will be lots of resizing that involve memory allocation, deallocation and copies. I would prefer a bigger size than the compression chunk size, e.g. 4M. I will hardcode it for now, and we can improve later.

c++/include/orc/Writer.hh
xndai: Ok, I just use rowIndexStride(0) for disabling index.

c++/include/orc/Writer.hh
xndai: I might have looked at the wrong file. Will remove it.

xndai:
Hi @omalley , updated again according to the latest feedback. Pls take another look.

c++/include/orc/Common.hh
```diff
@@ -27,6 +27,30 @@
 #include <string>
 
 namespace orc {
+
+  class FileVersion {
```
omalley: You should have 2 instance values for HIVE_0_11 and HIVE_0_12. With the current implementation, you'll need to throw if they use anything but HIVE_0_11, since that corresponds to RLEv1. We'll also need to have Reader.getFormatVersion() return this type too, but that can be done later.

c++/src/ColumnWriter.cc
omalley: Ok.

c++/src/Vector.cc
```diff
@@ -35,6 +35,7 @@ namespace orc {
                                           hasNulls(false),
                                           memoryPool(pool) {
     // PASS
```
omalley: I was using "// PASS" to mean it was intentionally empty. Since you added a statement, you should remove the comment also.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,228 @@
...
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace orc {
+
+  // classes that hold data members so we can maintain binary compatibility
+  struct WriterOptionsPrivate;
+
+  enum EncodingStrategy {
```
xndai: It's not used. I will remove it.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,228 @@
...
+    /**
+     * Get the data compression block size.
+     * @return if not set, return default value.
+     */
+    uint64_t getCompressionBlockSize() const;
+
+    /**
+     * Set row index stride.
+     */
+    WriterOptions& setRowIndexStride(uint64_t stride);
```
xndai: Sure.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,228 @@
...
+     * Get the error stream.
+     * @return if not set, return std::err.
+     */
+    std::ostream * getErrorStream() const;
+
+    /**
+     * Set whether or not to write statistics (file statistics,
+     * stripe statistics, etc.)
+     */
+    WriterOptions& setEnableStats(bool enable);
```
xndai: Will do.

c++/include/orc/Writer.hh
```diff
@@ -0,0 +1,228 @@
...
+  class Writer {
+  public:
+    virtual ~Writer();
+
+    /**
+     * Create a row batch for writing the columns into this file.
+     * @param size the number of rows to read
+     * @return a new ColumnVectorBatch to write into
+     */
+    virtual ORC_UNIQUE_PTR<ColumnVectorBatch> createRowBatch(uint64_t size
```
xndai: Will remove that.

c++/include/orc/Common.hh
xndai: I see. I will add the reader method as well.

c++/src/Vector.cc
xndai: Will do.

c++/include/orc/Writer.hh
xndai: Actually after giving a second look, I prefer keeping this method based on two reasons - 1. Reader has this method too. we probably want to be consistent. 2. Type::createRowBatch requires passing in memory pool object. In a lot of cases, we want to use the same memory pool that's used by writer. This method will ensure that.

xndai:
@omalley Update change again, please take a look. I also reduce the initial stream memory capacity from 4M to 1M, and greatly reduce the pre-allocated memory for footers and ps. This should alleviate the concern on writer memory consumption. We can figure out a long term solution later.

omalley:
Ok, I committed this., Thanks @xndai! I did make a small change to rename ColumnStatisticsImplBase to MutableColumnStatistics, which seemed a bit easier to remember.

PR #130, opened 2017-06-07
ORC-212 ORC-213 Fix timestamp pattern and added new test

spasam:
ORC-212: JSON convert should properly convert Timestamp string to java.sql.Timestamp
ORC-213: Fix Timestamp regex in JsonSchemaFinder

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/json/JsonSchemaFinder.java
+++ b/java/tools/src/java/org/apache/orc/tools/json/JsonSchemaFinder.java
@@ -58,7 +58,7 @@ public class JsonSchemaFinder {
   private static final Pattern TIMESTAMP_PATTERN =
       Pattern.compile("^[\"]?([0-9]{4}[-/][0-9]{2}[-/][0-9]{2})[T ]" +
           "([0-9]{2}:[0-9]{2}:[0-9]{2})" +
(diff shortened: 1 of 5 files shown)


java/tools/src/java/org/apache/orc/tools/json/JsonSchemaFinder.java
```diff
@@ -58,7 +58,7 @@
   private static final Pattern TIMESTAMP_PATTERN =
       Pattern.compile("^[\"]?([0-9]{4}[-/][0-9]{2}[-/][0-9]{2})[T ]" +
           "([0-9]{2}:[0-9]{2}:[0-9]{2})" +
-          "(([ ][-+]?[0-9]{2}([:][0-9]{2})?)|Z)?[\"]?$");
+          "(([ ]?[-+]?[0-9]{2}([:][0-9]{2})?)|Z)?[\"]?$");
```
omalley: I'd propose that this should be "( [-+]?|[-+])" instead of "[ ]?[-+]?". It doesn't seem right to accept 
"2016-12-30T12:34:5608"

spasam: @omalley Good catch. Let me tweak this!

omalley:
Can you please create a jira for this issue? Thanks! http://orc.apache.org/bugs

PR #131, opened 2017-06-16
ORC-199. Add convert from CSV.

omalley:
This patch incorporates Carter's converter into the tool jar. I haven't made any tests yet.

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/convert/ConvertTool.java
+++ b/java/tools/src/java/org/apache/orc/tools/convert/ConvertTool.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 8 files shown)


spasam:
@omalley Am interested in testing this with PostgreSQL dump output (CSV and TEXT). Is something blocking you from merging this? Thanks

cartershanklin:
I think Owen mentioned he had incorporated the code into another tool, if so we should close the ticket with the right pointer.

The original project had customizable null strings and strict mode that were put there with Postgres in mind. Interested to hear your experiences when you try it outl

java/tools/src/java/org/apache/orc/tools/convert/ConvertTool.java
```diff
@@ -84,7 +225,22 @@ static CommandLine parseOptions(String[] args) throws ParseException {
     options.addOption(
         Option.builder("o").longOpt("output").desc("Output filename")
             .hasArg().build());
-    CommandLine cli = new GnuParser().parse(options, args);
+    options.addOption(
```
spasam: Could you document the default values for these options in the help text?

java/tools/src/java/org/apache/orc/tools/convert/ConvertTool.java
```diff
@@ -18,53 +18,178 @@
...
+        schemaFinder.addFile(file.getReader(file.filesystem.open(file.path)));
+      } else if (file.format == Format.ORC) {
+        System.err.println("Merging schema from " + file.path);
+        Reader reader = OrcFile.createReader(file.path,
+            OrcFile.readerOptions(conf)
+                .filesystem(file.filesystem));
+        schemaFinder.addSchema(reader.getSchema());
+      }
     }
     return schemaFinder.getSchema();
```
spasam: This is throwing NPE if no command line arguments are specified except for CSV file:

```
Exception in thread "main" java.lang.NullPointerException
	at org.apache.orc.tools.json.JsonSchemaFinder.getSchema(JsonSchemaFinder.java:321)
	at org.apache.orc.tools.convert.ConvertTool.buildSchema(ConvertTool.java:75)

```

spasam:
Finally was able to test this end to end. I had to tweak CsvReader. Added customer converters for boolean and timestamp. I can upload changes in a separate pull request.

Ideally, if **json-schema** Driver command line option is changed to **schema** and CSV schema is determined on the fly (assuming there is a header line with column names), this would be awesome.

PR #132, opened 2017-06-16
ORC-202. Add writer implementation enum to file format

omalley:


java/core/src/java/org/apache/orc/OrcFile.java
```diff
@@ -108,66 +108,118 @@ public int getMinor() {
...
-    ORC_101(5),    // bloom filters use utf8
-    ORC_135(6), // timestamp stats use utc
+    // Java ORC Writer
+    ORIGINAL(WriterImplementation.ORC_JAVA, 0),
+    HIVE_8732(WriterImplementation.ORC_JAVA, 1), // corrupted stripe/file maximum column statistics
+    HIVE_4243(WriterImplementation.ORC_JAVA, 2), // use real column names from Hive tables
+    HIVE_12055(WriterImplementation.ORC_JAVA, 3), // vectorized writer
+    HIVE_13083(WriterImplementation.ORC_JAVA, 4), // decimal writer updating present stream wrongly
+    ORC_101(WriterImplementation.ORC_JAVA, 5),    // bloom filters use utf8
+    ORC_135(WriterImplementation.ORC_JAVA, 6), // timestamp stats use utc
```
dain: Can you add these comments to the protobuf file also?

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -3039,9 +3039,45 @@ public void testExpansion() throws Exception {
...
+        OrcFile.WriterVersion.from(OrcFile.WriterImplementation.ORC_JAVA, 99));
+    assertEquals(OrcFile.WriterVersion.ORIGINAL,
+        OrcFile.WriterVersion.from(OrcFile.WriterImplementation.ORC_JAVA, 0));
+    assertEquals(OrcFile.WriterVersion.HIVE_4243,
+        OrcFile.WriterVersion.from(OrcFile.WriterImplementation.ORC_JAVA, 2));
+    assertEquals(OrcFile.WriterVersion.FUTURE,
+        OrcFile.WriterVersion.from(OrcFile.WriterImplementation.ORC_CPP, 99));
+    assertEquals(OrcFile.WriterVersion.ORC_CPP_ORIGINAL,
+        OrcFile.WriterVersion.from(OrcFile.WriterImplementation.ORC_CPP, 0));
+    assertEquals(OrcFile.WriterVersion.ORC_CPP_ORIGINAL,
```
dain: Add a test for the Presto version?

proto/orc_proto.proto
```diff
@@ -221,15 +227,29 @@ message PostScript {
...
+  //   0 = original
   //   1 = HIVE-8732 fixed
   //   2 = HIVE-4243 fixed
   //   3 = HIVE-12055 fixed
   //   4 = HIVE-13083 fixed
   //   5 = ORC-101 fixed
   //   6 = ORC-135 fixed
+  //
+  // Version of the ORC C++ writer:
+  //   0 = original
```
dain: We likely need to start these with version 6 (maybe jump to 10) so that old readers don't think this is one of the over versions of the ORC_JAVA writer.

java/core/src/java/org/apache/orc/OrcFile.java
omalley: Sure.

proto/orc_proto.proto
omalley: In the Java, I made the test for "includes" treat the other writers as always having the bug fixed. So, PRESTO_ORIGINAL will automatically pass all of the checks looking for specific ORC or Hive jiras (HIVE-4243 or ORC-101).

That does mean that if all three implementations are broken in the same way, we'll need a test that looks like:
    if (version.includes(WriterVersion.ORC-999) && version.includes(WriterVersion.ORC-1000)
        && version.includes(WriterVersion.PRESTO-1001) {
        ... some future bug is fixed
    }

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
omalley: Sure

proto/orc_proto.proto
dain: I'm thinking of the previous releases of the ORC reader where they don't know about the writer implementation field, so they will assume that Presto original is ORC_JAVA original.  Moving forward, I think this design works.

omalley: *doh* Yes, of course.

omalley:
Ok, I've addressed all of Dain's feedback. :)

java/core/src/java/org/apache/orc/impl/OrcTail.java
```diff
@@ -70,8 +70,11 @@ public long getFileModificationTime() {
 
   public OrcFile.WriterVersion getWriterVersion() {
     OrcProto.PostScript ps = fileTail.getPostscript();
+    OrcProto.Footer footer = fileTail.getFooter();
+    OrcFile.WriterImplementation writer =
+        OrcFile.WriterImplementation.from(footer.getWriter());
     return (ps.hasWriterVersion()
```
dain: I suggest we make `writerVersion` required for the new writer implementations, so you would just throw in that case...  This is similar to the reason we are starting with writer version 6

proto/orc_proto.proto
```diff
@@ -221,15 +227,32 @@ message PostScript {
...
-  //   4 = HIVE-13083 fixed
-  //   5 = ORC-101 fixed
-  //   6 = ORC-135 fixed
+
+  // The version of the writer that wrote the file. This number is
+  // updated when we make fixes or large changes to the writer so that
+  // readers can detect whether a given bug is present in the data.
+  //
+  // These numbers are assigned from 6 per a writer so that readers that
+  // predate ORC-202 treat the new writers correctly.
```
dain: maybe note that writer version is required for everything except for the original version of ORC_JAVA.

java/core/src/java/org/apache/orc/impl/OrcTail.java
omalley: writerVersion is required for anything except the pre-HIVE-8732 Java ORC files. Otherwise, as you say, the readers will interpret it as one of the pre-HIVE-8732 files. I can simplify this condition, because if it isn't present it will get the default value of 0, which is correct.

The current code in this pull request, would make broken files that have writer = presto, writerVersion = 0 as future. I guess I could check for the case of presto or orc_cpp and version < 6 and throw an exception.

dain: That is what I would prefer.  In my mind the file would be corrupt and we should fail.

proto/orc_proto.proto
omalley: Ok, I tried to make the comment in the protobuf clearer.

xndai:
@omalley Sorry, I didn't see this earlier. One minor comment - should we define an enum for writers in protobuf? This way we get consistent writer code across board.

omalley:
There are a couple of problems with enums in Protobuf... The main is that if it is an unknown value (from a future version) it becomes unset. It is important that we can tell the difference between an unset value (orc java writer) and some future writer implementation. It is also useful to be able to print the value of the unknown enum.

Obviously I wasn't aware of the problems when I wrote the original version of orc_proto.proto. I probably should go through and convert them all to integers.

xndai:
I see. We can work around the parsing issue through a default value. But still not able to print out the value if it's out of range.

omalley:
Actually, you can't work around it unless you wrap the enum in a separate message. The problem is that a reader can't tell the difference from the orc files produced by ORC 1.4 (no writer set) and an unknown writer (writer mapped to the default). They will both show up as null. You would need to wrap a message around the enum to make it work.

xndai:
That makes sense. I add the writer code in c++ writer. We might need a similar change for c++ reader. Will do it in a separate PR.

PR #134, opened 2017-06-20
Orc 17

AnatoliShein:
In this pull request I added **LIBHDFS++** library for reading files from HDFS to **ORC** project. 

Libhdfs++ is located in orc/c++/lib/libhdfspp and by default builds as a light-weight library without examples, tests, and tools (and by this avoids dependencies on JDK, valgrind and gmock). However, if the flag **-DHDFSPP_LIBRARY_ONLY=FALSE** is passed to cmake, then it will build the examples, tests, and tools as well.

Libhdfs++ depends on protobuf libraries in orc/c++/libs/protobuf-2.6.0 and is searching the system for packages Doxygen, OpenSSL, CyrusSASL, GSasl, and Threads dynamically (however only OpenSSL and Threads are required).

The folder libhdfspp also includes a script pull_hdfs.sh which pulls the latest changes from Libhdfs++ Hadoop branch to ORC, and generates file 'imported_timestamp' with the timestamp and the information about the latest commit.

I also updated all the ORC tools to automatically use Libhdfs++ to read ORC files on HDFS if their path begins with 'hdfs://'.

Please review.


c++/src/OrcHdfsFile.cc
```diff
@@ -0,0 +1,170 @@
...
+
+      if (!buf) {
+        throw ParseError("Buffer is null");
+      }
+
+      hdfs::Status status;
+      char input_buffer[BUF_SIZE];
+      do{
+        //Reading file chunks
+        status = file->PositionRead(input_buffer, sizeof(input_buffer), offset, &last_bytes_read);
```
xndai: Why do we need an intermedia buffer here? Can we just pass in "buf" and "length" into PositionRead()? This way we don't need following memcpy() and can also simply the logic here.

c++/src/OrcHdfsFile.cc
```diff
@@ -0,0 +1,170 @@
...
+      hdfs::FileHandle *file_raw = nullptr;
+      status = fs->Open(uri->get_path(), &file_raw);
+      if (!status.ok()) {
+        throw ParseError("Can't open " + uri->get_path() + ". " + status.ToString());
+      }
+      //wrapping file_raw into a unique pointer to guarantee deletion
+      std::unique_ptr<hdfs::FileHandle> file_handle(file_raw);
+      file = std::move(file_handle);
+      //transferring the ownership of FileSystem to HdfsFileInputStream to avoid premature deallocation
+      file_system = fs;
```
xndai: Not quite understand why we need a local variable "fs". You could just initialize this->file_system. Also "fs" should have been a unique_ptr as it doesn't share ownership with anyone.

c++/src/OrcHdfsFile.cc
```diff
@@ -0,0 +1,170 @@
...
+      }
+
+      hdfs::FileHandle *file_raw = nullptr;
+      status = fs->Open(uri->get_path(), &file_raw);
+      if (!status.ok()) {
+        throw ParseError("Can't open " + uri->get_path() + ". " + status.ToString());
+      }
+      //wrapping file_raw into a unique pointer to guarantee deletion
+      std::unique_ptr<hdfs::FileHandle> file_handle(file_raw);
+      file = std::move(file_handle);
```
xndai: You could combine these two lines into file.reset(file_raw);

AnatoliShein: Yep, good catch.

c++/src/OrcHdfsFile.cc
AnatoliShein: This makes sense, we don't really need a shared pointer here. Just fixed it.

c++/src/OrcHdfsFile.cc
AnatoliShein: I used intermediate buffer because it was used for PositionRead in a similar case. So I just tried without this buffer and it worked.

c++/src/OrcHdfsFile.cc
```diff
@@ -121,39 +116,24 @@ namespace orc {
...
-          } else {
-            memcpy((char*) buf + total_bytes_read, input_buffer, last_bytes_read);
-            total_bytes_read += last_bytes_read;
-            offset += last_bytes_read;
-          }
-        } else {
-          throw ParseError("Error reading the file: " + status.ToString());
-        }
-      } while (true);
+      status = file->PositionRead(buf, static_cast<size_t>(length), static_cast<off_t>(offset), &last_bytes_read);
```
xndai: you still need to put this in a loop in case last_bytes_read is less than length.

c++/src/OrcHdfsFile.cc
```diff
@@ -66,22 +64,22 @@ namespace orc {
         options = config->GetOptions();
       }
       hdfs::IoService * io_service = hdfs::IoService::New();
-      //Wrapping fs into a shared pointer to guarantee deletion
-      std::shared_ptr<hdfs::FileSystem> fs(hdfs::FileSystem::New(io_service, "", options));
-      if (!fs) {
+      //Wrapping file_system into a unique pointer to guarantee deletion
+      file_system = std::unique_ptr<hdfs::FileSystem>(hdfs::FileSystem::New(io_service, "", options));
+      if (!file_system) {
```
xndai: are you sure this would work? Should it be if (file_system.get() != nullptr) ?

c++/src/OrcHdfsFile.cc
```diff
@@ -34,15 +34,13 @@
 #include "common/hdfs_configuration.h"
 #include "common/configuration_loader.h"
 
-#define BUF_SIZE 1048576 //1 MB
-
 namespace orc {
 
   class HdfsFileInputStream : public InputStream {
```
xndai: do you have corresponding UTs for this new class?

c++/src/OrcHdfsFile.cc
AnatoliShein: Yes, unique pointer actually supports this notation (thanks to [operator bool](http://en.cppreference.com/w/cpp/memory/unique_ptr/operator_bool))

c++/src/OrcHdfsFile.cc
AnatoliShein: I don't currently have the UTs for this class because to test it we need a hadoop cluster, and in order to fire one up we would need to bring the entire hadoop tree into ORC (which is actually much larger than ORC), and it can make things very heavy.

c++/src/OrcHdfsFile.cc
AnatoliShein: This makes sense. I just added a loop here.

c++/src/OrcHdfsFile.cc
xndai: Unfortunately unique_ptr can be redefined as auto_ptr if platform doesn't support unique_ptr (see orc_config.hh). We talked about removing this redefine, but it hasn't been done yet. So I'd suggest we stick to file_system.get() != nullptr for now.

c++/src/OrcHdfsFile.cc
```diff
@@ -123,17 +123,21 @@ namespace orc {
...
       hdfs::Status status;
-      status = file->PositionRead(buf, static_cast<size_t>(length), static_cast<off_t>(offset), &last_bytes_read);
-      if(!status.ok()) {
-        throw ParseError("Error reading the file: " + status.ToString());
-      }
+      size_t total_bytes_read = 0;
+      size_t last_bytes_read = 0;
+      
+      do {
+        status = file->PositionRead(buf, static_cast<size_t>(length) - total_bytes_read, static_cast<off_t>(offset + total_bytes_read), &last_bytes_read);
```
xndai: Make sure to wrap on 80 characters.

c++/src/OrcHdfsFile.cc
xndai: Does HDFS C++ provide a mocked service interface? My concern is the testability of HdfsFileInputStream.

c++/src/OrcHdfsFile.cc
AnatoliShein: I changed it here, however '!fs' notation is also used in many other parts of the libhdfs++ library, like examples, tests, tools, and c bindings, so hopefully this does not create too much trouble.

c++/src/OrcHdfsFile.cc
AnatoliShein: Just fixed the whole file to wrap on 80 chars.

c++/src/OrcHdfsFile.cc
AnatoliShein: Currently we don't actually have a mocked service interface in libhdfs++.

jamesclampffer: For a mock that does real network stuff libhdfs++ uses the C++/JVM bindings over the minidfscluster (java namenode and datanode in a single process) for its unit tests.  That seems like a very large dependency to bring into this project.  It'd be possible to write a mock that goes over TCP in C/C++, I can think of 2 ways:
1) Capture network traffic of a read with tcpdump/wireshark for a given request and send the response back.  This would be really brittle if IO request sizes changed and pretty ugly to store a bunch of binary data.
2) Implement at least some of the server side HDFS RPC protocol and DataTransferProtocol to make a mock namenode and datanode.  This is possible but would be a considerable amount of work that wouldn't be as useful for libhdfs++ since it already has the minidfscluster.

c++/libs/libhdfspp/CMakeLists.txt
```diff
@@ -0,0 +1,248 @@
...
+SET(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMake" ${CMAKE_MODULE_PATH})
+
+# If there's a better way to inform FindCyrusSASL.cmake, let's make this cleaner:
+SET(CMAKE_PREFIX_PATH "${CMAKE_PREFIX_PATH};${CYRUS_SASL_DIR};${GSASL_DIR}")
+
+find_package(Doxygen)
+find_package(OpenSSL REQUIRED)
+find_package(Protobuf REQUIRED)
+find_package(CyrusSASL)
+find_package(GSasl)
```
jamesclampffer: I think you're going to need a flag to make sure it doesn't attempt to find and link against GSasl since the Apache and GPL licenses aren't compatible.  Eventually GSasl will be removed from libhdfs++ for the same reason.

majetideepak:
Can we add libhdfspp as a tarball and add a line to untar it?
I am going to make these changes for other libraries in the lib folder as well.

omalley:
Sorry, this patch is really out of date with the master. Can you rebase it?

AnatoliShein:
Hi @omalley ,
Yes, I am working on rebasing it right now. Running into some protobuf problems, but it should not be too bad.

AnatoliShein:
Rebased the branch and resolved the protobuf issues, but still need to move Libhdfspp out of the libs and fix other samll issues. Testing if it passes CI-tests now.

omalley:
Unfortunately, this doesn't build on MacOS. :(

I'll take some time this week to figure out what needs to be fixed. I've gone ahead and rebased it to the current master and squashed the commits in https://github.com/omalley/orc/tree/pr/134

AnatoliShein:
Hi @omalley ,
What error messages do you get when trying to build on MacOS?

omalley:
It looks like it is a "long" vs "long long" problem.

````
/Users/owen/work/code/orc/build/libhdfspp_ep-prefix/src/libhdfspp_ep/lib/common/configuration.cc:85:12: error: no viable conversion from returned value of type 'optional<long>' to function return type 'optional<long long>'
    return result;
           ^~~~~~
/Users/owen/work/code/orc/build/libhdfspp_ep-prefix/src/libhdfspp_ep/third_party/tr2/optional.hpp:427:13: note: candidate constructor not viable: no known conversion from 'std::experimental::optional<long>' to 'std::experimental::nullopt_t' for 1st argument
  constexpr optional(nullopt_t) noexcept : OptionalBase<T>() {};
            ^
/Users/owen/work/code/orc/build/libhdfspp_ep-prefix/src/libhdfspp_ep/third_party/tr2/optional.hpp:429:3: note: candidate constructor not viable: no known conversion from 'std::experimental::optional<long>' to 'const std::experimental::optional<long long> &' for 1st argument
  optional(const optional& rhs)
  ^
/Users/owen/work/code/orc/build/libhdfspp_ep-prefix/src/libhdfspp_ep/third_party/tr2/optional.hpp:438:3: note: candidate constructor not viable: no known conversion from 'std::experimental::optional<long>' to 'std::experimental::optional<long long> &&' for 1st argument
  optional(optional&& rhs) noexcept(is_nothrow_move_constructible<T>::value)
  ^
/Users/owen/work/code/orc/build/libhdfspp_ep-prefix/src/libhdfspp_ep/third_party/tr2/optional.hpp:447:13: note: candidate constructor not viable: no known conversion from 'std::experimental::optional<long>' to 'const long long &' for 1st argument
  constexpr optional(const T& v) : OptionalBase<T>(v) {}
            ^
/Users/owen/work/code/orc/build/libhdfspp_ep-prefix/src/libhdfspp_ep/third_party/tr2/optional.hpp:449:13: note: candidate constructor not viable: no known conversion from 'std::experimental::optional<long>' to 'long long &&' for 1st argument
  constexpr optional(T&& v) : OptionalBase<T>(constexpr_move(v)) {}
            ^
1 error generated.
make[5]: *** [lib/common/CMakeFiles/common_obj.dir/configuration.cc.o] Error 1
make[4]: *** [lib/common/CMakeFiles/common_obj.dir/all] Error 2
make[3]: *** [all] Error 2
````

**Comment by jamesclampffer** [comment]
Those errors should go away after HDFS-10787 is finished (or applied to the PR temporarily) since it exposes an API to get at the configuration without using std::tr2::optional.  We had to suppress a bunch of errors coming from optional when it was added to libhdfs++; eventually I'd like to get rid of the remaining uses of it in the implementation as well.

**Comment by majetideepak** [comment]
I am adding a Travis test for OSX os. It should ease catching OSX issues.

**Comment by majetideepak** [comment]
Can you check if `thread_local` is supported by the platform, and then disable libhdfspp on those platforms?

**Comment by jamesclampffer** [comment]
@majetideepak Is there a way to get the OSX CI to use a generic LLVM/Clang install rather than the one that gets shipped with OSX?

Apparently they apply a patch to Clang to break thread_local, my understanding is they don't want to break the ABI if they ever get around to making builtin thread local specialized for their OS.
```
        .Case("cxx_thread_local",
-                 LangOpts.CPlusPlus11 && PP.getTargetInfo().isTLSSupported() &&
-                 !PP.getTargetInfo().getTriple().isOSDarwin())
+                 LangOpts.CPlusPlus11 && PP.getTargetInfo().isTLSSupported())
```

Otherwise this won't get coverage on OSX CI builds.

**Comment by omalley** [comment]
.. Owen

> On Aug 18, 2017, at 07:14, jamesclampffer <notifications@github.com> wrote:
> 
> @majetideepak Is there a way to get the OSX CI to use a generic LLVM/Clang install rather than the one that gets shipped with OSX?
> 

Don't do that. If it doesn't compile with the "real" Mac OS compiler, you'll break my dev platform.

**Comment by AnatoliShein** [comment]
Hi all! It looks like thread_local feature is by default supported by Xcode 8.0 (and above) which runs on OS X El Capitan (10.11.5) or newer. Is it possible to update the osx_image of the CI-run from "xcode6.4" to at least "xcode8.0"? For older versions we can just display a warning message and not build libhdfspp.

**Comment by AnatoliShein** [comment]
Looks like [THIS](https://docs.travis-ci.com/user/reference/osx/#OS-X-Version) is the list of available CI images for OS X. I will try osx_image: xcode8.3 now and see what happens.

**Comment by omalley** [comment]
There are a lot of warnings in libhdfs that clang reports.

It is currently failing in:
[ 91%] Linking CXX shared library libhdfspp.dylib

with link errors about sasl and protobuf.

**Comment by AnatoliShein** [comment]
Hi @omalley , the linking should be fixed with the new commit, and we will fix the warnings shortly.

**Comment by AnatoliShein** [comment]
Hi guys, I now fixed all the build errors and warnings for OS X (with Xcode 8+), and I would like some additional feedback on this pull request if you have any. Thanks!
(@omalley , @majetideepak , @xndai, @jamesclampffer )

File: c++/src/OrcHdfsFile.cc [hunk 9]
```diff
@@ -0,0 +1,172 @@
...
+          }
+        }
+      }
+
+      if (file_system.get() == nullptr) {
+        throw ParseError("Can't connect the file system. ");
+      }
+
+      hdfs::FileHandle *file_raw = nullptr;
+      status = file_system->Open(uri.get_path(), &file_raw);
```
xndai: nit: you may want to cache "status" within the class.

c++/src/OrcHdfsFile.cc
```diff
@@ -0,0 +1,172 @@
...
+      }
+      totalLength = stat_info.length;
+    }
+
+    uint64_t getLength() const override {
+      return totalLength;
+    }
+
+    uint64_t getNaturalReadSize() const override {
+      return 1048576; //1 MB
```
xndai: nit: Use const variable?

c++/include/orc/OrcFile.hh
```diff
@@ -103,12 +103,18 @@ namespace orc {
   };
 
   /**
-   * Create a stream to a local file.
+   * Create a stream to a local file or HDFS file if path begins with "hdfs://"
```
xndai: why do you want readLocalFile() to handle hdfs files? You already have readHdfsFile() which does exactly that.

AnatoliShein: Yes, I guess this is a little confusing. Currently I am calling readHdfsFile from readLocalFile if it starts with "hdfs://". To fix this I will add a new function readFile which will call either readHdfsFile or readLocalFile.

c++/src/OrcHdfsFile.cc
AnatoliShein: This sounds good, I will update this.

c++/src/OrcHdfsFile.cc
AnatoliShein: I am not quite sure why we will want to cache "status", since it is just used for the return values of HDFS operations (success or error message). Could you please explain?

xndai: My feeling is that you may want to return other meta info in the future. But it's your call.

xndai:
LGTM

AnatoliShein:
Now ORC can be built without LIBHDFSPP like this:
cmake -DBUILD_LIBHDFSPP=off ..

omalley:
Ok, this patch breaks the docker scripts on at least centos7, debian8, and ubuntu16 by adding new dependencies for the c++ build. I'm going through and fixing them.

gxkevin:
my hdfs path just like this:  hdfs://yycluster02/hive_warehouse/rs_hago_client.db/peta/test/orc_test/my-file.orc

gxkevin:
my hdfs path just like this: hdfs://yycluster02/hive_warehouse/rs_hago_client.db/peta/test/orc_test/my-file.orc

but, I got the error 
    '''
        [ERROR ][Async Runtime ][Fri Oct  9 17:05:09 2020][Thread id = 140438677890880][/home/gaoxiang2/workspace/online_code/orc/build/libhdfspp_ep-prefix/src/libhdfspp_ep/lib/common/namenode_info.cc:170]    Unabled to resolve endpoints for host: yycluster02 port: -1
[ERROR ][RPC           ][Fri Oct  9 17:05:09 2020][Thread id = 140438677890880][/home/gaoxiang2/workspace/online_code/orc/build/libhdfspp_ep-prefix/src/libhdfspp_ep/lib/rpc/rpc_engine.cc:191]    RpcEngine::AsyncRpcCommsError called; status="No endpoints provided" conn=0xf07340 reqs=1
[WARN  ][RPC           ][Fri Oct  9 17:05:09 2020][Thread id = 140438651348736][/home/gaoxiang2/workspace/online_code/orc/build/libhdfspp_ep-prefix/src/libhdfspp_ep/lib/rpc/rpc_engine.cc:202]    RpcEngine::RpcCommsError called; status="No endpoints provided" conn=0xf07340 reqs=1
Can't connect to yycluster02:. No endpoints provided
    '''

PR #135, opened 2017-07-10
ORC-204: Update and use CMake ExternalProject_Add to build compression libraries

majetideepak:
Including the whole source of external libraries adds bloat and unnecessary.
It also is not useful if clients prefer to use their own third-party libraries.
In this PR
1) The libraries are updated to the most recent ones.
2) Compressions library sources are deleted. Instead, compressed tar files of newer versions are added. (Future patch will download the tar files instead)
3) CMake `ExternalProject_Add`  is used to build the libraries

Files to review
```
cmake_modules/ThirdpartyToolchain.cmake
c++/libs/CMakeLists.txt
CMakeLists.txt
```
Rest of the changes are just deletes of the libraries
I will file a JIRA to enable users the ability to provide their own libraries.


cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -0,0 +1,90 @@
...
+if (${UPPERCASE_BUILD_TYPE} EQUAL "RELEASE")
+   set (SNAPPY_CXXFLAGS "CXXFLAGS='-DNDEBUG -O2'")
+endif ()
+
+ExternalProject_Add (snappy_ep
+  CONFIGURE_COMMAND ./configure "--prefix=${SNAPPY_PREFIX}" ${SNAPPY_CXXFLAGS}
+  BUILD_IN_SOURCE 1
+  BUILD_COMMAND ${MAKE}
+  INSTALL_DIR ${SNAPPY_PREFIX}
+  URL ${SNAPPY_SRC_URL}
```
wgtmac: Just curious if it is better to put an HTTP URL here to download compression libraries? In this approach, we can reduce the repo size and easy to bump library version in the future.

majetideepak: +1. I mentioned using HTTP URL(download the tar files) as future work because I wasn't sure of its impact on current users. I will update to use URLs in this patch if there are no concerns.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -13,17 +13,18 @@ set (SNAPPY_HOME "${SNAPPY_PREFIX}")
 set (SNAPPY_INCLUDE_DIRS "${SNAPPY_PREFIX}/include")
 set (SNAPPY_STATIC_LIB_NAME snappy)
 set (SNAPPY_STATIC_LIB "${SNAPPY_PREFIX}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}${SNAPPY_STATIC_LIB_NAME}${CMAKE_STATIC_LIBRARY_SUFFIX}")
-set (SNAPPY_SRC_URL "${CMAKE_SOURCE_DIR}/c++/libs/snappy-${SNAPPY_VERSION}.tar.gz")
+set (SNAPPY_SRC_URL "https://github.com/google/snappy/releases/download/1.1.4/snappy-${SNAPPY_VERSION}.tar.gz")
```
wgtmac: this line contains a hard-coded version 1.1.4

majetideepak: Good catch! Fixing it.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -0,0 +1,90 @@
+set (LZ4_VERSION "1.7.5")
+set (SNAPPY_VERSION "1.1.3")
+set (ZLIB_VERSION "1.2.11")
```
omalley: Let's put these variable sets in the top level CMakeLists.txt file.

majetideepak:
Fixed packaging static compression libraries. Only the source is available for Snappy 1.1.6 and I could not `install`(build works fine) this release. CMake support was introduced but only shared libraries are being built. Snappy 1.1.5 and 1.1.6 are the same. There is no performance difference between the current 1.1.4 and 1.1.6 
 https://github.com/google/snappy/releases

omalley:
Ok, I committed this with only a slight tweak to also filter out the shared libraries on MacOS.

PR #138, opened 2017-07-25
ORC-91: Use hdfs v-blocks instead of zero-padding stripes

t3rmin4t0r:


java/core/src/java/org/apache/orc/impl/HadoopShims.java
```diff
@@ -124,6 +126,43 @@
    */
   public TextReaderShim getTextReaderShim(InputStream input) throws IOException;
 
+  
+  /** 
+   * Block filler shim - make sure the DFS blocks ends at this offset.
+   */
+  public interface BlockFillerShim {
```
omalley: I think it would be better to create a HadoopShims_2_5 that has the simple fill method and make HadoopShimsCurrent have the full featured one.

omalley:
Ok, I've rebased this patch on top of ORC-234 in https://github.com/apache/orc/pull/166 . @t3rmin4t0r can you please review it?

PR #139, opened 2017-07-25
ORC-216: VectorRowBatch creation from TypeDescription invalidates batch size

prasanthj:
TypeDescription.createRowBatch() always resets the size to 0.
https://github.com/apache/orc/blob/1ef78bbfc7e834deca973ee8f9517c8575737fcf/java/core/src/java/org/apache/orc/TypeDescription.java#L669-L680


omalley:
This isn't a bug. The batch has no rows when created and thus batch.size should be 0. batch.getMaxSize() will return the capacity.

PR #140, opened 2017-07-25
ORC-217: Duplicate rat plugins in pom.xml

prasanthj:
Rat plugin is duplicated in pom.xml


prasanthj:
@omalley can you please take a look? small change

PR #142, opened 2017-07-25
[ORC-218] Cache timezone information in the library.

grundprinzip:
Right now, for every lookup of a timezone, the library will go to
disk and parse the timezone file. While the results are cached,
doing these system calls should be avoided in environments with
restricted system calls.

This patch loads all the time zone files into the binary at
build time and instead of accessing disk for the TZ
information will simply load the buffer from memory. In addition,
this allows the user to compile the library with a static
timezone to be assumed for local use. This comes in handy to avoid
yet another set of system calls to assume the local timezone.

The following CMake flags control this feature:

  * EMBEDED_TZ_DB - disables this feature if set to OFF
  * STATIC_TZ=Val - sets a specific timezone

### DIFF
--- a/c++/file2array.sh
+++ b/c++/file2array.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License
+
+# Convert file to a C++11 vector of unsigned char
+#
+set -e
+
+arrname=""
+modifiers=""
+
+usage () {
+  echo "Usage: $0 [options] [input file] " >&2
+  echo "Options:" >&2
+  echo "  -v <c array variable name>" >&2
+  echo "     Name of C variable in output file. Must be provided." >&2
+  echo "  -m <array variable modifiers>" >&2
+  echo "     Modifiers for C variable in output file. Default is const." >&2
+  exit 1
+}
+
+while getopts "m:v:" opt; do
+  case $opt in
+    m)
+      if [[ $modifiers != "" ]]; then
+        echo "-m specified twice" >&2
+        usage
+      fi
+      modifiers=$OPTARG
+      ;;
+    v)
+      if [[ $arrname != "" ]]; then
+        echo "-v specified twice" >&2
+        usage
+      fi
+      arrname=$OPTARG
+      ;;
+    \?)
+      echo "Invalid option: -$OPTARG" >&2
+      usage
+  esac
+done
+shift $((OPTIND - 1))
+
+infile=$1
+if [[ $# > 1 ]]; then
+  echo "Too many remaining arguments: $@" >&2
+  usage
+fi
+
+if [ -z "$arrname" ]; then
+  echo "-v not provided or empty." >&2
+  usage
+fi
+
+if [ -z "$modifiers" ]; then
+  # Default is const with global linking visibility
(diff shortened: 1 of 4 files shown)


c++/file2array.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+# Copyright 2017 Amazon.com and it's affiliates
+# Copyright 2015 Cloudera Inc.
+# Copyright 2013 University of Chicago and Argonne National Laboratory
```
majetideepak: I am not sure if this Copyright information and the Apache License can co-exist.

c++/src/CMakeLists.txt
```diff
@@ -125,6 +125,83 @@ include_directories (
   ${LZ4_INCLUDE_DIRS}
   )
 
+# To avoid reading the Timezone database from disk, we load the file during the
+# build and inject them as binary into the library. This can increase the size
+# of library but avoids doing system calls at runtime. The behavior can be
+# disabled by using the CMake flag NO_EMBEDDED_TZ_DB
+if (NOT DEFINED EMBEDDED_TZ_DB)
+  set(EMBEDDED_TZ_DB ON)
```
majetideepak: This feature should be OFF by default.

majetideepak: We won't need the ``if`` check then

c++/src/CMakeLists.txt
```diff
@@ -125,6 +125,83 @@ include_directories (
...
+
+  if (TZ_LIST_RC)
+    message(FATAL_ERROR "Could not parse Timezone database.")
+  endif()
+
+  set(FILE2ARRAY "${CMAKE_CURRENT_SOURCE_DIR}/../file2array.sh")
+
+  # For each zone generate a target
+  # Add each target to dependency on Timezone.cc
+  set(TZ_DEPS "")
```
majetideepak: hoist this initialization out of the `if (EMBEDDED_TZ_DB)`?
That way we can safely add it as a dependency to orc library on line 214

majetideepak:
Please add another travis-ci test with EMBEDDED_TZ_DB=ON to ensure this change gets tested

c++/src/CMakeLists.txt
xndai: A side effect of this is when the time zone info files are updated, the Orc lib has to be recompiled and redistributed.

c++/src/Timezone.cc
```diff
@@ -735,18 +739,36 @@ namespace orc {
...
   const Timezone& getTimezoneByName(const std::string& zone) {
+#ifdef USE_TZ_DB
+    {
+      std::lock_guard<std::mutex> timezone_lock(timezone_mutex);
+      std::map<std::string, std::shared_ptr<Timezone> >::iterator itr =
+          timezoneCache.find(zone);
+      if (itr != timezoneCache.end()) {
+        return *(itr->second).get();
+      }
+      timezoneCache[zone] = std::shared_ptr<Timezone>(new TimezoneImpl(zone, TZ_DATABASE[zone]));
```
xndai: why don't we pre populate the whole time zone map so you don't need to acquire a mutex every time accessing the map?

c++/src/CMakeLists.txt
grundprinzip: Done

c++/src/CMakeLists.txt
grundprinzip: Done

c++/src/Timezone.cc
grundprinzip: Done

omalley:
The build is failing when the cache is enabled. I'd guess the clean up is causing a segfault.

omalley:
For most environments, this is a bad idea. Those zone files are patched all of the time. Embedding them into a library will create problems. However, it does provide a way out for environments that either have very old zone files or no copies at all.

grundprinzip:
There's some funky stuff that's going on with the way how the code get's compiled in Travis, that I need to reproduce. The unit tests with the same combination of parameters works for me locally. Still working on it.

omalley:
You may find the docker scripts useful for testing in other linux platforms.

.. Owen

PR #144, opened 2017-07-30
[ORC-219] Boolean and timestamp converter for CSV.

spasam:
Added support for boolean and timestamp readers for CSV conversion.
Fixed deprecated JUnit imports in tests
Fixed deprecated GnuParser/Options
Removed unused imports
Upgraded JUnit to 4.12

Test CSV convert with Posgres TSV output that has boolean and timestamp

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/convert/CsvReader.java
+++ b/java/tools/src/java/org/apache/orc/tools/convert/CsvReader.java
@@ -7,18 +7,27 @@ import org.apache.hadoop.hive.ql.exec.vector.DecimalColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.DoubleColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.LongColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.StructColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.TimestampColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
 import org.apache.hadoop.hive.serde2.io.HiveDecimalWritable;
 import org.apache.hadoop.io.IntWritable;
 import org.apache.orc.RecordReader;
 import org.apache.orc.TypeDescription;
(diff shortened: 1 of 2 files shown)


java/core/pom.xml
```diff
@@ -57,6 +57,10 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
```
omalley: Let's keep this dependency in the tools module rather than core.

java/core/src/java/org/apache/orc/RecordReader.java
```diff
@@ -20,11 +20,15 @@
 import java.io.IOException;
 
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
+import org.threeten.bp.format.DateTimeFormatter;
 
 /**
  * A row-by-row iterator for ORC files.
  */
 public interface RecordReader extends AutoCloseable {
+  DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter.ofPattern(
```
omalley: This should move the the CSV converter in tools.

java/bench/src/java/org/apache/orc/bench/Driver.java
```diff
@@ -23,7 +23,6 @@
 import org.apache.commons.cli.HelpFormatter;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
-import org.apache.hadoop.conf.Configuration;
```
omalley: In general, you shouldn't fix imports in files that your patch isn't touching. Go ahead and leave them in this patch.

java/pom.xml
```diff
@@ -637,7 +637,7 @@
       <dependency>
         <groupId>junit</groupId>
         <artifactId>junit</artifactId>
-        <version>4.11</version>
+        <version>4.12</version>
```
omalley: Did you need this version change or was it an unassociated change? As with fixing the imports, you don't want to tie random changes into a single patch.

PR #145, opened 2017-08-01
ORC-221: RecordReaderImpl::evaluatePredicateMinMax should check wheth…

rajeshbalamohan:
ORC-221: RecordReaderImpl::evaluatePredicateMinMax should check whether predicate.getLiteralList() is empty


prasanthj:
+1

PR #149, opened 2017-08-08
ORC-224: Implement column writers of primitive types

wgtmac:
1. Implementation of ColumnWriters for float, double, tinyint, boolean, string, char, varchar, binary, date, timestamp and decimal types.
2. Only support RleVersion1. RleVersion2 and dictionary encoding for string types are not implemented in this patch.
3. Added tests in TestWriter.cc.

### DIFF
--- a/c++/src/ColumnWriter.cc
+++ b/c++/src/ColumnWriter.cc
@@ -179,14 +179,16 @@ namespace orc {
     notNullEncoder->recordPosition(rowIndexPosition.get());
   }
 
-  void ColumnWriter::resetIndex() {
-    // clear row index
-    rowIndex->clear_entry();
-    rowIndexEntry->clear_positions();
-    rowIndexEntry->clear_statistics();
+  void ColumnWriter::reset() {
+    if (enableIndex) {
+      // clear row index
+      rowIndex->clear_entry();
+      rowIndexEntry->clear_positions();
+      rowIndexEntry->clear_statistics();
 
-    // write current positions
-    recordPosition();
+      // write current positions
+      recordPosition();
+    }
   }
 
   class StructColumnWriter : public ColumnWriter {
@@ -222,7 +224,7 @@ namespace orc {
     virtual void writeIndex(
       std::vector<proto::Stream> &streams) const override;
 
-    virtual void resetIndex() override;
+    virtual void reset() override;
 
   private:
     std::vector<ColumnWriter *> children;
@@ -355,11 +357,11 @@ namespace orc {
     }
   }
 
-  void StructColumnWriter::resetIndex() {
-    ColumnWriter::resetIndex();
+  void StructColumnWriter::reset() {
+    ColumnWriter::reset();
 
     for (uint32_t i = 0; i < children.size(); ++i) {
-      children[i]->resetIndex();
+      children[i]->reset();
     }
   }
 
@@ -427,14 +429,16 @@ namespace orc {
     // update stats
     IntegerColumnStatisticsImpl* intStats =
       dynamic_cast<IntegerColumnStatisticsImpl*>(colIndexStatistics.get());
+    bool hasNull = false;
     for (uint64_t i = 0; i < numValues; ++i) {
       if (notNull == nullptr || notNull[i]) {
         intStats->increase(1);
         intStats->update(data[i], 1);
-      } else if (!intStats->hasNull()) {
-        intStats->setHasNull(true);
+      } else if (!hasNull) {
+        hasNull = true;
       }
     }
+    intStats->setHasNull(hasNull);
   }
 
   void IntegerColumnWriter::flush(std::vector<proto::Stream>& streams) {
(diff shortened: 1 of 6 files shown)


wgtmac:
@omalley @majetideepak Please take a look when you have time. Thanks!

majetideepak:
@wgtmac I will take a look at this today/tomorrow. Thanks!

majetideepak:
@wgtmac Sorry for taking longer. I will definitely try to complete this as soon as possible.

wgtmac:
@majetideepak No worries. Thanks for looking into this!

c++/src/ColumnWriter.cc
```diff
@@ -468,25 +472,1099 @@ namespace orc {
...
+                                      false,
+                                      rleVersion,
+                                      memPool);
+
+    if (enableIndex) {
+      recordPosition();
+    }
+  }
+
+  static int64_t formatNano(int64_t nanos) {
```
majetideepak: comments here will help.

c++/src/ColumnWriter.cc
```diff
@@ -468,25 +472,1099 @@ namespace orc {
...
+        // TimestampVectorBatch already stores data in UTC
+        int64_t millsUTC = secs[i] * 1000 + nanos[i] / 1000000;
+        tsStats->increase(1);
+        tsStats->update(millsUTC);
+      } else if (!tsStats->hasNull()) {
+        tsStats->setHasNull(true);
+      }
+    }
+
+    for (uint64_t i = 0; i < numValues; ++i) {
```
majetideepak: can this loop be fused with the above loop? Applicable to other Writers below as well?

c++/src/ColumnWriter.cc
```diff
@@ -468,25 +472,1099 @@ namespace orc {
...
+      if (!notNull || notNull[i]) {
+        dataStream->write(data[i], static_cast<size_t>(length[i]));
+      }
+    }
+
+    StringColumnStatisticsImpl* strStats =
+        dynamic_cast<StringColumnStatisticsImpl*>(colIndexStatistics.get());
+
+    bool hasNull = false;
+    for (uint64_t i = 0; i < numValues; ++i) {
```
majetideepak: Can this loop be fused with the above loop?

c++/src/ColumnWriter.cc
```diff
@@ -468,25 +472,1099 @@ namespace orc {
...
+      if (!notNull || notNull[i]) {
+        strStats->update(data[i],
+                         static_cast<size_t>(length[i]));
+        strStats->increase(1);
+      } else if (!hasNull) {
+        hasNull = true;
```
majetideepak: Using a separate `hasNull` variable approach is not being used for the `TimestampColumnWriter`. Will be nice to see be consistent.

c++/src/ColumnWriter.cc
```diff
@@ -468,25 +472,1099 @@ namespace orc {
...
+
+    const char* notNull = decBatch.hasNulls ?
+                          decBatch.notNull.data() + offset : nullptr;
+    const Int128* values = decBatch.values.data() + offset;
+
+    for (uint64_t i = 0; i < numValues; ++i) {
+      if (!notNull || notNull[i]) {
+        Int128 val = zigZagInt128(values[i]);
+        char* data = buffer;
+        while (true) {
```
majetideepak: some comments here will help.

c++/test/TestWriter.cc
```diff
@@ -209,5 +209,612 @@ namespace orc {
     }
     EXPECT_FALSE(rowReader->next(*batch));
   }
-}
 
+  TEST(Writer, writeStringAndBinaryColumn) {
```
majetideepak: These tests can definitely be improved by using a templated test class

majetideepak: See google typed tests. Those should help.

wgtmac: It is better to keep them in different test cases for following reasons:
1. It seems that they share the same logic in testing but they differ in the data construction and verification parts for different types. It has the same rationale with the design of ColumnWriter classes.
2. We may have different design/implementations for some types in the future. It is much easier to test them if we de-couple them into separate tests.

wgtmac:
@majetideepak Thanks for taking time to review my previous change! I have committed a new one based on your feedback. Please review them when you have time. Thanks!

wgtmac:
Hi @omalley. Please help review this patch when you get the chance, thanks!

wgtmac:
Hi @omalley @prasanthj , can anyone of you review and/or merge this patch as @majetideepak has already approved this a few days ago? Thanks!

c++/src/ColumnWriter.cc
```diff
@@ -468,25 +472,1099 @@ namespace orc {
...
+      recordPosition();
+    }
+  }
+
+  void BooleanColumnWriter::add(ColumnVectorBatch& rowBatch,
+                                uint64_t offset,
+                                uint64_t numValues) {
+    ColumnWriter::add(rowBatch, offset, numValues);
+
+    LongVectorBatch& byteBatch = dynamic_cast<LongVectorBatch&>(rowBatch);
```
jamesclampffer: It looks like this is following existing convention from the reader where return values for reinterpret_cast aren't checked but I'd think about adding an orc::InvalidArgument exception.  Then you can catch the std::bad_cast exception here (or check for nullptr when assigning to a pointer) and rethrow with a more descriptive message.

c++/src/ColumnWriter.cc
```diff
@@ -468,25 +472,1099 @@ namespace orc {
...
+              doubleData[i]));
+          for (size_t j = 0; j < bytes; ++j) {
+            data[j] = static_cast<char>(((*intBits) >> (8 * j)) & 0xff);
+          }
+        } else {
+          const int64_t* intBits =
+            reinterpret_cast<const int64_t*>(&(doubleData[i]));
+          for (size_t j = 0; j < bytes; ++j) {
+            data[j] = static_cast<char>(((*intBits) >> (8 * j)) & 0xff);
+          }
```
jamesclampffer: I think the intent of this and the block on line 731 would be a lot more clear if you made a helper function that took input and output pointers and a type width to handle an endian agnostic copy.

c++/src/ColumnWriter.cc
wgtmac: Done

c++/src/ColumnWriter.cc
wgtmac: Done

wgtmac:
@jamesclampffer @majetideepak @omalley I've updated the PR according to the feedback. Please review it again when you have time, thanks!

wgtmac:
@majetideepak Just a reminder that this one needs your help. Thanks!

majetideepak:
@wgtmac Sorry for the delay. I was tied up last week. I will definitely work on this today/tomorrow.

PR #151, opened 2017-08-10
ORC-226 Support getWriterId in c++ reader interface

xndai:
Add new interface for reader to retrieve writer ID.

Change-Id: I87939e448aa5eab1bc7ed728404ddbf41334d809

### DIFF
--- a/c++/include/orc/Reader.hh
+++ b/c++/include/orc/Reader.hh
@@ -288,6 +288,18 @@ namespace orc {
     virtual uint64_t getCompressionSize() const = 0;
 
     /**
+     * Get ID of writer that generated the file.
+     * @return UNKNOWN_WRITER if the writer ID is undefined
+     */
+    virtual WriterId getWriterId() const = 0;
+
+    /**
+     * Get the writer id value when getWriterId() returns an unknown writer.
+     * @return the integer value of the writer ID.
+     */
+    virtual int getWriterIdValue() const = 0;
+
+    /**
      * Get the version of the writer.
      * @return the version of the writer.
      */
--- a/c++/include/orc/Common.hh
+++ b/c++/include/orc/Common.hh
@@ -63,6 +63,13 @@ namespace orc {
     std::string toString() const;
   };
 
+  enum WriterId {
+    ORC_JAVA_WRITER = 0,
+    ORC_CPP_WRITER,
+    PRESTO_WRITER,
+    UNKNOWN_WRITER = INT32_MAX
+  };
+
   enum CompressionKind {
     CompressionKind_NONE = 0,
     CompressionKind_ZLIB = 1,
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -417,6 +417,26 @@ namespace orc {
     return footer->numberofrows();
   }
 
+  WriterId ReaderImpl::getWriterId() const {
+    WriterId id = WriterId::UNKNOWN_WRITER;
+    if (footer->has_writer()) {
+      id = static_cast<WriterId>(footer->writer());
+      if (id != WriterId::ORC_JAVA_WRITER &&
+          id != WriterId::ORC_CPP_WRITER &&
+          id != WriterId::PRESTO_WRITER) {
+        id = WriterId::UNKNOWN_WRITER;
+      }
+    }
+    return id;
+  }
+
+  int ReaderImpl::getWriterIdValue() const {
+    if (footer->has_writer()) {
+      return static_cast<int>(footer->writer());
+    }
+    throw std::logic_error("Writer id is not availiable.");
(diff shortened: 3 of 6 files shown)


c++/include/orc/Reader.hh
```diff
@@ -288,6 +288,17 @@ namespace orc {
...
     /**
+     * Get ID of writer that generated the file.
+     * Current availiable Orc writers:
+     * 0 = ORC Java
+     * 1 = ORC C++
+     * 2 = Presto
+     * @param id out parameter for writer id
+     * @return true if writer id is availiable, false if otherwise
+     */
+    virtual bool getWriterId(uint32_t & id) const = 0;
```
omalley: If it isn't there, it means ORC Java. Let's make an enum that matches this:

    enum WriterId { ORC_Java = 0, ORC_CPP = 1, ORC_Presto = 2, ORC_FUTURE = INT_MAX }

Any unknown code should return ORC_FUTURE.

   virtual WriterId getWriterId() const = 0;

xndai: Will there be any files generated by old Presto writer without an ID? Also if we choose to return ORC_FUTURE, there's no way to determine the actual writer ID.

omalley: No, it won't. Dain from the Presto team was the one that suggested adding the field.

omalley: Re-reading your question, I should clarify that the Presto writer hasn't been released yet.

xndai: Ok, I see. But I am still not sure about returning ORC_FUTURE. The caller may want to obtain the exact writer ID for debugging purpose. It can choose to ignore if it wants. We just give them one more option. What do you think?

omalley: Maybe we should print an error message with the actual id number if it isn't recognized. That would let you debug.

xndai: Then we need a logger in Orc library which we don't have now. What's the problem of returning actual value instead of ORC_FUTURE?

xndai: @omalley can we close on this by just returning the actual code? Or you have other concerns? Thx.

omalley: Actually, we do have the error stream, but we don't need to use it. I'd prefer not to use side effects in the parameters, because it is easy to misread in the code.

For the code and most users, the important part is which writer it is if known. Only diagnostic tools will care about the integer value of unknown ones. How about:

    enum WriterId { ORC_JAVA_WRITER, ORC_CPP_WRITER, PRESTO_WRITER, UNKNOWN_WRITER}

    class Reader {
    
       WriterId getWriterId();
       /**
         * For unknown writer ids, get the value.
         */
       int getUnknownWriterIdValue();

xndai: ok, that sounds good. I will update it in my next change.

xndai:
@omalley please take another look. thanks.

xndai:
Can someone help review and close on this? Thx. @omalley @majetideepak

c++/include/orc/Reader.hh
```diff
@@ -288,6 +288,18 @@ namespace orc {
...
+     * Get ID of writer that generated the file.
+     * @return UNKNOWN_WRITER if the writer ID is undefined
+     */
+    virtual WriterId getWriterId() const = 0;
+
+    /**
+     * Get the writer id value when getWriterId() returns an unknown writer.
+     * @return the integer value of the writer ID.
+     */
+    virtual int getUnknownWriterIdValue() const = 0;
```
majetideepak: Should this be `getWriteIdValue`? I think the implementation returns both known and unknown ids?

xndai: Sounds good. I will update it. Thanks.

xndai:
@majetideepak pls take another look. Thx!

majetideepak:
@xndai  can you please squash your commits?

xndai:
@majetideepak how do I do that? :)

ajayyadava:
@xndai You can use the following command to squash your last K commits into 1:
git reset --soft HEAD~K && git commit
e.g. to squash last 2 commits into one, replace K with 2
git reset --soft HEAD~2 && git commit

I will advise that you create a separate branch from the branch you want to commit and try it on that first.

xndai:
Squash commit. Thanks @ajayyadava @majetideepak

omalley:
Ok, this is generally good. A couple of points that I'll fix as part of committing:

* the API should use specific types (eg. uint32_t instead of int)
* since the enums are for serialization, I think it will be clearer if we assign explicit values for them
* if the footer doesn't have a writerId, it is ORC_JAVA_WRITER and not an exception

PR #154, opened 2017-08-11
ORC-228 Make MemoryManagerImpl.ROWS_BETWEEN_CHECKS configurable

ekoifman:


java/core/src/java/org/apache/orc/impl/MemoryManagerImpl.java
```diff
@@ -81,6 +81,7 @@ public Thread getOwner() {
    */
   public MemoryManagerImpl(Configuration conf) {
     double maxLoad = OrcConf.MEMORY_POOL.getDouble(conf);
+    ROWS_BETWEEN_CHECKS = OrcConf.ROWS_BETWEEN_CHECKS.getLong(conf);
```
prasanthj: validate values here? and set default value in case of invalid values and maybe log?

ekoifman:
@prasanthj  could you review please

java/core/src/java/org/apache/orc/impl/MemoryManagerImpl.java
ekoifman: default is set OrcConf - 5000
what would be reasonable range of valid values?  [1,?]

prasanthj: I don't think we want to support too larger interval for this. Having very high value means prolonging the memory check which is bad (flush often as opposed to don't flush and fail). 
1 to 10000 may be good range. Also please make a note in the description that keeping too low value is for testing only and can cause too early flushes in some cases and generate sub-optimal orc files.

ekoifman:
added validation, fixed comments

PR #159, opened 2017-08-18
ORC-175: ORC-232: add jmh-generator-annprocess in pom.xml. improve zlib perfomance using isal api

iamhumanbeing:
after add jmh-generator-annprocess in pom.xml, BenchmarkList was created


java/bench/src/java/org/apache/orc/bench/CompressionKind.java
```diff
@@ -53,6 +54,8 @@ public OutputStream create(OutputStream out) throws IOException {
         return new GZIPOutputStream(out);
       case SNAPPY:
         return new SnappyCodec().createOutputStream(out);
+      case ISAL:
```
t3rmin4t0r: ISAL is just the library for gzip? That doesn't need a new codec - have you tried LD_PRELOAD to load ISAL instead of libz?

iamhumanbeing: 1. igzip is only part of the ISAL.
2. igzip only support level 0 - level 1 compression, it can not replace libz totally.
3. igzip's API can not replace ligz's API directly.

10110346:
LGTM

omalley:
Is the ISA-L zlib support sufficient to read and write the ORC files with zlib compression? I agree with Gopal that it doesn't feel like a separate compression codec.

I'm don't think it is a good idea to build in support for proprietary compression formats. If it is just an optimization on performance, that might be workable. As a unique codec, it isn't.

iamhumanbeing:
@omalley: How about we add a compiling option which use ISA-L ZLIB support for ZLIB compression&decompression? this option is just an optimization on performance.

xndai:
@iamhumanbeing did you compare it with zstd? Based on my experience, zstd is way better than igzip. I would expect a similar result with ISA-L. It doesn't seem to be adding a lot of value if we plan to support zstd in near future.

PR #160, opened 2017-08-21
ORC-233 Allow `orc.include.columns` to be empty

ajayyadava:


java/mapreduce/src/java/org/apache/orc/mapred/OrcInputFormat.java
```diff
@@ -58,7 +58,7 @@
    */
   public static boolean[] parseInclude(TypeDescription schema,
                                        String columnsStr) {
-    if (columnsStr == null ||
+    if (StringUtils.isBlank(columnsStr) ||
```
dongjoon-hyun: BTW, is this enough for ORC column selection to return all null fields?
It looks to me that it returns all value fields. I mean without column selection.

ajayyadava: Hi @dongjoon-hyun,

Thank you for reviewing the patch and pointing me in the right direction. I had earlier fixed only the exception. Based on your comment, a bit of reading through the code and the spark test that you have linked in the JIRA my understanding is that the desired behavior is to not select/project any columns when the `orc.include.columns` is a blank string. I have added a test case for what I understand is the desired behavior. 

I have left the behavior for null(default) value to be as before which seems bit inconsistent(null & blank should not be different) but I am not sure on why it was changed and if it's ok to modify that behavior.

I am completely new to ORC and this is my first patch so my understanding might be completely off. Please feel free to point me in the right direction. Thanks!

java/mapreduce/src/test/org/apache/orc/mapreduce/TestMapreduceOrcOutputFormat.java
```diff
@@ -153,11 +153,55 @@ public void testColumnSelection() throws Exception {
...
+    }
+    assertEquals(false, reader.nextKeyValue());
+  }
+
+
+
+    /**
+     * Make sure that the writer ignores the OrcKey
+     * @throws Exception
+     */
```
dongjoon-hyun: Could you remove the redundant blank here in line 201 ~ 204?

java/mapreduce/src/test/org/apache/orc/mapreduce/TestMapreduceOrcOutputFormat.java
```diff
@@ -153,11 +153,55 @@ public void testColumnSelection() throws Exception {
...
+  public void testColumnSelectionBlank() throws Exception {
+    String typeStr = "struct<i:int,j:int,k:int>";
+    OrcConf.MAPRED_OUTPUT_SCHEMA.setString(conf, typeStr);
+    conf.set("mapreduce.output.fileoutputformat.outputdir", workDir.toString());
+    conf.setInt(OrcConf.ROW_INDEX_STRIDE.getAttribute(), 1000);
+    conf.setBoolean(OrcOutputFormat.SKIP_TEMP_DIRECTORY, true);
+    TaskAttemptID id = new TaskAttemptID("jt", 0, TaskType.MAP, 0, 1);
+    TaskAttemptContext attemptContext = new TaskAttemptContextImpl(conf, id);
+    OutputFormat<NullWritable, OrcStruct> outputFormat =
+            new OrcOutputFormat<OrcStruct>();
```
dongjoon-hyun: nit: indentation? Maybe, 4 blanks instead of 8.

java/mapreduce/src/test/org/apache/orc/mapreduce/TestMapreduceOrcOutputFormat.java
```diff
@@ -153,11 +153,55 @@ public void testColumnSelection() throws Exception {
...
+    RecordWriter<NullWritable, OrcStruct> writer =
+            outputFormat.getRecordWriter(attemptContext);
```
dongjoon-hyun: ditto.

java/mapreduce/src/test/org/apache/orc/mapreduce/TestMapreduceOrcOutputFormat.java
```diff
@@ -153,11 +153,55 @@ public void testColumnSelection() throws Exception {
...
+      row.setFieldValue(0, new IntWritable(r));
+      row.setFieldValue(1, new IntWritable(r * 2));
+      row.setFieldValue(2, new IntWritable(r * 3));
+      writer.write(nada, row);
+    }
+    writer.close(attemptContext);
+
+    conf.set(OrcConf.INCLUDE_COLUMNS.getAttribute(), "");
+    FileSplit split = new FileSplit(new Path(workDir, "part-m-00000.orc"),
+            0, 1000000, new String[0]);
```
dongjoon-hyun: indentation.

java/mapreduce/src/test/org/apache/orc/mapreduce/TestMapreduceOrcOutputFormat.java
```diff
@@ -153,11 +153,55 @@ public void testColumnSelection() throws Exception {
...
+    RecordReader<NullWritable, OrcStruct> reader =
+            new OrcInputFormat<OrcStruct>().createRecordReader(split,
```
dongjoon-hyun: ditto.

java/mapreduce/src/test/org/apache/orc/mapreduce/TestMapreduceOrcOutputFormat.java
```diff
@@ -153,11 +153,55 @@ public void testColumnSelection() throws Exception {
...
+                    attemptContext);
+    // the sarg should cause it to skip over the rows except 1000 to 2000
+    for (int r = 0; r < 3000; ++r) {
+      assertEquals(true, reader.nextKeyValue());
+      row = reader.getCurrentValue();
+      assertEquals(null, ((IntWritable) row.getFieldValue(0)));
+      assertEquals(null, row.getFieldValue(1));
+      assertEquals(null, ((IntWritable) row.getFieldValue(2)));
```
dongjoon-hyun: This test case looks good to me.

dongjoon-hyun:
Hi, @omalley .
Could you review this PR?

PR #162, opened 2017-08-22
ORC-231 Configurable capability to overwrite the file if it exists.

ajayyadava:
Added a boolean flag to configure the overwrite behavior. This change is backward compatible as the default value of the flag is false.


omalley:
I undid the inadvertent change to the import lines. Please configure your IDE to not introduce wildcard imports for this project.

PR #163, opened 2017-08-29
ORC-162. Handle 0 byte files as empty ORC files.

omalley:
Treat 0 byte files as an empty ORC file with schema of struct<>.

### DIFF
--- a/java/mapreduce/src/java/org/apache/orc/mapred/OrcInputFormat.java
+++ b/java/mapreduce/src/java/org/apache/orc/mapred/OrcInputFormat.java
@@ -24,6 +24,7 @@ import com.esotericsoftware.kryo.io.Output;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.hive.ql.io.sarg.SearchArgument;
 import org.apache.hadoop.hive.ql.io.sarg.SearchArgumentImpl;
 import org.apache.hadoop.io.WritableComparable;
@@ -32,6 +33,7 @@ import org.apache.hadoop.mapred.FileSplit;
 import org.apache.hadoop.mapred.RecordReader;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 
 import org.apache.hadoop.io.NullWritable;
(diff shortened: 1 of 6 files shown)


dain:
We were considering doing this internally and then we ran into a production bug where files got truncated to zero bytes.  Since empty files are illegal we could find all of the effected partitions easily, but without this you would be stuck.  A good work around would be to require empty files to end with ".empty" but I'm not sure you can do that with M/R.

dain:
Also this is a backwards incompatible change, so we would, at the very least, need to do the trick where it is disabled by default in the writer until the reader is rolled out everywhere.

omalley:
The problem is that Hive is doing this across the board. See HIVE-13040.

Making the reader not throw is ok, if slightly incompatible. This patch doesn't change the writer to write such files.

electrum:
We had the same use case of making empty bucket creation more efficient. Encoding the fact that the file is intentionally empty in the name provides a good safeguard against storage system problems that can cause files to be truncated (unfortunately far too common).

omalley:
I agree that a filename encoding would be a nice safe guard, but it doesn't work since Hive isn't using that convention. (Hive made the change in Hive's OrcInputFormat so it didn't move over to the ORC project.)

prasanthj:
Hive creates empty files only for MR to support bucketed joins. Tez doesn't create empty bucket files anymore. Hive currently discards empty files during split generation. We can do similar thing in Orc's version of OrcInputFormat (or add EmptyFilePathPattern to ignore 0 length files or files <= MAGIC.length). Creating splits for empty is anyway useless. As far as calling the Reader's directly with a empty file path, we can treat it as empty file with struct<>.

omalley:
@prasanthj There are customers out there with millions of zero byte ORC files in their Hive warehouses. We need to have the reader not throw when they read them with Spark, etc. Rather than patch each context where Readers may be created, I'd rather fix the core Reader.

prasanthj:
I agree that reader should gracefully handle 0 length files like what this patch does instead of throwing. In addition to that we should also avoid creating splits for 0 length files. Spinning up tasks for reading 0 length files is wasteful.

omalley:
@prasanthj Ok, I added the code that will cause the ORC input formats to not generate splits for the empty files.

java/mapreduce/src/java/org/apache/orc/mapred/OrcInputFormat.java
```diff
@@ -151,4 +153,26 @@ public static void setSearchArgument(Configuration conf,
...
+  /**
+   * Filter out the 0 byte files, so that we don't generate splits for the
+   * empty ORC files.
+   * @param job the job configuration
+   * @return a list of files that need to be read
+   * @throws IOException
+   */
+  protected FileStatus[] listStatus(JobConf job) throws IOException {
+    FileStatus[] result = super.listStatus(job);
+    List<FileStatus> ok = new ArrayList<>(result.length);
```
prasanthj: Instead of checking this after retrieving all FileStatus objects, it will be better if a PathFilter can be passed to listStatus() so that we will only get non-zero files. Getting 1000s of 0 length files and filtering here seems wasteful.

omalley: That would involve doing a second getStatus. Filters are only given the name.

prasanthj: Make sense. Just noticed filter gets applied after listStatus anyway.

PR #166, opened 2017-09-05
ORC-91: Use hdfs v-blocks instead of zero-padding stripes

omalley:
Ok, this is my rebasing of Gopal's patch to use the zero padding after my patch for ORC-234.


omalley:
Changes:
* Renamed HadoopShims_2_2 to HadoopShimsPre2_3 and moved to shims module.
* Removed TextReaderShim since we aren't using it any more.
* Pulled the padding method into the HadoopShim as padStreamToBlock.
* Added HadoopShimsPre2_7.
* Refactored a bit so that the shim classes share most of the code, while making sure that older shims never reference new shims.
* Fixed some broken javadoc.

PR #168, opened 2017-09-16
Install missing Statistics.hh header file

halayli:
After installing orc, I couldn't include any of the header files because Statistics.hh is missing in the installed include directory.


majetideepak:
Can you file a JIRA and append the JIRA number to the PR? Thanks.

omalley:
Since it was a simple fix, I went ahead and created the jira for it.

PR #169, opened 2017-09-18
[WIP] ORC-203 Modify the StringStatistics to trim the minimum and maximum v…

ajayyadava:
…alues.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
@@ -21,6 +21,7 @@ import java.sql.Date;
 import java.sql.Timestamp;
 import java.util.TimeZone;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.hive.serde2.io.HiveDecimalWritable;
 import org.apache.hadoop.hive.common.type.HiveDecimal;
 import org.apache.hadoop.hive.serde2.io.DateWritable;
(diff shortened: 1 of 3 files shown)


ajayyadava:
To make sure that maximum and minimum are always correct, we need to store intermediate large strings and trim them only at the end. 

If anyone has a better approach in mind, then please do let me know.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -608,8 +609,23 @@ public void merge(ColumnStatisticsImpl other) {
       OrcProto.StringStatistics.Builder str =
         OrcProto.StringStatistics.newBuilder();
       if (getNumberOfValues() != 0) {
-        str.setMinimum(getMinimum());
-        str.setMaximum(getMaximum());
+        String value = getMinimum();
+        if (value != null && value.length() > 1024) {
```
sadikovi: I suggest you make it a constant somewhere and reference that.

I would also consider making it configurable with default value of `1024`.

xndai:
I don't understand why you need to trim the strings. Protobuf doesn't support strings over 1024 characters?

dain:
@xndai if the min or max happens to be a multi megabyte value it can be really expensive for the reader.  Additionally, for filtering the first few bytes are the most valuable (they establish the range).

xndai:
@dain if the concern is the performance, should we also limit the string length when generate stats in writer path, which in my opinion is more costly? I think if we keep the string compare up to 1024 bytes, we won't need to trim at the end.

omalley:
Ok, starting with the representation. I'd suggest it look like:

    message StringStatistics {
      optional string minimum = 1;
      optional string maximum = 2;
      // sum will store the total length of all strings in a stripe
      optional sint64 sum = 3;
      // If the minimum or maximum value was longer than 1024 bytes, store a lower or upper
      // bound instead of the minimum or maximum values above.
      optional string lowerBound = 4;
      optional string upperBound = 5;
    }

Now obviously the lowerBound can just be the string truncated (at a utf8 character boundary!) to at most 1024 bytes. The upperBound is the same with the last code point increased by one.

In the StringStatisticsImpl, I'd keep two boolean flags as to whether it is a real value or an approximation for minimum or maximum. The value comparison is the same, since unless the current value is less than the lower bound, it won't change the lower bound and the same is true for the upper bound. If the new minimum/maximum is not truncated, the corresponding wasTruncated flag should be cleared. When merging, the flag follows the value. In the corner case of two identical values where one was truncated, the non-truncated one is the result.

We should end up with four methods for each:
* String getMinimum();
* String getLowerBound();
* String getMaximum();
* String getUpperBound();

If we only have a lower bound, getMinimum should be null and the same with upper bound and getMaximum. getLowerBound and getUpperBound should match getMinimum and getMaximum, if no truncation was done.

ajayyadava:
Thank you, everyone, for your comments! Sorry for being away for a while. I will start working on incorporating the feedback this week.

PR #178, opened 2017-10-14
ORC-254: Add badge for latest released version to README.md

prasanthj:


README.md
```diff
@@ -25,6 +25,9 @@ The current build status:
 ![master build status](https://travis-ci.org/apache/orc.svg?branch=master)</a>
 * <a href="https://travis-ci.org/apache/orc/pull_requests">Pull Requests</a>
 
+Latest release: <a href="http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.apache.orc%22">
```
omalley: We probably should link to https://orc.apache.org/docs/releases.html instead. 

We could change the link to maven central from the release description, which is here https://github.com/apache/orc/blob/master/site/_includes/release_description.md

PR #182, opened 2017-10-24
ORC-259. Make RAT and findbugs optional during the build.

omalley:


alanfgates:
Major +1.  This will speed up the build nicely yet still make sure both rat and findbugs are run before code is committed.

PR #185, opened 2017-11-01
ORC-261: [C++] Fix installation to include all header files

jcrist:
Addresses [issue 261](https://issues.apache.org/jira/projects/ORC/issues/ORC-261).

- Fixes installation to include all necessary header files in `include/orc`
- Removes a few unnecessary includes

### DIFF
--- a/tools/src/CMakeLists.txt
+++ b/tools/src/CMakeLists.txt
@@ -10,10 +10,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# TODO:
+# - orc-metadata relies on the protobuf routines, meaning protobuf and
+#   binary_dir/c++/src still need to be included
+# - timezone-dump relies on non-public timezone routines. I *think* this
+#   executable can just be removed, as it looks like it was written for testing
+#   alone.
+
 include_directories (
   ${PROJECT_SOURCE_DIR}/c++/include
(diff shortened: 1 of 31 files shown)


tools/src/CMakeLists.txt
```diff
@@ -10,10 +10,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# TODO:
+# - orc-metadata relies on the protobuf routines, meaning protobuf and
+#   binary_dir/c++/src still need to be included
+# - timezone-dump relies on non-public timezone routines. I *think* this
+#   executable can just be removed, as it looks like it was written for testing
+#   alone.
```
jcrist: If these issues can be resolved, then the `include_directories` below can just be:

```
include_directories (
   ${PROJECT_SOURCE_DIR}/c++/include
   ${PROJECT_BINARY_DIR}/c++/include
)
```

meaning the executables use the same interface a user of this library would get. Whether that means the protobuf headers are included in the build, or the interface is expanded to make `orc-metadata` not rely on the protobuf headers directly I'm not sure.

Either way this probably doesn't need to be done now, it's just a potential hygiene thing.

jcrist:
Note that this is the first time I've touched C++ code outside undergrad, so there's a decent chance I've done something wrong here. Everything seems to pass locally for me though.

omalley:
+1 This looks good.

PR #186, opened 2017-11-03
ORC-187: Simplify BitFieldReader to only support single bits

rajeshbalamohan:


java/core/src/java/org/apache/orc/impl/BitFieldReader.java
```diff
@@ -162,53 +92,32 @@ public void seek(PositionProvider index) throws IOException {
...
+    final int availableBits = 8 - (currentIdx + 1);
+    if (totalBits <= availableBits) {
+      currentIdx += totalBits;
     } else {
-      totalBits -= bitsLeft;
-      input.skip(totalBits / 8);
-      current = input.next();
-      bitsLeft = (int) (8 - (totalBits % 8));
+      final long bitsToSkip = (totalBits - availableBits);
+      input.skip(Math.min(1, bitsToSkip / 8));
```
t3rmin4t0r: Looks a bit odd - the min(1) might be a problem if 

availableBits = 0 and items=3, it will end up skip(1) where it should really be doing skip(0).

t3rmin4t0r:
LGTM - +1

omalley:
Since this is removing the old functionality, you should remove the old constructor that takes the bitWidth. With this change, someone who passed in a non-one width would be surprised with the result.

t3rmin4t0r:
@asfgit merge

PR #189, opened 2017-11-08
ORC-210 Add new encodings and benchmarks for new double encoding.

omalley:
This extends Teddy's pull request by adding:

    * Extended write benchmark.
    * Added read benchmark
    * Added new encodings:
        + plainV2, which is the same data format but faster implementation
        + fpcV2, which is the standard FPC implementation
        + flip, which rotates the bytes in a 8x8 matrix
        + split, which creates rle streams for sign, exponent, and mantissa
    * Added new datasets.


omalley:
The sizes of the different data sets in the different compression:

https://drive.google.com/file/d/11bzDaqJiL7CQDg7TVo1PvQqXauj_DgIQ/view?usp=sharing

java/bench/src/java/org/apache/orc/bench/floating/DoubleReadBenchmark.java
```diff
@@ -0,0 +1,251 @@
...
+          dataSet, compression);
+      DataInputStream inStream = new DataInputStream(new FileInputStream(path));
+      int footerSize = inStream.readInt();
+      byte[] footerBuffer = new byte[footerSize];
+      inStream.readFully(footerBuffer);
+      footer = OrcProto.StripeFooter.parseFrom(footerBuffer);
+      streams.clear();
+      for(OrcProto.Stream stream: footer.getStreamsList()) {
+        StreamName name = new StreamName(stream.getColumn(), stream.getKind());
+        CompressionCodec codec = WriterImpl.createCodec(compression);
```
t3rmin4t0r: Looks like this is using the old ZLIB codecs.

CompressionCodec codec = getCustomizedCodec(OrcProto.Stream.Kind.DATA); ?

omalley:
Here is a spread sheet that is sorted by data set and then by read+hdfs time. The read+hdfs time assumes 15mb/sec from hdfs.

https://docs.google.com/spreadsheets/d/1bE1j-AaUY7Xq_uh1nqX1Jf7qXvfL1crp2Y8Hl2M4Y-E/edit?usp=sharing

omalley:
@xndai the units for the sizes are bytes / 1000 values so uncompressed is 8000.

The time units are ms / 1m values .

winningsix:
@omalley  Thank you for the patch and initial performance result. The result looks very promising. With current result, looks like different encoding (FPC V1, FPC V2 and split) fits for different data set while split encoding is more compression friendly.  Not sure whether we can implement it as what RLE did. Dynamically determine the encoding based on the characteristics of data set. Any thoughts on this?

PR #190, opened 2017-11-15
ORC-264 schema evolution columns name comparising is case unaware

piyush-mukati:


owen-hortonworks:
Unfortunately, this would break the non-sql clients. Looking more into Hive's behavior, it currently does:
* lowercases all column names
* preserves all struct field names

That makes a mess. I'd propose that we add a new option for readers that makes schema evolution lowercase everything and Hive should always set that option.

piyush-mukati:
@owen-hortonworks  I have added the suggested change, please review.

PR #191, opened 2017-11-15
ORC-265: [C++] Add documentation for C++ build support

majetideepak:
I verified the modified page by building locally.

### DIFF
--- a/site/_docs/building.md
+++ b/site/_docs/building.md
@@ -70,3 +70,20 @@ To build:
 % mvn package
 ~~~
 
+## Building just C++
+
+~~~ shell
+% mkdir build
+% cd build
+% cmake .. -DBUILD_JAVA=OFF
+% make package test-out
+~~~
+
+## Specify third-party libraries for C++ build
+
+~~~ shell
+% mkdir build
+% cd build
+% cmake .. -DSNAPPY_HOME=<PATH> -DZLIB_HOME=<PATH> -DLZ4_HOME=<PATH> -DGTEST_HOME=<PATH> -DPROTOBUF_HOME=<PATH>
+% make package test-out
+~~~


site/_docs/building.md
```diff
@@ -70,3 +70,25 @@ To build:
...
+~~~ shell
+% mkdir build
+% cd build
+% GTEST_HOME=<PATH to Google Test library> \
+  SNAPPY_HOME=<PATH to Snappy library> \
+  ZLIB_HOME=<PATH to ZLIB library> \
+  LZ4_HOME=<PATH to LZ4 library> \
+  PROTOBUF_HOME=<PATH to Protobuf library> \
+  cmake  .. -DBUILD_JAVA=OFF
+% make package test-out
```
owen-hortonworks: I'd suggest using:

cmake .. -DSNAPPY_HOME=&lt;PATH&gt; -DZLIB_HOME=&lt;PATH&gt; ...

to be more consistent.

majetideepak: Hi Owen, the current setup depends on `<LIBNAME>_HOME` being environmental variables and not CMake variables.

wgtmac: I need to explicitly set OPENSSL_ROOT_DIR on my mac, otherwise I will end up with the following error message:

CMake Error at /Applications/CMake.app/Contents/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR (missing: OPENSSL_INCLUDE_DIR)
Call Stack (most recent call first):
  /Applications/CMake.app/Contents/share/cmake-3.7/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /Applications/CMake.app/Contents/share/cmake-3.7/Modules/FindOpenSSL.cmake:385 (find_package_handle_standard_args)
  CMakeLists.txt:41 (find_package)

majetideepak: @wgtmac  That is correct. We manually set it in travis-ci testing as well.
 ``cmake -DOPENSSL_ROOT_DIR=`brew --prefix openssl` ..``
I will check with Anatoli Shein if we can fix this / or add documentation here. Thanks!

owen-hortonworks: I'm using cmake 3.5.1, but it works for me. Why would you need environment variables instead of cmake variables?

owen-hortonworks: Actually, looking at the ThirdpartyToolchain.cmake, it goes the other way. It is looking at the environment variable and setting the cmake variable. Setting the cmake variable directly works fine.

majetideepak: That is correct. The recent changes made did put the dependency on CMake variables. Fixed the comments.

PR #192, opened 2017-11-21
ORC-266: [C++] Cleanup cmake scripts

jcrist:
This was an attempt to fix using local versions of thirdparty libraries (as added in https://github.com/apache/orc/pull/170) so that installing doesn't pull in every file matching not matching `*.so|*.dylib` in the containing `lib` folder (as it does now). However, it kind of spiraled out in scope as more issues developed. Apologies for dumping this here without prior discussion.

---

A summary of the backwards compatible changes:

- Cleanup handling of `find_package` scripts to:
  - Set consistent variable names, and also match conventions in other apache projects
  - Set consistent variable names in both `*_HOME` specified and vendored paths
  - Provide error messages when not found, and only search when `*_HOME` is specified
  - Allow specifying `*_HOME` paths in other cmake projects using orc as an `ExternalProject`. This allows the wrapping project to more easily specify location of third party libraries with orc (e.g. `snappy`).

- Add an option to not build the cpp tests, and only install/search for gtest if turned on

- Add an option to not build the tools

---

A few potentially more contentious changes:

- Don't install third party libraries, even when vendored. The current behavior installs the thirdparty libs even if not vendored, which feels wrong to me. It also accidentally copies every file in the `lib` directory that isn't a shared library, which is definitely a bug.

  At a minimum, the thirdparty libraries that aren't vendored shouldn't be installed, and the accidental copying fixed. I'd argue that no thirdparty library should be installed, as it may clobber existing versions of e.g. `protobuf.a`. Downstream projects may handle the linking and distribution of thirdparty libraries themselves.

  If people disagree, perhaps a flag `INSTALL_THIRDPARTY_LIBS=on` could be added, to allow downstream projects to turn this off?

- Don't install `NOTICE` and `LICENSE`. When running `make install` currently, these get installed in the root directory, which is not where they should go. Since the header files contain license headers already, and other apache projects don't install `NOTICE` and `LICENSE` when running `make install`, I think this should be fine?

Addresses https://issues.apache.org/jira/browse/ORC-266.


jcrist:
As an example, in a downstream c++ project I'm building `liborc.a` using the following `ExternalProject` setup:

```python
set (ORC_CMAKE_ARGS -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
                    -DCMAKE_INSTALL_PREFIX=${ORC_PREFIX}
                    -DBUILD_LIBHDFSPP=OFF
                    -DBUILD_JAVA=OFF
                    -DBUILD_TOOLS=OFF
                    -DBUILD_CPP_TESTS=OFF
                    -DPROTOBUF_HOME=${PROTOBUF_HOME}
                    -DLZ4_HOME=${LZ4_HOME}
                    -DSNAPPY_HOME=${SNAPPY_HOME}
                    -DZLIB_HOME=${ZLIB_HOME})

ExternalProject_Add(orc_ep
                    GIT_REPOSITORY "https://github.com/apache/orc"
                    GIT_TAG ${ORC_VERSION}
                    BUILD_BYPRODUCTS ${ORC_STATIC_LIB}
                    CMAKE_ARGS ${ORC_CMAKE_ARGS})
```

where `protobuf`, `lz4`, `snappy`, and `zlib` are all already existing dependencies.

jcrist:
Ping @majetideepak for discussion and review.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -10,19 +10,46 @@
...
 # ----------------------------------------------------------------------
 # Snappy
 
-set (SNAPPY_HOME "$ENV{SNAPPY_HOME}")
-find_package (Snappy)
-if (NOT SNAPPY_FOUND)
+if (NOT "${SNAPPY_HOME}" STREQUAL "")
+  find_package (Snappy REQUIRED)
+  set(SNAPPY_VENDORED FALSE)
+else ()
```
majetideepak: I would like the library to be vendored in the case where `SNAPPY_HOME` is set, but `SNAPPY_FOUND` is inferred as `false`

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -41,32 +68,35 @@ if (NOT SNAPPY_FOUND)
...
+  set (SNAPPY_VENDORED TRUE)
 endif ()
-include_directories (SYSTEM ${SNAPPY_INCLUDE_DIRS})
+
+include_directories (SYSTEM ${SNAPPY_INCLUDE_DIR})
 add_library (snappy STATIC IMPORTED)
 set_target_properties (snappy PROPERTIES IMPORTED_LOCATION ${SNAPPY_STATIC_LIB})
-set (SNAPPY_LIBRARIES snappy)
-add_dependencies (snappy snappy_ep)
-install(DIRECTORY ${SNAPPY_PREFIX}/lib DESTINATION .
```
majetideepak: AFAIK `cpack` depends on `install` as well to make a package. So I would vote for your other option of using `INSTALL_THIRDPARTY_LIBS=on`.

cmake_modules/ThirdpartyToolchain.cmake
jcrist: As I wrote it, we error out if `SNAPPY_HOME` is set but not found (since `REQUIRED` is set). I think this makes more sense - as a user if I specify `SNAPPY_HOME` I'd rather get an error than silently download and build a new version if it fails to find it.

majetideepak: Ok. Makes sense. Thanks.

cmake_modules/ThirdpartyToolchain.cmake
jcrist: Done. Went with `INSTALL_VENDORED_LIBS`, which defaults to on.

jcrist:
Opened issue https://issues.apache.org/jira/browse/ORC-266. I think this should be good now?

PR #193, opened 2017-11-27
ORC-267: INSTALL NOTICES and LICENSE to binary tarball

majetideepak:


CMakeLists.txt
```diff
@@ -105,6 +105,11 @@ INCLUDE(ThirdpartyToolchain)
 set (EXAMPLE_DIRECTORY ${CMAKE_SOURCE_DIR}/examples)
 
 add_subdirectory(c++)
+
+install(
+  FILES LICENSE NOTICE
+  DESTINATION .)
```
jcrist: If I do `make install`, these just get placed in my root directory. Other apache projects (e.g. arrow, parquet-cpp) don't install the `LICENSE` or `NOTICE` (the header files all have the license information), so why is this needed? This is why I removed these in #192, sorry if there was a good reason for installing these.

omalley:
Legally, the binary artifacts from the Apache projects are supposed to include them. For tarballs it is traditionally at the root of the project. Especially for RPM and DEB packaging, you wouldn't want them at the root. 

How about we move it to doc/orc/{NOTICE,LICENSE}?

jcrist:
From http://www.apache.org/dev/apply-license.html#new it looks like they recommend storing `LICENSE` and `NOTICE` in the root of the tar archive of a distribution (as you've done here). Unfortunately cpack doesn't make it easy to distinguish between files in `make install` and `make package`.

Neither `parquet-cpp` nor `arrow` include `LICENSE` or `NOTICE` on `make install` or `make package` (`parquet-cpp` only, arrow doesn't use `cpack`). For at least the python wrappers for these libraries, the actual release artifacts are generated via an external tool that adds in the LICENSE stuff.

This separation of build/package makes sense to me, but may not for this project (not sure how y'all handle releases). Perhaps the LICENSE/NOTICE could be added to the `tar.gz` file after `make package` manually via another make target? Adding to `doc/orc/{NOTICE,LICENSE}` might be equally fine? I'm not really sure what's best here, either is fine with me.

omalley:
Yeah, I agree. I'd rather leave things consistent and share/doc/orc makes sense.

Arguably we should also add the html for the site there too.

PR #196, opened 2017-12-02
ORC-270: fix target_link_libraries for tool-test

rip-nsk:
gmock depends on pthread library, but use ${GTEST_LIBRARIES} similar to c++\test\CMakeLists.txt


majetideepak:
Thanks for the patch!

PR #197, opened 2017-12-05
ORC-271: [C++] Improve HAS_PRE_1970 code

rip-nsk:


rip-nsk:
Change in HAS_POST_2038 shows the bug in osx/clang build:
/Users/travis/build/apache/orc/tools/test/TestMatch.cc:149: Failure
      Expected: expectedLine
      Which is: "{\"time\": \"2038-05-05 12:34:56.1\", \"date\": \"2038-12-25\"}"
To be equal to: line
      Which is: "{\"time\": \"2038-05-05 11:34:56.1\", \"date\": \"2038-12-25\"}"
wrong output at row 0

omalley:
Your HAS_POST_2038 doesn't do the correct test.

The point of the HAS_POST_2038 isn't that gmtime fails, it is that the timezone information continues in 2038. MacOS is the last hold out using the old 32 bit timezone files.

So the current test uses Americas/Los_Angeles and checks to see if daylight savings happens in 2038. Your test for HAS_POST_2038 erroneously passes on MacOS.

rip-nsk:
'Simplify HAS_POST_2038 check' commit is reverted.

PR #201, opened 2017-12-13
Orc 250 - Create sha256 mask

moresandeep:
Masking strategy that masks String, Varchar, Char and Binary types
as SHA 256 hash.

For String type:
 All string type of any length will be converted to 64 length SHA256 hash.

For Varchar type:
 For Varchar type, max-length property will be honored i.e.
 if the length is less than max-length then the SHA256 hash will be truncated
 to max-length. If max-length is greater than 64 then the output is the sha256
 length, which is 64.

For Char type:
 For Char type, the length of mask will always be equal to specified max-length.
 If the given length (max-length) is less than SHA256 hash length (64)
 the mask will be truncated.
 If the given length (max-length) is greater than SHA256 hash length (64)
 then the mask will be padded by blank spaces.

For Binary type:
 All Binary type of any length will be converted to 64 length SHA256 hash.


moresandeep:
No idea why the cpp build is failing, there are no changes to cpp side in this PR.

PR #202, opened 2017-12-14
Orc-280: [C++] use Z_PREFIX if zlib requires it

rip-nsk:


rip-nsk:
linux clang build broken due to new version of compiler (5.0.0)
This is not related to proposed changes.

omalley:
Which platforms need this? I assume it would only be needed when you are using the system zlib.

rip-nsk:
Any platform, if corresponding zlib used (by ZLIB_HOME) to build ORC.

omalley:
Do any of the platforms build their vendor zlib library with Z_PREFIX? Or is it just custom installations?

rip-nsk:
It is custom installation. In my environment I have to use prebuilt static "zprefixed" zlib.
But it is general case since custom zlib can be used via ZLIB_HOME.

PR #207, opened 2018-01-04
Move install location of LICENSE/NOTICE

jcrist:
Currently these are put in the root directory on `make install`, which is not where they should go. Moves them to `share/doc/orc/`, as discussed in https://github.com/apache/orc/pull/193#issuecomment-349063277.


omalley:
I committed this as https://issues.apache.org/jira/browse/ORC-288.

PR #208, opened 2018-01-10
ORC-250 - Create sha256 mask

moresandeep:
Masking strategy that masks String, Varchar, Char and Binary types
as SHA 256 hash.

**For String type:**
All string type of any length will be converted to 64 length SHA256 hash.

**For Varchar type:**
For Varchar type, max-length property will be honored i.e.
if the length is less than max-length then the SHA256 hash will be truncated
to max-length. If max-length is greater than 64 then the output is the sha256
length, which is 64.

**For Char type:**
For Char type, the length of mask will always be equal to specified max-length.
If the given length (max-length) is less than SHA256 hash length (64)
the mask will be truncated.
If the given length (max-length) is greater than SHA256 hash length (64)
then the mask will be padded by blank spaces.

**For Binary type:**
All Binary type of any length will be converted to 64 length SHA256 hash.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/mask/SHA256MaskFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/mask/SHA256MaskFactory.java
@@ -1,0 +1,290 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.orc.impl.mask;
+
+import org.apache.hadoop.hive.ql.exec.vector.BytesColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.ColumnVector;
+import org.apache.orc.DataMask;
+import org.apache.orc.TypeDescription;
+
+import javax.xml.bind.DatatypeConverter;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+
+/**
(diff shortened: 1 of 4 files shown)


java/core/src/java/org/apache/orc/impl/mask/SHA256MaskFactory.java
```diff
@@ -0,0 +1,290 @@
...
+       * we use the hash length (64 bytes) always.
+       */
+      if (schema.getMaxLength() < hash.length) {
+        targetLength = schema.getMaxLength();
+      }
+
+    }
+
+    /* For type char */
+    if (schema.getCategory() == TypeDescription.Category.CHAR) {
```
omalley: Probably a switch would be clearer here.

java/core/src/java/org/apache/orc/impl/mask/SHA256MaskFactory.java
```diff
@@ -0,0 +1,290 @@
...
+   * @param target
+   */
+  void maskBinary(final BytesColumnVector source, final int row,
+      final BytesColumnVector target) {
+
+    final ByteBuffer sourceBytes = ByteBuffer
+        .wrap(source.vector[row], source.start[row], source.length[row]);
+
+    // take SHA-256 Hash and convert to HEX
+    byte[] hash = DatatypeConverter
```
omalley: Let's put the hash back as bytes rather than hex digits for the binary types.

java/core/src/java/org/apache/orc/impl/mask/SHA256MaskFactory.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    // ensure we have enough space, if the masked data is the same size
+    target.ensureValPreallocated(targetLength);
+    byte[] outputBuffer = target.getValPreallocatedBytes();
+    int outputOffset = target.getValPreallocatedStart();
+
+    if (targetLength > hash.length) {
+
+      System.arraycopy(hash, 0, outputBuffer, 0, hash.length);
```
omalley: You don't need to copy the output bytes, just assign the hash variable into the target by setting:
* target.vector[r] = hash
* target.start[r] = 0
* target.length[r] = targetLength

java/core/src/java/org/apache/orc/impl/mask/SHA256MaskFactory.java
moresandeep: Sure, I'll use a Switch statement here.

moresandeep:
Hello @omalley 
Thank you for the review !
I have incorporated most of the changes you proposed, except the one about copying output bytes.
The thing is I will be needing another buffer for case where the targetLength > hash.length to pad the hash with blanks. So, I am wondering if we can keep that one as it is, I will do this in the maskBinary function where I do not need to pad. Let me know your thoughts. 
Also I have added hardened some test cases.
Thanks again.

moresandeep:
Updated the PR with suggested changes.

PR #212, opened 2018-01-19
ORC-293: [C++] Fix RleEncoderV1 for case when sizeof(long) < sizeof(int64_t)

rip-nsk:
and minor related changes.

### DIFF
--- a/c++/include/orc/Common.hh
+++ b/c++/include/orc/Common.hh
@@ -69,7 +69,7 @@ namespace orc {
     UNKNOWN_WRITER = INT32_MAX
   };
 
-  enum CompressionKind {
+  enum CompressionKind : std::int64_t {
     CompressionKind_NONE = 0,
     CompressionKind_ZLIB = 1,
     CompressionKind_SNAPPY = 2,
@@ -84,7 +84,7 @@ namespace orc {
    */
   std::string compressionKindToString(CompressionKind kind);
 
-  enum WriterVersion {
+  enum WriterVersion : std::int64_t {
     WriterVersion_ORIGINAL = 0,
     WriterVersion_HIVE_8732 = 1,
     WriterVersion_HIVE_4243 = 2,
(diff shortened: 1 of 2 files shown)


c++/src/RLEv1.cc
```diff
@@ -148,7 +148,7 @@ void RleEncoderV1::write(int64_t value) {
         numLiterals += 1;
       } else {
         numLiterals -= static_cast<int>(MINIMUM_REPEAT - 1);
-        long base = literals[numLiterals];
+        int64_t base = literals[numLiterals];
```
rip-nsk: 'auto' is more suitable, but it seems its usage is avoided here

c++/src/RLEv1.cc
```diff
@@ -169,11 +169,11 @@ void RleEncoderV1::writeVslong(int64_t val) {
 
 void RleEncoderV1::writeVulong(int64_t val) {
   while (true) {
-    if ((val & ~0x7f) == 0) {
+    if ((val & ~BASE_128_MASK) == 0) {
```
wgtmac: I don't think we need change here as ~0x7f will be implicitly converted to int64_t.
Same for the below.

rip-nsk: It is added to match code in RleDecoderV1::readLong

c++/include/orc/Common.hh
```diff
@@ -69,7 +69,7 @@ namespace orc {
     UNKNOWN_WRITER = INT32_MAX
   };
 
-  enum CompressionKind {
+  enum CompressionKind : std::int64_t {
```
xndai: The definition of CompressionKind_MAX already makes CompressionKind a int64 type. Enum base type is a C++ 11 feature and should at least be configurable through orc-config.hh

rip-nsk: What is the reason for these constants?

xndai: We can potentially get rid of this enum and use proto::CompressionKind instead. It can be done in a separate change.

xndai:
+1

c++/include/orc/Common.hh
rip-nsk: I mean that constants CompressionKind_MAX and WriterVersion_MAX are not used and can be removed.

PR #213, opened 2018-01-23
ORC-278 - Create in memory KeyProvider class

moresandeep:
This PR addresses ORC-278 by creating an in-memory implementation of HadoopShims.KeyProvider interface which can be used for testing.

### DIFF
--- a/java/core/src/java/org/apache/orc/InMemoryKeystore.java
+++ b/java/core/src/java/org/apache/orc/InMemoryKeystore.java
@@ -1,0 +1,374 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.orc;
+
+import org.apache.orc.impl.HadoopShims;
+
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import java.io.IOException;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.Key;
+import java.security.NoSuchAlgorithmException;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * This is an in-memory implementation of {@link HadoopShims.KeyProvider}.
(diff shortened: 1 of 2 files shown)


java/core/src/java/org/apache/orc/InMemoryKeystore.java
```diff
@@ -0,0 +1,402 @@
...
+  }
+
+  /**
+   * Function that takes care of adding the key.
+   * NOTE: Here, if the key already exists with same name then this function overwrite it
+   * this is by design and is used by {@link #rollNewVersion(String, byte[])} function.
+   * Also note, this function is not thread safe, thread safety is the
+   * responsibility of the calling function.
+   *
+   * @param keyName
```
omalley: Please add comments about the parameters.

java/core/src/java/org/apache/orc/InMemoryKeystore.java
```diff
@@ -0,0 +1,403 @@
...
+      rwl.writeLock().unlock();
+    }
+  }
+
+}
+
+/**
+ * This class contains the meta-data and the material for the key.
+ */
+class KeyVersion implements HadoopShims.KeyMetadata {
```
omalley: You probably want this to be a static inner class.

java/core/src/java/org/apache/orc/InMemoryKeystore.java
```diff
@@ -0,0 +1,403 @@
...
+   */
+  private final Map<String, KeyVersion> keys;
+
+  /**
+   * A map that store the keyName (without version)
+   * and metadata associated with it.
+   */
+  private final Map<String, HadoopShims.KeyMetadata> meta = new HashMap<>();
+
+  private final ReentrantReadWriteLock rwl = new ReentrantReadWriteLock(true);
```
omalley: I probably wouldn't have bothered with the read-write lock, but it is fine.

java/core/src/java/org/apache/orc/InMemoryKeystore.java
```diff
@@ -0,0 +1,403 @@
...
+   * new data.
+   *
+   * @param keyName the name of a key
+   * @return metadata for the current version of the key
+   */
+  @Override
+  public HadoopShims.KeyMetadata getCurrentKeyVersion(final String keyName) {
+
+    /* prevent the material from leaking */
+    final HadoopShims.KeyMetadata key = meta.get(keyName);
```
omalley: This isn't necessary. Just return the meta.get(keyName). Once the secrets are in the process, you really can't protect the secrets.

java/core/src/java/org/apache/orc/InMemoryKeystore.java
```diff
@@ -0,0 +1,403 @@
...
+
+    } else {
+      System.arraycopy(masterKey, 0, buffer, 0, algorithm.keyLength());
+    }
+
+    final KeyVersion key = new KeyVersion(keyName, version, algorithm, buffer);
+
+    keys.put(buildVersionName(keyName, version), key);
+    /* our metadata also contains key material, but it should be fine for testing */
+    meta.put(keyName, key);
```
omalley: If there isn't already an entry for keyName, you should put the new one. If it is a higher version, it should update it. If it is a lower version, it should not.

java/core/src/test/org/apache/orc/TestInMemoryKeystore.java
```diff
@@ -0,0 +1,111 @@
...
+
+  @Test
+  public void testDuplicateKeyNames() {
+    try {
+      memoryKeystore.addKey("key128", EncryptionAlgorithm.AES_128, "exception".getBytes());
+      Assert.fail("Keys with same name cannot be added.");
+    } catch (IOException e) {
+      Assert.assertTrue(StringUtils.contains(e.toString(), "Key key128 already exists."));
+    }
+
```
omalley: You should add a test case with multiple versions of a key.

moresandeep:
Hello @omalley 
Thanks for the review, the new PR should incorporate the changes you suggested.

omalley:
Sandeep, I started making some changes and ended up making pretty substantial changes.

Can you look at my changes? https://github.com/omalley/orc/tree/orc-278

moresandeep:
Hello @omalley 
The changes look good, thanks !

PR #218, opened 2018-02-14
ORC-301. `extractFileTail` should open a file in `try` statement

dongjoon-hyun:
This PR moves the file open position to prevent opened file leakage more robustly.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ReaderImpl.java
@@ -516,13 +516,11 @@ public class ReaderImpl implements Reader {
 
   protected OrcTail extractFileTail(FileSystem fs, Path path,
       long maxFileLength) throws IOException {
-    FSDataInputStream file = null;
     ByteBuffer buffer;
     OrcProto.PostScript ps;
     OrcProto.FileTail.Builder fileTailBuilder = OrcProto.FileTail.newBuilder();
     long modificationTime;
-    try {
-      file = fs.open(path);
+    try (FSDataInputStream file = fs.open(path)) {
       // figure out the size of the file using the option or filesystem
       long size;
       if (maxFileLength == Long.MAX_VALUE) {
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Hi, @omalley .
Could you review this PR?

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
```diff
@@ -516,12 +516,13 @@ OrcTail buildEmptyTail() {
 
   protected OrcTail extractFileTail(FileSystem fs, Path path,
       long maxFileLength) throws IOException {
-    FSDataInputStream file = fs.open(path);
+    FSDataInputStream file = null;
     ByteBuffer buffer;
     OrcProto.PostScript ps;
     OrcProto.FileTail.Builder fileTailBuilder = OrcProto.FileTail.newBuilder();
     long modificationTime;
     try {
```
prasanthj: Maybe replace with try-with-resource block?

dongjoon-hyun: Thank you for review, @prasanthj .
Ya. I thought to use `try-with-resource` first, but it needs more change to print out the log at [Line 610](https://github.com/apache/orc/pull/218/files#diff-90ceebe6c0bfa3e98b4536455a60b498R610).

dongjoon-hyun: I think this is the minimal change while keeping the current behavior.

prasanthj: Line 610 doesn't add much value as close() will throw IOException anyway which should be sufficient to know the reason.

dongjoon-hyun: So, you mean it's okay to remove the following in this PR, right?
```
catch (IOException ex) {
  LOG.error("Failed to close the file after another error", ex);
}
```

prasanthj: Yeah that's right.

dongjoon-hyun:
Thank you, @prasanthj . I updated according to your advice.

prasanthj:
lgtm, +1

dongjoon-hyun:
Thank you so much, @prasanthj !

dongjoon-hyun:
Thank you for merging, @prasanthj !

PR #221, opened 2018-02-28
ORC-308. Add function to get subtypes by name.

omalley:
This just adds the new function, which will be useful for specifying properties for columns.

Previously the parser was allowing "." in column names, which doesn't match Hive. I suspect we added that to support schemas from JSON. Now that we support backquotes for column names, that would be a better fix.


omalley:
I just extended this by adding TypeDescription.getSubtypes, which takes a string and returns a List<TypeDescription>.

PR #222, opened 2018-03-01
ORC-310 better error handling for codec

sershe:


java/core/src/java/org/apache/orc/impl/WriterImpl.java
```diff
@@ -645,6 +645,8 @@ public void appendUserMetadata(List<OrcProto.UserMetadataItem> userMetadata) {
     return ReaderImpl.deserializeStats(builder.getStatisticsList());
   }
 
+  // TODO: remove this
```
omalley: Remove the changes in this file. We do not want a guice dependency.

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
```diff
@@ -472,8 +473,9 @@ public static OrcTail extractFileTail(ByteBuffer buffer, long fileLength, long m
         .setPostscript(ps)
         .setFooter(footer)
         .setFileLength(fileLength);
+      isCodecError = false;
     } finally {
```
omalley: try {} finally is pretty broken. You are better off using the try-with-resources statement.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -1353,6 +1353,8 @@ public static String encodeTranslatedSargColumn(int rootColumn, Integer indexInS
     return result;
   }
 
+  // TODO: remove this
```
omalley: Revert the changes in this file.

sershe:
Updated the JIRA with motivation. 
I will rename the flag; there isn't a better way primarily because of the TODOs you are asking me to remove - we don't have access to information about errors if the user and the return-er of the codec are two different pieces of code at different times.
There's also no good way to check the state of the codec to see if it's valid or not. Hadoop ZLib codec may just throw NPE in reset when it's invalid, so the "Safely" catch block will handle that case. However it's nice to give user an opportunity to be proactive about this, esp. in the case of the get-use-return usage pattern

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
sershe: existing code

java/core/src/java/org/apache/orc/impl/WriterImpl.java
sershe: Removed the attribute, clarified the TODO

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -306,15 +330,24 @@ public void releaseBuffer(ByteBuffer buffer) {
...
     public DataReader clone() {
+      if (this.file != null) {
+        throw new UnsupportedOperationException(
+            "Cannot clone a DataReader that is already opened");
+      }
       try {
-        return (DataReader) super.clone();
+        DefaultDataReader clone = (DefaultDataReader) super.clone();
+        // Make sure we don't share the same codec between two readers.
+        clone.codec = OrcCodecPool.getCodec(clone.compressionKind);
```
prasanthj: can this be moved inside super.clone()?

sershe: super.clone() is Object.clone(), so no

PR #223, opened 2018-03-02
ORC-311: Fix null points when StripeFooter corrupts

stiga-huang:
The c++ codes do not check streams got from the StripeStreamsImpl::getStream. It may return null pointers for corrupt StripeFooter, which crash the c++ Reader finally.

Test files are attached to the JIRA.


omalley:
Clearly this change is failing on clang. Can you fix it? Most of the Linux distros have a clang package to install.

stiga-huang:
@omalley Sure, done.

stiga-huang:
Hi @omalley, @majetideepak,
I rebase and update the PR to simply check streams in the constructors.

PR #225, opened 2018-03-03
ORC-313,ORC-317: Check types in footer

stiga-huang:
We need to verify the subtype count of LIST and MAP types. Otherwise,
the ill types will lead the c++ reader to crash.

The java reader also has this problem.

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -897,6 +897,34 @@ namespace orc {
     return REDUNDANT_MOVE(postscript);
   }
 
+  void checkTypes(int& index, const proto::Footer& footer) {
+    if (index >= footer.types_size())
+      throw ParseError(std::string("Footer is corrupt that it lost types(") +
+          std::to_string(index) + ")");
+    const proto::Type& type = footer.types(index);
+
+    // ORC-313: check subtype count of LIST and MAP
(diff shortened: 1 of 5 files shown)


stiga-huang:
With the bug of ORC-317, I think we can check the protobuf Types at the end of orc::readFooter for

- ORC-313: check the subtype count of LIST and MAP
- ORC-317: check that the type tree is valid (not containing loop)

Will update this PR later.

stiga-huang:
Hi @omalley @majetideepak, this PR is ready for review. Could you have a look?

majetideepak:
@stiga-huang can you open a separate PR or each JIRA? Thanks.

stiga-huang:
@majetideepak OK. When I separate the PR, I found orc-test failed in TestWriter.writeTimestamp (ORC-320). I might need to fix that first.

stiga-huang:
@majetideepak I opened #229 for ORC-313

majetideepak:
@stiga-huang  I will look into these this week. Can you fix the title of this PR? Thanks!

stiga-huang:
@majetideepak To be clean, I open #231 for ORC-317. I'd like to close this PR. Thanks for your review!

PR #226, opened 2018-03-05
ORC-314: Check scale range when parsing decimals

stiga-huang:
The scale of decimals may be out of range in corrupt files. We should check this to avoid crashing.
The attached file in the JIRA can reproduce this bug.


stiga-huang:
Some tests are failed. Will look into them one by one.

PR #227, opened 2018-03-07
ORC-318. Change KeyProvider API to separate createLocalKey

omalley:
Separate createLocalKey from decryptLocalKey in HadoopShims.KeyProvider.

It also:
- changes HadoopShims.KeyMetadata from an interface to a concrete class
- creates a NullKeyProvider for Hadoop versions < 2.7


omalley:
Ok, I significantly updated this patch.
* I added tests for both the InMemoryKeyProvider and the Hadoop-based KeyProvider.
* I allowed the user to pass in a Random implementation for the KeyProvider. That allows me to control the random for tests.
* I added a lot of documentation about what I was doing and why.
* I fixed an issue where the HadoopShimsFactory wasn't creating the right version of the shim.
* I changed the InMemoryKeyProvider to use AES/CBC/NoPadding to encrypt and decrypt the local keys. Thus like the AWS KMS, the operations for encryption and decryption aren't symmetric.

PR #229, opened 2018-03-12
ORC-313: Check subtype count of LIST, MAP and UNION types

stiga-huang:
We need to verify the subtype count of LIST, MAP and UNION types. Otherwise, the ill types will lead the c++ reader to crash. The java reader also has this problem.

Files attached to the JIRA can reproduce the crash.


stiga-huang:
@omalley @majetideepak could you have a look at these? We need this bug be fixed to integrate the c++ library into Impala (ORC-315).

PR #231, opened 2018-03-15
ORC-317: [C++] check that indices in the protobuf type tree are valid

stiga-huang:
A corrupt file may contain loops in the protobuf type tree. We should check this before invoking `convertType`. The file attached to the Jira can reproduce this bug. Tests are added for coverage.

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -901,11 +901,11 @@ namespace orc {
   // when we convert the proto::Types to TypeImpls.
   void checkProtoTypeIds(const proto::Footer &footer) {
     std::stringstream msg;
-    uint32_t maxId = footer.types_size();
-    for (uint32_t i = 0; i < maxId; ++i) {
+    int maxId = footer.types_size();
+    for (int i = 0; i < maxId; ++i) {
       const proto::Type& type = footer.types(i);
       for (int j = 0; j < type.subtypes_size(); ++j) {
-        uint32_t subTypeId = type.subtypes(j);
+        int subTypeId = static_cast<int>(type.subtypes(j));
         if (subTypeId <= i) {
           msg << "Footer is corrupt: malformed link from type " << i << " to "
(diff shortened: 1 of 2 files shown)


stiga-huang:
@omalley @majetideepak could you have a look at these? We need this bug be fixed to integrate the c++ library into Impala (ORC-315).

c++/src/Reader.cc
```diff
@@ -897,6 +897,27 @@ namespace orc {
     return REDUNDANT_MOVE(postscript);
   }
 
+  // ORC-317: check that indices in the type tree are valid, so we won't crash
+  // when we convert the proto::Types to TypeImpls.
+  void checkProtoTypeIds(int &index, const proto::Footer &footer) {
```
omalley: I'd suggest only passing in the footer here and check all of the types in a loop.

c++/src/Reader.cc
```diff
@@ -897,6 +897,27 @@ namespace orc {
...
+    if (index >= footer.types_size())
+      throw ParseError(std::string("Footer is corrupt that it lost types(") +
+          std::to_string(index) + ")");
+    const proto::Type& type = footer.types(index);
+
+    int origin_index = index;
+    for (int i = 0; i < type.subtypes_size(); ++i) {
+      int proto_index = static_cast<int>(type.subtypes(i));
```
omalley: For each type:
   - For each subtype value:
      + ensure it is larger than the parent type id
      + ensure it is larger than the previous subtype
      + ensure it is lower or equal to than the max type id

That should prevent any failures, even with malformed ORC file footers.

stiga-huang: That's cool!

omalley:
Actually, I just committed it. Thanks!

I did adjust the comment on the code to make it more consistent and remove the jira reference. The git history does that better if we need to track the history.

PR #236, opened 2018-03-26
ORC-327. Fix java unit test cases to pass when -Duser.timezone is set.

omalley:
It moves the SimpleDateFormat constructor, which stores the timezone, into the place where the timezone is controlled.

It also sets the testOldTimestamp to explicitly use America/Los_Angeles.

Finally, it uses string dates instead of longs in TestFileDump.testDataDump.


prasanthj:
+1

PR #240, opened 2018-04-07
ORC-331: [C++] Initial support of Windows/MSVC

rip-nsk:


xndai:
We should also add an MSVC build in travis. Otherwise there's no way to verify it.

rip-nsk:
This is the beginning of the changes required for MSVC.
Build at this stage will not be successful.

majetideepak:
We must integrate AppVeyor (https://www.appveyor.com/) for Windows. Jira here https://issues.apache.org/jira/browse/ORC-334

PR #245, opened 2018-04-12
ORC-161: Proposal for new decimal encodings and statistics.

wgtmac:
New decimal encoding proposal is added into docs for discussion.


site/_docs/encodings.md
```diff
@@ -123,6 +127,41 @@ DIRECT_V2     | PRESENT         | Yes      | Boolean RLE
...
+When precision is no greater than 18, decimal values can be fully
+represented by 64-bit signed integers which are stored in DATA stream
+and use signed integer RLE.
+
+Encoding      | Stream Kind     | Optional | Contents
+:------------ | :-------------- | :------- | :-------
+DECIMAL       | PRESENT         | Yes      | Boolean RLE
+              | DATA            | No       | Signed Integer RLE v1
+DECIMAL_V2    | PRESENT         | Yes      | Boolean RLE
+              | DATA            | No       | Signed Integer RLE v2
```
prasanthj: Will there be a new RLE version in ORC 2.0? Since we already have 2 RLE versions this is confusing (<0.11 version uses RLE v1 and >=0.12 uses RLE v2). 
If this is newer RLE version then we should probably rename it to RLE v3.
If there will not be a new RLE version then we should probably not use the old RLE v1.

site/_docs/encodings.md
```diff
@@ -123,6 +127,41 @@ DIRECT_V2     | PRESENT         | Yes      | Boolean RLE
...
+When precision is greater than 18, decimal value is split into two
+parts: a signed integer stores higher 64 bits and an unsigned integer
+stores lower 64 bits. Therefore, a DATA stream is utilized to store
+the higher 64-bit signed integer of decimal values and a SECONDARY
+stream holds the lower 64-bit unsigned integer of decimal values.
+Both streams use RLE and are not optional in this case.
+
+Encoding      | Stream Kind     | Optional | Contents
+:------------ | :-------------- | :------- | :-------
+DECIMAL       | PRESENT         | Yes      | Boolean RLE
```
t3rmin4t0r: This is when Decimal v1 can be retired from the encodings.

site/_docs/encodings.md
```diff
@@ -123,6 +127,41 @@ DIRECT_V2     | PRESENT         | Yes      | Boolean RLE
...
+              | DATA            | No       | Signed Integer RLE v1
+              | SECONDARY       | No       | Unsigned Integer RLE v1
+DECIMAL_V2    | PRESENT         | Yes      | Boolean RLE
+              | DATA            | No       | Signed Integer RLE v1
```
t3rmin4t0r: RLEv2 for all integers

site/_docs/file-tail.md
```diff
@@ -249,12 +249,26 @@ For booleans, the statistics include the count of false and true values.
...
+which have better performance.
 
 ```message DecimalStatistics {
  optional string minimum = 1;
  optional string maximum = 2;
  optional string sum = 3;
+ // for precision <= 18 
+ optional sint64 minimum64 = 4;
+ optional sint64 maximum64 = 5;
+ optional sint64 sum64 = 6;
```
t3rmin4t0r: sum64 needs to be Int128

site/_docs/encodings.md
```diff
@@ -123,6 +127,41 @@ DIRECT_V2     | PRESENT         | Yes      | Boolean RLE
...
+              | SECONDARY       | No       | Unsigned Integer RLE v2
```
t3rmin4t0r: Moving sign bits to the lower bits has ease of use advantage (decimal64 is just suppressing the high bit stream)

site/_docs/encodings.md
wgtmac: For decimals, current Integer RLE will be used. As you have explained, I agree that we should not use old RLE v1.

site/_docs/encodings.md
wgtmac: Yep.

site/_docs/encodings.md
wgtmac: Sorry, a copy-and-paste mistake. Will fix it.

site/_docs/file-tail.md
wgtmac: Make sense, will combine them as sum128.

site/_docs/encodings.md
wgtmac: This would be hacky since we use int64_t and uint64_t to represent Int128 in C++. I can force to use signed integer RLE for uint64_t integers. Not sure if java can do the same thing.

site/_docs/file-tail.md
```diff
@@ -249,12 +249,25 @@ For booleans, the statistics include the count of false and true values.
...
+string representation is deprecated and DecimalStatistics uses integers
+which have better performance.
 
 ```message DecimalStatistics {
  optional string minimum = 1;
  optional string maximum = 2;
  optional string sum = 3;
+  message Int128 {
+   repeated sint64 highBits = 1;
+   repeated uint64 lowBits = 2;
```
majetideepak: shouldn't this be sint64 as well since we are using uint64 for the SECONDARY stream?

site/_docs/encodings.md
xndai: I think we should keep RLE v1 as an option. The C++ writer currently does not support RLE v2 (we are working on it). We don't want the new decimal writer to have dependency on that.

majetideepak: @xndai Vertica is interested in getting RLE v2 for C++ as well. Do you think we can collaborate on getting this in quickly?

wgtmac: @majetideepak We are already working on it and doing test & benchmark. Will contribute back  but may not be that soon.

site/_docs/file-tail.md
wgtmac: Here I was aligning with C++ orc::Int128's implementation to avoid many casts.

site/_docs/encodings.md
```diff
@@ -109,10 +109,20 @@ DIRECT_V2     | PRESENT         | Yes      | Boolean RLE
...
+128 varint. The scale is stored in the SECONDARY stream as an signed
+integer.
+
+In ORC 2.0, DECIMAL encoding is introduced and totally remove scale
+stream as all decimal values use the same scale. When precision is
+no greater than 18, decimal values can be fully represented by DATA
+stream which stores 64-bit signed integers. When precision is greater
+than 18, we use a 128-bit signed integer to store the decimal value.
+DATA stream stores the higher 64 bits and SECONDARY stream holds the
+lower 64 bits. Both streams use signed integer RLE v2.
```
dain: Why split the data across two streams?  This means 2 IOs (or one large coalesced IO) to read the values (assuming no nulls).  Instead, can't we put all 128 bits in one stream?

wgtmac: The main problem is that we don't have 128-bit integer RLE on hand.

wgtmac:
After second thought, I added back DECIMAL_V1 to support RLE v1 in decimal encoding. The reason is that in our testing, we found RLEv1 + zstd may be the best combination than others in terms of both compression ration and encoding/decoding speed.

prasanthj:
"we found RLEv1 + zstd may be the best combination than others in terms of both compression ration and encoding/decoding speed."

do you have experimental numbers for this?

site/_docs/encodings.md
t3rmin4t0r: The multiple-stream + row-group stride problems for IO were discussed by Owen.

The disk layout is what matters for IO, not the logical stream separation.

site/_docs/encodings.md
t3rmin4t0r: Some part of this discussion is about the new ORC format and existing reader compatibility is not a requirement, until we switch to the new format as a default.

wgtmac:
Will provide them after comprehensive benchmark.

site/_docs/encodings.md
```diff
@@ -109,10 +109,20 @@ DIRECT_V2     | PRESENT         | Yes      | Boolean RLE
...
-encoded base 128 varint. The scale is stored in the SECONDARY stream
-as an signed integer.
+bits plus a sign bit.
+
+DIRECT and DIRECT_V2 encodings of decimal columns stores the integer
+representation of the value as an unbounded length zigzag encoded base
+128 varint. The scale is stored in the SECONDARY stream as an signed
+integer.
+
+In ORC 2.0, DECIMAL_V1 and DECIMAL_V2 encodins are introduced and
```
omalley: In ORCv2, we'll just pick a RLE and not leave it pickable.

In terms of the encoding names, I'm a bit torn. My original inclination would be to use DECIMAL64 and DECIMAL128 as encoding names. However, It would be nice to have the ability to use dictionaries, so we'd need dictionary forms of them too. Thoughts?

site/_docs/encodings.md
```diff
@@ -122,6 +132,12 @@ DIRECT        | PRESENT         | Yes      | Boolean RLE
 DIRECT_V2     | PRESENT         | Yes      | Boolean RLE
               | DATA            | No       | Unbounded base 128 varints
               | SECONDARY       | No       | Unsigned Integer RLE v2
+DECIMAL_V1    | PRESENT         | Yes      | Boolean RLE
+              | DATA            | No       | Signed Integer RLE v1
+              | SECONDARY       | Yes      | Signed Integer RLE v1
```
omalley: We probably need a RLE128 that can just encode a int128 directly. Then we can just use the DATA stream directly.

site/_docs/file-tail.md
```diff
@@ -249,12 +249,25 @@ For booleans, the statistics include the count of false and true values.
...
-For decimals, the minimum, maximum, and sum are stored.
+For decimals, the minimum, maximum, and sum are stored. In ORC 2.0,
+string representation is deprecated and DecimalStatistics uses integers
+which have better performance.
 
 ```message DecimalStatistics {
  optional string minimum = 1;
  optional string maximum = 2;
  optional string sum = 3;
+  message Int128 {
```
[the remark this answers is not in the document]
wgtmac: Good suggestion!

site/_docs/encodings.md
wgtmac: As I have said, I don't see any obvious benefit to use RLEv2 and abandon RLEv1 in our experiment. So I don't think it is a good idea not to provide an option to choose a RLE version.

site/_docs/encodings.md
wgtmac: We may use LEB128. Thoughts?

site/_docs/encodings.md
wgtmac: We may also add dictionary encodings for integer and floating types. I think we can still use DICTIONARY_Vx and it has unique meaning for different data types.

xndai:
+1 for Gang's proposal.

ddrinka:
@wgtmac what did you end up doing with all your production data waiting for the decimal changes?

wgtmac:
@ddrinka We have done some experiments to apply encodings (RLEv1 and RLEv2) to decimals with precision <= 18 and remove scale streams. Below is an example to read/write 1GB TPC-H lineitem which has 4 columns with decimal(12,2) among all 16 columns.

  | Base128 decimal + RLEv1 integer | Base128 decimal + RLEv2 integer | RLEv1 for decimal & integer | RLEv2 for decimal & integer
-- | -- | -- | -- | --
File Size | 159,598,526 | 151,705,690 | 160,636,343 | 149,565,596
Writer Time | 7.24717 | 8.82114 | 4.64885 | 6.51180
Reader Time | 1.51090 | 1.75568 | 1.42397 | 1.73446

We can see the speed of RLEv1 is promising but RLEv2 achieves better compression ratio. It is a trade off to choose between RLEv1 and RLEv2 for decimals for now. I am not sure if anyone is working on RLEv3 for ORCv2. We are also evaluating other integer encoding schemes and will propose if we really think there is a working solution.

xndai:
I think these are good numbers. They show even with current RLE scheme we could see noticible  improvements with the new decimal format. The RLEv3 is not even on the horizon. Can we start working on this change with RLEv1 and v2? @omalley

PR #247, opened 2018-04-12
ORC-339. Reorganize the ORC file format specification.

omalley:


site/specification/ORCv2.md
```diff
@@ -0,0 +1,1032 @@
...
+* Create RLEv3:
+   * 64 and 128 bit variants
+   * Zero suppression
+   * Evaluate the rle subformats
+* Group stripe data into stripelets to enable Async IO for reads.
+* Reorder stripe data into (stripe metadata, index, dictionary, data)
+* Stop sorting dictionaries and record the sort order separately in the index.
+* Remove use of RLEv1 and RLEv2.
+* Remove non-utf8 bloom filter.
+* Use numeric value for decimal bloom filter.
```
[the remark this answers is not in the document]
omalley: Agreed

omalley:
Any other comments?

site/specification/ORCv2.md
```diff
@@ -0,0 +1,1032 @@
+---
+layout: page
+title: Evolving Draft for ORC Specification v2
+---
+
+This specification is rapidly evolving and should only be used for
+developers on the project.
+
+# TO DO items
```
xndai: Is this a final list of v2 or we are still working on it? I have one proposal to add to ORC v2, which is what I call "clustered index". Basically the writer can specify a sorting property on one or more columns, then we create an index section in ORC file with keys being the column(s) value and the value is the row number. To reduce the size of index, each row group has one entry in the clustered index. This will enable new range scan pattern when reader provides upper bound and lower bound of column(s) values. 

I can write up a detailed proposal for this.

site/specification/ORCv2.md
```diff
@@ -0,0 +1,1032 @@
...
+
+The list of things that we plan to change:
+
+* Create a decimal representation with fixed scale using rle.
+* Create a better float/double encoding that splits mantissa and
+  exponent.
+* Create a dictionary encoding for float, double, and decimal.
+* Create RLEv3:
```
wgtmac: Is there anyone working on it already?

site/specification/ORCv2.md
omalley: This is absolutely a work in progress. I'd love to see your proposal.

site/specification/ORCv2.md
omalley: I have some thoughts, but haven't done any work on it.

site/specification/ORCv2.md
```diff
@@ -0,0 +1,1032 @@
...
+   * 64 and 128 bit variants
```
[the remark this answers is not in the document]
omalley: No, I mean that the new RLEv3 needs to support both 64 bit and 128 bit integers. The specification of course could just define the spec using 128 bits, but the Java and C++ implementation would likely be split.

PR #249, opened 2018-04-14
[ORC-341] Support time zone as a parameter for Java reader and writer

jcamachor:


wgtmac:
Why do we need this change?

jcamachor:
@wgtmac , see discussion in https://issues.apache.org/jira/browse/HIVE-12192 for more context.

wgtmac:
AFAIK, I don't think ORC has any issue in HIVE-12192. What ORC guarantees is that we should always get same wall clock time representation w/o timezone. Current Java implementation leverages java.sql.Timestamp which uses local timezone and that's why writer and reader always use  timestamp values in local timezone. Unless we add a new TimestampColumnVector which enforces UTC timezone to adopt your change here.

omalley:
@jcamachor I'd suggest a much simpler API:

- Instead of passing in the reader timezone, make a boolean option to useUtcForTimestamp.
- Extend TimestampColumnVector to have a boolean isUTC field.
- The TimestampTreeWriter can use the isUTC in the ColumnVector to determine if it is in UTC.
- The reader can set isUTC appropriately based on the option.

Thoughts?

omalley:
Also note that the C++ reader already uses UTC for its TimestampColumnVector. :)

jcamachor:
@omalley , it seems like a good idea, let me explore it and refresh the PR. I will adapt HIVE-19226 to these new changes too.

@wgtmac , I understand you are suggesting that this can be fixed only from Hive side? Problem is that existing ORC files should still be read properly, hence you would need to recognize old vs new ORC files. In addition, you will apply displacement twice when reading/writing, in Hive and in ORC. It seems to me the cleaner solution is just being able to point to ORC that timestamp data is in UTC from Java reader/writer. FWIW, change to stringify in TimestampColumnVector is needed indeed.

wgtmac:
@jcamachor Yes I was just meaning ORC doesn't have problems in dealing with timestamps itself. Definitely using UTC everywhere makes things way easier.

wgtmac:
For reader, can we set useUTCTImestamp in the function TimestampTreeReader::nextVector? For writer, it is caller's responsibility to set useUTCTImestamp before calling TimestampTreeWriter::writeBatch. Does this help? @jcamachor

jcamachor:
Pushed a new commit with the changes.

We would still need a storage-api release for the ```TimestampColumnVector``` changes.

wgtmac:
@jcamachor You are right. WriterOptions/WriterContext are ideal places to set this kind of values.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -975,6 +992,8 @@ public void nextVector(ColumnVector previousVector,
       TimestampColumnVector result = (TimestampColumnVector) previousVector;
       super.nextVector(previousVector, isNull, batchSize);
 
+      // TODO: If context.isUseUTCTimestamp(), set TimestampColumnVector.useUTC to true
```
wgtmac: I think it is better to set storage-api to 3.0.0 and fix this TODO in this patch as well.

jcamachor: There is a vote going on for a storage-api release, then next week I can rebase the patch to consume it and hopefully we can check it in. Thanks @wgtmac !

jcamachor:
I have just updated the patch now that we have moved to the new storage-api version. I will run some tests with Hive locally asap and will get back confirming that everything is working as expected.

jcamachor:
I have been testing the patch from Hive and everything seems to be working as expected.

I have rebased the patch and merge both commits. Also, I had to extend my changes to the newly created ```WriterImplV2```.

@omalley , @wgtmac , could you take a final look and merge if it is OK? Thanks

java/core/src/java/org/apache/orc/impl/writer/TimestampTreeWriter.java
```diff
@@ -54,9 +57,20 @@ public TimestampTreeWriter(int columnId,
     if (rowIndexPosition != null) {
       recordPosition(rowIndexPosition);
     }
-    this.localTimezone = TimeZone.getDefault();
-    // for unit tests to set different time zones
-    this.baseEpochSecsLocalTz = Timestamp.valueOf(BASE_TIMESTAMP_STRING).getTime() / MILLIS_PER_SECOND;
+    if (writer.isUseUTCTimestamp()) {
+      this.localTimezone = TimeZone.getTimeZone("UTC");
```
wgtmac: We'd better change its name to this.writeTimezone to avoid confusion in the future.
Same for localDateFormat and baseEpochSecsLocalTz below.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -990,6 +1007,10 @@ public void nextVector(ColumnVector previousVector,
       TimestampColumnVector result = (TimestampColumnVector) previousVector;
       super.nextVector(previousVector, isNull, batchSize);
 
+      if (context.isUseUTCTimestamp()) {
+        result.setIsUTC(true);
```
wgtmac: result.setIsUTC(context.isUseUTCTimestamp());

Just in case result is in UTC but context.isUseUTCTimestamp() is false.

java/core/src/java/org/apache/orc/impl/writer/TimestampTreeWriter.java
```diff
@@ -28,7 +28,9 @@
 import org.apache.orc.impl.SerializationUtils;
 
 import java.io.IOException;
-import java.sql.Timestamp;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.TimeZone;
 
 public class TimestampTreeWriter extends TreeWriterBase {
```
wgtmac: We should also change writeBatch function below.

The input vector.isUTC may be true while writer.isUseUTCTimestamp() is false; vice versa. In this case, we need to convert them to correct writer timezone.

java/core/src/java/org/apache/orc/OrcFile.java
```diff
@@ -320,6 +321,16 @@ public ReaderOptions fileMetadata(final FileMetadata metadata) {
     public FileMetadata getFileMetadata() {
       return fileMetadata;
     }
+
+    public ReaderOptions useUTCTimestamp(boolean value) {
+      useUTCTimestamp = value;
+      return this;
+    }
+
+    public boolean isUseUTCTimestamp() {
```
omalley: This should be getUseUTCTimestamp.

java/core/src/java/org/apache/orc/OrcFile.java
```diff
@@ -320,6 +321,16 @@ public ReaderOptions fileMetadata(final FileMetadata metadata) {
     public FileMetadata getFileMetadata() {
       return fileMetadata;
     }
+
+    public ReaderOptions useUTCTimestamp(boolean value) {
```
omalley: This should also cause the TimestampStatistics to use UTC.

java/core/src/java/org/apache/orc/OrcFile.java
```diff
@@ -761,6 +782,10 @@ public boolean getWriteVariableLengthBlocks() {
     public HadoopShims getHadoopShims() {
       return shims;
     }
+
+    public boolean isUseUTCTimestamp() {
```
omalley: Rename this to getUseUTCTimestamp.

java/core/src/java/org/apache/orc/impl/writer/WriterImplV2.java
```diff
@@ -373,7 +379,11 @@ private void flushStripe() throws IOException {
       OrcProto.StripeFooter.Builder builder =
           OrcProto.StripeFooter.newBuilder();
       if (writeTimeZone) {
-        builder.setWriterTimezone(TimeZone.getDefault().getID());
+        if (useUTCTimeZone) {
+          builder.setWriterTimezone(TimeZone.getTimeZone("UTC").getID());
```
omalley: I'd be tempted to just use setWriterTimezone("UTC"), because we'll already fail if UTC is called something else.

java/core/src/java/org/apache/orc/impl/writer/TimestampTreeWriter.java
```diff
@@ -54,9 +57,20 @@ public TimestampTreeWriter(int columnId,
...
     }
-    this.localTimezone = TimeZone.getDefault();
-    // for unit tests to set different time zones
-    this.baseEpochSecsLocalTz = Timestamp.valueOf(BASE_TIMESTAMP_STRING).getTime() / MILLIS_PER_SECOND;
+    if (writer.isUseUTCTimestamp()) {
+      this.localTimezone = TimeZone.getTimeZone("UTC");
+    } else {
+      this.localTimezone = TimeZone.getDefault();
+    }
+    this.localDateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
```
omalley: It sucks that there isn't a simpler way in Java to do this.

jcamachor:
@wgtmac , @omalley , thanks for the feedback. I think I have addressed all your points with last two commits, could you take another look? Thanks

PR #252, opened 2018-04-18
ORC-345 Create and use a new Decimal64StatisticsImpl on the write path

omalley:
Currently the DecimalTreeWriter spends a lot of time maintaining the statistics for decimal objects. For short decimals (p <= 18) we can use longs to track the min, max, and sum.

### DIFF
--- a/java/bench/src/java/org/apache/orc/bench/DecimalBench.java
+++ b/java/bench/src/java/org/apache/orc/bench/DecimalBench.java
@@ -1,0 +1,270 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.orc.bench;
+
+import com.google.gson.JsonStreamParser;
+import org.apache.avro.file.DataFileReader;
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.io.DatumReader;
+import org.apache.avro.mapred.FsInput;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.TrackingLocalFileSystem;
+import org.apache.hadoop.hive.ql.exec.vector.ColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.Decimal64ColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.DecimalColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
+import org.apache.hadoop.hive.ql.io.parquet.read.DataWritableReadSupport;
+import org.apache.hadoop.hive.ql.io.parquet.read.ParquetRecordReaderWrapper;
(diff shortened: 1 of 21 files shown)


wgtmac:
Do we need some test on Decimal64StatisticsImpl, especially to make sure overflow checking is correct.

omalley:
Ok, with this and ORC-344, the difference in writing small vs large decimals to a null file system is huge:

Benchmark           (precision)  Mode  Cnt       Score      Error  Units
DecimalBench.write            8  avgt   30   36914.791 ±  866.408  us/op
DecimalBench.write           19  avgt   30  240789.318 ± 5146.880  us/op

So I'm seeing a 6x speed up when using precision = 8 with the new code path instead of p = 19 and the old code path.

java/bench/src/java/org/apache/orc/bench/DecimalBench.java
```diff
@@ -0,0 +1,270 @@
...
+            new Decimal64Loader(precision, 2) :
+            new DecimalLoader(precision, 2);
+      readCsvData(total_amount, root, "total_amount", conf);
+      batch = schema.createRowBatchV2();
+    }
+  }
+
+  @Benchmark
+  public void write(OutputState state) throws Exception {
+    Writer writer = OrcFile.createWriter(new Path("null"),
```
prasanthj: Should the writer creation moved out of the @Benchmark method? I think we might unnecessarily measure IO creating writer.

java/bench/src/java/org/apache/orc/bench/DecimalBench.java
```diff
@@ -0,0 +1,270 @@
...
+        OrcFile.writerOptions(state.conf)
+            .fileSystem(state.fs)
+            .setSchema(state.schema)
+            .compress(CompressionKind.NONE));
+    int r = 0;
+    int batchSize = state.batch.getMaxSize();
+    while (r < state.total_amount.length) {
```
prasanthj: Should this go into @OperationsPerInvocation annotation to measure the time per operation correctly?

java/bench/src/java/org/apache/orc/bench/DecimalBench.java
```diff
@@ -0,0 +1,270 @@
...
+        if (child.getCategory() == TypeDescription.Category.DECIMAL) {
+          include[child.getId()] = true;
+        }
+      }
+    }
+  }
+
+  @Benchmark
+  public void read(Blackhole blackhole, InputState state) throws Exception {
+    Reader reader = OrcFile.createReader(state.path,
```
prasanthj: Similarly as above. Reader out of benchmark method? Probably during setup.

java/core/src/test/org/apache/orc/impl/TestColumnStatisticsImpl.java
```diff
@@ -85,4 +86,53 @@ public void testOldTimestamps() throws IOException {
...
+        OrcProto.DecimalStatistics.newBuilder();
+    decimalBuilder.setMaximum("1000.0");
+    decimalBuilder.setMinimum("1.010");
+    decimalBuilder.setSum("123456789.123456");
+    pb.setDecimalStatistics(decimalBuilder);
+    pb.setHasNull(false);
+    pb.setNumberOfValues(3);
+
+    // the base case doesn't overflow
+    DecimalColumnStatistics stats1 = (DecimalColumnStatistics)
```
prasanthj: Can you also add a test for merge with opposite signs (adding negative and positive so that overflow if condition in merge() gets covered) and also one that overflows?

java/bench/src/java/org/apache/orc/bench/DecimalBench.java
omalley: I started that way, but the problem is that I do want to measure the close, since a lot of the work is deferred until then, as part of the benchmark and it felt wrong to have mismatched open/close calls.

java/bench/src/java/org/apache/orc/bench/DecimalBench.java
omalley: Since I'm holding the operations constant, it doesn't add much. The operation is just larger (create file, write 1m decimals, close).

java/bench/src/java/org/apache/orc/bench/DecimalBench.java
omalley: Ok, I'll move creating the Reader out, but the RecordReader has the same issue as the writer. A lot of the work is done in the constructor that we need to count.

omalley:
I should comment while adding the tests that it became clear that I didn't need the sign change logic, because (10**18 - 1 )* 2 < 2**63. So by enforcing the 10**18 limit, I didn't need the more complicated sign change detection.

PR #253, opened 2018-04-19
ORC-346: [C++] Add one second when writing negative Timestamp's with non zero nanos …

rip-nsk:
…to match the reader code.

### DIFF
--- a/c++/src/ColumnWriter.cc
+++ b/c++/src/ColumnWriter.cc
@@ -1199,6 +1199,10 @@ namespace orc {
         tsStats->increase(1);
         tsStats->update(millsUTC);
 
+        if (secs[i] < 0 && nanos[i] != 0) {
+          secs[i] += 1;
+        }
+
         secs[i] -= timezone.getEpoch();
         nanos[i] = formatNano(nanos[i]);
       } else if (!hasNull) {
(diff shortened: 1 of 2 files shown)


rip-nsk:
I'm not sure that it is correct fix, IMO, TimestampColumnReader should not do this:
        if (secsBuffer[i] < 0 && nanoBuffer[i] != 0) {
          secsBuffer[i] -= 1;
        }
but it seems necessary to match the java reader.

omalley:
*Sigh* This is caused by the fact that Java defined Timestamp badly and then Hive copied those semantics into the ColumnVector.

We really need to fix these semantics in ORCv2.

I'll also comment that it would probably be nicer to our users if we weren't modifying the data in the columnVector they passed into us, but clearly we are already doing that.

c++/src/ColumnWriter.cc
```diff
@@ -1199,6 +1199,10 @@ namespace orc {
         tsStats->increase(1);
         tsStats->update(millsUTC);
 
+        if (secs[i] < 0 && nanos[i] != 0) {
+          secs[i] += 1;
```
wgtmac: If secs[i] is -1 and we add 1 to it here. On the reader side, the if branch will simply fail to minus 1:
` if (secsBuffer[i] < 0 && nanoBuffer[i] != 0) { secsBuffer[i] -= 1; } `  This results in an 1 second shift of the original data. I canot find any good workaround but throw an exception if -1 has been found. @rip-nsk @omalley

rip-nsk: possible solutions 
- allow negative nanos in orc stream to cover the second before unix epoch.
- introduce new Timestamp type, w/o shift and deprecate old one

PR #254, opened 2018-04-21
ORC-351: [C++] Increase buffer in Decimal*ColumnWriter to required size

rip-nsk:


majetideepak:
Can you add a test as well? Thanks.

c++/src/ColumnWriter.cc
```diff
@@ -1453,7 +1453,7 @@ namespace orc {
                      uint64_t numValues) override;
 
   private:
-    char buffer[16];
+    char buffer[20];
```
wgtmac: 19 should be sufficient, right?

rip-nsk: Yes, both buffers are "aligned" to 2 bytes

PR #257, opened 2018-04-27
ORC-49. Create RLE-based encoding for short decimals in ORCv2.

omalley:
Create the first pass of the new decimal encoding for ORCv2. For small decimals, it uses RLEv2 and the old format for long decimals.


omalley:
Part of what we absolutely need in RLEv3 is zero suppression for the whole row batch. So that if you have integers like:

- 1000
- 2000
- 3000
- 10000
- 4000
- 1000

it will record 3 for the number of zeros and encode:

- 1
- 2
- 3
- 10
- 4
- 1

PR #259, opened 2018-04-30
ORC-248. PhysicalFsWriter sometimes passes negative padding down to shims

omalley:
The PhysicalFsWriter sometimes pass down negative sizes for padding, which leads to exceptions.

My fix:
* Add a new option to enable writing variable length HDFS blocks "orc.write.shortened.blocks", which is off by default.
* Modified the shim to shortenBlock so that the PhysicalFsWriter can track the offsets.
* shortenBlock isn't gated by the padding tolerance, although it is called when the new stripe would cross a block boundary.

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -152,7 +152,10 @@ public enum OrcConf {
   OVERWRITE_OUTPUT_FILE("orc.overwrite.output.file", "orc.overwrite.output.file", false,
     "A boolean flag to enable overwriting of the output file if it already exists.\n"),
   IS_SCHEMA_EVOLUTION_CASE_SENSITIVE("orc.schema.evolution.case.sensitive", "orc.schema.evolution.case.sensitive", true,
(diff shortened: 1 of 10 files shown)


java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -152,7 +152,9 @@
   OVERWRITE_OUTPUT_FILE("orc.overwrite.output.file", "orc.overwrite.output.file", false,
     "A boolean flag to enable overwriting of the output file if it already exists.\n"),
   IS_SCHEMA_EVOLUTION_CASE_SENSITIVE("orc.schema.evolution.case.sensitive", "orc.schema.evolution.case.sensitive", true,
-          "A boolean flag to determine if the comparision of field names in schema evolution is case sensitive .\n")
+          "A boolean flag to determine if the comparision of field names in schema evolution is case sensitive .\n"),
+  WRITE_SHORTENED_BLOCKS("orc.write.shortened.blocks", null, false,
```
t3rmin4t0r: Might as well call it what it is "orc.write.variable.length.blocks"

PR #263, opened 2018-05-02
ORC-357: [C++] Use orc::InputStream in getTimezoneByFilename

rip-nsk:


c++/src/Timezone.cc
```diff
@@ -698,40 +694,15 @@ namespace orc {
...
-                            filename + " - " + strerror(errno));
-      }
-      posn += static_cast<size_t>(ret);
-    }
-    if (close(in) == -1) {
-      std::stringstream err;
-      err << "failed to close " << filename << " - " << strerror(errno);
-      throw TimezoneError(err.str());
+    try {
+      ORC_UNIQUE_PTR<InputStream> file = readFile(filename);
```
rip-nsk: 'auto' is more suitable here

PR #265, opened 2018-05-04
ORC-334: [C++] Add AppVeyor support for integration on windows

rip-nsk:


rip-nsk:
AppVeyor after this PR:
[https://ci.appveyor.com/project/rip-nsk/orc/build/1.0.32]

c++/test/CreateTestFiles.cc
```diff
@@ -31,8 +31,8 @@
 void writeCustomOrcFile(const std::string& filename,
                         const orc::proto::Metadata& metadata,
                         const orc::proto::Footer& footer,
-                        const std::vector<uint>& version,
-                        uint writerVersion) {
+                        const std::vector<unsigned>& version,
```
omalley: We generally prefer to use explicitly sized types, so uint64_t or uint32_t would be better.

c++/src/Timezone.cc
```diff
@@ -710,7 +710,11 @@ namespace orc {
    * Get the local timezone.
    */
   const Timezone& getLocalTimezone() {
+#ifdef _MSC_VER
+    return getTimezoneByName("UTC");
```
majetideepak: Can you comment on why we look at `UTC` for windows instead of `LOCAL_TIMEZONE`?

omalley: Because there isn't a direct windows equivalent.

Here is what Java writes about it-
https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/time-zone002.html

There is a registry setting, but it uses a non-standard set of names *doh*. The easiest way would be to create a Java jar and run that to determine the local timezone.

rip-nsk: http://unicode.org/cldr/data/common/supplemental/windowsZones.xml - here is a mapping of windows <=> *nix time zones, 
and here - https://github.com/google/cctz - "timezone" library which probably is a best choice (BTW https://github.com/google/cctz/issues/53).
Anyway,  it is non trivial work which I think does not related to orc directly.
I believe it should be customer decision how to pass/extract timestamps from orc library and any timezone conversions should be done on customer side.

wgtmac: Recently we have change to UTC everywhere in C++ code. Is it better to delete getLocalTimezone function to avoid confusion?

rip-nsk: getLocalTimezone is still used in RowReaderImpl

wgtmac: Just double checked, we should still use local timezone if writer timezone is not found to keep backward compatibility.

rip-nsk: Why it can't be UTC in this (or any other) case?

wgtmac: If we cannot find writer timezone from the stripe footer, we assume the writer and reader are in the same timezone. TimestampColumnReader tries to convert the timestamp to UTC to get the same wall clock time as the writer. Unless the files are written under UTC, otherwise we may get wrong timestamps if we set local timezone to UTC by default.

rip-nsk: So, proposed solution (local == UTC) will work until we read of the file created using java writer.

majetideepak: I also think that it is better to leave the conversion to the client/customer. We should ideally change the conversion for Nix* systems but cannot due to backward compatibility. We should be able to converge both Java and C++ timestamps in ORC 2.0.
For Windows, since this is the first official build support, we should be okay to use "UTC" and document this behavior.
I will merge this PR end of today if there are no objections.

majetideepak:
+1 LGTM

PR #268, opened 2018-05-14
ORC-363 Enable zstd decompression in ORC Java reader

xndai:
1. Upgrade aircompressor lib to 0.11
2. Enable Zstd decompression in Java reader
3. Zstd compression is still not availiable. Will throw illegal argument
exception if writer uses zstd compression.

Change-Id: Ic5492b09af4e1e51215e62ed29233729857c45ac


xndai:
The current solution is not perfect. But at least it gives us some ability to read zstd Orc files, which I believe is important from the compatibility perspective - our in-house system has zstd Orc that would like to be consumed by Hive, Spark, etc. I am not sure when the zstd compressor will be available. It's probably another 6 months or a year.

If we enable zstd on C++ reader/writer first. Then we enable Java reader to consume zstd Orc from C++ writer. Would you consider that as end-to-end test?

xndai:
Hi all, I'd like to bring this up again. It's almost another three months, and we haven't seen the zstd java library yet. I would suggest we move forward to add zstd support in C++ reader/writer. And at the same time, enable Java reader by this change.

t3rmin4t0r:
I'm in favour of using the org.apache.hadoop.io.compress.ZStandardCodec if it is available from Hadoop.

xndai:
Close this PR since @wgtmac is working on this through ORC-136.

PR #274, opened 2018-05-25
ORC-368:[C++] Reader must return default version 0.11 instead of error for missing version values

czxrrr:


majetideepak:
+1 LGTM

PR #275, opened 2018-05-30
ORC-371: [C++] Disable Libhdfspp build when Cyrus SASL is not found

majetideepak:


majetideepak:
CC @jamesclampffer

jamesclampffer:
Rather than skipping the entire libhdfs++ build you could set the NO_SASL parameter if openssl or cyrus is missing to build libhdfs++ without support for strong authentication.

At one point someone added a partial implementation of SASL DIGEGT-MD5 directly in libhdfs++ and I think that may try to use SSL's RNG to generate nonces (and looks like NO_SASL might not prevent that).  If that breaks the build with NO_SASL set we can get rid of that code in libhdfs++.  Even if that implementation worked it'd prevent the library from being FIPS 140-2 compliant which is a deal breaker for my work.

jamesclampffer:
@majetideepak it looks like the build still fails with this in place.  OrcHdfsFile.cc includes hdfspp.h which won't exist if libhdfs++ isn't extracted.

```
<cut>/apache_orc/build$ cmake .. -DBUILD_JAVA=OFF
-- No build type selected, default to ReleaseWithDebugInfo
-- compiler GNU version 4.9.4
-- Could NOT find CYRUS_SASL (missing: CYRUS_SASL_SHARED_LIB CYRUS_SASL_INCLUDE_DIR) 
-- WARNING: Libhdfs++ library was not built because the required CyrusSASL library was not found
-- Configuring done
-- Generating done
-- Build files have been written to: <cut>/apache_orc/build
<cut>:~/apache_orc/build$ make package
[  7%] Built target googletest_ep
[ 14%] Built target snappy_ep
[ 21%] Built target zlib_ep
[ 28%] Built target lz4_ep
[ 35%] Built target protobuf_ep
[ 36%] Building CXX object c++/src/CMakeFiles/orc.dir/OrcHdfsFile.cc.o
<cut>/apache_orc/c++/src/OrcHdfsFile.cc:32:27: fatal error: hdfspp/hdfspp.h: No such file or directory
 #include "hdfspp/hdfspp.h"
                           ^
compilation terminated.
c++/src/CMakeFiles/orc.dir/build.make:693: recipe for target 'c++/src/CMakeFiles/orc.dir/OrcHdfsFile.cc.o' failed

```

PR #277, opened 2018-06-03
ORC-372: Enable valgrind for C++ travis-ci tests

majetideepak:


majetideepak:
There are indeed valgrind failures. I will push a followup patch to fix these.

majetideepak:
@wgtmac, @xndai  can you take a look at this patch? ZLIB compression code seems to be leaking memory. Thanks!

c++/src/TypeImpl.cc
```diff
@@ -258,31 +258,34 @@ namespace orc {
 
     case STRUCT: {
       StructVectorBatch *result = new StructVectorBatch(capacity, memoryPool);
+      std::unique_ptr<ColumnVectorBatch> return_value = std::unique_ptr<ColumnVectorBatch>(result);
```
wgtmac: maybe we can directly merge line 260 and 261 to a single statement. same for below.

c++/src/TypeImpl.cc
```diff
@@ -498,54 +506,54 @@ namespace orc {
...
+    std::vector<std::pair<std::string, ORC_UNIQUE_PTR<Type> > > res =
       TypeImpl::parseType(input, 0, input.size());
     if (res.size() != 1) {
       throw std::logic_error("Invalid type string.");
     }
-    return ORC_UNIQUE_PTR<Type>(res[0].second);
+    return std::move(res[0].second);
   }
 
   Type* TypeImpl::parseArrayType(const std::string &input,
```
wgtmac: is it better to change return type to unique_ptr as well to be consistent? same for below.

c++/src/Compression.cc
```diff
@@ -282,6 +285,11 @@ DIAGNOSTIC_PUSH
     }
   }
 
+  void ZlibCompressionStream::end() {
+    (void)deflateEnd(&strm);
+  }
+
+
```
wgtmac: remove this line

c++/src/TypeImpl.cc
majetideepak: we need `result` of type `StructVectorBatch` to access fields on line 263

c++/src/Compression.cc
```diff
@@ -199,13 +199,16 @@ namespace orc {
                           uint64_t blockSize,
                           MemoryPool& pool);
 
+    ~ZlibCompressionStream() { end(); }
```
omalley: You need an override on this or the mac complains.

c++/test/CMakeLists.txt
```diff
@@ -64,7 +64,12 @@ target_link_libraries (create-test-files
   protobuf
 )
 
-add_test (NAME orc-test COMMAND orc-test)
+if (TEST_VALGRIND_MEMCHECK)
+  add_test (orc-test
```
omalley: You have trailing spaces here.

c++/src/Compression.cc
majetideepak: Thanks!

PR #278, opened 2018-06-03
ORC-251: Extend InStream and OutStream to support encryption.

omalley:
This patch:
* Adds a method to Codec to get the CompressionKind.
* Creates StreamOptions for both InStream and OutStream to gather together the parameters they need.
* Extends InStream and OutStream to handle encryption.
* Changes InStream to use DiskRangeList instead of List<DiskRange>.
* Creates CryptoUtils with a method to create an IV based on the stream name.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/CryptoUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/CryptoUtils.java
@@ -31,13 +31,14 @@ public class CryptoUtils {
   private static final int STRIPE_ID_LENGTH = 3;
   private static final int MIN_COUNT_BYTES = 8;
 
-  private static final SecureRandom random = new SecureRandom();
   static final int MAX_COLUMN = 0xffffff;
   static final int MAX_KIND = 0xffff;
   static final int MAX_STRIPE = 0xffffff;
 
   /**
    * Create a unique IV for each stream within a single key.
+   * The top bytes are set with the column, stream kind, and stripe id and the
+   * lower 8 bytes are always 0.
    * @param name the stream name
    * @param stripeId the stripe id
    * @return the iv for the stream
--- a/java/core/src/java/org/apache/orc/impl/CryptoUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/CryptoUtils.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.orc.impl;
+
+import org.apache.orc.EncryptionAlgorithm;
+import java.security.SecureRandom;
+
+/**
+ * This class has routines to work with encryption within ORC files.
+ */
+public class CryptoUtils {
+
+  private static final int COLUMN_ID_LENGTH = 3;
+  private static final int KIND_LENGTH = 2;
+  private static final int STRIPE_ID_LENGTH = 3;
+  private static final int MIN_COUNT_BYTES = 8;
+
+  private static final SecureRandom random = new SecureRandom();
+  static final int MAX_COLUMN = 0xffffff;
+  static final int MAX_KIND = 0xffff;
+  static final int MAX_STRIPE = 0xffffff;
+
+  /**
+   * Create a unique IV for each stream within a single key.
+   * @param name the stream name
+   * @param stripeId the stripe id
+   * @return the iv for the stream
+   */
+  public static byte[] createIvForStream(EncryptionAlgorithm algorithm,
+                                         StreamName name,
+                                         int stripeId) {
+    byte[] iv = new byte[algorithm.getIvLength()];
+    int columnId = name.getColumn();
+    if (columnId < 0 || columnId > MAX_COLUMN) {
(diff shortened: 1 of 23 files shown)


java/core/src/java/org/apache/orc/impl/CryptoUtils.java
```diff
@@ -0,0 +1,77 @@
...
+ * This class has routines to work with encryption within ORC files.
+ */
+public class CryptoUtils {
+
+  private static final int COLUMN_ID_LENGTH = 3;
+  private static final int KIND_LENGTH = 2;
+  private static final int STRIPE_ID_LENGTH = 3;
+  private static final int MIN_COUNT_BYTES = 8;
+
+  private static final SecureRandom random = new SecureRandom();
```
prasanthj: unused?

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -173,33 +203,208 @@ private static ByteBuffer allocateBuffer(int size, boolean isDirect) {
...
+    }
+
+    /**
+     * We are seeking to a new range, so update the cipher to change the IV
+     * to match. This code assumes that we only support encryption in CTR mode.
+     * @param offset where we are seeking to in the stream
+     */
+    void changeIv(long offset) {
+      int blockSize = cipher.getBlockSize();
+      long encryptionBlocks = offset / blockSize;
```
prasanthj: Add +1 if there is extra?

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -173,33 +203,208 @@ private static ByteBuffer allocateBuffer(int size, boolean isDirect) {
...
+      this.name = name;
+      algorithm = options.algorithm;
+      key = options.key;
+      iv = options.iv;
+      cipher = algorithm.createCipher();
+    }
+
+    /**
+     * We are seeking to a new range, so update the cipher to change the IV
+     * to match. This code assumes that we only support encryption in CTR mode.
```
prasanthj: Comment about how the bottom 8 bytes of IV will be filled (block counter) will be helpful.

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -401,84 +596,149 @@ private String rangeString() {
...
+
+    @Override
+    public StreamOptions clone() {
+      try {
+        StreamOptions clone = (StreamOptions) super.clone();
+        if (clone.codec != null) {
+          // Make sure we don't share the same codec between two readers.
+          clone.codec = OrcCodecPool.getCodec(codec.getKind());
+        }
+        return clone;
```
prasanthj: why is this clone required? incomplete clone anyways

java/core/src/java/org/apache/orc/impl/writer/StreamOptions.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.apache.orc.CompressionCodec;
+import org.apache.orc.EncryptionAlgorithm;
+
+import java.security.Key;
+
+/**
+ * The compression and encryption options for writing a stream.
+ */
+public class StreamOptions {
```
prasanthj: Is this duplicate class? I saw StreamOptions class above as well.

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -55,112 +60,137 @@ public long getStreamLength() {
...
       currentOffset += 1;
-      return 0xff & range.get();
+      return 0xff & decrypted.get();
+    }
+
+    protected void setCurrent(DiskRangeList newRange, boolean isJump) {
+      currentRange = newRange;
+      if (newRange != null) {
+        decrypted = newRange.getData().slice();
+        decrypted.position((int) (currentOffset - newRange.getOffset()));
```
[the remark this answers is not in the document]
omalley: No, it is moving the ByteBuffer position to the correct spot so that it matches currentOffset, which is relative to the stream.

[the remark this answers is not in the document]
omalley: But I'll add a comment to the code.

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -55,112 +60,137 @@ public long getStreamLength() {
...
-          return;
-        }
-        ++i;
-      }
-      // if they are seeking to the precise end, go ahead and let them go there
-      int segments = bytes.size();
-      if (segments != 0 && desired == bytes.get(segments - 1).getEnd()) {
+      // If we are seeking inside of the current range, just reposition.
+      if (currentRange != null && desired >= currentRange.getOffset() &&
+          desired < currentRange.getEnd()) {
```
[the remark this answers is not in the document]
omalley: No, the currentRange.getEnd() is the position after the range (offset + length), so to see if an offset is in the currentRange, I need to use <.

java/core/src/java/org/apache/orc/impl/CryptoUtils.java
omalley: oops. Removed.

java/core/src/java/org/apache/orc/impl/InStream.java
omalley: No, we always start decryption on a block boundary and throw away the leading bytes. See InStream.EncryptionState.changeIv(long offset) and the "extra" and "wasted" variables.

java/core/src/java/org/apache/orc/impl/writer/StreamOptions.java
omalley: It is org.apache.orc.impl.InStream.StreamOptions and org.apache.orc.impl.writer.StreamOptions. I guess they should be named the same way, but they are separate classes.

java/core/src/java/org/apache/orc/impl/InStream.java
omalley: Ok, I added a comment in CryptoUtils.createIvForStream. In InStream, the IV could be anything they provide although ORC will always have 0 in the bottom 8 bytes.

java/core/src/java/org/apache/orc/impl/InStream.java
omalley: *sigh* It is Cloneable, because it is used as a field in DefaultDataReader. The clone is correct, I believe. We really need a better solution for CompressionCodec than the current Cloneable and CompressionCodecPool mess. We need to remove the compression parameters out of Codec and then they are completely stateless again and can be trivially pooled.

PR #279, opened 2018-06-04
ORC-373: Option to disable dictionary encoding

prasanthj:
disables dictionary encoding at the time of string tree writer initialization.


omalley:
Prasanth, rather than adding the flush count, which is really test code in the main source, how about creating a StringTreeWriter and catching the stream directly. I'd propose this - https://github.com/omalley/orc/blob/8a782d204f071e2bd16bcfd591a54319fadff132/java/core/src/test/org/apache/orc/TestStringDictionary.java#L251

prasanthj:
lgtm

PR #280, opened 2018-06-06
ORC-376: Add Ubuntu18 docker file

omalley:
We need to be able to test gcc 7 and ubuntu 18.04.


omalley:
I'll also comment that the default openjdk is 10, which does not work. I've set the docker file to use openjdk 8 instead, which does work. With openjdk 10 maven crashes with a NullPointerException trying to compile ORC.

PR #281, opened 2018-06-06
ORC-375: Update libhdfs to compile with gcc7.

omalley:
This changes the code in the libhdfs tarball, but we need this change applied upstream.


jamesclampffer:
@omalley I have a patch posted at https://issues.apache.org/jira/browse/HDFS-13534 but need another hadoop committer to review before I can push it.  Do you want to take a look?

omalley:
I accidentally pushed the temporary commit to branch-1.5. I've done a force push to undo the change.

PR #289, opened 2018-06-27
ORC-384 fix memory leak when loading non-ORC files

martinrupp:
Change-Id: Ia32f687f12f66c26b807200f90dd7ddcc5f27f8d


majetideepak:
+1 LGTM. I will check this in by end of today.

PR #290, opened 2018-07-11
ORC-386 Add spark benchmarks.

omalley:
I also refactored all of the old benchmarks to reduce the common code. I also split it into three modules so that I could
separate the common code, the code that depends on hive, and the code that depends on spark. Avoiding building an uber
jar that has both hive and spark made life much easier.


omalley:
@dongjoon-hyun if you have feedback on these benchmarks, that would be great.

PR #296, opened 2018-07-27
[ORC-391][c++] parseType does not accept underscore in the field name

czxrrr:


majetideepak:
JIRA name and a test please!

majetideepak:
+1 LGTM. Thanks.

PR #299, opened 2018-08-07
ORC-203 - Update StringStatistics to trim long strings to 1024 characters & record they were trimmed

moresandeep:
Reopening the PR.


java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -683,6 +787,150 @@ public int hashCode() {
...
+
+        for(int i=0; i<5; i++) {
+
+          byte b = text[truncateLen];
+          /* check for the beginning of 1,2,3,4,5 bytes long char */
+          offset = getCharLength(b);
+
+          /* found beginning of a valid char */
+          if(offset > 0) {
+            byte[] truncated = Arrays.copyOfRange(text, 0, (truncateLen));
```
omalley: If you do:
Text result = new Text();
result.set(text, 0, truncateLen);
return result;

you'll have 1 less copy of the bytes.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -683,6 +787,150 @@ public int hashCode() {
...
+                byteBuffer.limit());
+
+            final byte[] result = new byte[lastCharPosition + bytes.length];
+
+            /* copy truncated array minus last char */
+            System.arraycopy(text, 0, result, 0, lastCharPosition);
+            /* copy last char */
+            System.arraycopy(bytes, 0, result, lastCharPosition, bytes.length);
+
+            return new Text(result);
```
omalley: A better pattern is:

Text result = new Text();
result.setCapacity(lastCharPosition + bytes.length);
result.set(text, 0, lastCharPosition);
result.append(bytes, 0, bytes.length);
return result;

moresandeep:
Hello @omalley 
Thanks for the review ! I made the suggested changes, the changes are on a new commit (there are two commits), I am doing this to save the review history, let me know if you prefer single commit.

omalley:
Please make a couple more changes:
* Revert the change to java/bench/README.md.
* We should add a new WriterVersion (ORC_203) so that if there is a bug in the new code, we can try to recover in the reader.

moresandeep:
@omalley I updated the PR with the suggested changes, updated the WriterVersion and put back README.

omalley:
I simplified the code to get the upper bound. I also added 4 test cases to test the different number of bytes in the utf-8 characters.
See https://github.com/omalley/orc/tree/orc-203

PR #300, opened 2018-08-10
[ORC-394][C++] Add addUserMetadata() function to C++ write

czxrrr:
This commit aims to provide API to add Usermetadata into FileFooter.


c++/test/TestWriter.cc
```diff
@@ -1170,5 +1170,52 @@ namespace orc {
     }
   }
 
+  TEST_P(WriterTest, writeUserMetadata) {
```
majetideepak: can you add the user metadata check to an existing test?

xndai:
Looks good. Thanks for adding this.

PR #304, opened 2018-08-28
ORC-397. Allow selective disabling of dictionary encoding.

omalley:
I've forward ported the patch from Mithun.


omalley:
Oh, rather than fix WriterImplV2 I simplified it to remove the duplicated code.

omalley:
I pushed an update that simplified the test case a bit.

java/core/src/test/org/apache/orc/TestStringDictionary.java
```diff
@@ -409,4 +411,77 @@ public void testTooManyDistinctV11AlwaysDictionary() throws Exception {
...
+
+    TypeDescription schema =
+        TypeDescription.fromString("struct<shortString:string,longString:string>");
+
+    Writer writer = OrcFile.createWriter(
+        testFilePath,
+        OrcFile.writerOptions(conf).setSchema(schema)
+            .compress(CompressionKind.NONE)
+            .bufferSize(10000)
+            .directEncodingColumns("longString"));
```
wgtmac: Is it better to support specifying columns which use dictionary encoding?

omalley: We shouldn't change the default behavior of deciding between direct/dictionary encoding based on the data. With the parameter being the columns that are being forced to be direct, the default is the empty string, which is straight-forward.

Does that make sense?

wgtmac: That makes sense. I will also port current dictionary encoding to C++ writer shortly.
BTW, we plan to do some testing about global dictionary which is shared by all stripes in that file. Can we come up with a design in ORC V2? I can propose a prototype after gathering certain experiment results.

PR #305, opened 2018-08-31
ORC-399. Move Java compiler to version 8.

omalley:


omalley:
It looks like the Calcite pom were pulling in some jar that is missing from Maven Central.

PR #308, opened 2018-09-13
Covert tool should create a lowercase schema

ddrinka:
Mixed-case struct field names don't work in Hive.  There should be a way to convert a camel-cased JSON document into ORC without having to pre-process the JSON.

This pull request is a proof-of-concept which generates two schemas, one using the default case which is provided to the JsonReader as usual, and another schema which is lower cased and is provided to OrcFile.

TypeDescription is immutable and non-trivial to manually clone using public accessors, so to make the idea clear, I do the conversion at schema ingest rather than where it's provided to OrcFile.  The downside of this approach is that automatic schema detection doesn't benefit from these changes.  A more experienced implementer could certainly do better.

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/convert/ConvertTool.java
+++ b/java/tools/src/java/org/apache/orc/tools/convert/ConvertTool.java
@@ -51,6 +51,7 @@ public class ConvertTool {
 
   private final List<FileInformation> fileList;
   private final TypeDescription schema;
+  private final TypeDescription lowerSchema;
   private final char csvSeparator;
   private final char csvQuote;
   private final char csvEscape;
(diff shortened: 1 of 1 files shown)


omalley:
Actually, this is doesn't match my experience. Hive doesn't support uppercase letters in the top level column names, but it does support them in sub-structs. Furthermore, ORC from other systems can handle the uppercase letters.

You might also try setting "orc.schema.evolution.case.sensitive" to false to get name-based matching of the ORC types that isn't case sensitive.

In terms of this patch, can you add a parameter that allows the user to downcase the column names?

omalley:
Thinking about this more, I'd like to suggest that if the user asks for a lowercase schema that when the fields of a struct are parsed, they are lowercased there. That will prevent "aa" and "AA" both being converted to "aa", but not being merged into a single field.

PR #309, opened 2018-09-15
ORC-403: [C++] Add checks to avoid invalid offsets in InputStream

stiga-huang:
We should check offsets and lengths got from the protobuf objects. The attached files in the JIRA can reveal the errors.


c++/src/Reader.cc
```diff
@@ -498,6 +498,12 @@ namespace orc {
       const proto::Stream& stream = currentStripeFooter.streams(i);
       uint64_t length = static_cast<uint64_t>(stream.length());
       if (static_cast<StreamKind>(stream.kind()) == StreamKind::StreamKind_ROW_INDEX) {
+        if (offset + length > fileLength) {
```
omalley: This check is really good, but it would also be nice to check that the stream is within the stripe, although you'd need to pass that in also. We could  pass the whole proto::StripeInformation in.

stiga-huang: Yes! I'll add more strict checks.

PR #319, opened 2018-10-09
ORC-415: [C++] Fix writing ColumnStatistics

wgtmac:
Fix two issues below:
1. A new batch may override previous batch's has_null info of colIndexStatistics if the new batch has no null but the previous batch has at least one null values.
2. If ColumnStatistics does not have any not-null data, it has no sum/min/max infos and this results in writing generic ColumnStatistics in the proto-buf serialization.

### DIFF
--- a/c++/src/ColumnWriter.cc
+++ b/c++/src/ColumnWriter.cc
@@ -290,8 +290,10 @@ namespace orc {
           colIndexStatistics->increase(1);
         } else if (!hasNull) {
           hasNull = true;
-          colIndexStatistics->setHasNull(true);
         }
+      }
+      if (hasNull) {
+        colIndexStatistics->setHasNull(true);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


c++/src/ColumnWriter.cc
```diff
@@ -280,20 +280,20 @@ namespace orc {
...
       colIndexStatistics->increase(numValues);
     } else {
+      bool hasNull = false;
       const char* notNull = structBatch->notNull.data() + offset;
       for (uint64_t i = 0; i < numValues; ++i) {
         if (notNull[i]) {
           colIndexStatistics->increase(1);
         } else if (!hasNull) {
           hasNull = true;
+          colIndexStatistics->setHasNull(true);
```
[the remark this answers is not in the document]
wgtmac: I agree. Now I have moved them out of the for-loop. Please check again. Thanks!

c++/src/ColumnWriter.cc
```diff
@@ -1094,7 +1109,7 @@ namespace orc {
       throw InvalidArgument("Failed to cast to BinaryColumnStatisticsImpl");
     }
 
-    bool hasNull = false;
+    bool hasNull = true;
```
majetideepak: why is this initialized to `true`?

wgtmac: sorry about that. fixed...

PR #320, opened 2018-10-10
ORC-416: Avoid opening data reader when there is no stripe

dongjoon-hyun:
Currently, `RecordReaderImpl` invokes `dataReader.open` in the middle of the constructor. We can postpone this until `advanceToNextRow` reads the stripe at the end of the constructor.
```
this.dataReader.open();
```
This will reduce the chance of potential open-file leakages due to IOException and OOM during `RecordReaderImpl` construction. Also, we can avoid it completely when there is no stripe in the file.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -249,7 +249,6 @@ public class RecordReaderImpl implements RecordReader {
               .withZeroCopy(zeroCopy)
               .build());
     }
-    this.dataReader.open();
     firstRow = skippedRows;
     totalRowCount = rows;
     Boolean skipCorrupt = options.getSkipCorruptRecords();
--- a/java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
+++ b/java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
@@ -25,6 +25,7 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.atLeastOnce;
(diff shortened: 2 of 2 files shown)


java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -249,7 +249,6 @@ protected RecordReaderImpl(ReaderImpl fileReader,
               .withZeroCopy(zeroCopy)
               .build());
     }
-    this.dataReader.open();
```
dongjoon-hyun: At line 273, `advanceToNextRow(reader, 0L, true);` will invoke this via `readStripeFooter`.

dongjoon-hyun:
Hi, @omalley .
Could you review this PR?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
zsxwing: @dongjoon-hyun If a read operation after the open method is interrupted, will dataReader be closed automatically (such as in a `finally` block)?

dongjoon-hyun: For now, this PR doesn't aim to change `close` behavior. This only removes the unnecessary opened period.

dongjoon-hyun:
Thank you for review and merging, @omalley ! Also, thank you for reivew, @zsxwing .

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
zsxwing: So this is not going to resolve https://issues.apache.org/jira/browse/SPARK-23390 Right?

dongjoon-hyun: For Spark thing, let's talk on Spark-side.

PR #325, opened 2018-10-17
ORC-419: Ensure to call `close` at RecordReaderImpl constructor exception

dongjoon-hyun:
Currently, there is no way to call `RecordReaderImpl.close` if `RecordReader.rows` fails at creating `RecordReaderImpl`. This PR aims to try to invoke `close` in the constructor in case of Exceptions.

```java
  @Override
  public RecordReader rows(Options options) throws IOException {
    LOG.info("Reading ORC rows from " + path + " with " + options);
    return new RecordReaderImpl(this, options);
  }
```


dongjoon-hyun:
Thank you so much!

PR #326, opened 2018-10-19
ORC-420: [C++] Implement string dictionary encoding for C++ writer

wgtmac:
Add dictionary encoding support for string, char and varchar types. Under the hood, std::map is used to be the sorted-dictionary implementation.

### DIFF
--- a/c++/src/ColumnWriter.cc
+++ b/c++/src/ColumnWriter.cc
@@ -204,6 +204,10 @@ namespace orc {
     }
   }
 
+  void ColumnWriter::writeDictionary() {
+    // PASS
+  }
+
   class StructColumnWriter : public ColumnWriter {
   public:
     StructColumnWriter(
@@ -236,6 +240,8 @@ namespace orc {
 
     virtual void writeIndex(
       std::vector<proto::Stream> &streams) const override;
+
+    virtual void writeDictionary() override;
 
     virtual void reset() override;
 
@@ -379,6 +385,12 @@ namespace orc {
 
     for (uint32_t i = 0; i < children.size(); ++i) {
       children[i]->reset();
+    }
+  }
+
+  void StructColumnWriter::writeDictionary() {
+    for (uint32_t i = 0; i < children.size(); ++i) {
+      children[i]->writeDictionary();
     }
   }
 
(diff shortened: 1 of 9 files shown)


wgtmac:
@majetideepak Can you take a look at this and #301 when you have time? Thanks!

c++/src/ColumnWriter.cc
```diff
@@ -831,6 +843,121 @@ namespace orc {
...
+
+    // return total length of strings in the dictioanry
+    uint64_t length() const;
+
+    void clear();
+
+  private:
+    struct LessThan {
+      bool operator()(const DictEntry& left, const DictEntry& right) const {
+        int ret = memcmp(left.data, right.data, std::min(left.length, right.length));
```
majetideepak: Why not use `strcmp`?

c++/src/ColumnWriter.cc
```diff
@@ -831,6 +843,121 @@ namespace orc {
...
+    uint64_t totalLength;
+  };
+
+  // insert a new string into dictionary, return its insertion order
+  size_t StringDictionary::insert(const char * str, size_t len) {
+    auto ret = dict.insert({DictEntry(str, len), dict.size()});
+    if (ret.second) {
+      // make a copy to internal storage
+      data.push_back(std::vector<char>(len));
+      memcpy(data.back().data(), str, len);
```
majetideepak: `strcmp` ?

c++/src/ColumnWriter.cc
```diff
@@ -831,6 +843,121 @@ namespace orc {
...
+    std::vector<std::vector<char>> data;
+    uint64_t totalLength;
+  };
+
+  // insert a new string into dictionary, return its insertion order
+  size_t StringDictionary::insert(const char * str, size_t len) {
+    auto ret = dict.insert({DictEntry(str, len), dict.size()});
+    if (ret.second) {
+      // make a copy to internal storage
+      data.push_back(std::vector<char>(len));
```
majetideepak: why not populate the vector first and push_back?

c++/src/ColumnWriter.cc
```diff
@@ -831,6 +843,121 @@ namespace orc {
...
+        int ret = memcmp(left.data, right.data, std::min(left.length, right.length));
+        if (ret != 0) {
+          return ret < 0;
+        }
+        return left.length < right.length;
+      }
+    };
+
+    std::map<DictEntry, size_t, LessThan> dict;
+    std::vector<std::vector<char>> data;
```
majetideepak: Won't it simplify things if we use `std::string` instead of `std::vector<char>`?

c++/src/ColumnWriter.cc
```diff
@@ -831,6 +843,121 @@ namespace orc {
...
+
+    for (size_t i = 0; i != idxBuffer.size(); ++i) {
+      idxBuffer[i] = static_cast<int64_t>(
+        mapping[static_cast<size_t>(idxBuffer[i])]);
+    }
+  }
+
+  // get dict entries in insertion order
+  std::vector<const StringDictionary::DictEntry *>
+  StringDictionary::entriesInInsertionOrder() const {
```
majetideepak: make this `void StringDictionary::getEntriesInInsertionOrder(std::vector<const StringDictionary::DictEntry *>&)` ?
Avoids a `move` and don't have to depend on the compiler to optimize the return value.

c++/src/ColumnWriter.cc
wgtmac: I was using strcmp but it will stop when '\0' is met which will lead to incorrect result.

c++/src/ColumnWriter.cc
wgtmac: sure, will do that.

c++/src/ColumnWriter.cc
wgtmac: yes, I can change std::vector<std::vector<char>> to std::vector<std::string> once your review is done.

c++/src/ColumnWriter.cc
```diff
@@ -831,6 +843,121 @@ namespace orc {
...
+  void StringDictionary::flush(AppendOnlyBufferedStream * dataStream,
+                               RleEncoder * lengthEncoder) const {
+    for (auto it = dict.cbegin(); it != dict.cend(); ++it) {
+      dataStream->write(it->first.data, it->first.length);
+      lengthEncoder->write(static_cast<int64_t>(it->first.length));
+    }
+  }
+
+  // reorder input index buffer from insertion order to dictionary order
+  void StringDictionary::reorder(std::vector<int64_t>& idxBuffer) const {
```
majetideepak: It took me a while to understand that we are actually calculating the dictionary indexes of the input values. Is this correct?
Can we name this as `getIndices`?

c++/src/ColumnWriter.cc
```diff
@@ -850,27 +977,76 @@ namespace orc {
...
+     */
+    StringDictionary dictionary;
+    // whether or not dictionary checking is done
+    bool doneDictionaryCheck;
+    // whether or not it should be used
+    bool useDictionary;
+    // keys in the dictionary should not exceed this ratio
+    double dictSizeThreshold;
+    // record index of insertion order in the dictionary for not-null rows
+    std::vector<int64_t> idxInDictBuffer;
```
majetideepak: Shouldn't this live inside `StringDictionary`?

majetideepak: In my opinion, the StringColumnWriter must not have any state corresponding to the dictionary encoding.

wgtmac: sure, I will try to move these variables as many as I can to StringDictionary.

c++/src/ColumnWriter.cc
wgtmac: Yes. I will add more comments here. Thanks for your review!

c++/src/ColumnWriter.cc
wgtmac: std::string does not work here because its constructor will terminate copying when '\0' is met. this is caught by my test cases. so I decide to keep the old implementation.

wgtmac:
@majetideepak Please review my latest commit. Thanks a lot!

c++/src/ColumnWriter.cc
xndai: Again I think this is to work around strings that contains null terminators '\0'.

c++/src/ColumnWriter.cc
xndai: This reordering really complicates things a lot. As discussed, we are moving forward to make sorted dictionary optional. If that's the case, I am not sure if we want to stick to a sorted dictionary for now. The current reader does not rely on, or take advantage of, the sorting property anyway.

wgtmac:
@majetideepak Do you have the chance to review it again? Thanks!

majetideepak:
@wgtmac I will take a look this today/tomorrow. I missed the previous notification.

PR #327, opened 2018-10-20
ORC-421: Separate compression options from the CompressionCodec.

omalley:
Signed-off-by: Owen O'Malley <omalley@apache.org>

### DIFF
--- a/java/core/src/java/org/apache/orc/CompressionCodec.java
+++ b/java/core/src/java/org/apache/orc/CompressionCodec.java
@@ -23,26 +23,40 @@ import java.util.EnumSet;
 
 public interface CompressionCodec {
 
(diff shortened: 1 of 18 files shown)


omalley:
Basically, this patch does:
- Makes CompressionCodec.Option that controls compression options
- CompressionCodec.compress takes a CompressionCodec.Option
- Removes CompressionCodec.modify so that we don't clone the codec via modify
- Makes the StreamOptions.withCodec take a CompressionCodec.Option object
- Fix all of the callers to use the OutStream constructor that takes the StreamOptions.
- WriterImpl.getCustomizedCodec now returns a CompressionCodec.Option instead of a CompressionCodec.

PR #333, opened 2018-10-31
ORC-429: [C++] Refactor code in TypeImpl.cc

fangzheng:
1. In convertType() function: in the case of proto::Type_Kind_STRUCT, two vectors are created but never used. They shall be removed.

2. In TypeImpl::parseType() function: the function calls input.substr() to copy the substring before parsing it. This string copy can be avoided by directly parsing on the input string.


xndai:
Change looks good to me. I will commit this for you if I don't see any other feedback.

fangzheng:
Thanks, Xiening!

PR #337, opened 2018-11-08
ORC-434: Fix incorrect documentation for Date/Timestamp ColumnStatistics

wgtmac:
The old specification says Date and Timestamp ColumnStatistics stores the difference from epoch 1/1/2015. This is incorrect and should be changed to UNIX epoch time.

### DIFF
--- a/site/specification/ORCv0.md
+++ b/site/specification/ORCv0.md
@@ -297,7 +297,7 @@ message DecimalStatistics {
 ```
 
 Date columns record the minimum and maximum values as the number of
-days since the epoch (1/1/2015).
+days since the UNIX epoch (1/1/1970 in UTC).
 
 ```
 message DateStatistics {
@@ -308,7 +308,7 @@ message DateStatistics {
 ```
 
 Timestamp columns record the minimum and maximum values as the number of
-milliseconds since the epoch (1/1/2015).
+milliseconds since the UNIX epoch (1/1/1970 00:00:00 in UTC).
 
 ```
 message TimestampStatistics {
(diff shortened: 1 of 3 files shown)


wgtmac:
@omalley Can you please help verify this doc change? Thanks!

site/specification/ORCv0.md
```diff
@@ -297,7 +297,7 @@ message DecimalStatistics {
 ```
 
 Date columns record the minimum and maximum values as the number of
-days since the epoch (1/1/2015).
+days since the UNIX epoch (1/1/1970 in UTC).
```
omalley: This is correct.

site/specification/ORCv0.md
```diff
@@ -308,7 +308,7 @@ message DateStatistics {
 ```
 
 Timestamp columns record the minimum and maximum values as the number of
-milliseconds since the epoch (1/1/2015).
+milliseconds since the UNIX epoch (1/1/1970 00:00:00 in UTC).
```
omalley: This is wrong. Timestamps are millis since 1/1/2015, until we added minimumUtc and maximumUtc in ORC-135 f2b8b799c .

wgtmac: Thanks for pointing this out! I think ORC-135 belongs to ORCv1 only; therefore I will only fix ORCv1.md and leave ORCv0.md as is.

wgtmac:
Though we can still write v0 format, ORC-135 fix is not in it. IIUC, I believe it is OK not to mention this in ORCv0 doc.

PR #346, opened 2018-12-04
ORC-445: [C++] Code improvements in RLEV2Util.

fangzheng:


c++/src/RLEV2Util.hh
```diff
@@ -23,85 +23,24 @@
...
-    } else if (n <= 40) {
-      return 40;
-    } else if (n <= 48) {
-      return 48;
-    } else if (n <= 56) {
-      return 56;
-    } else {
-      return 64;
-    }
+    return ClosestFixedBitsMap[n];
```
xndai: what if n > 64?

fangzheng: @xndai Hi Xiening, thanks for the catch. When computing patch width, it is possible for the input n to be larger than 64. I have added boundary check in this function and getClosestAlignedFixedBits().

wgtmac:
Can you please provide some perf data to confirm the refactoring is effective?

c++/src/RLEV2Util.hh
```diff
@@ -23,85 +23,32 @@
 
 namespace orc {
   extern const uint32_t FBSToBitWidthMap[FixedBitSizes::SIZE];
+  extern const uint32_t ClosestFixedBitsMap[65];
+  extern const uint32_t ClosestAlignedFixedBitsMap[65];
+  extern const uint32_t BitWidthToFBSMap[65];
 
   inline uint32_t decodeBitWidth(uint32_t n) {
     return FBSToBitWidthMap[n];
```
wgtmac: make sure this won't be out of bound.

fangzheng:
@wgtmac Hi Gang, thanks for the comments. I'll add performance measurements later this week.

c++/src/RLEV2Util.hh
fangzheng: Hi Gang, I thought this over and don't feel we need to add boundary check here for two reasons:
1. in current RLEv2 encoder and decoder code, all the callers of decodeBitWidth() pass in a integer that only has the lowest 5 bits set, so the input n is always less than FixedBitSizes::SIZE (32).
2. Since this function is used to decode the 5-bit length code, it would be a programming error for a caller to pass in any value that is >= 32. In this case, returning 64 (as the original implementation does) is not helping. That would just silently mask the error (potentially file content corruption) and is likely to cause further error when decoding with the wrong bit width.

PR #347, opened 2018-12-04
Orc 446: [C++] Add possibility to use prebuilt (windows) zstd lib

rip-nsk:


c++/src/Compression.cc
```diff
@@ -32,6 +32,10 @@
 
 #include "wrap/snappy-wrapper.h"
 
+#ifndef ZSTD_CLEVEL_DEFAULT
+#define ZSTD_CLEVEL_DEFAULT 3
```
xndai: Where is this new macro used?

PR #348, opened 2018-12-12
ORC-422 - Fix issue with Predicate push down when lower/upper bounds are set

moresandeep:
This PR fixes a bug where Predicate push down will not return proper values when upper and lower bounds are set for StringStatisticsImpl.
This patch also contains removing some unused imports as part of cleanup.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
@@ -35,8 +35,13 @@ import org.apache.orc.StringColumnStatistics;
 import org.apache.orc.TimestampColumnStatistics;
 import org.apache.orc.TypeDescription;
 
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.sql.Date;
 import java.sql.Timestamp;
+import java.util.Arrays;
 import java.util.TimeZone;
 
 public class ColumnStatisticsImpl implements ColumnStatistics {
--- a/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
@@ -677,7 +677,7 @@ public class ColumnStatisticsImpl implements ColumnStatistics {
 
     @Override
     public String getMaximum() {
-      /* if we have upper bound is set (in case of truncation)
+      /* if we have upper bound set (in case of truncation)
       getMaximum will be null */
       if(isUpperBoundSet) {
         return null;
--- a/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
@@ -35,13 +35,8 @@ import org.apache.orc.StringColumnStatistics;
 import org.apache.orc.TimestampColumnStatistics;
 import org.apache.orc.TypeDescription;
 
-import java.nio.ByteBuffer;
-import java.nio.CharBuffer;
-import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 import java.sql.Date;
 import java.sql.Timestamp;
-import java.util.Arrays;
 import java.util.TimeZone;
 
 public class ColumnStatisticsImpl implements ColumnStatistics {
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -52,7 +52,6 @@ import org.apache.orc.TypeDescription;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.hive.common.io.DiskRange;
 import org.apache.hadoop.hive.common.io.DiskRangeList;
 import org.apache.hadoop.hive.common.io.DiskRangeList.CreateHelper;
 import org.apache.hadoop.hive.common.type.HiveDecimal;
@@ -359,7 +358,7 @@ public class RecordReaderImpl implements RecordReader {
     } else if (index instanceof DoubleColumnStatistics) {
       return ((DoubleColumnStatistics) index).getMaximum();
     } else if (index instanceof StringColumnStatistics) {
-      return ((StringColumnStatistics) index).getMaximum();
+      return ((StringColumnStatistics) index).getUpperBound();
     } else if (index instanceof DateColumnStatistics) {
       return ((DateColumnStatistics) index).getMaximum();
     } else if (index instanceof DecimalColumnStatistics) {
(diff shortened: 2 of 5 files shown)


java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -35,13 +35,8 @@
 import org.apache.orc.TimestampColumnStatistics;
 import org.apache.orc.TypeDescription;
 
-import java.nio.ByteBuffer;
-import java.nio.CharBuffer;
-import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 import java.sql.Date;
 import java.sql.Timestamp;
-import java.util.Arrays;
```
dongjoon-hyun: Shall we exclude these unrelated changes? It makes backports difficult sometimes.

java/core/src/test/org/apache/orc/TestColumnStatistics.java
```diff
@@ -37,7 +37,6 @@
 import java.sql.Timestamp;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.Arrays;
```
dongjoon-hyun: ditto.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
moresandeep: Good point, but do we need to carry on with un-used imports for upcoming releases, they can accumulate over time. Also, for backporting it could be  relatively easy to add these dependencies (I am assuming manual backports). Let me know if you feel strongly about it.

java/core/src/test/org/apache/orc/TestColumnStatistics.java
moresandeep: Same as above.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
dongjoon-hyun: We had better split that kind of cosmetic commits and critical bug fix commits.

java/core/src/test/org/apache/orc/TestColumnStatistics.java
dongjoon-hyun: Usually, it's better to clean up this if you change the actual code in this file.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
moresandeep: @dongjoon-hyun I reverted these changes.

java/core/src/test/org/apache/orc/TestColumnStatistics.java
moresandeep: @dongjoon-hyun, the changes are undone.

moresandeep:
Hello @omalley I am updating a work in progress patch for you to look at. I have a bunch of lines commented in there which I'll take out in the final patch and cleanup the code, apologies for the rough code, I don't want to delay this patch further. 

Basically, the crux of the patch is in file RecordReaderImpl line # 325 and # 335, what I am doing here is depending on whether upperbound or lowerbound is set for corner cases where the predicate literal equals upperbound or lowerbound I am picking the appropriate value. 
e.g. 
```
/* since min value is truncated when we have compare=0, it means the predicate string is BEFORE the min value*/
    else if (minCompare == 0 && isLowerBoundSet) {
      return Location.BEFORE;
    }
```
and 
   ```
 /* if upperbound is set then location here will be AFTER */
    else if (maxCompare == 0 && isUpperBoundSet) {
      return Location.AFTER;
    }
```
I also, tried to write up UnitTests to test all the corner case scenarios that I could think. I could not find a way test greater than since there is no Operator 'GREATER_THAN', code suggest using 'NOT' Operator, is there an example that I can look at for something like NOT LESS_THAN.

File: java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java [hunk 3]
```diff
@@ -318,16 +315,25 @@ public long getNext() {
    * @param <T> the type of the comparision
    * @return the location of the point
    */
-  static <T> Location compareToRange(Comparable<T> point, T min, T max) {
+  static <T> Location compareToRange(Comparable<T> point, T min, T max, boolean isLowerBoundSet, boolean isUpperBoundSet) {
```
omalley: I suggest that we instead do:
```static <T> Location compareToRange(Comparable<T> point, T min, T max, T lowerBound, T upperBound)```
so the normal case would look like:
```compareToRange(min, max, min, max);```
or in the more general case:
```compareToRange(stats.getMin(), stats.getMax(), stats.getLowerBound(), stat.getUpperBound());```

File: java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java [hunk 4]
```diff
@@ -341,7 +347,7 @@ public long getNext() {
    * @return the object for the maximum value or null if there isn't one
    */
   static Object getMax(ColumnStatistics index) {
-    return getMax(index, false);
+    return getMax(index, false, null);
```
omalley: This seems suspicious.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -577,19 +645,21 @@ private static boolean shouldEvaluateBloomFilter(PredicateLeaf predicate,
   private static TruthValue evaluatePredicateMinMax(PredicateLeaf predicate, Object predObj,
       Object minValue,
       Object maxValue,
-      boolean hasNull) {
+      boolean hasNull,
```
omalley: I think you should make the same change here.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -577,19 +645,21 @@ private static boolean shouldEvaluateBloomFilter(PredicateLeaf predicate,
...
+      boolean isLowerBoundSet,
+      boolean isUpperBoundSet) {
     Location loc;
 
     switch (predicate.getOperator()) {
       case NULL_SAFE_EQUALS:
-        loc = compareToRange((Comparable) predObj, minValue, maxValue);
+        loc = compareToRange((Comparable) predObj, minValue, maxValue, isLowerBoundSet, isUpperBoundSet);
```
omalley: You also need to deal with the IN sarg.

moresandeep: This should be fixed in the new commit.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
moresandeep: Should be fixed in the new commit.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
moresandeep: Sure, fixed this.

moresandeep:
Hello @omalley can you take a look at the new changes which addresses the fixes.

omalley:
I just committed this. Thanks, Sandeep!

PR #349, opened 2018-12-13
ORC-189. Add timestamp with local timezone.

omalley:


java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -1634,6 +1634,30 @@ public int hashCode() {
...
+    TimestampInstantStatisticsImpl() {
+    }
+
+    TimestampInstantStatisticsImpl(OrcProto.ColumnStatistics stats) {
+      super(stats);
+    }
+
+    @Override
+    public void updateTimestamp(Timestamp value) {
+      updateTimestamp(value.getTime());
```
prasanthj: do we always return the timestamp stats in local timezone? If so PPD on timestamp will still be broken right?

java/core/src/java/org/apache/orc/impl/writer/TimestampInstantTreeWriter.java
```diff
@@ -0,0 +1,160 @@
...
+      if (vector.noNulls || !vector.isNull[0]) {
+        // ignore the bottom three digits from the vec.time field
+        final long secs = vec.time[0] / MILLIS_PER_SECOND;
+        final int newNanos = vec.nanos[0];
+        // set the millis based on the top three digits of the nanos
+        long millis = secs * MILLIS_PER_SECOND + newNanos / 1_000_000;
+        if (millis < 0 && newNanos > 999_999) {
+          millis -= MILLIS_PER_SECOND;
+        }
+        indexStatistics.updateTimestamp(millis);
```
prasanthj: there seems to be some code duplication here. except for millis and seconds rest all looks the same between both timestamp writers. should the common code be factored into a common method? in case of any fixes will be easier to fix in one place. similarly for the readers.

java/core/src/java/org/apache/orc/impl/writer/TimestampInstantTreeWriter.java
```diff
@@ -0,0 +1,160 @@
...
+    this.isDirectV2 = isNewWriteFormat(writer);
+    this.seconds = createIntegerWriter(writer.createStream(id,
+        OrcProto.Stream.Kind.DATA), true, isDirectV2, writer);
+    this.nanos = createIntegerWriter(writer.createStream(id,
+        OrcProto.Stream.Kind.SECONDARY), false, isDirectV2, writer);
+    if (rowIndexPosition != null) {
+      recordPosition(rowIndexPosition);
+    }
+    DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+    dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
```
prasanthj: nit: UTC timezone can be static somewhere. repeated in multiple places.

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -2886,6 +2898,7 @@ public static boolean canConvert(TypeDescription fileType, TypeDescription reade
       case LONG:
       case DOUBLE:
       case TIMESTAMP:
+      case TIMESTAMP_INSTANT:
```
prasanthj: conversion between TIMESTAMP and TIMESTAMP_INSTANT seems to be missing?

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
wgtmac: I think this is correct. We will always compare the timestamp instant for the new type.

jcamachor:
@omalley , @prasanthj , would it be possible to get this in? Then we can implement the missing part in Hive to support instant semantics in ORC.

omalley:
Ok, I rewrote this patch.

Changes:
* The timestamp TreeReader/TreeWriter use the same code with a boolean on the constructor.
* Fixed the SchemaEvolution to work with the instant timestamps.
* Added a set of test cases to test the schema evolution. That brought up another set of problems:
  * The schema evolution wasn't being tested and wasn't documented.
  * Some of the conversions were very confusing.
  * I changed the previous semantics for timestamp -> long to use seconds instead of millis from 1970. This was based on the fact that timestamp -> decimal and double used seconds.
  * The conversions to timestamp were ignoring the option to useUtc.
  * The conversion to/from decimal were going through double and string, which loses precision.
  * The float conversions exactly mirror the double conversions, so I removed the timestamp ones rather than fixing them too. I created a new jira to refactor all of them.
* instant timestamp <-> string uses 'yyyy-mm-dd hh:mm:ss.s zone' where zone is the zone id. 
* statistics now depends on the type of the column, which took a little refactoring

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -1995,31 +1520,33 @@ public void nextVector(ColumnVector previousVector,
...
+                   context.getUseUTCTimestamp();
     }
 
     @Override
     public void setConvertVectorElement(int elementNum) {
-      long millis =
-          DateWritable.daysToMillis((int) longColVector.vector[elementNum]);
-      timestampColVector.set(elementNum, new Timestamp(millis));
+      long days = longColVector.vector[elementNum];
+      long millis = days * 24 * 60 * 60 * 1000;
```
prasanthj: "L" to avoid millis becoming negative?

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -1833,70 +1380,38 @@ public void nextVector(ColumnVector previousVector,
...
-      // The TimestampColumnVector will set the entry to null when a null timestamp is passed in.
-      timestampColVector.set(elementNum, timestampValue);
+      double seconds = doubleColVector.vector[elementNum];
+      if (!useUtc) {
+        seconds = SerializationUtils.convertFromUtc(local, seconds);
+      }
+      long wholeSec = (long) Math.floor(seconds);
+      timestampColVector.time[elementNum] = wholeSec * 1000;
+      timestampColVector.nanos[elementNum] =
+          1_000_000 * (int) Math.round((seconds - wholeSec) * 1000);
```
prasanthj: similarly add "L" to avoid int overflows.

PR #354, opened 2019-01-02
ORC-452: Support converting JSON object to MAP type

stiga-huang:
Add `MapColumnConverter` to support converting JSON object to MAP type.
Since JSON keys can only be strings, we only support MAP type with string keys.

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/convert/JsonReader.java
+++ b/java/tools/src/java/org/apache/orc/tools/convert/JsonReader.java
@@ -20,6 +20,7 @@ package org.apache.orc.tools.convert;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
 import com.google.gson.JsonStreamParser;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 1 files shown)


stiga-huang:
@omalley Would you please have a look for this.

stiga-huang:
Hi @wgtmac, do you have time to review this? Is there anyone I can invite too?

java/tools/src/java/org/apache/orc/tools/convert/JsonReader.java
```diff
@@ -222,6 +225,40 @@ public void convert(JsonElement value, ColumnVector vect, int row) {
     }
   }
 
+  static class MapColumnConverter implements JsonConverter {
+    private JsonConverter keyConverter;
+    private JsonConverter valueConverter;
+
+    public MapColumnConverter(TypeDescription schema) {
+      TypeDescription keyType = schema.getChildren().get(0);
+      if (keyType.getCategory() != TypeDescription.Category.STRING)
```
wgtmac: Why it can't support other types?

wgtmac: Got it. Thanks!

PR #356, opened 2019-01-18
ORC-461. Replace `git-wip-us` with `gitbox`

dongjoon-hyun:
This PR aims to replace the obsolete `git-wip-us` to `gitbox` URLs because `git-wip-us` is removed now.

PR #357, opened 2019-01-18
ORC-461. Replace `git-wip-us` with `gitbox`

dongjoon-hyun:
This PR aims to replace the obsolete git-wip-us to gitbox URLs because git-wip-us is removed now. And, this is a generate result from #356 .


dongjoon-hyun:
Since #356 is merged, I'll merge this generated result together.

PR #363, opened 2019-01-29
ORC-466: Show `mode` for `NoSuchPaddingException`

dongjoon-hyun:
Since `NoSuchPaddingException` is thrown when a particular padding mechanism is requested but is not available in the environment, we had better show `mode` instead of `algorithm` in the exception message.


prasanthj:
+1

dongjoon-hyun:
Thank you, @prasanthj .

PR #366, opened 2019-02-22
ORC-470: [C++] Fix BooleanRleDecoderImpl::skip() read over end of stream

wgtmac:
If boolean RLE input stream skips all remaining boolean values, we will get an exception since BooleanRLEDecoder still tries to read one last byte from underlying input stream even if the requested number of bits are multiple of 8 and then it reads over the end of stream. This patch fixes this by checking if BooleanRLEDecoder does not need to read one last byte.

### DIFF
--- a/c++/test/TestByteRle.cc
+++ b/c++/test/TestByteRle.cc
@@ -18,6 +18,7 @@
 
 #include "Adaptor.hh"
 #include "ByteRLE.hh"
+#include "Compression.hh"
 #include "OrcTest.hh"
 #include "wrap/gtest-wrapper.h"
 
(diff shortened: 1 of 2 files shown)


c++/test/TestByteRle.cc
```diff
@@ -1420,4 +1421,31 @@ TEST(BooleanRle, seekBoolAndByteRLE) {
   rle->next(value, 1, nullptr);
   EXPECT_EQ(num[45], value[0]);
   }
+
+  TEST(BooleanRle, seekAndSkipToEnd) {
+    const unsigned char buffer[] = {
```
xndai: Please add some comments regarding the layout of the test data, so the purpose of this test can be easily understood.

wgtmac: DONE

xndai: I would suggest using encoder to generate the test data in this case. It would be much easier to understand. The rest of the file uses literal strings for test. But those are legacy tests created before encoder was existed.

PR #367, opened 2019-02-22
ORC-469:[C++] Add EncodedStringVectorBatch to expose string dictionary in VectorBatch

yuruiz:


c++/include/orc/Vector.hh
```diff
@@ -51,6 +51,8 @@ namespace orc {
     // whether there are any null values
     bool hasNulls;
 
+    bool isEncoded;
```
wgtmac: add some comments to briefly describe what it means.

c++/include/orc/Vector.hh
```diff
@@ -113,6 +115,36 @@ namespace orc {
...
 
+  struct StringDictionary {
+    StringDictionary(MemoryPool& pool);
+    DataBuffer<char> dictionaryBlob;
+
+    // Offset for each dictionary key entry.
+    DataBuffer<int64_t> dictionaryOffset;
+
+    void getValueByIndex(int64_t index, char*& valPtr, int64_t& length)
+    {
```
wgtmac: move the curly brace to the the previous line to follow the coding style. same applies to code below

c++/src/ColumnWriter.cc
```diff
@@ -840,15 +840,15 @@ namespace orc {
   /**
    * Implementation of increasing sorted string dictionary
    */
-  class StringDictionary {
+  class StringDictionaryImpl {
```
wgtmac: Rename it to SortedStringDictionary?

c++/include/orc/Reader.hh
```diff
@@ -180,6 +180,16 @@ namespace orc {
      */
     RowReaderOptions& forcedScaleOnHive11Decimal(int32_t forcedScale);
 
+    /**
+     * Set enable encoding block mode.
+     */
+    RowReaderOptions& setEnableLazyDecoding(bool enable);
```
xndai: If this is for dictionary encoding, can we be more specific here? For example, setEnableLazyDictionaryDecoding?

c++/include/orc/Type.hh
```diff
@@ -64,7 +64,8 @@ namespace orc {
      * Create a row batch for this type.
      */
     virtual ORC_UNIQUE_PTR<ColumnVectorBatch> createRowBatch(uint64_t size,
-                                                             MemoryPool& pool
+                                                             MemoryPool& pool,
+                                                             bool encoded = false
```
xndai: rename into "dictionaryEncoded"?

c++/include/orc/Vector.hh
```diff
@@ -113,6 +115,35 @@ namespace orc {
     DataBuffer<int64_t> length;
   };
 
+  struct StringDictionary {
+    StringDictionary(MemoryPool& pool);
+    DataBuffer<char> dictionaryBlob;
+
+    // Offset for each dictionary key entry.
+    DataBuffer<int64_t> dictionaryOffset;
```
xndai: uint32_t? I assume 4G of dictionary size is more than enough.

c++/include/orc/Vector.hh
```diff
@@ -113,6 +115,35 @@ namespace orc {
...
+
+    void getValueByIndex(int64_t index, char*& valPtr, int64_t& length) {
```
xndai: index as uint32_t?

c++/src/ColumnReader.cc
```diff
@@ -582,6 +579,20 @@ namespace orc {
...
+                                                  uint64_t numValues,
+                                                  char* notNull) {
+    ColumnReader::next(rowBatch, numValues, notNull);
+    notNull = rowBatch.hasNulls ? rowBatch.notNull.data() : nullptr;
+    rowBatch.isEncoded = true;
+
+    EncodedStringVectorBatch& batch = dynamic_cast<EncodedStringVectorBatch&>(rowBatch);
+    batch.dictionary = this->dictionary;
+
+    // Length buffer is reused to save dictionary entry ids
```
xndai: You are not reusing, but create a new index vector, don't you?

c++/src/ColumnWriter.cc
```diff
@@ -840,15 +840,15 @@ namespace orc {
   /**
    * Implementation of increasing sorted string dictionary
    */
-  class StringDictionary {
+  class SortedStringDictionary {
```
xndai: This renaming doesn't seem to relate to the purpose of this change. Suggest separate it.

c++/test/TestColumnReader.cc
```diff
@@ -434,6 +434,89 @@ TEST(TestColumnReader, testDictionaryWithNulls) {
   }
 }
 
+TEST(TestColumnReader, testEncodedVectorWithNulls) {
```
xndai: Add test to cover case when dictionary encoding is not used, but lazy decoding option is set. Also cover composite types, such as list, map, etc.

c++/include/orc/Reader.hh
```diff
@@ -180,6 +180,16 @@ namespace orc {
      */
     RowReaderOptions& forcedScaleOnHive11Decimal(int32_t forcedScale);
 
+    /**
+     * Set enable encoding block mode.
```
xndai: Add more comments to explain what is "encoding block mode".

c++/include/orc/Vector.hh
```diff
@@ -113,6 +115,35 @@ namespace orc {
...
+      }
+
+      int64_t* offsetPtr = dictionaryOffset.data();
+
+      valPtr = dictionaryBlob.data() + offsetPtr[index];
+      length = offsetPtr[index + 1] - offsetPtr[index];
+    }
+  };
+
+  struct EncodedStringVectorBatch : public ColumnVectorBatch {
```
xndai: Add comment to explain the layout of EncodedStringVectorBatch.

c++/include/orc/Reader.hh
yuruiz: this is not only meant for dictionary decoding. This flag controls weather we use ColumnReader next() or nextEncoded() so I think we should leave it as is.

c++/include/orc/Type.hh
yuruiz: Same as above. I think we better leave it as encoded for future support like RLE lazy decoding.

c++/include/orc/Vector.hh
yuruiz: Right now the RleDecoder only support int64_t*

c++/include/orc/Vector.hh
yuruiz: same as above, currently the RleDecoder only support int64_t, so we make it int64_t here as well.

c++/src/ColumnReader.cc
yuruiz: yes

c++/src/ColumnWriter.cc
yuruiz: This change is to avoid the StringDictionary name conflicts cause we already declared StringDictionary in Vector.hh

PR #371, opened 2019-03-06
ORC-475: ORC reader should lazily get filesystem

prasanthj:


java/core/src/java/org/apache/orc/impl/ReaderImpl.java
```diff
@@ -365,6 +364,14 @@ public ReaderImpl(Path path, OrcFile.ReaderOptions options) throws IOException {
     this.schema = OrcUtils.convertTypeFromProtobuf(this.types, 0);
   }
 
+  protected FileSystem getFileSystem() throws IOException {
```
t3rmin4t0r: Two calls to getFileSystem() will do this 2x - have this init this.fileSystem inline.

t3rmin4t0r: Also use the options.filesystem if it provided

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
```diff
@@ -341,7 +338,9 @@ public ReaderImpl(Path path, OrcFile.ReaderOptions options) throws IOException {
     } else {
       OrcTail orcTail = options.getOrcTail();
       if (orcTail == null) {
-        tail = extractFileTail(fs, path, options.getMaxLength());
+        this.fileSystem = getFileSystem();
+        options.filesystem(fileSystem);
```
t3rmin4t0r: options fs object got lost - it's not clear that was null before.

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
prasanthj: fixed it.

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
prasanthj: fixed it.

PR #372, opened 2019-03-06
[ORC-476] Make SearchAgument kryo buffer size configurable

dhruve:


dhruve:
@omalley Can you review this. If the changes are acceptable, I will have them against the master branch as well. Thanks.

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -136,6 +136,10 @@
           "being the first column, 1 being the next, and so on. ."),
   KRYO_SARG("orc.kryo.sarg", "orc.kryo.sarg", null,
       "The kryo and base64 encoded SearchArgument for predicate pushdown."),
+  KRYO_SARG_BUFFER("orc.kryo.sarg.buffer", null, 4096,
```
omalley: How about rather than introducing two options, we just introduce KRYO_SARG_BUFFER and define a reasonable max? I'd propose that the buffer default to 8192. We can set the max to 16777216, which means in the worst case the buffer will double 11 times.

dhruve:
@omalley I see you merged the PR and also backported it to 1.5 and 1.4. Thank you.

omalley:
After I thought about it, I realized that I wanted this to land in the 1.5.5 release candidate that I was rolling. I therefore went ahead and made the changes that I was proposing. I hope it is ok with you and it satisfies your use case.

dhruve:
I am okay with the changes being merged and our use case will be satisfied with the change. I however don't know if we could ever end up with a sarg exceeding more than 16M, we would have to pull up another fix.

PR #377, opened 2019-03-15
ORC-474: Add a clear() method to ColumnVectorBatch

rixed:
So that users can merely call it after a batch is written to empty it
recursively before adding new elements for the next batch.

Without that method I can think of no better way to reset a batch than to overload all the ColumnVectorBatch to add this method.


rixed:
rebased

PR #379, opened 2019-04-03
ORC-488: [C++] Support BloomFilter in C++ library

wgtmac:
1. Support C++ writer to enable bloom filter of specific columns.
2. Support C++ reader to get bloom filter of specific columns.
3. Only bloom filter of UTF8 type is supported.

### DIFF
--- a/c++/src/BloomFilter.cc
+++ b/c++/src/BloomFilter.cc
@@ -1,0 +1,333 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "BloomFilter.hh"
+#include "Murmur3.hh"
+
+namespace orc {
+
+  constexpr uint64_t BITS_OF_LONG = 64;
+  constexpr uint8_t  SHIFT_6_BITS = 6;
+  constexpr uint8_t  SHIFT_3_BITS = 3;
+
+  static bool isLittleEndian() {
+    static union { uint32_t i; char c[4]; } num = { 0x01020304 };
+    return num.c[0] == 4;
+  }
+
+  /**
+   * Implementation of BitSet
+   */
+  BitSet::BitSet(uint64_t numBits) {
+    mData.resize(static_cast<size_t>(ceil(
+      static_cast<double>(numBits) / BITS_OF_LONG)), 0);
+  }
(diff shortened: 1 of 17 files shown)


majetideepak:
@wgtmac  I will take a look this week. Sorry for not doing this earlier.

c++/src/BloomFilter.cc
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  DIAGNOSTIC_POP
+
+  void BloomFilterImpl::addHash(uint64_t hash64) {
+    int32_t hash1 = static_cast<int32_t>(hash64 & 0xffffffff);
+    int32_t hash2 = static_cast<int32_t>(hash64 >> 32);
+
+    for (int32_t i = 1; i <= static_cast<int32_t>(mNumHashFunctions); ++i) {
+      int32_t combinedHash = hash1 + i * hash2;
```
majetideepak: should this be uint64_t to get more hash space? This is being casted to `uint64_t` anyway below.

wgtmac: I think combinedHash is deliberately designed to use 32 bits according to the "Bloom Filter Index" section in https://orc.apache.org/specification/ORCv1/. Not sure if it will have any problem if I change it to uint64_t

c++/src/BloomFilter.cc
```diff
@@ -0,0 +1,337 @@
...
+  /**
+   * Helper functions
+   */
+  void checkArgument(bool expression, const std::string& message) {
+    if (!expression) {
+      throw std::logic_error(message);
+    }
+  }
+
+  uint64_t optimalNumOfHashFunctions(uint64_t expectedEntries, uint64_t numBits) {
```
majetideepak: Java uses `int` for the return value? Should we use `int32_t` ?

c++/src/BloomFilter.cc
```diff
@@ -0,0 +1,337 @@
...
+    double n = static_cast<double>(expectedEntries);
+    return std::max<uint64_t>(1ull, static_cast<uint64_t>(
+      std::round(static_cast<double>(numBits) / n * std::log(2.0))));
+  }
+
+  uint64_t optimalNumOfBits(uint64_t expectedEntries, double fpp) {
```
majetideepak: `int32_t`?

c++/src/Reader.cc
```diff
@@ -1061,6 +1061,64 @@ namespace orc {
                                                   postscriptLength));
   }
 
+  std::map<uint32_t, BloomFilterIndex>
```
majetideepak: Why not return a `std::vector<BloomFilterIndex>` ?

wgtmac: Bloom filters are enabled for part of columns in the file. Therefore returning a vector containing all columns is not reasonable. Returning a map allows users to figure out which column has bloom filter index inside.

c++/src/BloomFilter.hh
```diff
@@ -0,0 +1,201 @@
...
+
+    // compute k hash values from hash64 and check bits
+    bool testHash(uint64_t hash64) const;
+
+    void serialize(proto::BloomFilter& bloomFilter) const;
+
+  private:
+    static constexpr double DEFAULT_FPP = 0.05;
+    uint64_t mNumBits;
+    uint64_t mNumHashFunctions;
```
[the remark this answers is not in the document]
wgtmac: done

c++/src/BloomFilter.cc
wgtmac: done

c++/src/BloomFilter.cc
wgtmac: done

PR #385, opened 2019-04-19
ORC-483 Update protobuf for column encryption.

omalley:
Updates the protobuf and the v1 spec for column encryption. 

After this gets approved, I'll update the v2 spec.

### DIFF
--- a/proto/orc_proto.proto
+++ b/proto/orc_proto.proto
@@ -120,6 +120,14 @@ message Stream {
     ROW_INDEX = 6;
     BLOOM_FILTER = 7;
     BLOOM_FILTER_UTF8 = 8;
+    // Virtual stream kinds to allocate space for encrypted index and data.
+    ENCRYPTED_INDEX = 9;
+    ENCRYPTED_DATA = 10;
+
+    // stripe statistics streams
+    STRIPE_STATISTICS = 100;
+    // A virtual stream kind that is used for setting the encryption IV.
+    FILE_STATISTICS = 101;
   }
   optional Kind kind = 1;
   optional uint32 column = 2;
@@ -142,11 +150,34 @@ message ColumnEncoding {
   optional uint32 bloomEncoding = 3;
 }
 
+message StripeEncryptionVariant {
+  repeated Stream streams = 1;
+  repeated ColumnEncoding encoding = 2;
+}
+
+// each stripe looks like:
+//   index streams
+//     unencrypted
+//     variant 1..N
+//   data streams
+//     unencrypted
+//     variant 1..N
+//   footer
+
 message StripeFooter {
   repeated Stream streams = 1;
   repeated ColumnEncoding columns = 2;
   optional string writerTimezone = 3;
+  // one for each column encryption variant
+  repeated StripeEncryptionVariant encryption = 4;
 }
+
+// the file tail looks like:
+//   encrypted stripe statistics: ColumnarStripeStatistics (order by variant)
+//   stripe statistics: Metadata
+//   footer: Footer
+//   postscript: PostScript
+//   psLen: byte
 
 message Type {
   enum Kind {
@@ -178,11 +209,29 @@ message Type {
 }
 
 message StripeInformation {
+  // the global file offset of the start of the stripe
   optional uint64 offset = 1;
+  // the number of bytes of index
   optional uint64 indexLength = 2;
+  // the number of bytes of data
   optional uint64 dataLength = 3;
+  // the number of bytes in the stripe footer
   optional uint64 footerLength = 4;
+  // the number of rows in this stripe
   optional uint64 numberOfRows = 5;
+  // If this is present, the reader should use this value for the encryption
+  // stripe id for setting the encryption IV. Otherwise, the reader should
+  // use one larger than the previous stripe's encryptStripeId.
+  // For unmerged ORC files, the first stripe will use 1 and the rest of the
+  // stripes won't have it set. For merged files, the stripe information
+  // will be copied from their original files and thus the first stripe of
+  // each of the input files will reset it to 1.
+  // Note that 1 was choosen, because protobuf v3 doesn't serialize
+  // primitive types that are the default (eg. 0).
+  optional uint64 encryptStripeId = 6;
+  // For each encryption variant, the new encrypted local key to use
+  // until we find a replacement.
+  repeated bytes encryptedLocalKeys = 7;
 }
 
 message UserMetadataItem {
@@ -190,12 +239,90 @@ message UserMetadataItem {
   optional bytes value = 2;
 }
 
+// StripeStatistics (1 per a stripe), which each contain the
+// ColumnStatistics for each column.
+// This message type is only used in ORC v0 and v1.
 message StripeStatistics {
   repeated ColumnStatistics colStats = 1;
 }
 
+// This message type is only used in ORC v0 and v1.
 message Metadata {
   repeated StripeStatistics stripeStats = 1;
+}
+
+// In ORC v2 (and for encrypted columns in v1), each column has
+// their column statistics written separately.
+message ColumnarStripeStatistics {
+  // one value for each stripe in the file
+  repeated ColumnStatistics colStats = 1;
+}
+
+enum EncryptionAlgorithm {
+  UNKNOWN_ENCRYPTION = 0;  // used for detecting future algorithms
+  AES_CTR_128 = 1;
+  AES_CTR_256 = 2;
+}
+
+message FileStatistics {
+  repeated ColumnStatistics column = 1;
+}
+
+// How was the data masked? This isn't necessary for reading the file, but
+// is documentation about how the file was written.
+message DataMask {
+  // the kind of masking, which may include third party masks
+  optional string name = 1;
+  // parameters for the mask
+  repeated string maskParameters = 2;
+  // the unencrypted column roots this mask was applied to
+  repeated uint32 columns = 3 [packed = true];
+}
+
+// Information about the encryption keys.
+message EncryptionKey {
+  optional string keyName = 1;
+  optional uint32 keyVersion = 2;
+  optional EncryptionAlgorithm algorithm = 3;
+}
+
+// The description of an encryption variant.
+// Each variant is a single subtype that is encrypted with a single key.
+message EncryptionVariant {
+  // the column id of the root
+  optional uint32 root = 1;
+  // The master key that was used to encrypt the local key, referenced as
+  // an index into the Encryption.key list.
+  optional uint32 key = 2;
+  // the encrypted key for the file footer
+  optional bytes encryptedKey = 3;
+  // the stripe statistics for this variant
+  repeated Stream stripeStatistics = 4;
+  // encrypted file statistics as a FileStatistics
+  optional bytes fileStatistics = 5;
+}
+
+// Which KeyProvider encrypted the local keys.
+enum KeyProviderKind {
+  UNKNOWN = 0;
+  HADOOP = 1;
+  AWS = 2;
+  GCP = 3;
+  AZURE = 4;
+}
+
+message Encryption {
+  // all of the masks used in this file
+  repeated DataMask mask = 1;
+  // all of the keys used in this file
+  repeated EncryptionKey key = 2;
+  // The encrypted variants.
+  // Readers should prefer the first variant that the user has access to
+  // the corresponding key. If they don't have access to any of the keys,
+  // they should get the unencrypted masked data.
+  repeated EncryptionVariant variants = 3;
+  // How are the local keys encrypted?
+  optional KeyProviderKind keyProvider = 4;
 }
 
 message Footer {
@@ -214,6 +341,12 @@ message Footer {
   // 2 = Presto
   // 3 = Scritchley Go from https://github.com/scritchley/orc
   optional uint32 writer = 9;
+
+  // information about the encryption in this file
+  optional Encryption encryption = 10;
+
+  // the number of bytes in the encrypted stripe statistics
+  optional uint64 stripeStatisticsLength = 11;
 }
 
 enum CompressionKind {
@@ -249,10 +382,11 @@ message PostScript {
   //   1 = HIVE-8732 fixed (fixed stripe/file maximum statistics &
   //                        string statistics use utf8 for min/max)
   //   2 = HIVE-4243 fixed (use real column names from Hive tables)
-  //   3 = HIVE-12055 fixed (vectorized writer implementation)
+  //   3 = HIVE-12055 added (vectorized writer implementation)
   //   4 = HIVE-13083 fixed (decimals write present stream correctly)
   //   5 = ORC-101 fixed (bloom filters use utf8 consistently)
   //   6 = ORC-135 fixed (timestamp statistics use utc)
+  //   7 = ORC-14 added (column encryption)
   //
   // Version of the ORC C++ writer:
   //   6 = original
--- a/site/specification/ORCv1.md
+++ b/site/specification/ORCv1.md
@@ -45,6 +45,13 @@ incorporates the Protobuf definition from the
 reader is encouraged to review the Protobuf encoding if they need to
 understand the byte-level encoding
 
+The sections of the file tail are (and their protobuf message type):
+* encrypted stripe statistics: list of ColumnarStripeStatistics
+* stripe statistics: Metadata
+* footer: Footer
+* postscript: PostScript
+* psLen: byte
+
 ## Postscript
 
 The Postscript section provides the necessary information to interpret
@@ -123,6 +130,16 @@ message Footer {
  repeated ColumnStatistics statistics = 7;
  // the maximum number of rows in each index entry
  optional uint32 rowIndexStride = 8;
+ // Each implementation that writes ORC files should register for a code
+ // 0 = ORC Java
+ // 1 = ORC C++
+ // 2 = Presto
+ // 3 = Scritchley Go from https://github.com/scritchley/orc
+ optional uint32 writer = 9;
+ // information about the encryption in this file
+ optional Encryption encryption = 10;
+ // the number of bytes in the encrypted stripe statistics
+ optional uint64 stripeStatisticsLength = 11;
 }
 ```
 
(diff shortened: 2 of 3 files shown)


site/specification/ORCv1.md
```diff
@@ -123,6 +123,16 @@ message Footer {
...
+ // Each implementation that writes ORC files should register for a code
+ // 0 = ORC Java
+ // 1 = ORC C++
+ // 2 = Presto
+ // 3 = Scritchley Go from https://github.com/scritchley/orc
+ optional uint32 writer = 9;
+ // information about the encryption in this file
+ optional Encryption encryption = 10;
+ // the number of bytes in the encrypted stripe statistics
+ optional uint64 stripeStatisticsLength = 11;
```
wgtmac: rename to encryptedStripeStatisticsLength ?

site/specification/ORCv1.md
```diff
@@ -369,6 +392,174 @@ message Metadata {
...
+the user and the writer only needs to define which columns and
+encryption keys to use. When reading an ORC file, if the user has
+access to the keys, they will get the real data. If they do not have
+the keys, they will get the masked data.
+
+Each encrypted column in each file will have a random local key
+generated for it. Thus, even though all of the decryption happens
+locally in the reader, a malicious user that stores the key only
+enables access that column in that file. The local keys are encrypted
+by the Hadoop or Ranger Key Management Server (KMS). The encrypted
```
wgtmac: Is it supposed to work with Hadoop ecosystem only?

site/specification/ORCv1.md
```diff
@@ -369,6 +392,174 @@ message Metadata {
...
+
+## Stream Encryption
+
+Our encryption is done using AES/CTR. CTR is a mode that has some very
+nice properties for us:
+
+* It is seeded so that identical data is encrypted differently.
+* It does not require padding the stream to the cipher length.
+* It allows readers to seek in to a stream.
+* The IV does not need to be randomly generated.
```
wgtmac: I'd suggest to put full text of initialization vector (IV) here instead in line 547.

site/specification/ORCv1.md
```diff
@@ -369,6 +392,174 @@ message Metadata {
...
+the masked unencrypted data. Encryption variants encrypt a subtree of
+columns and use a single local key. The initial version of encryption
+support only allows the two variants, but this may be extended later
+and thus readers should use the first variant of a column that the
+reader has access to.
+
+```
+message EncryptionVariant {
+  // the column id of the root
+  optional uint32 root = 1;
```
wgtmac: What is the root used for?

proto/orc_proto.proto
```diff
@@ -178,18 +208,34 @@ message Type {
...
 
 message UserMetadataItem {
   optional string name = 1;
   optional bytes value = 2;
 }
 
+// The ORC 1 stripe statistics are organized as a sequence of
+// StripeStatistics (1 per a stripe), which each contain the
+// ColumnStatistics for each column.
 message StripeStatistics {
```
wgtmac: Should we mark it as [deprecated=true] in ORC v2?

site/specification/ORCv1.md
omalley: My thought was that in ORCv2 we should move all of the stripe statistics into that section, so that all of the stripe statistics are split up by column. So that is why I didn't include the encrypted in the name. What do you think in that context?

site/specification/ORCv1.md
omalley: No, it should work with anything that will encrypt local keys for us. In the Java side, look at the KeyProvider API in the Hadoop shims module. It was designed to work with either the cloud specific KMS's such as Amazon's or the on-prem Hadoop or Ranger KMS. 

Your point however, remains very relevant however. We should also encode how the local keys were encoded. It looks like we need another enum in the Key that looks like:

```
enum KeyProviderKind {
  UNKNOWN = 0;
  HADOOP_KMS = 1;
  AWS_KMS = 2;
  GCP_KMS = 3;
  AZURE_KEY_VAULT = 4;
}
```

site/specification/ORCv1.md
omalley: Ok.

site/specification/ORCv1.md
omalley: Root is the column id of that variant.

proto/orc_proto.proto
omalley: We can't really deprecate, since we'll need to leave the definition for ORC v0 and v1 readers.

But yes, we should make it as only used in v0 and v1.

site/specification/ORCv1.md
omalley: Given a type tree, this is the root of a subtree that is encrypted by a local key.

site/specification/ORCv1.md
```diff
@@ -369,6 +392,192 @@ message Metadata {
...
+    HADOOP = 1;
+    AWS = 2;
+    GCP = 3;
+    AZURE = 4;
+  }
+  optional KeyProviderKind keyProvider = 4;
+}
+```
+
+When ORC is using the Hadoop or Ranger KMS, it generates a random local key
```
majetideepak: Does this mean we need to include the KMS APIs for each provider (Ranger, AWS, etc.) inside the ORC library?

site/specification/ORCv1.md
```diff
@@ -149,6 +164,14 @@ message StripeInformation {
  optional uint64 footerLength = 4;
  // the number of rows in the stripe
  optional uint64 numberOfRows = 5;
+ // If this is present, it resets the stripe id with respect to
+ // encryption until the next occurrance of encryptStripeId. Otherwise,
+ // readers should use the previous stripe id + 1.  This should only
```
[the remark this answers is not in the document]
omalley: Yes, I'll try to make that more clear.

site/specification/ORCv1.md
```diff
@@ -149,6 +164,14 @@ message StripeInformation {
...
+ // occur in the first stripe and the boundaries of merged ORC files.
+ optional uint64 encryptStripeId = 6;
+ // For each encryption variant, the new encrypted key to use until we
+ // find a replacement.
+ repeated bytes encryptedKeys = 7;
```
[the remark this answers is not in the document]
omalley: Ok.

site/specification/ORCv1.md
omalley: Ranger uses the Hadoop KMS API, so we already have it. For the others, I'd propose that we build separate modules (one per a cloud provider) so that we don't end up in a dependency tangle.

site/specification/ORCv1.md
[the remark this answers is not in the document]
omalley: Actually, how about encryptedLocalKeys?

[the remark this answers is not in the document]
majetideepak: +1 for `encryptedLocalKeys`

site/specification/ORCv1.md
omalley: After further reflection, I moved this to the Encryption section, so it becomes a file level property. I don't think we need to support a use case where some of the keys are encrypted/decrypted by different KeyProviders.

site/specification/ORCv1.md
```diff
@@ -369,6 +397,194 @@ message Metadata {
...
+  AES_256 = 2;
+}
+```
+
+## Encryption Variants
+
+Each encrypted column is written as two variants:
+
+* encrypted unmasked - for users with access to the key
+* unencrypted masked - for all other users
```
xndai: Where are "unencrypted masked" stream recorded? I think they should be at StripeFooter.steams so the old reader can consume.

proto/orc_proto.proto
```diff
@@ -178,26 +208,121 @@ message Type {
...
+  // the column id of the root
+  optional uint32 root = 1;
+  // the key that encrypted this variant
+  optional uint32 key = 2;
+  // the encrypted key for the file footer
+  optional bytes encryptedKey = 3;
+  // the stripe statistics for this variant
+  repeated Stream stripeStatistics = 4;
+  // encrypted file statistics as a FileStatistics
+  optional bytes fileStatistics = 5;
```
xndai: You add a FILE_STATISTICS kind of stream. But you actually save it as a byte array in protobuf serialization. Why?

site/specification/ORCv1.md
```diff
@@ -369,6 +397,194 @@ message Metadata {
...
+However, it is critical for CTR that we never reuse an initialization
+vector (IV) with the same local key.
+
+For data in the footer, use the number of stripes in the file as the
+stripe id. This guarantees when we write an intermediate footer in to
+a file that we don't use the same IV.
+
+Additionally, we never reuse a local key for new data. For example, when
+merging files, we don't reuse local key from the input files for the new
+file tail, but always generate a new local key.
```
xndai: I wonder what would happen in append scenario. I think we will just rotate IV by increasing stripe id.

proto/orc_proto.proto
```diff
@@ -178,26 +208,121 @@ message Type {
...
+  optional EncryptionAlgorithm algorithm = 3;
+}
+
+// The description of an encryption variant.
+// Each variant is a single subtype that is encrypted with a single key.
+message EncryptionVariant {
+  // the column id of the root
+  optional uint32 root = 1;
+  // the key that encrypted this variant
+  optional uint32 key = 2;
```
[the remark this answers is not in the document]
omalley: This is an index into Encryption.key list and thus describes the master key. I'll improve the comment.

proto/orc_proto.proto
```diff
@@ -178,26 +208,121 @@ message Type {
...
+// their column statistics written separately.
+message ColumnarStripeStatistics {
+  // one value for each stripe in the file
+  repeated ColumnStatistics colStats = 1;
+}
+
+enum EncryptionAlgorithm {
+  UNKNOWN_ENCRYPTION = 0;  // used for detecting future algorithms
+  AES_128 = 1;
+  AES_256 = 2;
```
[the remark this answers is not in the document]
omalley: Ok, although it seems unlikely we'll add a different encryption mode in the short term.

proto/orc_proto.proto
```diff
@@ -178,26 +208,121 @@ message Type {
...
+  // For unmerged ORC files, the first stripe will use 1 and the rest of the
+  // stripes won't have it set. For merged files, the stripe information
+  // will be copied from their original files and thus the first stripe of
+  // each of the input files will reset it to 1.
+  // Note that 1 was choosen, because protobuf v3 doesn't serialize
+  // primitive types that are the default (eg. 0).
+  optional uint64 encryptStripeId = 6;
+  // For each encryption variant, the new encrypted local key to use
+  // until we find a replacement.
+  repeated bytes encryptedLocalKeys = 7;
```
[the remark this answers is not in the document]
omalley: Generally encryptedLocalKeys will only be set on the first stripe. All following stripes will use the encryptedLocalKeys of the previous stripe unless this field is defined.

proto/orc_proto.proto
```diff
@@ -142,12 +149,35 @@ message ColumnEncoding {
...
   repeated ColumnEncoding columns = 2;
   optional string writerTimezone = 3;
+  // one for each column encryption variant
+  repeated StripeEncryptionVariant encryption = 4;
 }
 
+// the file tail looks like:
+//   encrypted stripe statistics: ColumnStripeStatistics (order by variant)
+//   stripe statistics: Metadata
+//   footer: Footer
```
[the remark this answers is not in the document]
omalley: The encrypted file statistics are serialized directly into the EncryptionVariant.fileStatistics.

proto/orc_proto.proto
```diff
@@ -120,6 +120,13 @@ message Stream {
     ROW_INDEX = 6;
     BLOOM_FILTER = 7;
     BLOOM_FILTER_UTF8 = 8;
+    // Allocate space for the encrypted stuff
+    ENCRYPTED_INDEX = 9;
+    ENCRYPTED_DATA = 10;
+
+    // statistics streams
+    STRIPE_STATISTICS = 100;
+    FILE_STATISTICS = 101;
```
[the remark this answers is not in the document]
omalley: The file statistics are left in line, but for setting the encryption IV we need a Stream.Kind for the file statistics.

For v1, the unencrypted statistics stay the same. For v2, the unencrypted stripe statistics will become streams. Encrypted stripe statistics will always be streams.

proto/orc_proto.proto
```diff
@@ -178,26 +208,121 @@ message Type {
...
+
+message Encryption {
+  // all of the masks used in this file
+  repeated DataMask mask = 1;
+  // all of the keys used in this file
+  repeated EncryptionKey key = 2;
+  // The encrypted variants.
+  // Readers should prefer the first variant that the user has access to
+  // the corresponding key. If they don't have access to any of the keys,
+  // they should get the unencrypted data.
```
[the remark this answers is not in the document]
omalley: ok.

site/specification/ORCv1.md
```diff
@@ -369,6 +397,194 @@ message Metadata {
...
+
+* encrypted unmasked - for users with access to the key
+* unencrypted masked - for all other users
+
+The changes to the format were done so that old ORC readers will read
+the masked unencrypted data. Encryption variants encrypt a subtree of
+columns and use a single local key. The initial version of encryption
+support only allows the two variants, but this may be extended later
+and thus readers should use the first variant of a column that the
+reader has access to.
```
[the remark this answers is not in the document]
omalley: The use case would be where you needed multiple levels of access. For example, if you wanted some of your data scientists to be able to see the actual data, others were only allowed to see the sha256 so that they could join against other tables, but general users couldn't see the data at all.

| Key | data |
| ---| ----- |
| pii | actual data |
| sha-pii | sha256 of the data |
| none | nulls |

site/specification/ORCv1.md
omalley: Naturally

proto/orc_proto.proto
omalley: I need the kind to set the IV because those bytes are encrypted.

site/specification/ORCv1.md
omalley: The goal is to allow merging files without re-encrypting the data in the stripe. By putting the encryption stripe id and encrypted local keys in to the stripe information, the merge will do the right thing. (Assuming that the entire file is merged, since the writer is only setting the encryption information in the first stripe of each file. If someone builds a merge tool that merges partial files, they'll need to take care to set the properties on the transitions.)

site/specification/ORCv1.md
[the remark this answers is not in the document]
xndai: I am not sure if we should be doing this at Orc level. The scenario mentioned here can also be achieved from the application level by adding an additional column. That way user has much more flexibility to choose how they would expose information at different security level. Multiple encryption variants complicate the reader/writer quite a bit IMHO.

proto/orc_proto.proto
```diff
@@ -120,6 +120,14 @@ message Stream {
     ROW_INDEX = 6;
     BLOOM_FILTER = 7;
     BLOOM_FILTER_UTF8 = 8;
+    // Virtual stream kinds to allocate space for encrypted index and data.
+    ENCRYPTED_INDEX = 9;
+    ENCRYPTED_DATA = 10;
+
+    // stripe statistics streams
+    STRIPE_STATISTICS = 100;
```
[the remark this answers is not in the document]
omalley: // In ORC v2 (and for encrypted columns in v1), each column has
// their column statistics written separately.
message ColumnarStripeStatistics {
  // one value for each stripe in the file
  repeated ColumnStatistics colStats = 1;
}

site/specification/ORCv1.md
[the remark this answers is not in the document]
omalley: I'm not proposing adding this in the initial release of column encryption, but I'd like to leave the possibility open. Given that we already have 2 variants, having overlapping encryption variants is very straightforward.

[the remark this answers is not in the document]
omalley: Additionally, with users that I've talked to about this feature, some of them could really use it. 

Note that Hive & Ranger together provide this dynamically, if the user is only using Hive Server 2.

proto/orc_proto.proto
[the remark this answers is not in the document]
omalley: My thought was that stripe statistics can get large, but file statistics don't. So rather than go through another level of stream lookup and reading, we could just inline it.

omalley:
Any additional feedback?

site/specification/ORCv1.md
xndai: My question is specifically about appending scenario - you open an existing Orc file and append data at the tail. But I guess it would be very similar to merge case.

xndai:
I add a few more comments to clarify a couple of questions. Other than that, looks good to me. Thanks.

proto/orc_proto.proto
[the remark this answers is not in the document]
omalley: There will be one stream per a column. Each stream will be a single ColumnarStripeStatistics. Let me try to clarify the spec.

site/specification/ORCv1.md
omalley: Yes, it would be very similar to the merge case.

site/specification/ORCv1.md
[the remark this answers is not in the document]
omalley: When the reader starts planning a stripe read, it will try to decrypt the local keys. If it has access to the key for a given variant, it will use the streams & encodings for all of the columns under that variant's root column. So if the schema looks like: struct<a:int,b:struct<ba:string,bb:int>,c:int> , with "b" being encrypted. If the reader doesn't have access to the key, they'll just use the streams & encodings from the StripeFooter. If it does have access to the key, it will use the streams & encodings for "a", and "c" from the StripeFooter and "b", "ba", and "bb" from the StripeEncryptionVariant. Because the streams are separated by column, it is easy to do that filtering.

Let's get the first version released with column encryption and build more experience using it in production. That will help us understand what will help users the most.

xndai:
LGTM. +1.

proto/orc_proto.proto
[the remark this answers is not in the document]
sunriseLe: file statistics mean column statistics in file level？If so, file statistics have already existed in the File Footer.
For entrypted column, column statistcs will be compressed, encrypted. To initial IV, you introduced  a new steam kind FILE_STATISTICS？
> Encrypted stripe statistics will always be streams,
Where Encrypted stripe statistics will store? Still in file footer ？

proto/orc_proto.proto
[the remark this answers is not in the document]
sunriseLe: > The encrypted file statistics are serialized directly into the EncryptionVariant.fileStatistics.
When will EncryptionVarian.fileStatistics be not empty？ Or where is the encryption variant for file statistics stored？

PR #386, opened 2019-04-20
ORC-485. Add API to write encrypted files.

omalley:
These changes provide most of the API changes for column encryption.

### DIFF
--- a/java/core/src/java/org/apache/orc/EncryptionVariant.java
+++ b/java/core/src/java/org/apache/orc/EncryptionVariant.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.orc;
+
+import java.io.IOException;
+import java.security.Key;
+
+/**
+ * Information about a key used for column encryption in an ORC file.
+ */
+public interface EncryptionVariant extends Comparable<EncryptionVariant> {
+
+  /**
+   * Get the encryption key for this column.
+   * @return the encryption key.
+   */
+  EncryptionKey getKey();
+
+  /**
+   * Get the root column for this encryption.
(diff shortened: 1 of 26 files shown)


omalley:
Any comments on this before I commit it?

java/core/src/java/org/apache/orc/EncryptionVariant.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Information about a key used for column encryption in an ORC file.
+ */
+public interface EncryptionVariant extends Comparable<EncryptionVariant> {
+
+  /**
+   * Get the encryption key for this column.
+   * @return the encryption key.
+   */
+  EncryptionKey getKey();
```
prasanthj: use getColumnKey() to be more descriptive? Also the scope of these keys are stripe or file? Can it be explicitly called out?

java/core/src/java/org/apache/orc/EncryptionVariant.java
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Get the encryption variant id within the file.
+   */
+  int getVariantId();
+
+  /**
+   * Get the local key for the footer.
+   * @return the local decrypted key or null if it isn't available
+   */
+  Key getFooterKey() throws IOException;
```
prasanthj: getStripeFooterKey() or if file level getFileFooterKey()?

java/core/src/java/org/apache/orc/Reader.java
```diff
@@ -114,7 +114,7 @@
    * Get the statistics about the columns in the file.
    * @return the information about the column
    */
-  ColumnStatistics[] getStatistics();
+  ColumnStatistics[] getStatistics() throws IOException;
```
prasanthj: is this required? when does this throw exception?

java/pom.xml
```diff
@@ -642,6 +642,11 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
+        <groupId>org.jetbrains</groupId>
```
prasanthj: nit: required? can we use Objects.requireNotNull instead?

omalley: I really want an annotation to mark the intent and help with static analysis. I really don't need a run time check.

java/core/src/java/org/apache/orc/EncryptionVariant.java
omalley: They are for the entire file, I'll update the name.

java/core/src/java/org/apache/orc/EncryptionVariant.java
omalley: It is the file footer, so I'll update this name too.

java/core/src/java/org/apache/orc/Reader.java
omalley: Good catch. I can't add a throw here without breaking backwards compatibility. In general, we won't throw here, but if the key provider is flaky, we can get an exception. I guess I'll make it an unchecked exception.

java/core/src/java/org/apache/orc/EncryptionVariant.java
omalley: More precisely, it is the key description so I updated the name to getKeyDescription.

PR #389, opened 2019-04-26
ORC-496: [C++] Fix integer overflow in skip of column reader

wgtmac:
skip() function is the StringDirectColumnReader and DoubleColumnReader use size_t to compute total size to skip; however, the value is casted to int when calling underlying input stream.


xndai:
LGTM. +1

PR #390, opened 2019-04-29
ORC-497 - Fix build failures for maven 3.6.x

moresandeep:
While building the java side, with -Panalyze option, ORC build fails, this is because 'findbugs-maven-plugin' version 3.0.3 is not compatible with maven 3.6.X, upgrading 'findbugs-maven-plugin' to 3.0.5 (latest) fixes the issue. The patch is tested with older maven version (3.3.9)


moresandeep:
@omalley Would be great if you can take a look at this !

PR #392, opened 2019-05-08
[ORC-499][C++] Change databuffer resize and refill previous data

czxrrr:
[C++] CSVFileImport.cc has error when DataBuffer running out and doubling size

CSVFileImport was using Databuffer::reserve() but this will not change the size.

Even the size was changed, the new buffer has totally different address, which will make previous stringBatch->data[i], (i is the already finished ones) stores the wrong address.

### DIFF
--- a/tools/src/CSVFileImport.cc
+++ b/tools/src/CSVFileImport.cc
@@ -90,7 +90,13 @@ void fillStringValues(const std::vector<std::string>& data,
     } else {
       batch->notNull[i] = 1;
       if (buffer.size() - offset < col.size()) {
-        buffer.reserve(buffer.size() * 2);
+        char* old_buffer = buffer.data();
+        buffer.resize(buffer.size() * 2);
+        char* new_buffer = buffer.data();
+        for (uint64_t refill_index = 0; refill_index < i; ++refill_index){
+          stringBatch->data[refill_index] = stringBatch->data[refill_index] - old_buffer + new_buffer;
+        }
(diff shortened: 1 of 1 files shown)


tools/src/CSVFileImport.cc
```diff
@@ -90,7 +90,13 @@ void fillStringValues(const std::vector<std::string>& data,
...
       batch->notNull[i] = 1;
       if (buffer.size() - offset < col.size()) {
-        buffer.reserve(buffer.size() * 2);
+        char* old_buffer = buffer.data();
+        buffer.resize(buffer.size() * 2);
+        char* new_buffer = buffer.data();
+        for (uint64_t refill_index = 0; refill_index < i; ++refill_index){
+          stringBatch->data[refill_index] = stringBatch->data[refill_index] - old_buffer + new_buffer;
+        }
+        std::cout << std::endl;
```
wgtmac: remove this line

tools/src/CSVFileImport.cc
```diff
@@ -90,7 +90,13 @@ void fillStringValues(const std::vector<std::string>& data,
     } else {
       batch->notNull[i] = 1;
       if (buffer.size() - offset < col.size()) {
-        buffer.reserve(buffer.size() * 2);
+        char* old_buffer = buffer.data();
+        buffer.resize(buffer.size() * 2);
```
wgtmac: you may also consider if this size is enough or not

PR #393, opened 2019-05-20
ORC-501: [C++]Move blob ownership from StringColumnReader to StringVectorBatch to avoid data overriden

yuruiz:
- Move blob ownership from StringColumnReader to StringVectorBatch to avoid data overriden
- Enforce memcpy on StringColumnReader to make sure string blob is owned by external user

### DIFF
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -122,6 +122,8 @@ namespace orc {
     DataBuffer<char*> data;
     // the length of each string
     DataBuffer<int64_t> length;
+    // string blob
+    DataBuffer<char> blob;
   };
 
   struct StringDictionary {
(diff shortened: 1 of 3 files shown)


c++/include/orc/Vector.hh
```diff
@@ -122,6 +122,8 @@ namespace orc {
     DataBuffer<char*> data;
     // the length of each string
     DataBuffer<int64_t> length;
+    // string blob
+    DataBuffer<char> blob;
```
wgtmac: How about make it private? Though we may need to add StringDirectColumnReader as a friend class.

yuruiz: I would prefer to make it public since exposing the blob memory make it easier data conversion, and external caller would be able to access this memory via data array any way.

wgtmac: OK, make sense

PR #394, opened 2019-05-23
ORC-506: [C++] fix compiler warnings

rip-nsk:


omalley:
Let me kick off a docker run to check test various linux variants.

PR #395, opened 2019-05-28
ORC-398 col stats list map

moresandeep:
Hello @omalley, 
This is my initial stab at this JIRA, let me know what you think.


moresandeep:
The test failure seems unrelated to patch
`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.21.0:test (default-test) on project orc-shims: Unable to generate classpath: org.apache.maven.artifact.resolver.ArtifactResolutionException: Unable to get dependency information for org.apache.maven.surefire:surefire-junit4:jar:2.21.0: Failed to retrieve POM for org.apache.maven.surefire:surefire-junit4:jar:2.21.0: Could not transfer artifact org.apache.maven.surefire:surefire-junit4:pom:2.21.0 from/to central (https://repo.maven.apache.org/maven2): repo.maven.apache.org: unknown error
[ERROR] org.apache.maven.surefire:surefire-junit4:jar:2.21.0`

is there a way to re-trigger Travis CI build ?

proto/orc_proto.proto
```diff
@@ -74,6 +74,13 @@ message BinaryStatistics {
   optional sint64 sum = 1;
 }
 
+// ORC-398: Statistics for list and map
+message CollectionStatistics {
+  optional uint64 minCollectionLength = 1;
+  optional uint64 maxCollectionLength = 2;
+  optional uint64 childCount = 3;
```
wgtmac: Can we avoid adding new fields in the file meta by aggregating the stats directly from child col stats?

java/core/src/java/org/apache/orc/AggregateStatistics.java
```diff
@@ -0,0 +1,49 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.orc;
+
+/**
+ * Define aggregate statistics, Min, Max and Sum.
+ */
+public interface AggregateStatistics extends ColumnStatistics {
```
wgtmac: Is it only intended for integer values?

moresandeep:
> Do you have any design doc or some context for adding this?

This is the reason: https://issues.apache.org/jira/browse/ORC-398
This is not a big change to warrant a design doc, this change is similar to https://issues.apache.org/jira/browse/ORC-305

proto/orc_proto.proto
moresandeep: Not sure what you mean by 

> aggregating the stats directly from child col stats

We could use IntegerStatistics but then that would be misleading, this specifically deals with Maps and Lists and the idea is to have min, max, and sum of the number of children for Map and List.

java/core/src/java/org/apache/orc/AggregateStatistics.java
moresandeep: Yes it is only intended for integer values, since we are interested in min, max and sum.

java/core/src/java/org/apache/orc/CollectionColumnStatistics.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.orc;
+
+/**
+ * Statistics for all of collections such as Map and List.
+ */
+public interface CollectionColumnStatistics extends AggregateStatistics {
+
```
omalley: I'd move the functions down here. I'd suggest:
long getMinimumChildren();
long getMaximumChildren();
long getTotalChildren();

proto/orc_proto.proto
omalley: I think we are better leaving it separate.

java/core/src/java/org/apache/orc/CollectionColumnStatistics.java
moresandeep: Thanks @omalley the PR has been updates with the changes.

java/core/src/java/org/apache/orc/AggregateStatistics.java
moresandeep: Thanks @omalley, the PR has been updates with the changes.

moresandeep:
@omalley thanks for the review, I updates the PR with suggested changes, let me know what you think.

java/core/src/java/org/apache/orc/AggregateStatistics.java
```diff
@@ -0,0 +1,49 @@
+/**
```
omalley: This should only have /* and not /**.

java/core/src/java/org/apache/orc/CollectionColumnStatistics.java
```diff
@@ -0,0 +1,47 @@
+/**
```
omalley: This one too.

java/core/src/java/org/apache/orc/AggregateStatistics.java
omalley: We don't need this file.

java/core/src/java/org/apache/orc/CollectionColumnStatistics.java
```diff
@@ -0,0 +1,47 @@
...
+   * Get maximum number of children in the collection.
+   * @return the maximum children count
+   */
+  long getMaximumChildren();
+
+  /**
+   * Is the total defined? If the sum overflowed the counter this will be false.
+   * @return is the total no. of children available
+   */
+  boolean isTotalDefined();
```
omalley: We don't need isTotalDefined. It isn't possible to overflow the long behind it.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -170,19 +166,215 @@ public int hashCode() {
...
+    protected boolean hasMinimum = false;
+    protected boolean overflow = false;
+
+    CollectionColumnStatisticsImpl() {
+      super();
+    }
+
+    CollectionColumnStatisticsImpl(OrcProto.ColumnStatistics stats) {
+      super(stats);
+      OrcProto.IntegerStatistics intStat = stats.getIntStatistics();
```
omalley: This should be the collectionStatistics.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -170,19 +166,215 @@ public int hashCode() {
...
+   */
+  private static final class IntegerColumnStatisticsImpl extends ColumnStatisticsImpl
       implements IntegerColumnStatistics {
 
-    private long minimum = Long.MAX_VALUE;
-    private long maximum = Long.MIN_VALUE;
-    private long sum = 0;
-    private boolean hasMinimum = false;
-    private boolean overflow = false;
+    protected long minimum = Long.MAX_VALUE;
```
omalley: We don't need this change.

moresandeep: Yup, no need after refactoring it back.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
moresandeep: yes, this was an oversight, sorry :(

java/core/src/java/org/apache/orc/CollectionColumnStatistics.java
moresandeep: Sure, will take it out.

java/core/src/java/org/apache/orc/CollectionColumnStatistics.java
moresandeep: Sure, will take it out.

java/core/src/java/org/apache/orc/AggregateStatistics.java
moresandeep: right, this was the artifact of refactoring, will remove it.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -170,7 +165,198 @@ public int hashCode() {
...
+            minimum = otherColl.minimum;
+          }
+          if (otherColl.maximum > maximum) {
+            maximum = otherColl.maximum;
+          }
+        }
+
+        overflow |= otherColl.overflow;
+        if (!overflow) {
+          boolean wasPositive = sum >= 0;
```
omalley: the sum for the lengths will always be positive (assuming you can't overflow).

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -170,7 +165,198 @@ public int hashCode() {
...
+      maximum = Long.MIN_VALUE;
+      sum = 0;
+      overflow = false;
+    }
+
+    @Override
+    public void merge(ColumnStatisticsImpl other) {
+      if (other instanceof CollectionColumnStatisticsImpl) {
+        CollectionColumnStatisticsImpl otherColl = (CollectionColumnStatisticsImpl) other;
+        if (!hasMinimum) {
```
omalley: I'd suggest replacing with count != 0

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -179,10 +365,10 @@ public int hashCode() {
     private boolean hasMinimum = false;
     private boolean overflow = false;
 
-    IntegerStatisticsImpl() {
+    IntegerColumnStatisticsImpl() {
```
omalley: Let's not change the class name.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -357,6 +527,22 @@ public int hashCode() {
       result = 31 * result + (overflow ? 1 : 0);
       return result;
     }
+
+    @Override
```
omalley: Or move the methods around.

java/tools/src/java/org/apache/orc/tools/JsonFileDump.java
```diff
@@ -352,6 +353,12 @@ private static void writeColumnStatistics(JSONWriter writer, ColumnStatistics cs
           writer.key("sum").value(((DecimalColumnStatistics) cs).getSum());
         }
         writer.key("type").value(OrcProto.Type.Kind.DECIMAL);
+      } else if (cs instanceof CollectionColumnStatistics) {
+        writer.key("minChildren").value(((CollectionColumnStatistics) cs).getMinimumChildren());
+        writer.key("maxChildren").value(((CollectionColumnStatistics) cs).getMaximumChildren());
+        writer.key("totalChildren").value(((CollectionColumnStatistics) cs).getTotalChildren());
+        /* What about MAP Type ? */
+        writer.key("type").value(OrcProto.Type.Kind.LIST);
```
omalley: I'd suggest removing the type key rather than putting out a potentially incorrect list.

proto/orc_proto.proto
```diff
@@ -74,6 +74,13 @@ message BinaryStatistics {
   optional sint64 sum = 1;
 }
 
+// ORC-398: Statistics for list and map
```
omalley: I don't think annotating the jira number helps here.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -170,7 +165,198 @@ public int hashCode() {
...
+      if (overflow != that.overflow) {
+        return false;
+      }
+
+      return true;
+    }
+
+    @Override
+    public int hashCode() {
+      int result = super.hashCode();
```
omalley: I'd suggest using a simpler hashCode if the count is zero rather than hashing in the Long MIN and MAX.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -170,7 +165,198 @@ public int hashCode() {
...
 
-  private static final class IntegerStatisticsImpl extends ColumnStatisticsImpl
+  /**
+   * Column statistics for List and Map types.
+   */
+  private static final class CollectionColumnStatisticsImpl extends ColumnStatisticsImpl
+      implements CollectionColumnStatistics {
+
+    protected long minimum = Long.MAX_VALUE;
+    protected long maximum = Long.MIN_VALUE;
```
omalley: I'd suggest using 0 here, so that we don't accidentally use a negative length.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -170,7 +165,198 @@ public int hashCode() {
...
+    protected long sum = 0;
+    protected boolean hasMinimum = false;
```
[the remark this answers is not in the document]
moresandeep: Sure, took it out.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -170,7 +165,198 @@ public int hashCode() {
...
+      /*
+       * Here, minimum = minCollectionLength
+       * maximum = maxCollectionLength
+       * sum = childCount
+       */
+      if (!hasMinimum) {
+        hasMinimum = true;
+        minimum = length;
+        maximum = length;
+      } else if (length < minimum) {
```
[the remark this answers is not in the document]
moresandeep: Fixed it.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
moresandeep: agreed, I will simplify this.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -170,7 +165,198 @@ public int hashCode() {
...
+      super();
+    }
+
+    CollectionColumnStatisticsImpl(OrcProto.ColumnStatistics stats) {
+      super(stats);
+      OrcProto.CollectionStatistics collStat = stats.getCollectionStatistics();
+      if (collStat.hasMinChildren()) {
+        hasMinimum = true;
+        minimum = collStat.getMinChildren();
+      }
```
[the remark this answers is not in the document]
moresandeep: Fixed it.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
moresandeep: Fixed it.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
moresandeep: Sorry, this was aftereffect of refactoring. Fixed it.

java/tools/src/java/org/apache/orc/tools/JsonFileDump.java
moresandeep: Fixed it.

proto/orc_proto.proto
moresandeep: Fixed it.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
moresandeep: Right, Fixed it.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
moresandeep: Sure, fixed it.

PR #396, opened 2019-05-29
ORC-510: Cleanup StreamOptions and CompressionCodec.Options

omalley:
The writer's stream options should be cleaned up. There are a couple of related items:

* We end up changing the IV's for streams and we need to notify the stream when we do so, therefore it is better to make an explicit API to change them.
* CryptoUtils should match with functions to modify an IV rather than forcing an allocation.
* CompressionCodec.Options only needs to be copied when we have tuning options.
* CompressionCodec would be nice to have as Closeable where that returns it to the pool.


omalley:
Any comments?

PR #397, opened 2019-06-01
ORC-512: Fix C++ build on CentOS 6.

omalley:


omalley:
The only serious change was the one where I needed to drop the default of {} on getBloomFilters. Other than that, it was all dealing with the older C++ from centos6.

Changes:
* use ORC_NOEXCEPT instead of noexcept
* use orc::to_string instead of std::to_string to work around the strange types
* add a new test for constexpr
* if the compiler doesn't have constexpr use const instead, which isn't perfect.
* explicitly deal with the unique_ptr to shared_ptr conversion

PR #399, opened 2019-06-06
ORC-514: [C++] Malformed RowIndex stream meta exception when read statistics from stripe with bloom filter

rip-nsk:


c++/src/Reader.cc
```diff
@@ -530,6 +531,9 @@ namespace orc {
           (*indexStats)[column].push_back(entry.statistics());
         }
       }
+      if (streamKind != StreamKind::StreamKind_ROW_INDEX) {
```
omalley: Can you make this an else?

rip-nsk: sure.
is WriterImpl::writeStripe() works correctly? probably bloom filter streams should be added to the indexLength, not dataLength.

wgtmac: You are right. WriterImpl::writeStripe() should be fixed to include bloom filter index.

c++/src/Reader.cc
```diff
@@ -529,6 +530,11 @@ namespace orc {
           const proto::RowIndexEntry& entry = rowIndex.entry(j);
           (*indexStats)[column].push_back(entry.statistics());
         }
+      } else if (streamKind == StreamKind::StreamKind_BLOOM_FILTER_UTF8) {
+        // ORC-514
```
wgtmac: I'd suppose to fix those affected files instead of doing these here.

PR #412, opened 2019-07-12
ORC-363: Enable zStandard codec

wgtmac:
Support ZSTD codec for java ORC writer and reader. The implementation under the hood leverages airlift compressor library.

### DIFF
--- a/java/core/pom.xml
+++ b/java/core/pom.xml
@@ -97,11 +97,6 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>io.airlift</groupId>
-      <artifactId>slice</artifactId>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
 
   <build>
--- a/java/core/pom.xml
+++ b/java/core/pom.xml
@@ -97,6 +97,11 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.airlift</groupId>
+      <artifactId>slice</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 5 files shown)


java/core/pom.xml
```diff
@@ -97,6 +97,11 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.airlift</groupId>
+      <artifactId>slice</artifactId>
+      <scope>test</scope>
```
omalley: Unfortunately, this isn't a test scope, but an actual dependence. I tried to use a zstd file and it failed with a class not found exception.

I opened https://github.com/airlift/slice/issues/119 to discuss the problem with the slice maintainers.

As it stands, we can't add this to the build for Java. The best we can do is add an additional non-default maven profile that will pull in the GPL infected code. *sigh*

java/core/src/java/org/apache/orc/impl/WriterImpl.java
```diff
@@ -275,6 +277,9 @@ public static CompressionCodec createCodec(CompressionKind kind) {
       case LZ4:
         return new AircompressorCodec(kind, new Lz4Compressor(),
             new Lz4Decompressor());
+      case ZSTD:
+        return new AircompressorCodec(kind, new ZstdCompressor(),
+            new ZstdDecompressor());
```
omalley: We need a similar fix in WriterImpl (line 537).

java/core/pom.xml
omalley: They replied that jot-core is actually GPL with a classpath exemption. Unfortunately, that is still [category X](https://www.apache.org/legal/resolved.html#category-x) from Apache's point of view. We could either:
* Create a non-default maven profile that enables zstd and pulls in the GPL'ed code.
* Extend the shim layer to detect whether jot-core is available on the classpath and enable the zstd code at runtime. If we go that way, I'd recommend having the aircompressor block jot-core rather than slice and add jot-core as an optional dependency.

omalley: Ok, looking through the Apache legal discussions, I found the relevant thread about using [GPL+CPE in Apache projects](https://issues.apache.org/jira/browse/LEGAL-336).

wgtmac: @omalley So I still have to choose either way you provided above, right?

omalley: We either need to wait until aircompressor releases with https://github.com/airlift/aircompressor/pull/100 or we need to use one of the approaches above.

omalley: The summary of aircompressor#100 is that they remove the dependency on slice completely by pulling a hash function over.

PR #415, opened 2019-07-27
ORC-529: Allow configuration and table properties to control encryption.

omalley:
We need to allow the configuration to control encryption. This PR changes a few things:
* Factors a lot of the parsing of type names out of TypeDescription into the new ParserUtils.
* Changes the WriterOptions to have a string for encryption and one for masking.
* Uses annotations on the types based on the options. The user can also use type annotations to control encryption.
* Allows control of the KeyProvider using an ORC property.
* Adds a cache for the KeyProviders.
* Moves KeyProvider interface out of HadoopShims.
* Adds a service loader for ORC key providers so that we can use ORC key providers even if the Hadoop version doesn't support them.
* Fixes the TypeDecription.clone and equals for type attributes.
* Tests both the Hadoop KeyProvider API as well as the in memory key provider.


jsnorman:
hi omalley , currently i want to use the orc encryption and mask feature, but i can not find any manual about how to used in hive? 
i create a hive table in hive shell like this
create table testorc (
  name string
) stored as orc tblproperties ("orc.mask"="sha256:name",
"orc.encrypt"="pii:name");
since i want to test the column 'name' can be masked, but i failed.
can you provide tutorial about how to use this feature.

daddy-sunday:
I also have this requirement. Do you know how to use it

dongjoon-hyun:
I didn't tried it in Hive, but Apache ORC provides a document for Apache Spark.
- https://orc.apache.org/docs/spark-config.html
```sql
CREATE TABLE encrypted (
  ssn STRING,
  email STRING,
  name STRING
)
USING ORC
OPTIONS (
  hadoop.security.key.provider.path "kms://http@localhost:9600/kms",
  orc.key.provider "hadoop",
  orc.encrypt "pii:ssn,email",
  orc.mask "nullify:ssn;sha256:email"
)
```

PR #416, opened 2019-07-28
ORC-538. Fix a few problems with the type attributes.

omalley:
Fix a few problems with type annotations:
* clone() needs to copy them.
* setAttribute with null should clear them.
* equals should compare them
* file dump should print them.
* json file dump should make them nicer
* add test cases for the file dump commands.
* change the json write schema to be tree-like rather than flat.

### DIFF
--- a/java/core/src/java/org/apache/orc/TypeDescription.java
+++ b/java/core/src/java/org/apache/orc/TypeDescription.java
@@ -31,6 +31,7 @@ import org.apache.hadoop.hive.ql.exec.vector.TimestampColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.UnionColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
 import org.apache.orc.impl.SchemaEvolution;
+import org.jetbrains.annotations.NotNull;
 
 import java.io.Serializable;
 import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


java/core/src/java/org/apache/orc/TypeDescription.java
```diff
@@ -476,7 +476,11 @@ public TypeDescription withScale(int scale) {
    * @return this for method chaining
    */
   public TypeDescription setAttribute(String key, String value) {
-    attributes.put(key, value);
+    if (value == null) {
```
edgarRd: This behavior should be documented in the JavaDoc as well, otherwise it is not very clear. Additionally, should there be a `removeAttribute(...)` method instead of this? 
Seems like the `setAttribute` semantics would be to set a value to `null` if a `null` is assigned to a key rather than removing the attribute with that key.

java/tools/src/java/org/apache/orc/tools/FileDump.java
```diff
@@ -293,6 +293,39 @@ private static void printMetaData(List<String> files, Configuration conf,
...
+    if (children != null) {
+      switch (type.getCategory()) {
+        case STRUCT:
+          List<String> fields = type.getFieldNames();
+          for(int c = 0; c < children.size(); ++c) {
+            printTypeAnnotations(children.get(c), prefix + "." + fields.get(c));
+          }
+          break;
+        case MAP:
+          printTypeAnnotations(children.get(0), prefix + "._key");
```
edgarRd: Is it guaranteed that if `children != null` the list is non-empty?

omalley: For MAP and LIST, we can assume that they have 2 and 1 children respectively.

java/core/src/java/org/apache/orc/TypeDescription.java
omalley: I agree that I need to update the Javadoc. 

What is the semantics of having a null value as a type attribute? All of the examples of type attributes that I've seen, we always want a value. If we allow null values, that means that we need to put null checks everywhere we pull a value out. If we guarantee that they aren't null, that simplifies things.

So I guess we could have it throw on getting a null value and add the removeAttribute if you think that is cleaner. Thoughts?

edgarRd: I think `removeAttribute` is more clear and I agree with just with the non-null guarantee. Thanks for the changes!

edgarRd:
Lgtm overall. Thanks for pushing this fixes!

PR #425, opened 2019-08-31
ORC-526: Respect second fractions in orc-tools convert

yuokada:
## Todo
- [x] Unittests like https://github.com/apache/orc/blob/79f7595e25b770f3012c4df6a69e1895b6fb4006/java/tools/src/test/org/apache/orc/tools/TestJsonFileDump.java#L91-L154


yuokada:
@omalley Could you take a look if you get a chance?

PR #427, opened 2019-09-03
ORC-550: Use default system time zone if it is not present in the fil…

jcamachor:
…e footer


java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -919,7 +919,11 @@ protected TimestampTreeReader(int columnId, InStream presentStream, InStream dat
         this.readerTimeZone = TimeZone.getDefault();
       }
       if (context.getWriterTimezone() == null || context.getWriterTimezone().isEmpty()) {
-        this.base_timestamp = getBaseTimestamp(readerTimeZone.getID());
+        if (instantType) {
```
omalley: This is an unnecessary change. readerTimeZone is a utc if it is an instant or default if it isn't.

jcamachor: `readerTimeZone` may also be UTC if `context.getUseUTCTimestamp()` is true. However, in that case, if writer timezone is not present, we want to use default timezone to keep backwards compatible behavior.

omalley: No, in that case, the reader timezone should still be utc. Otherwise the upper layer will misinterpret the data.

jcamachor: The reader time zone setting does not change with this patch, it is still UTC.
Before ORC-341 went in, when writer time zone was not present in file footer, we were assuming that the file was written in system default time zone.
The incorrect assumption made in that patch was that if we used UTC time zone for the reader of a file and time zone was not present, we could just assume UTC for the writer since no shifting would be needed. However, this is not necessarily true with time zones with DST.
This patch restores the correct behavior and adds tests for these cases (the tests include files where time zone is missing from the file footer).

omalley: Ah, I understand, thank you. Using the reader timezone in the case of a missing timezone didn't work right and it has to be the default timezone.

PR #431, opened 2019-09-16
ORC-554: Float to timestamp schema evolution handles time/nanoseconds incorrectly

abstractdog:


java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -1411,7 +1411,7 @@ public void setConvertVectorElement(int elementNum) {
       long wholeSec = (long) Math.floor(seconds);
       timestampColVector.time[elementNum] = wholeSec * 1000;
       timestampColVector.nanos[elementNum] =
-          1_000_000 * (int) Math.round((seconds - wholeSec) * 1000);
+          Math.max(0, 1_000_000 * (int) Math.round((seconds - wholeSec) * 1000));
```
jcamachor: @abstractdog , I am confused on whether this is the correct fix for the issue. Is this negative value because of long overflow? @omalley , @prasanthj , @t3rmin4t0r , what would be the expected behavior in this case for ORC? Setting timestamp to `null`? Using `Long.MAX_VALUE`?

abstractdog: @jcamachor: yes, this was about long overflow, so I agree, it should be handled accordingly in my opinion, the next commit is about this:
https://github.com/apache/orc/pull/431/commits/9c2b9098597aa44ca2264e31e1b6ac8bd0659467
please take a look @omalley, @prasanthj , @t3rmin4t0r

omalley: The overflow case should become null. I know it feels weird, because in programming we use "null" to mean "no value." In SQL it means "error value," which you can see when you accidentally write "col1 = null" and get no matches.

abstractdog: thanks @omalley! so, do you mean that isNull[elementNum] = true is enough for the overflow case (without any other assignments in the timestamp col vector)?

omalley: you need that as well as cv.noNulls = false

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -1409,9 +1409,16 @@ public void setConvertVectorElement(int elementNum) {
...
       }
       long wholeSec = (long) Math.floor(seconds);
-      timestampColVector.time[elementNum] = wholeSec * 1000;
-      timestampColVector.nanos[elementNum] =
-          1_000_000 * (int) Math.round((seconds - wholeSec) * 1000);
+
+      if (seconds >= Long.MAX_VALUE || seconds <= Long.MIN_VALUE) { // overflow
+        timestampColVector.time[elementNum] = 0L;
+        timestampColVector.nanos[elementNum] = 0;
+        timestampColVector.isNull[elementNum] = true;
```
omalley: You also need to set timestampColVector.noNulls = false.

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -1409,9 +1409,16 @@ public void setConvertVectorElement(int elementNum) {
...
+      } else {
+        timestampColVector.time[elementNum] = wholeSec * 1000;
+        timestampColVector.nanos[elementNum] =
+            1_000_000 * (int) Math.round((seconds - wholeSec) * 1000);
```
omalley: I'm worried that your patch has removed the protection keeping the value above 0.

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -1409,9 +1409,16 @@ public void setConvertVectorElement(int elementNum) {
         seconds = SerializationUtils.convertFromUtc(local, seconds);
       }
       long wholeSec = (long) Math.floor(seconds);
-      timestampColVector.time[elementNum] = wholeSec * 1000;
-      timestampColVector.nanos[elementNum] =
-          1_000_000 * (int) Math.round((seconds - wholeSec) * 1000);
+
+      if (seconds >= Long.MAX_VALUE || seconds <= Long.MIN_VALUE) { // overflow
```
omalley: I think this should be seconds * 1000, since otherwise the time value can overflow.

java/core/src/test/org/apache/orc/impl/TestSchemaEvolution.java
```diff
@@ -2265,4 +2265,72 @@ public void testEvolutionToTimestamp() throws Exception {
...
+    TimestampColumnVector t1 = (TimestampColumnVector) batchTimeStamp.cols[0];
+
+    OrcFile.ReaderOptions options = OrcFile.readerOptions(conf);
+    Reader.Options rowOptions = new Reader.Options().schema(readerSchema);
+
+    try (Reader reader = OrcFile.createReader(testFilePath, options);
+      RecordReader rows = reader.rows(rowOptions)) {
+      assertTrue(rows.nextBatch(batchTimeStamp));
+
+      assertTrue(String.format("nanos should be > 0, instead it's: %d", t1.nanos[0]),
```
omalley: This test case is much too loose. Let me take a pass at strengthening it.

omalley:
Ok, take a look at https://github.com/omalley/orc/tree/orc-554 to see if it looks ok.

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
abstractdog: agree, let's multiple first and then check

java/core/src/test/org/apache/orc/impl/TestSchemaEvolution.java
abstractdog: yeah, test case is cleaner and covers+checks lots of values and use cases now, +1

abstractdog:
double-checked testcase locally, it's working, I would only rename floatToTimeStampPositiveOverflow -> floatToTimeStampOverflow (similar to double)

LGTM

PR #433, opened 2019-10-04
ORC-361: Remove the single thread restriction on the MemoryManagerImpl.

omalley:
The current MemoryManagerImpl assumes that only a single thread is writing ORC files. I've changed the design so that:
* There is a single MemoryManagerImpl object for all of the threads.
* The only expensive operations are when new writers are created or closed.
* There is a new method (checkMemory) that is used to determine if the scale has changed. It uses an atomic long to detect changes without contention.
* The rows between memory checks is moved into the writer.


omalley:
I simplified the new MemoryManager implementation some more.

PR #435, opened 2019-10-10
ORC-557: Fix the issue when orc-tool opens the huge file over 4GB

yuokada:
I have confirmed this patch works on my laptop.
But I don't have enough confidence to this patch.


omalley:
Thank you for your problem report and fix. I think we need to generalize the fix a little bit to walk through the input buffers for uncompressed streams. I'll upload a revised fix and the corresponding test.

omalley:
Sorry, I referenced this PR by mistake on the fix for ORC-552. I know this is still open.

omalley:
@yuokada can you please try out my fix on #439?

yuokada:
@omalley I've tried your PR. It looks good to me.

omalley:
Ok, this was fixed by 9891eae .

PR #439, opened 2019-10-21
ORC-557 Fix problem reading large header with uncompressed streams.

omalley:
This is my generalization of @yuokada 's patch for #435 .


omalley:
This patch pulls some shared code between CompressedStream and UncompressedStream into their parent class InStream. I'd fixed CompressedStream, but had a corresponding problem in UncompressedStream. By sharing the code, there is less duplicated code and they both are fixed.

PR #440, opened 2019-10-22
ORC-561: Avoid -Weverything

pitrou:


pitrou:
Well, the clang developers themselves advocate against using it:
https://quuxplusone.github.io/blog/2018/12/06/dont-use-weverything/

Regardless, as long as ORC-480 is fixed, I don't really care about this one.

omalley:
That is a fair point. Maybe I was too hasty on this. I thought you had turned off all of the warnings rather than moving from -Weverything to -Wall.

PR #441, opened 2019-10-22
ORC-480: [C++] Add an option to succeed build despite warnings

pitrou:
The extensive warnings that are set are useful for CI and development,
but a nuisance for external packagers and dependent projects which
can encounter failing builds because of warnings in ORC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,10 @@ option(INSTALL_VENDORED_LIBS
     "Install vendored thirdparty libraries along with liborc"
     ON)
 
+option(STOP_BUILD_ON_WARNING
+    "Fail the build on C++ warnings"
+    ON)
+
 # Make sure that a build type is selected
 if (NOT CMAKE_BUILD_TYPE)
   message(STATUS "No build type selected, default to ReleaseWithDebugInfo")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -49,6 +49,10 @@ option(INSTALL_VENDORED_LIBS
     "Install vendored thirdparty libraries along with liborc"
     ON)
 
+option(ENABLE_CPP_WARNINGS
```
omalley: Actually, I'd propose that we make it: STOP_BUILD_ON_WARNING
and default it to ON

CMakeLists.txt
```diff
@@ -74,20 +78,25 @@ if (NOT MSVC)
...
   set (CXX11_FLAGS "-std=c++11")
-  set (WARN_FLAGS "-Weverything -Wno-c++98-compat -Wno-missing-prototypes")
-  set (WARN_FLAGS "${WARN_FLAGS} -Wno-c++98-compat-pedantic -Wno-padded")
-  set (WARN_FLAGS "${WARN_FLAGS} -Wno-covered-switch-default")
-  set (WARN_FLAGS "${WARN_FLAGS} -Wno-missing-noreturn -Wno-unknown-pragmas")
-  set (WARN_FLAGS "${WARN_FLAGS} -Wno-gnu-zero-variadic-macro-arguments")
-  set (WARN_FLAGS "${WARN_FLAGS} -Wconversion")
-  if (CMAKE_HOST_APPLE AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "11.0")
-    set (WARN_FLAGS "${WARN_FLAGS} -Wno-c++2a-compat")
+  if (ENABLE_CPP_WARNINGS)
```
omalley: Use 
if (STOP_BUILD_ON_WARNING) 
  set (WARN_FLAGS "${WARN_FLAGS) -Werror")
endif ()

We absolutely should get the warnings even if they are building a non-dev build.

CMakeLists.txt
pitrou: Agreed. I'll post an update tomorrow.

pitrou: Done.

PR #442, opened 2019-10-24
ORC-562:Don't wrap readerSchema in acidSchema, if readerSchema is alr…

laszlopinter86:
…eady acid

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
+++ b/java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
@@ -86,10 +86,11 @@ public class SchemaEvolution {
     this.isOnlyImplicitConversion = true;
     this.fileSchema = fileSchema;
     this.isAcid = checkAcidSchema(fileSchema);
+    boolean readerSchemaIsAcid = readerSchema == null ? false : checkAcidSchema(readerSchema);
     this.includeAcidColumns = options.getIncludeAcidColumns();
-    this.readerColumnOffset = isAcid ? acidEventFieldNames.size() : 0;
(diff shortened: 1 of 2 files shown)


java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
```diff
@@ -87,7 +87,8 @@ public SchemaEvolution(TypeDescription fileSchema,
     this.fileSchema = fileSchema;
     this.isAcid = checkAcidSchema(fileSchema);
     this.includeAcidColumns = options.getIncludeAcidColumns();
-    this.readerColumnOffset = isAcid ? acidEventFieldNames.size() : 0;
+    this.readerColumnOffset = (isAcid && readerSchema != null && checkAcidSchema(readerSchema)) || !isAcid
```
omalley: I'd suggest keeping the checkAcidSchema(readerSchema) rather than calling it multiple times. This expression needs to be simplified.

java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
```diff
@@ -87,7 +87,8 @@ public SchemaEvolution(TypeDescription fileSchema,
...
+        ? 0 : acidEventFieldNames.size();
     if (readerSchema != null) {
       if (isAcid) {
```
omalley: You probably want to cut out here with if (isAcid && !isReaderAcid) rather than change addEventSchema.

laszlopinter86:
@omalley I changed the implementation and added new test.

omalley:
I've committed this.

PR #443, opened 2019-10-25
ORC-563: ORC-540 could break schema evolution on PPD codepaths

abstractdog:


abstractdog:
I need to create unit test for column removal case too

abstractdog:
could you please take a look @omalley , @prasanthj ?

omalley:
This seems to be the same issue as #452 .

I've tried #452 's fix with your test case and it passes.

PR #455, opened 2019-11-26
ORC-27: Add support for proleptic Gregorian calendar

omalley:
For better support of dates before 1600AD. This is a work in progress, I still need to fix the predicate push down. 

It would also be good to have the C++ writer set the proleptic calendar flag.

### DIFF
--- a/proto/orc_proto.proto
+++ b/proto/orc_proto.proto
@@ -340,6 +340,15 @@ message Encryption {
   optional KeyProviderKind keyProvider = 4;
 }
 
+enum CalendarKind {
+  UNKNOWN_CALENDAR = 0;
+   // The Java default calendar changes from Julian to Gregorian
+   // in 1583.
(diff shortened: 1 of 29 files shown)


omalley:
The CI will be broken until we release storage api 2.7.1.

omalley:
The predicate push down and file and stripe statistics work now. I think this patch is pretty close.

jcamachor:
LGTM, +1

proto/orc_proto.proto
```diff
@@ -340,6 +340,15 @@ message Encryption {
   optional KeyProviderKind keyProvider = 4;
 }
 
+enum CalendarKind {
+  UNKNOWN_CALENDAR = 0;
+   // The Java default calendar changes from Julian to Gregorian
+   // in 1583.
```
abstractdog: just noticed, actually it's 1582, we can fix this minor comment issue later

PR #456, opened 2019-12-02
ORC-573: [C++] Fix several issues in Int128 and Decimal

wgtmac:
Before implementing predicate pushdown for C++ reader, we need to fix several issues related to decimal types.

1. Support Decimal::toString() to trim trailing zeros.
2. Int128::toDecimalString has an issue when it deals with substr operation which may lead to wrong string representation of decimal. Fix it to handle scale correctly.
3. Apply DecimalColumnWriter to trim trailing zeros before writing to bloom filter.

### DIFF
--- a/c++/include/orc/Common.hh
+++ b/c++/include/orc/Common.hh
@@ -274,6 +274,30 @@ namespace orc {
     FUTURE = INT32_MAX
   };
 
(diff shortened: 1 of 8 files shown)


xndai:
Can you please add a unit test? Thanks.

xndai:
LGTM

PR #457, opened 2019-12-02
ORC-556: Do not check for TypeDescription attribute equality during schema evolution check

shardulm94:
During schema evolution checks, for a field with the same category, if the read schema has some attributes but the file schema does not, then the code will try to use `ConvertTreeReaderFactory` to create a converter, but will eventually fail because it sees that the category is the same https://github.com/apache/orc/blob/master/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java#L1669. Since the `ConvertTreeReaderFactory` does not require/understand these attribute values, we should probably not be checking them when deciding whether conversion is required. 

@omalley Can you take a look?

### DIFF
--- a/java/core/src/findbugs/exclude.xml
+++ b/java/core/src/findbugs/exclude.xml
@@ -65,4 +65,9 @@
     <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE"/>
     <Class name="org.apache.orc.TestStringDictionary"/>
   </Match>
+  <Match>
+    <Bug pattern="EQ_UNUSUAL"/>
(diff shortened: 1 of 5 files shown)


omalley:
This looks good, but can you add some javadoc and a unit test?

shardulm94:
@omalley PR is ready for another round of review

omalley:
It is failing findbugs with:

`[INFO] org.apache.orc.TypeDescription.equals(Object) is unusual [org.apache.orc.TypeDescription] At TypeDescription.java:[line 405] EQ_UNUSUAL`

If you are building with Maven, you can use `-Panalyze` to run findbugs.

You could either suppress the warning or refactor equals to make it happy.

PR #458, opened 2019-12-04
ORC-574: Use const references for string statistics min and max to avoid copy construction

dzanter:
… std::string alloc/delete

due to the getMaximum/Minimum returning std:string copies.

Changing the getMaximum/getMinimum methods to return const vals will prevent these alloc/copy/deletes from occurring.

This results in about a 30% cpu boost to write.

### DIFF
--- a/c++/include/orc/Statistics.hh
+++ b/c++/include/orc/Statistics.hh
@@ -282,13 +282,13 @@ namespace orc {
      * Get the minimum value for the column.
      * @return minimum value
      */
-    virtual std::string getMinimum() const = 0;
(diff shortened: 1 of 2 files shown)


xndai:
LGTM. Thanks.

xndai:
@dzanter Can you please reformat and update the to be something like -

ORC-574: Use const references for string statistics min and max to avoid copy construction

I will merge after it's updated.

omalley:
I committed this, but referenced the wrong issue.

omalley:
For the record, the commit was c210565e9e559feb9ec64ffea967a9a36e9d06dc.

PR #461, opened 2019-12-17
ORC-40: [C++] Support building SearchArgument

wgtmac:
1. Implement Literal class to handle literal values in a predicate of different data types.
2. Port TruthValue, ExpressionTree, PredicateLeaf, and SearchArgument from Java hive-storage-api.
3. Port SargsApplier from Java ORC reader.

### DIFF
--- a/c++/include/orc/sargs/Literal.hh
+++ b/c++/include/orc/sargs/Literal.hh
@@ -1,0 +1,128 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef ORC_LITERAL_HH
+#define ORC_LITERAL_HH
+
+#include "orc/Int128.hh"
+#include "orc/Vector.hh"
+
+namespace orc {
+
+  /**
+   * Possible types for predicates
+   */
+  enum class PredicateType {
+    LONG = 0, FLOAT, STRING, DATE, DECIMAL, TIMESTAMP, BOOLEAN
+  };
+
+  /**
+   * Represents a literal value in a predicate
+   */
+  class Literal {
+  public:
+    Literal(const Literal &r);
+    ~Literal();
+    Literal& operator=(const Literal& r);
+    bool operator==(const Literal& r) const;
+    bool operator!=(const Literal& r) const;
+
+    /**
+     * Create a literal of null value for a specific type
+     */
+    Literal(PredicateType type);
+
+    /**
+     * Create a literal of LONG type
+     */
+    Literal(int64_t val);
+
+    /**
+     * Create a literal of FLOAT type
+     */
+    Literal(double val);
+
+    /**
+     * Create a literal of BOOLEAN type
+     */
+    Literal(bool val);
+
+    /**
+     * Create a literal of Timestamp or DATE type
+     */
+    Literal(PredicateType type, int64_t val);
+
+    /**
+     * Create a literal of STRING type
+     */
+    Literal(const char * str, size_t size);
+
+    /**
+     * Create a literal of DECIMAL type
+     */
+    Literal(Int128 val, int32_t precision, int32_t scale);
+
+    /**
+     * Getters of a specific data type for not-null literals
+     */
+    int64_t getLong() const;
+    int64_t getDate() const;
+    int64_t getTimestamp() const;
+    double getFloat() const;
+    std::string getString() const;
+    bool getBool() const;
+    Decimal getDecimal() const;
+
+    /**
+     * Check if a literal is null
+     */
+    bool isNull() const { return mIsNull; }
+
+    PredicateType getType() const { return mType; }
+    std::string toString() const;
+    size_t getHashCode() const { return mHashCode; }
+
+  private:
+    size_t hashCode() const;
+
+    union LiteralVal {
+      int64_t IntVal;
+      double DoubleVal;
+      int64_t DateVal;
+      char * Buffer;
+      int64_t TimeStampVal;
+      Int128 DecimalVal;
+      bool BooleanVal;
+
+      // explicitly define default constructor
+      LiteralVal(): DecimalVal(0) {}
+    };
+
+  private:
+    LiteralVal mValue;     // data value for this literal if not null
+    PredicateType mType;   // data type of the literal
+    size_t mSize;          // size of mValue if it is Buffer
+    int32_t mPrecision;    // precision of decimal type
+    int32_t mScale;        // scale of decimal type
+    bool mIsNull;          // whether this literal is null
+    size_t mHashCode;      // precomputed hash code for the literal
+  };
+
+} // namespace orc
+
+#endif //ORC_LITERAL_HH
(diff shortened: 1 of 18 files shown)


wgtmac:
@omalley @majetideepak @xndai Can you take a look at this patch when you have time? Most of the new classes are ported from java ORC to support predicate pushdown of C++ reader. Thanks!

c++/src/sargs/ExpressionTree.cc
```diff
@@ -0,0 +1,192 @@
...
+        return leaves[mLeaf];
+      case Operator::CONSTANT:
+        return mConstant;
+      default:
+        throw std::invalid_argument("Unknown operator!");
+    }
+  }
+
+  std::string to_string(TruthValue truthValue) {
+    switch (truthValue) {
```
chunyang-wen: Can we use a map to store the relationship which uses the enum as the key?

c++/src/sargs/ExpressionTree.cc
```diff
@@ -0,0 +1,192 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ExpressionTree.hh"
+
+#include <cassert>
```
chunyang-wen: It's better to follow the sequence of header style of existing code.
+ system headers
+ third party headers
+ project headers

c++/src/sargs/ExpressionTree.cc
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  TruthValue ExpressionTree::evaluate(
+                                 const std::vector<TruthValue>& leaves) const {
+    TruthValue result;
+    switch (mOperator) {
+      case Operator::OR:
+      {
+        result = mChildren.at(0)->evaluate(leaves);
+        for (size_t i = 1; i < mChildren.size(); ++i) {
```
chunyang-wen: Is it better to introduce for short-circuit evaluation for `AND` and `OR` operators?

c++/src/sargs/Literal.cc
```diff
@@ -0,0 +1,321 @@
...
+    return *this;
+  }
+
+  std::string Literal::toString() const {
+    if (mIsNull) {
+      return "null";
+    }
+
+    std::ostringstream sstream;
+    std::string str;
```
chunyang-wen: move `std::string str` to the place where it is used.

c++/src/sargs/Literal.cc
```diff
@@ -0,0 +1,321 @@
...
+    }
+  }
+
+  bool Literal::operator!=(const Literal& r) const {
+    return !(*this == r);
+  }
+
+  int64_t Literal::getLong() const {
+    if (mIsNull) {
+      throw std::logic_error("cannot call getLong for null!");
```
chunyang-wen: Refact those two if-clauses into a function.

majetideepak:
@wgtmac I will take a look by the end of this week.

c++/test/TestPredicateLeaf.cc
```diff
@@ -0,0 +1,827 @@
...
+    PredicateLeaf pred1(
+      PredicateLeaf::Operator::NULL_SAFE_EQUALS,
+      PredicateType::BOOLEAN,
+      "x",
+      Literal(false));
+    EXPECT_EQ(TruthValue::NO,
+              evaluate(pred1, createBooleanStats(10, 10)));
+    EXPECT_EQ(TruthValue::YES_NO,
+              evaluate(pred1, createBooleanStats(10, 0)));
+
```
stiga-huang: It'd be better to also test "x <=> NULL" (x NULL_SAFE_EQUALS to NULL).

c++/src/sargs/ExpressionTree.cc
wgtmac: I don't think we are following this pattern.

c++/src/sargs/ExpressionTree.cc
wgtmac: Good catch, will do.

c++/include/orc/sargs/Literal.hh
```diff
@@ -0,0 +1,128 @@
...
+
+#include "orc/Int128.hh"
+#include "orc/Vector.hh"
+
+namespace orc {
+
+  /**
+   * Possible types for predicates
+   */
+  enum class PredicateType {
```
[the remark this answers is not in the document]
wgtmac: Yes, the concept of predicate stays unchanged across different systems. I will rename PredicateType to PredicateDataType to eliminate the confusion.

c++/include/orc/sargs/TruthValue.hh
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+#ifndef ORC_TRUTHVALUE_HH
+#define ORC_TRUTHVALUE_HH
+
+namespace orc {
+
+  enum class TruthValue {
+    YES, NO, IS_NULL, YES_NULL, NO_NULL, YES_NO, YES_NO_NULL
```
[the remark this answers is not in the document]
wgtmac: These values are copied from [hive-storage-api library](https://github.com/apache/hive/blob/master/storage-api/src/java/org/apache/hadoop/hive/ql/io/sarg/SearchArgument.java#L39) which unfortunately does not provide more comment than here. I will add some comment here later.

c++/src/sargs/PredicateLeaf.cc
```diff
@@ -0,0 +1,698 @@
...
+
+    auto trueCount = stats.bucketstatistics().count(0);
+    auto falseCount = stats.numberofvalues() - trueCount;
+    switch (op) {
+      case PredicateLeaf::Operator::IS_NULL:
+        return hasNull ? TruthValue::YES_NO : TruthValue::NO;
+      case PredicateLeaf::Operator::NULL_SAFE_EQUALS: {
+        if (literals.at(0).getBool()) {
+           return trueCount == 0 ?
+                  TruthValue::NO : TruthValue::YES_NO;
```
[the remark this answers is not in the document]
wgtmac: I think you are right, will fix it.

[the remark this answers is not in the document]
wgtmac: You are right. The original implementation is too conservative.

c++/src/sargs/PredicateLeaf.cc
```diff
@@ -0,0 +1,698 @@
...
+      return hasNull ? TruthValue::YES_NO_NULL : TruthValue::YES_NO;
+    }
+
+    auto trueCount = stats.bucketstatistics().count(0);
+    auto falseCount = stats.numberofvalues() - trueCount;
+    switch (op) {
+      case PredicateLeaf::Operator::IS_NULL:
+        return hasNull ? TruthValue::YES_NO : TruthValue::NO;
+      case PredicateLeaf::Operator::NULL_SAFE_EQUALS: {
+        if (literals.at(0).getBool()) {
```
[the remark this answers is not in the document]
wgtmac: I am not sure if this is a valid case. I suppose NULL will not appear as a literal in an NULL_SAFE_EQUALS expression.

c++/src/sargs/PredicateLeaf.cc
```diff
@@ -0,0 +1,698 @@
...
+                              , mColumnName(colName)
+                              , mLiterals(literals.begin(), literals.end()) {
+    mHashCode = hashCode();
+    validate();
+  }
+
+  void PredicateLeaf::validate() const {
+    switch (mOperator) {
+      case Operator ::IS_NULL:
+        if (mColumnName.empty()) {
```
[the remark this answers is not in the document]
wgtmac: It will check mLiterals inside and will miss the check if validate() is moved to the beginning.

c++/src/sargs/ExpressionTree.cc
wgtmac: I don't think map brings any benefit here.

c++/test/TestPredicateLeaf.cc
wgtmac: Done. I have resolved all issues. Can you take a look again? Thanks! @stiga-huang

stiga-huang:
LGTM. Thanks for addressing the comments!

majetideepak:
@wgtmac I plan to do a lazy review after merge for this. Since all of this is new code, I am not too worried about regressions. 
Thanks to @stiga-huang for giving a round of feedback.

majetideepak:
Feel free to merge this if its ready.

PR #467, opened 2020-01-11
ORC-588: Static field or method should be directly referred by its class

lamber-ken:
Static field or method should be directly referred by its class


lamber-ken:
> This is good, but the build is failing because you are hitting a find bugs warning about an unused field, which must be the utils field. Can you remove its definition in the patch.

Done, thanks for remind me. :)

PR #468, opened 2020-01-13
ORC-589: [C++] add checks about negative dictionary entry lengths

boroknagyz:
The library used to have a check for negative dictionary entry lengths.
However we don't have it on current master. It causes a problem because
it will use a negative value (casted to unsigned) to allocate a
DataBuffer. This commit adds the missing check.

Testing
* added a corrupt orc file and a unit test


c++/src/ColumnReader.cc
```diff
@@ -581,6 +581,8 @@ namespace orc {
     lengthDecoder->next(lengthArray + 1, dictSize, nullptr);
     lengthArray[0] = 0;
     for(uint32_t i = 1; i < dictSize + 1; ++i) {
+      if (lengthArray[i] < 0)
```
wgtmac: Please add curly brace here even for a single line if-statement.

c++/test/TestCorruptFiles.cc
```diff
@@ -0,0 +1,56 @@
...
+
+#include "Adaptor.hh"
+
+#include "wrap/gmock.h"
+#include "wrap/gtest-wrapper.h"
+
+namespace orc {
+
+  TEST(TestCorruptFiles, testNegativeDictEntryLengths) {
+
```
wgtmac: remove empty line

c++/src/ColumnReader.cc
boroknagyz: Done

c++/test/TestCorruptFiles.cc
boroknagyz: Done

c++/test/TestCorruptFiles.cc
```diff
@@ -0,0 +1,55 @@
...
+#include "orc/Reader.hh"
+
+#include "Adaptor.hh"
+
+#include "wrap/gmock.h"
+#include "wrap/gtest-wrapper.h"
+
+namespace orc {
+
+  TEST(TestCorruptFiles, testNegativeDictEntryLengths) {
```
stiga-huang: I think moving this into TestFileScan.testErrorHandling in TestFileScan.cc is simpler.

boroknagyz: Done

PR #470, opened 2020-01-17
Fix missing @Override annotation

lamber-ken:
Fix missing @Override annotation.


omalley:
I committed this as b594deb86b38a48af951084dc2a633eed3fb8110 without marking this PR as fixed.

PR #473, opened 2020-01-27
ORC-593 Row-level skipping

pgaref:
Proof of concept patch for row-level skipping using new nextVector method as part of TreeReader.
Currently tested on orc-1.5.8


omalley:
You should change the base of the pull request to be branch-1.5, so that it doesn't show up as a huge merge conflict.

PR #476, opened 2020-02-04
ORC-751: [C++] Implement Predicate Pushdown for C++ Reader

wgtmac:
1. Use RowReaderOptions to pass SearchArgument to enable PPD.
2. Modify RowReaderImpl::startNextStripe to seek to next matched row group
   based on PPD.
3. RowReaderImpl::next seeks to next matched row group based on PPD result.
4. RowReaderImpl::seekToRow also jumps to the 1st matched row group after
   target row.


wgtmac:
@stiga-huang Can you help review and verify this patch in Impala? Thanks!

luksan47:
Hi, @wgtmac, I've started to implement predicate pushdown in Impala, will test the PR that way.

wgtmac:
@luksan47 cool! Let me know if you have any feedback.

csringhofer:
I reviewed Norbert's implementation for Impala (https://gerrit.cloudera.org/#/c/15403/) and found some problems that are very hard to implement on the client side:
https://issues.apache.org/jira/browse/ORC-611
https://issues.apache.org/jira/browse/ORC-612

The issues are related to types that are generally problematic (TIMESTAMP, CHAR, VARCHAR). It is possible to disable predicate pushdown for this types, but in the long term it would be good to have a proper solution.

Can you look at the issues? I think that it would be much better to implement them in the ORC lib, because it is easy to create buggy implementations in clients.

wgtmac:
@csringhofer Thanks for letting me know! I have also noticed that I need to use lowerbound & upperbound in TimestampColumnStatistics to handle timezone difference. Will take a deeper look into these issues.

c++/src/Reader.cc
```diff
@@ -944,13 +1035,67 @@ namespace orc {
...
+    if (sargsApplier) {
+      uint64_t nextRowToRead = advanceToNextRowGroup(currentRowInStripe,
+                                                     rowsInCurrentStripe,
+                                                     footer->rowindexstride(),
+                                                     sargsApplier->getRowGroups());
+      if (currentRowInStripe != nextRowToRead) {
+        // it is guaranteed to be at start of a row group
+        currentRowInStripe = nextRowToRead;
+        if (currentRowInStripe < rowsInCurrentStripe) {
+          seekToRowGroup(static_cast<uint32_t>(currentRowInStripe / footer->rowindexstride()));
```
csringhofer: I think that there is a hidden performance regression here:
at the end of the call chain, e.g. ColumnReader::seekToRowGroup(), RleDecoderV2::seek(), ZlibDecompressionStream::seek(), the inputBuffer is reset to nullptr:
https://github.com/apache/orc/blob/028261ad6e26d8ae9f9483c8aeb76a85c7f1168b/c%2B%2B/src/Compression.cc#L573
This leads to calling ZlibDecompressionStream::readBuffer() and reading from the stream again, even if the whole stream was buffered previously to inputBuffer (which is usually the case in Impala, as we set the block size to 8 MB, but can also cause problems with the default 256KB).
In the worst case this can lead to reading the whole column O(num_of_row_groups) times. 

The solution could be to make ZlibDecompressionStream/BlockDecompressionStream::seek() smarter and keep the inputBuffer if the new position (which is the header byte of the new compression block) is still in it.

A further improvement could be to make ZlibDecompressionStream/BlockDecompressionStream::Skip() smarter to avoid decompressing all the skipped blocks in this case ( https://github.com/apache/orc/blob/028261ad6e26d8ae9f9483c8aeb76a85c7f1168b/c%2B%2B/src/Compression.cc#L545 ), but this is not a regression so can go to another patch.

wgtmac: You are right. Performance of seeking is another headache if you use large compression blocks. In our internal ORC setup, we have optmized this by finishing compression blocks at the end of every row group. In this way, seeking will never read compression blocks of other row groups.

c++/include/orc/Reader.hh
```diff
@@ -245,6 +251,11 @@ namespace orc {
      * What scale should all Hive 0.11 decimals be normalized to?
      */
     int32_t getForcedScaleOnHive11Decimal() const;
+
+    /**
+     * Get search argument for predicate push down
+     */
+    std::shared_ptr<SearchArgument> getSearchArgument() const;
```
luksan47: It would be useful for debugging to get statistics about the number of row groups the reader skipped because of the sargs. Could you add such a function to the interface?

wgtmac: Yes, that would be helpful. Will add it once I have received enough feedback and probably after ORC-611 & ORC-612 have been fixed.

c++/src/Reader.cc
csringhofer: I have created a Jira (ORC-614) for the optimization of seek(), and we put together an implementation with Norbert, he will upload it soon as pull request. It seems to solve the regressions we experienced.

About finishing compression blocks at row group end: I actually created a ticket (IMPALA-8449) to do something similar in Parquet:
In Parquet we came from the other direction - pages are both the unit of indexing and compression in Parquet, but they are not aligned, so unlike ORC, index entries can point to different row numbers for every column. This decreases decreases the efficiency of combining filters for different columns + makes the reader implementation very complex.

wgtmac:
@csringhofer @luksan47 @stiga-huang I have changed this PR to reflect the feedback. The fix for timestamp is addressed by a separate PR: [https://github.com/apache/orc/pull/543](https://github.com/apache/orc/pull/543). Please review it again. Sorry for the long delay.

wgtmac:
@dongjoon-hyun @pgaref This PR has been sleeping for a decade. Really appreciate it if you can help reviewing it once get the chance. Thanks!

c++/test/TestPredicatePushdown.cc
```diff
@@ -0,0 +1,183 @@
...
+    writer->close();
+
+    std::unique_ptr<InputStream> inStream(new MemoryInputStream (
+      memStream.getData(), memStream.getLength()));
+    ReaderOptions readerOptions;
+    options.setMemoryPool(pool);
+    auto reader = createReader(std::move(inStream), readerOptions);
+    EXPECT_EQ(3500, reader->getNumberOfRows());
+
+    // build search argument (x >= 300000 AND x < 600000)
```
pgaref, line 73: Can we also explain the data of every Pred in terms of row-groups (as it would be easier to understand).

c++/test/TestPredicatePushdown.cc
```diff
@@ -0,0 +1,183 @@
...
+      rowReaderOpts.setSearchArgument(std::move(sarg));
+      auto rowReader = reader->createRowReader(rowReaderOpts);
+
+      auto readBatch = rowReader->createRowBatch(2000);
+      auto& batch0 = dynamic_cast<StructVectorBatch&>(*readBatch);
+      auto& batch1 = dynamic_cast<LongVectorBatch&>(*batch0.fields[0]);
+      auto& batch2 = dynamic_cast<StringVectorBatch&>(*batch0.fields[1]);
+
+      EXPECT_EQ(true, rowReader->next(*readBatch));
+      EXPECT_EQ(1000, readBatch->numElements);
```
pgaref, line 148: I would assume 2 row-groups selected based on the comment above

c++/test/TestPredicatePushdown.cc
```diff
@@ -0,0 +1,183 @@
...
+        EXPECT_EQ(300 * i, batch1.data[i - 3000]);
+        EXPECT_EQ(std::to_string(10 * i),
+                  std::string(batch2.data[i - 3000], static_cast<size_t>(batch2.length[i - 3000])));
+      }
+
+      EXPECT_EQ(false, rowReader->next(*readBatch));
+      EXPECT_EQ(3500, rowReader->getRowNumber());
+
+      // test seek to 3rd row group but is adjusted to 4th row group
+      rowReader->seekToRow(2500);
```
pgaref, line 169: Test-seek should be on a separate test?

c++/src/Reader.cc
```diff
@@ -376,6 +419,14 @@ namespace orc {
     return forcedScaleOnHive11Decimal;
   }
 
+  bool RowReaderImpl::getPPDStats(std::pair<uint64_t, uint64_t>& stats) const {
+    if (!sargsApplier) {
```
pgaref, line 423: I would change this to if (sargsApplier) {} return false;

pgaref (commented):
Hey @wgtmac thanks for the reminder! 
Just went through the code and left some comments -- I believe we should add some more Tests here to increase confidence. Few example include: testPredEvalWithBooleanStats, testPredEvalWithIntStats etc. https://github.com/apache/orc/blob/949c7444ef98680026d9f2b93d7299b76cdfb030/java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java#L581

On another note, ORC-40 seems to be about  building SearchArgument so we should move this to a new ticket -- open ORC-751 for this

wgtmac:
Thanks @pgaref for the comments! I will fix them and then let you know.

dongjoon-hyun:
Thank you for pinging me, @wgtmac .

c++/src/Reader.hh
```diff
@@ -19,13 +19,14 @@
 #ifndef ORC_READER_IMPL_HH
 #define ORC_READER_IMPL_HH
 
+#include "orc/Exceptions.hh"
 #include "orc/Int128.hh"
 #include "orc/OrcFile.hh"
 #include "orc/Reader.hh"
 
 #include "ColumnReader.hh"
-#include "orc/Exceptions.hh"
```
dongjoon-hyun, line 27: The move from line 27 to 22 looks reasonable. BTW, do we have a C++ convention for `C++ include` statement order?

c++/src/ColumnReader.cc
```diff
@@ -1553,6 +1558,8 @@ namespace orc {
     ColumnReader::seekToRowGroup(positions);
     valueStream->seek(positions.at(columnId));
     scaleDecoder->seek(positions.at(columnId));
+    buffer = nullptr;
+    bufferEnd = nullptr;
```
dongjoon-hyun, line 1562: Just a question: The changes in this file look general instead of specific to this PPD. Previously, we don't need these?

dongjoon-hyun (commented):
I didn't compare the logic with Java reader yet, but I'm wondering if we keep the logic structure consistently in both readers.

c++/src/Reader.hh
wgtmac: Unfortunately I don't think we had a clear C++ coding style. Maybe we can use cpplint tool to enfoce it.

dongjoon-hyun: Thanks. No problem then. I'll ignore the C++ coding style.

c++/src/ColumnReader.cc
dongjoon-hyun: Then, can we have this change first as a separate PR?

dongjoon-hyun:
Since https://github.com/apache/orc/pull/645 is merged, could you rebase this to the master once more, @wgtmac ?

wgtmac:
> Since #645 is merged, could you rebase this to the master once more, @wgtmac ?

@dongjoon-hyun I'm working on it. Will let you know once ready.

c++/src/Reader.cc
```diff
@@ -376,6 +419,14 @@ namespace orc {
...
+      return false;
+    }
+    stats = sargsApplier->getStats();
```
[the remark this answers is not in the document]
wgtmac: I will do that in a separate patch as this patch does not evaluate stripe stats yet.

c++/src/sargs/SargsApplier.cc
```diff
@@ -25,7 +26,13 @@ namespace orc {
                                     const std::string& colName) {
     for (uint64_t i = 0; i != type.getSubtypeCount(); ++i) {
       if (type.getFieldName(i) == colName) {
-        return type.getSubtype(i)->getColumnId();
+        if (type.getKind() == CHAR || type.getKind() == VARCHAR) {
+          ///FIXME: disable PPD for char and varchar types as their rules
```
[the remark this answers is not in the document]
wgtmac: I don't have a clean solution here. So I have left a comment here to fix it in the future.

c++/test/TestPredicatePushdown.cc
wgtmac: Done

c++/test/TestPredicatePushdown.cc
wgtmac: The above comment for pred is slightly wrong. But still 2 row groups are selected.

c++/test/TestPredicatePushdown.cc
wgtmac: Done

wgtmac:
> Hey @wgtmac thanks for the reminder!
> Just went through the code and left some comments -- I believe we should add some more Tests here to increase confidence. Few example include: testPredEvalWithBooleanStats, testPredEvalWithIntStats etc.
> 
> https://github.com/apache/orc/blob/949c7444ef98680026d9f2b93d7299b76cdfb030/java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java#L581
> 
> On another note, [ORC-40](https://issues.apache.org/jira/browse/ORC-40) seems to be about building SearchArgument so we should move this to a new ticket -- open [ORC-751](https://issues.apache.org/jira/browse/ORC-751) for this

I have covered these kinds of cases in the TestPredicateLeaf.cc in a earlier commit. I can add more progressively to gain more confidence.

dongjoon-hyun:
Thank you for updating, @wgtmac . But, it looks a little weird to me~

.asf.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+# https://cwiki.apache.org/confluence/display/INFRA/git+-+.asf.yaml+features
+---
+github:
+  description: "Apache ORC - the smallest, fastest columnar storage for Hadoop workloads"
+  homepage: https://orc.apache.org/
+  enabled_merge_buttons:
+    merge: false
+    squash: true
+    rebase: true
```
dongjoon-hyun, line 24: This should not be here.

PR #481, opened 2020-02-13
ORC-599 Bumping up guava to 28.1-jre

pgaref:
Even though guava dependency is used only by Test classes it make sense to upgrade due to known security vulnerabilities  https://nvd.nist.gov/vuln/detail/CVE-2018-10237 and to reduce the dependency footprint e.g. Apache Hive is upgrading to 28.1-jre https://issues.apache.org/jira/browse/HIVE-21569


dongjoon-hyun:
Hi, All. This is backported to branch-1.6 for Apache ORC 1.6.8.

PR #483, opened 2020-02-14
ORC-601: Add more debug info to error messages in the scanner

luksan47:
This commit adds more debug info and unifies error messages in
the readers and related classes.

Tests:
  - Added test file to check for valid error message related to
    streams.

### DIFF
--- a/c++/src/ColumnReader.hh
+++ b/c++/src/ColumnReader.hh
@@ -58,6 +58,21 @@ namespace orc {
                               proto::Stream_Kind kind,
                               bool shouldStream) const = 0;
 
+    std::unique_ptr<SeekableInputStream>
+                    getStreamOrThrow(const Type& type,
+                                     proto::Stream_Kind kind,
+                                     bool shouldStream) const {
+      std::unique_ptr<SeekableInputStream> stream =
+          getStream(type.getColumnId(), kind, shouldStream);
+      if (stream == nullptr) {
+        std::stringstream buffer;
+        buffer << Stream_Kind_Name(kind) << " stream not found in column "
+          << type.getColumnId() << " of type " << type.toString();
+        throw ParseError(buffer.str());
+      }
+      return stream;
+    }
+
     /**
      * Get the memory pool for this reader.
      */
--- a/c++/src/io/InputStream.cc
+++ b/c++/src/io/InputStream.cc
@@ -56,6 +56,13 @@ namespace orc {
     // PASS
   }
 
+  void SeekableInputStream::throwFrom(const char* class_name) {
+    std::stringstream msg;
+    msg << "Read past end of stream in " << class_name
+        << "or some errors occured: " << this->getName();
+    throw ParseError(msg.str());
+  }
+
   SeekableArrayInputStream::~SeekableArrayInputStream() {
     // PASS
   }
--- a/c++/src/io/InputStream.cc
+++ b/c++/src/io/InputStream.cc
@@ -56,7 +56,7 @@ namespace orc {
     // PASS
   }
 
-  void SeekableInputStream::throwFrom(const char* class_name) {
+  void SeekableInputStream::throwParseErrorFrom(const char* class_name) {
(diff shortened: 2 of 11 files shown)


luksan47:
Realised that this PR also fixes ORC-592 with the tests.
Also found out that I missed the tests for dictionary encoding, gonna add them with my next commit.

c++/test/TestColumnReaderInvalidStripes.cc
```diff
@@ -0,0 +1,267 @@
+
```
wgtmac: Please add the license header here

c++/test/TestColumnReaderInvalidStripes.cc
```diff
@@ -0,0 +1,267 @@
...
+    public:
+      virtual ~MockStripeStreams() override;
+
+      std::unique_ptr<SeekableInputStream> getStream(uint64_t columnId,
+                                                    proto::Stream_Kind kind,
+                                                    bool stream)
+                                                    const override {
+      return std::unique_ptr<SeekableInputStream>
+              (getStreamProxy(columnId, kind, stream));
+    }
```
wgtmac: Fix alignment and style, same for below lines.

c++/src/io/InputStream.cc
```diff
@@ -56,6 +56,13 @@ namespace orc {
     // PASS
   }
 
+  void SeekableInputStream::throwFrom(const char* class_name) {
+    std::stringstream msg;
+    msg << "Read past end of stream in " << class_name
+        << " or some errors occured: " << this->getName();
+    throw ParseError(msg.str());
```
wgtmac: Either rename it to throwParseErrorFrom or make it more general to throw other exceptions

c++/src/ColumnReader.hh
```diff
@@ -58,6 +58,21 @@ namespace orc {
                               proto::Stream_Kind kind,
                               bool shouldStream) const = 0;
 
+    std::unique_ptr<SeekableInputStream>
+                    getStreamOrThrow(const Type& type,
```
wgtmac: I prefer to make it virtual and remove getStream interface.

luksan47: I did not want to remove it since I saw it used here without throwing error:
https://github.com/apache/orc/blob/ffb1a2b51f25b909a1d5da1090f8b471b8fca286/c%2B%2B/src/ColumnReader.cc#L54
Should we throw there or not?

How about making getStream protected while keeping getStreamOrThrow non-virtual?

c++/test/TestColumnReaderInvalidStripes.cc
luksan47: Is there any problem with the MOCK_CONST_METHODN styles? I found the alignment in TestColumnReader.cc pretty ugly.

c++/src/ColumnReader.hh
wgtmac: Definitely not throw there. How about adding a default input parameter in getStream to advise whether it should throw or not if stream is null?
` getStream(uint64_t columnId, proto::Stream_Kind kind, bool shouldStream, bool optional = false)`

wgtmac: Actually proto::Stream_Kind is sufficient to provide whether it is optional or not.

luksan47: Okay, so if I understand correctly here's how it should look like:
1. The interface should not change (or would only change in replacing columnId with type in order to give better debug information).
2. The check should happen in getStream implementation.
3. If the stream kind is not PRESENT, we should throw an exception (maybe through a private/protected function to get standard error messages).

What do you think?

wgtmac: I agree on most of them. For 3), I think it is better to pass `bool optional` explicitly. The reason is that not only PRESENT stream can be optional but also SECONDARY stream of decimal type in ORC 2.0.

PR #484, opened 2020-02-21
ORC-519: Correct encoding for Decimal columns in specification

luksan47:
Currently the ORC specifications state that the secondary stream of
decimal columns are Unsigned RLE, but according to the code (both
Java and C++), Signed RLE is used.


luksan47:
> @luksan47 Can you please also add the fix to ORCv0.md? Thanks!

@wgtmac Done.

PR #493, opened 2020-03-12
optimization debian package manager tweaks

Rajpratik71:
By default, Ubuntu or Debian based "apt" or "apt-get" system installs recommended but not suggested packages .

By passing "--no-install-recommends" option, the user lets apt-get know not to consider recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/docker/debian10/Dockerfile
+++ b/docker/debian10/Dockerfile
@@ -21,7 +21,7 @@ FROM debian:buster
 MAINTAINER Owen O'Malley <omalley@apache.org>
 
 RUN apt-get update
-RUN apt-get install -y \
+RUN apt-get --no-install-recommends install -y \
   cmake \
   gcc \
   g++ \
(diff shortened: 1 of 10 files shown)


dongjoon-hyun:
Thank you for making a PR, @Rajpratik71 . Could you file a ORC JIRA?

omalley:
This does in fact make the images smaller, but the build fails with aclocal not being installed in debian8. You probably got a cached image in docker, so you'll need to run the reinit.sh script to clear the caches before you test.

dongjoon-hyun:
According to the above discussion, I close this PR for now.
Thank you, @Rajpratik71 and @omalley .

PR #495, opened 2020-03-15
Int128: Undefined behaviour in left-shift

igaztanaga:
Left-shifting into the sign bit is Undefined Behaviour and UBSAN reports it executing ORC tests. If shiftint into the sign it is desied, the workaround is doing the shift in unsigned types, and casting it to signed. Casting an unsigned type to signed when the value is not representable by the signed type is Implementation Defined, but nearly all implementations just maintain all bits.

With this patch, UBSAN errors from sign bit shifts are removed. But UBSAN still reports other errors like:
/Src/RLE.cc:80:22: runtime error: left shift of negative value -290
../Src/BloomFilter.cc:246:15: runtime error: signed integer overflow: -1628351410 + -814175404 cannot be represented in type 'int'

### DIFF
--- a/c++/include/orc/Int128.hh
+++ b/c++/include/orc/Int128.hh
@@ -196,11 +196,11 @@ namespace orc {
     Int128& operator<<=(uint32_t bits) {
       if (bits != 0) {
         if (bits < 64) {
-          highbits <<= bits;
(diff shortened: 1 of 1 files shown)


igaztanaga:
Opening a more complete pull request (#496) to fix all detected UB

PR #497, opened 2020-03-16
ORC-611: Min/max column statistics precision fix

pgaref:
Column statistics have ms precision, while timestamps come with ns precision,
We should round-down min and round-up max column statistics to make sure PPD with ns precision works

Change-Id: I22e89b102ffc3aa9750a30bbfce5c847993bf927

### DIFF
--- a/java/core/src/java/org/apache/orc/TimestampColumnStatistics.java
+++ b/java/core/src/java/org/apache/orc/TimestampColumnStatistics.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
--- a/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
@@ -1638,8 +1638,14 @@ public class ColumnStatisticsImpl implements ColumnStatistics {
 
   private static class TimestampStatisticsImpl extends ColumnStatisticsImpl
       implements TimestampColumnStatistics {
-    private Long minimum = null;
-    private Long maximum = null;
+    private static final int DEFAULT_MIN_NANOS = 000_000;
+    private static final int DEFAULT_MAX_NANOS = 999_999;
+
+    private long minimum = Long.MAX_VALUE;
+    private long maximum = Long.MIN_VALUE;
+
+    private int minNanos = DEFAULT_MIN_NANOS;
+    private int maxNanos = DEFAULT_MAX_NANOS;
 
     TimestampStatisticsImpl() {
     }
@@ -1670,31 +1676,51 @@ public class ColumnStatisticsImpl implements ColumnStatistics {
         minimum = DateUtils.convertTime(timestampStats.getMinimumUtc(),
             writerUsedProlepticGregorian, convertToProlepticGregorian, true);
       }
+      if (timestampStats.hasMaximumNanos()) {
+        maxNanos = timestampStats.getMaximumNanos() - 1;
+      }
+      if (timestampStats.hasMinimumNanos()) {
+        minNanos = timestampStats.getMinimumNanos() - 1;
+      }
     }
 
     @Override
     public void reset() {
       super.reset();
-      minimum = null;
-      maximum = null;
+      minimum = Long.MAX_VALUE;
+      maximum = Long.MIN_VALUE;
+      minNanos = DEFAULT_MIN_NANOS;
+      maxNanos = DEFAULT_MAX_NANOS;
     }
 
     @Override
     public void updateTimestamp(Timestamp value) {
       long millis = SerializationUtils.convertToUtc(TimeZone.getDefault(),
           value.getTime());
-      updateTimestamp(millis);
+      // prune the last 6 digits for ns precision
+      updateTimestamp(millis, value.getNanos() % 1_000_000);
     }
 
     @Override
-    public void updateTimestamp(long value) {
-      if (minimum == null) {
+    public void updateTimestamp(long value, int nanos) {
+      if (minimum > maximum) {
         minimum = value;
         maximum = value;
-      } else if (minimum > value) {
-        minimum = value;
-      } else if (maximum < value) {
-        maximum = value;
+        minNanos = nanos;
+        maxNanos = nanos;
+      } else {
+        if (minimum >= value) {
+          if (minimum > value || nanos < minNanos) {
+            minNanos = nanos;
+          }
+          minimum = value;
+        }
+        if (maximum <= value) {
+          if (maximum < value || nanos > maxNanos) {
+            maxNanos = nanos;
+          }
+          maximum = value;
+        }
       }
     }
 
@@ -1702,19 +1728,31 @@ public class ColumnStatisticsImpl implements ColumnStatistics {
     public void merge(ColumnStatisticsImpl other) {
       if (other instanceof TimestampStatisticsImpl) {
         TimestampStatisticsImpl timestampStats = (TimestampStatisticsImpl) other;
-        if (minimum == null) {
-          minimum = timestampStats.minimum;
-          maximum = timestampStats.maximum;
-        } else if (timestampStats.minimum != null) {
-          if (minimum > timestampStats.minimum) {
+        if (count == 0) {
+          if (timestampStats.count != 0) {
+            minimum = timestampStats.minimum;
+            maximum = timestampStats.maximum;
+            minNanos = timestampStats.minNanos;
+            maxNanos = timestampStats.maxNanos;
+          }
+        } else if (timestampStats.count != 0) {
+          if (minimum >= timestampStats.minimum) {
+            if (minimum > timestampStats.minimum ||
+                minNanos > timestampStats.minNanos) {
+              minNanos = timestampStats.minNanos;
+            }
             minimum = timestampStats.minimum;
           }
-          if (maximum < timestampStats.maximum) {
+          if (maximum <= timestampStats.maximum) {
+            if (maximum < timestampStats.maximum ||
+                maxNanos < timestampStats.maxNanos) {
+              maxNanos = timestampStats.maxNanos;
+            }
             maximum = timestampStats.maximum;
           }
         }
       } else {
-        if (isStatsExists() && minimum != null) {
+        if (isStatsExists() && count != 0) {
           throw new IllegalArgumentException("Incompatible merging of timestamp column statistics");
         }
       }
@@ -1726,9 +1764,15 @@ public class ColumnStatisticsImpl implements ColumnStatistics {
       OrcProto.ColumnStatistics.Builder result = super.serialize();
       OrcProto.TimestampStatistics.Builder timestampStats = OrcProto.TimestampStatistics
           .newBuilder();
-      if (getNumberOfValues() != 0 && minimum != null) {
+      if (getNumberOfValues() != 0) {
         timestampStats.setMinimumUtc(minimum);
         timestampStats.setMaximumUtc(maximum);
+        if (minNanos != DEFAULT_MIN_NANOS) {
+          timestampStats.setMinimumNanos(minNanos + 1);
+        }
+        if (maxNanos != DEFAULT_MAX_NANOS) {
+          timestampStats.setMaximumNanos(maxNanos + 1);
+        }
       }
       result.setTimestampStatistics(timestampStats);
       return result;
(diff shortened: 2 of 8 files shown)


pgaref:
> Thanks for the quick fix!
> I didn't read through the tests carefully yet, I will do that later.

Hey @csringhofer -- thanks a lot for the comments!
I put some more thought into the patch and it seems that we can sort out the issue on the Reader side avoiding changing the Column stats that are written and having to support another writer version.

In the latest patch, whenever we are evaluating a predicate with Timestamps, I just widen the min/max range by 1 ms to make sure the underlying ns range is included -- this should solve the issue both on Java and C++ consumers using our Readers  (until we feel comfortable changing the column statistics on the writer side and remove this change on the reader).

Thoughts?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -405,7 +405,12 @@ static ValueRange getValueRange(ColumnStatistics index,
       TimestampColumnStatistics stats = (TimestampColumnStatistics) index;
       Timestamp min = useUTCTimestamp ? stats.getMinimumUTC() : stats.getMinimum();
       Timestamp max = useUTCTimestamp ? stats.getMaximumUTC() : stats.getMaximum();
-      return new ValueRange<>(predicate, min, max, stats.hasNull());
+      // To support predicates with ns precision we widen min and max ms range by 1 to ensure ns values are in range.
```
omalley: This is the wrong pattern, unfortunately. You'll need to follow the example of what was done for the string shortening. (We just store the first 1024 characters for min/max on strings.) The problem comes that if someone does a test that looks for equality, this change will return wrong results.

java/core/src/test/org/apache/orc/TestOrcTimestampPPD.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+    rows.close();
+    ColumnStatistics[] colStats = reader.getStatistics();
+
+    // Make sure PPD is now passing
+    Assert.assertEquals(SearchArgument.TruthValue.YES_NO, RecordReaderImpl.evaluatePredicate(colStats[0],
+        SearchArgumentFactory.newBuilder()
+            .equals("c", PredicateLeaf.Type.TIMESTAMP, Timestamp.valueOf("2037-01-01 00:00:00.001009")).build()
+            .getLeaves().get(0), null));
+  }
```
omalley: Please add a test with a timestamp equality and both approximate and precise min/max.

c++/src/sargs/PredicateLeaf.cc
```diff
@@ -536,11 +536,17 @@ namespace orc {
...
+          // To support predicates with ns precision we widen
+          // min and max ms range by 1 to ensure ns values are in range.
+          // e.g., for 1970-01-01 00:00:00.0010005 min
+          // becomes: 1970-01-01 00:00:00.000 and max: 1970-01-01 00:00:00.002
           result = evaluatePredicateRange(
             mOperator,
             literal2Timestamp(mLiterals),
-            stats.minimumutc(),
-            stats.maximumutc(),
+            minimum-1,
```
csringhofer: Decreasing the minimum is probably needed, as the writers round towards negative infinity.

java/core/src/test/org/apache/orc/TestOrcTimezonePPD.java
```diff
@@ -294,7 +294,7 @@ public void testTimestampMinMaxAndBloomFilter() throws Exception {
         .bufferSize(10000).bloomFilterColumns("ts"));
     assertEquals(writerTimeZone, TimeZone.getDefault().getID());
     List<String> ts = Lists.newArrayList();
-    ts.add("2007-08-01 00:00:00.0");
+    ts.add("2007-08-01 01:00:00.0");
```
csringhofer: What is the reason behind this 1 hour change? It doesn't seem relevant to sub-millisec issues.

java/core/src/test/org/apache/orc/TestOrcTimestampPPD.java
```diff
@@ -0,0 +1,173 @@
...
+        Assert.assertEquals(tslist.get(0), times.asScratchTimestamp(r));
+        Assert.assertEquals(tslist.get(0).getNanos(), times.asScratchTimestamp(r).getNanos());
+      }
+    }
+    rows.close();
+    ColumnStatistics[] colStats = reader.getStatistics();
+    Timestamp gotMin = ((TimestampColumnStatistics) colStats[0]).getMinimum();
+    assertEquals("1970-01-01 00:00:00.0", gotMin.toString());
+
+    Timestamp gotMax = ((TimestampColumnStatistics) colStats[0]).getMaximum();
```
csringhofer: A comment could explain that this is a bit buggy in the writer, but the reader can handle it.

java/core/src/test/org/apache/orc/TestOrcTimestampPPD.java
```diff
@@ -0,0 +1,173 @@
...
+    // Now reading
+    Reader reader = OrcFile.createReader(testFilePath, OrcFile.readerOptions(conf).filesystem(fs));
+
+    RecordReader rows = reader.rows();
+    batch = reader.getSchema().createRowBatch();
+    times = (TimestampColumnVector) batch.cols[0];
+    while (rows.nextBatch(batch)) {
+      for (int r = 0; r < batch.size; ++r) {
+        Assert.assertEquals(tslist.get(0), times.asScratchTimestamp(r));
+        Assert.assertEquals(tslist.get(0).getNanos(), times.asScratchTimestamp(r).getNanos());
```
csringhofer: I don't get this - why are the nanos always equal with the first element in 'tslist'? Is it always 0?

pgaref: The reason is all **tslist** Timestamps have the same value up to the supported MS precision -- just made this assumption for the test

java/core/src/test/org/apache/orc/TestOrcTimestampPPD.java
pgaref: A proper solution, in general, would include lower and upper bounds similar to what we do for string shortening (https://issues.apache.org/jira/browse/ORC-203) -- see Owens reply above.

To achieve that we cannot really avoid bumping the writer version so I will update the patch accordingly

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pgaref: Thanks for the comments @omalley ! Latest patch updates TimestampStatistics to store the lower 6 digits (that were initially ignored) in order to achieve nanosecond precision.

Also liked the idea proposed by @csringhofer and added 0 and 999_999 as DEFAULT_NANOS MIN and MAX respectively when these values are missing.

Please let me know what you think!

java/core/src/test/org/apache/orc/TestOrcTimestampPPD.java
pgaref: @csringhofer please let me know what you think about the latest changes!

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -1815,7 +1893,7 @@ public int hashCode() {
 
     @Override
     public void updateTimestamp(Timestamp value) {
-      updateTimestamp(value.getTime());
+      updateTimestamp(value.getTime(), value.getNanos());
```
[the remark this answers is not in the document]
pgaref: correct, this is now fixed

java/core/src/java/org/apache/orc/impl/writer/TimestampTreeWriter.java
```diff
@@ -111,7 +111,7 @@ public void writeBatch(ColumnVector vector, int offset,
         }
```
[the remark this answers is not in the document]
pgaref: Makes sense to me, thats why in getMinimum and getMaximum methods in TimestampStatistics I set the nanos after the utc conversion

java/core/src/test/org/apache/orc/TestOrcTimestampPPD.java
```diff
@@ -0,0 +1,208 @@
...
+    assertEquals("2037-01-01 00:00:00.0067891", gotMax.toString());
+
+    // PPD Equals with nano precision passing
+    PredicateLeaf pred = createPredicateLeaf(PredicateLeaf.Operator.EQUALS, PredicateLeaf.Type.TIMESTAMP, "c",
+        Timestamp.valueOf("2037-01-01 00:00:00.001109"), null);
+    Assert.assertEquals(SearchArgument.TruthValue.YES_NO, RecordReaderImpl.evaluatePredicate(colStats[0], pred, null));
+
+    // PPD Equals with ms precision NOT passing
+    pred = createPredicateLeaf(PredicateLeaf.Operator.EQUALS, PredicateLeaf.Type.TIMESTAMP, "c",
+        Timestamp.valueOf("2037-01-01 00:00:00"), null);
```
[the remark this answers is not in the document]
pgaref: Thanks, fixed

java/core/src/test/org/apache/orc/TestOrcTimestampPPD.java
```diff
@@ -0,0 +1,208 @@
...
+    Assert.assertEquals(SearchArgument.TruthValue.NO, RecordReaderImpl.evaluatePredicate(colStats[0], pred, null));
+
+    // PPD Equals with ns precision passing
```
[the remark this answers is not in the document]
pgaref: fixed

java/core/src/test/org/apache/orc/TestOrcTimestampPPD.java
```diff
@@ -0,0 +1,208 @@
...
+    pred = createPredicateLeaf(PredicateLeaf.Operator.LESS_THAN, PredicateLeaf.Type.TIMESTAMP, "c",
+        Timestamp.valueOf("2037-01-01 00:00:00.006789"), null);
+    Assert.assertEquals(SearchArgument.TruthValue.YES_NO, RecordReaderImpl.evaluatePredicate(colStats[0], pred, null));
+
+    // PPD Equals with ms precision passing
```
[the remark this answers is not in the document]
pgaref: fixed

java/core/src/java/org/apache/orc/TimestampColumnStatistics.java
```diff
@@ -47,4 +47,19 @@
    * @return maximum value in UTC
    */
   Timestamp getMaximumUTC();
+
+  /**
+   * Get the lowest digits of the minimum value for the column.
+   * The minimum value is truncated at ms scale.
+   * @return the 6 lowest digits of minimum value
+   */
+  int getMinimumNanos();
```
wgtmac: java.sql.Timestamp does support nanosecond. Is it better to use existing APIs and handle nanosecond transparently?

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -1670,46 +1675,88 @@ public int hashCode() {
...
       long millis = SerializationUtils.convertToUtc(TimeZone.getDefault(),
           value.getTime());
-      updateTimestamp(millis);
+      // prune the last 6 digits for ns precision
+      updateTimestamp(millis, value.getNanos() % 1_000_000);
     }
 
     @Override
-    public void updateTimestamp(long value) {
+    public void updateTimestamp(long value, int nanos) {
```
wgtmac: Just curious, can we directly use java.sql.Timestamp to store minimum & maximum in UTC? If that is true, we can deprecate updateTimestamp and make the comparison and merge process easier.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -1670,46 +1675,88 @@ public int hashCode() {
...
-          if (minimum > timestampStats.minimum) {
+        if (count == 0) {
+          if (timestampStats.count != 0) {
+            minimum = timestampStats.minimum;
+            maximum = timestampStats.maximum;
+            minNanos = timestampStats.minNanos;
+            maxNanos = timestampStats.maxNanos;
+          } else {
+            /* both are empty */
+            maximum = minimum = null;
```
wgtmac: the else branch can be deleted

pgaref: Makes sense, thanks

java/core/src/java/org/apache/orc/TimestampColumnStatistics.java
pgaref: We could do that, however I believe its more clear to handle the nanoseconds as separate methods to show how ns precision is handled in old Proto (default values) and new Proto with existing values.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
pgaref: That would be useful, havent checked but I believe this could be part of a followup ticket.

omalley:
Sorry for the long delay on this one. I made some changes that should be looked over. I pushed my changes to my [orc-611 branch](https://github.com/omalley/orc/tree/orc-611). The main one was changing the protobuf to used unsigned ints for min & max nanos. To ensure that we don't lose the ability to detect the old files (when we move to protobuf 3), I added one to the values so that the range of both fields is 1 to 1_000_000. (Protobuf 3 doesn't serialize the default value for fields, so a 0 would not be serialized.)

omalley:
@pgaref Can you take a look at my change to see if it looks reasonable?

pgaref:
> @pgaref Can you take a look at my change to see if it looks reasonable?

Hey @omalley thanks for the comment -- happy to take a look just give me a couple of days as I am currently OOO

omalley:
@pgaref when you get a chance, it would be great.

pgaref:
Hey @omalley just did a bit of polishing and squashed everything.
Thanks for the protobuf suggestion!  It does make sense to me and it also simplified the min/max calculation as a side effect! Latest patch LGTM

PR #498, opened 2020-03-17
ORC-613: Fix OrcMapredRecordReader when dealing with union of multiple structs with different schema

autumnust:
The bug is described in ORC-613 but briefly, in `OrcMapredRecordReader.java#nextStruct`, the determination of whether a previous object can be reused is buggy: It only checks top-level type (`OrcStruct`) but not the child schema under `OrcStruct`. So if dealing with following schema: 
`uniontype <struct<field0, field1>, struct<>>`

Let's say we have sequence of data like following : 
record0: {0: <field0, field1>}
record1: {1:<>}

Record1 will reuse the previous object's schema and results in incorrect output. 

Added a unit test that can reproduce the scenario.


autumnust:
@omalley  Can you help take a look ? Thanks.

PR #499, opened 2020-03-25
ORC-614: Implement efficient seek() in decompression streams

luksan47:
The current implementation of ZlibDecompressionStream::seek and
BlockDecompressionStream::seek resets the state of the decompressor
and the underlying file reader and throws away their buffers.

This commit introduces two optimizations which rely on reusing
the buffers that still contain useful data, and therefore reducing
the time spent reading/uncompressing the buffers again.

The first case is when the seeked position is already read
and decompressed into the output stream.

The second case is when the seeked position is already read from
the input stream, but has not been decompressed yet, ie. it's
not in the output stream.

Tests:
 - Run the ORC tests, and the Impala tests working on ORC tables.
 - The regression that #476 would cause is not present anymore.

### DIFF
--- a/c++/src/Compression.cc
+++ b/c++/src/Compression.cc
@@ -338,7 +338,8 @@ DIAGNOSTIC_PUSH
         inputBufferEnd = nullptr;
       } else {
         inputBufferEnd = inputBuffer + length;
-        inputBufferStartPosition = input->ByteCount() - length;
+        inputBufferStartPosition
+            = static_cast<size_t>(input->ByteCount() - length);
         inputBufferStart = inputBuffer;
       }
     }
@@ -379,7 +380,7 @@ DIAGNOSTIC_PUSH
     DecompressState state;
 
     // The starting and current position of the buffer for the uncompressed
-    // data.
+    // data. It either points to the data buffer or the underlying input stream.
     const char* outputBufferStart;
     const char* outputBuffer;
     // The original (ie. the overall) and the actual length of the uncompressed
@@ -477,8 +478,9 @@ DIAGNOSTIC_POP
     }
     if (state == DECOMPRESS_HEADER || remainingLength == 0) {
       readHeader();
-      headerPosition
-          = inputBufferStartPosition + (inputBuffer - inputBufferStart);
+      // Here we already read the three bytes of the header.
+      headerPosition = inputBufferStartPosition
+          + static_cast<size_t>(inputBuffer - inputBufferStart) - 3;
       saveBufferPositions = true;
     }
     if (state == DECOMPRESS_EOF) {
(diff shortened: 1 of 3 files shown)


luksan47:
FYI, I also opened since there is a lot of duplicate code: https://issues.apache.org/jira/browse/ORC-615

luksan47:
Sent in #500, if you could review it first, that would reduce the amount of code change (and possible mistakes) here as well.

wgtmac:
@luksan47  I am working on it. Will finish reviewing this PR by the end of this week.

luksan47:
Thanks, @wgtmac!

c++/src/Compression.cc
```diff
@@ -564,16 +591,47 @@ DIAGNOSTIC_POP
...
     remainingLength = 0;
-    inputBuffer = nullptr;
-    inputBufferEnd = nullptr;
+    if (seekedPosition < static_cast<uint64_t>(input->ByteCount())) {
+      // Case 2.: The input is buffered, but not yet decompressed. No need to
+      // force re-reading the inputBuffer, we just have to move it to the
+      // seeked position.
+      position.next(); // Skip the input level position.
+      inputBuffer
+          = inputBufferStart + (seekedPosition - inputBufferStartPosition);
```
wgtmac: If seekedPosition is smaller than inputBufferStartPosition, the inputBufferStart does not contain the data we need. This can happen when users call RowReader::seekToRow() in a reverse order.

c++/src/Compression.cc
```diff
@@ -457,6 +477,9 @@ DIAGNOSTIC_POP
     }
     if (state == DECOMPRESS_HEADER || remainingLength == 0) {
       readHeader();
+      headerPosition
```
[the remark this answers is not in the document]
luksan47: Kept the name since we look up this position, changed the value instead.

c++/src/Compression.cc
luksan47: Added condition.

wgtmac:
@luksan47 Please fix the travis build. It would be more confident if you can test cases for all 3 scenarios under seek.

pgaref:
Was reviewing https://github.com/apache/orc/pull/476 and bump into this -- shall we revive this? @luksan47  @wgtmac

wgtmac:
> Was reviewing #476 and bump into this -- shall we revive this? @luksan47 @wgtmac

Thanks for pinging me. I will try to do that.

wgtmac:
Hey @luksan47, are you still interested in this patch?

luksan47:
Hey @wgtmac, sorry but I don't have the capacity at the moment to work on this PR :(
You can use/copy my commits if you'd like to merge this.

wgtmac:
Opened a new PR https://github.com/apache/orc/pull/695 and thus closed this one. Thanks @luksan47

PR #500, opened 2020-03-26
ORC-615: Refactor decompression streams into common base class

luksan47:
This commit does not intend to change anything in the current
behaviour, its only purpose is to reduce the huge amount of code
duplication in Compression.cc by refactoring the common functionality
of ZlibDecompressionStream and BlockDecompressionStream under a
common base class.

Tests:
 - Run orc-tests successfully.

### DIFF
--- a/c++/src/Compression.cc
+++ b/c++/src/Compression.cc
@@ -721,12 +721,12 @@ DIAGNOSTIC_POP
                                 ) override;
   };
 
-  uint64_t SnappyDecompressionStream::decompress(const char *input,
+  uint64_t SnappyDecompressionStream::decompress(const char *_input,
                                                  uint64_t length,
                                                  char *output,
                                                  size_t maxOutputLength) {
     size_t outLength;
-    if (!snappy::GetUncompressedLength(input, length, &outLength)) {
+    if (!snappy::GetUncompressedLength(_input, length, &outLength)) {
       throw ParseError("SnappyDecompressionStream choked on corrupt input");
     }
 
@@ -734,7 +734,7 @@ DIAGNOSTIC_POP
       throw std::logic_error("Snappy length exceeds block size");
     }
 
-    if (!snappy::RawUncompress(input, length, output)) {
+    if (!snappy::RawUncompress(_input, length, output)) {
       throw ParseError("SnappyDecompressionStream choked on corrupt input");
     }
     return outLength;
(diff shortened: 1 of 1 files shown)


c++/src/Compression.cc
```diff
@@ -576,244 +499,201 @@ DIAGNOSTIC_POP
...
     std::ostringstream result;
     result << "zlib(" << input->getName() << ")";
     return result.str();
   }
 
-  class BlockDecompressionStream: public SeekableInputStream {
+  class BlockDecompressionStream: public DecompressionStream {
   public:
     BlockDecompressionStream(std::unique_ptr<SeekableInputStream> inStream,
                              size_t blockSize,
```
csringhofer: This argument is called blockSize here, but bufferSize in the definition

c++/src/Compression.cc
```diff
@@ -576,244 +499,201 @@ DIAGNOSTIC_POP
...
+    virtual std::string getName() const override;
+
+  protected:
+    virtual void NextDecompress(const void** data,
+                                int* size,
+                                size_t availableSize) override;
+  private:
+    const size_t blockSize;
+    z_stream zstream;
+    DataBuffer<char> buffer;
```
csringhofer: This could be also merged with outputDataBuffer.

c++/src/Compression.cc
```diff
@@ -576,244 +499,201 @@ DIAGNOSTIC_POP
...
+                            MemoryPool& pool);
+    virtual ~ZlibDecompressionStream() override;
+    virtual std::string getName() const override;
+
+  protected:
+    virtual void NextDecompress(const void** data,
+                                int* size,
+                                size_t availableSize) override;
+  private:
+    const size_t blockSize;
```
csringhofer: This could be removed by using outputDataBuffer.capacity() instead as in BlockDecompressionStream,

c++/src/Compression.cc
luksan47: Renamed to blockSize, I think it's more meaningful for BlockDecompressionStream.

c++/src/Compression.cc
luksan47: Done.

c++/src/Compression.cc
luksan47: Done.

PR #501, opened 2020-03-30
ORC-616: In Patched Base encoding, the value of headerThirdByte goes beyond the range of byte

ZorTsou:


java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
```diff
@@ -529,7 +529,7 @@ private void determineEncoding() {
       // fallback to DIRECT encoding.
       // The decision to use patched base was based on zigzag values, but the
       // actual patching is done on base reduced literals.
-      if ((brBits100p - brBits95p) != 0) {
+      if ((brBits100p - brBits95p) != 0 && Math.abs(min) < 1l << 56) {
```
prasanthj, line 532: can you make `(1L << 56)` as static final constant?

prasanthj: Can you also include unit test case that was mentioned in the description of the ticket?

ZorTsou:
> can you make `(1L << 56)` as static final constant?
> Can you also include unit test case that was mentioned in the description of the ticket?

@prasanthj Done

PR #503, opened 2020-04-22
ORC-622: Added BatchReader and TypeReader

pavibhai:
This is fix for [ORC-622](https://issues.apache.org/jira/browse/ORC-622)

Fixed the TreeReaderFactory and ConvertTreeReaderFactory
Made RecordReaderImpl use RootReader instead of TreeReader
removed nextBatch method from TreeReader
Rename RootReader -> BatchReader, rowReader -> rootType
Moved the batch commit logic into the BatchReader
StreamInformation public
Added getColumnId into TypeReader
commitBatch has been renamed to resetBatch in BatchReader

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
@@ -249,13 +249,13 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


omalley:
Ok, I started [tweaking](https://github.com/omalley/orc/commit/e73bdbdb00ca43f679fc994e7f6988d7b6a45fec) this.

I hope that is ok, Pavan.

I'd propose:
* making BatchReader an abstract class.
* making a new PrimitiveBatchReader that handles the special case of a primitive root type
* removing the type parameter on BatchReader

Thoughts?

PR #506, opened 2020-05-06
ORC-628. Add tool to count number of rows in ORC files under a directory

omalley:


java/tools/src/java/org/apache/orc/tools/Driver.java
```diff
@@ -116,6 +116,8 @@ public static void main(String[] args) throws Exception {
       ConvertTool.main(conf, options.commandArgs);
     } else if ("key".equals(options.command)) {
       KeyTool.main(conf, options.commandArgs);
+    } else if ("count".equals(options.command)) {
```
dongjoon-hyun, line 119: Could you add a description for `count` command after line 95?
- https://github.com/apache/orc/pull/506/files#diff-e0682227343a18a7d50605f939f02795R95

dongjoon-hyun:
I found that `orc-tools` uber jar has been broken due to ORC-599 and make a PR to fix it. Could you review that, please?
- https://github.com/apache/orc/pull/507

omalley:
Sorry about that! I wrote and tested the patch on branch 1.5, so it worked. Thanks for fixing master.

dongjoon-hyun:
Thank you, @omalley .

PR #508, opened 2020-05-12
ORC-629: PPD: Floating point NaN is not transitive across comparisons

pgaref:


java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -495,6 +495,14 @@ static TruthValue evaluatePredicateProto(OrcProto.ColumnStatistics statsProto,
                    " include ORC-517. Writer version: {}",
           predicate.getColumnName(), writerVersion);
       return TruthValue.YES_NO_NULL;
+    } else if (category == TypeDescription.Category.DOUBLE) {
+      DoubleColumnStatistics dstas = (DoubleColumnStatistics) cs;
+      if (!Double.isFinite(dstas.getMinimum()) || !Double.isFinite(dstas.getMaximum())
```
omalley, line 500: This check is overly broad. Positive and Negative infinity work fine with min and max. You should just check for NaN in the sum.

wgtmac: I found the [fix for C++](https://github.com/apache/orc/commit/f6554377828b66d1198fe720ae8b9a734255a2df
) has been merged already. Need to ensure they are consistent.

pgaref: Hey @wgtmac -- the logic is replicated across Cpp and Java versions.

Regarding the broad check, the way we compare Double Stats (without Double.compare) can also lead to wrong min/max values (see test case) -- this is also something we could improve. https://github.com/apache/orc/blob/ca33ce64bf1fa8b3696e2e44b32237d842c70df3/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java#L563

However, we could probably replace the infinity check with a Nan check.

wgtmac: Ok, I found the fix for java is merged too: https://github.com/apache/orc/commit/5dd3b29a2f4a97aa716b301c657c1853cc27e6c2. Would like to close this PR to make it less confusing.

PR #514, opened 2020-06-10
ORC-659: Initialize "next_in" before calling DeflateInit2

igaztanaga:

This PR aims to initialize `next_in` before calling DeflateInit2.

According to zlib documentation of DeflateInit2:

_"This is another version of deflateInit with more compression options. The fields **next_in**, zalloc, zfree and opaque must be initialized before by the caller."_

In an alpine docker DeflateInit2 was returning an error that seems to be solved with the patch.

Pass UT.

### DIFF
--- a/c++/src/Compression.cc
+++ b/c++/src/Compression.cc
@@ -292,6 +292,7 @@ DIAGNOSTIC_PUSH
     strm.zalloc = nullptr;
     strm.zfree = nullptr;
     strm.opaque = nullptr;
+    strm.next_in = nullptr;
 
     if (deflateInit2(&strm, level, Z_DEFLATED, -15, 8, Z_DEFAULT_STRATEGY)
         != Z_OK) {


dongjoon-hyun:
Hi, @igaztanaga .
Could you describe how to reproduce the error?
> In an alpine docker DeflateInit2 was returning an error that seems to be solved with the patch.

dongjoon-hyun:
Gentle ping, @igaztanaga ~

dongjoon-hyun:
+1, LGTM. Thank you, @igaztanaga .

dongjoon-hyun:
Welcome, @igaztanaga . I added to the Apache ORC contributor group and assigned ORC-659 to you.

igaztanaga:
Sorry, somehow I overlooked your notification 3 days ago. Sometimes the init function returned error before the change, but I'm not sure if that was the only reason. Since the documentation explicitly stated that next_in should be initialized before calling DeflateInit2, we think the error was, in part, caused by a uninitialized value bug. Thanks for merging.

dongjoon-hyun:
Thank you for replying, @igaztanaga . I merged it because I agree with you.

dongjoon-hyun:
This is now merged to `branch-1.6` and `branch-1.5` for next 1.6.4 and 1.5.11 release.

PR #516, opened 2020-06-17
Change logging level of creating new codec in OrcCodecPool

bmaidics:
When using Hive LLAP in a cloud environment, we observed a bunch of (>100000) non-useful logs: 

`impl.OrcCodecPool: Got brand-new codec ZLIB`

I think the log level here can be reduced to DEBUG.


prasanthj:
@bmaidics is there Jira for this ticket under ORC project?

PR #537, opened 2020-08-11
ORC-639: Improve zstd performance

igaztanaga:

This PR aims to improve ZSTD performance in branch-1.6. This is a backport of #511

Zstd's "Simple API" is used which initializes the compression/decompression context each time. This includes time to initialize all compression/decompression tables. Using the "Explicit context" API improves the performance significantly as compression/decompression tables are constructed once and reused each time "decompress" or "doBlockCompression" is called. We've noticed more than 15% time improvement in some applications with this change, so it seems that any ORC user using Zstd can benefit from this change.

Pass the UT

### DIFF
--- a/c++/src/Compression.cc
+++ b/c++/src/Compression.cc
@@ -1031,11 +1031,15 @@ DIAGNOSTIC_POP
                                                    capacity,
                                                    blockSize,
                                                    pool) {
-      // PASS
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Thank you, @igaztanaga !

PR #538, opened 2020-08-13
Fix Julian/Gregorian cutoff in a comment

findepi:


proto/orc_proto.proto
```diff
@@ -343,7 +343,7 @@ message Encryption {
 enum CalendarKind {
   UNKNOWN_CALENDAR = 0;
    // The Java default calendar changes from Julian to Gregorian
-   // in 1583.
+   // in October 1582.
```
dongjoon-hyun, line 346: Hi, @findepi . Is this the last one? Do we have more obsolete comments?

findepi: this is the only occurrence of "1583" i could find in the codebase

dongjoon-hyun: I'm wondering if the whole sentence makes sense. `The Java default calendar changes from Julian to Gregorian in October 1582`? It sounds like `Java` changes it in `October 1581`, doesn't it? Technically, 1582 is just a year when Gregorian was introduced.

Shall we remove it (the whole sentence) simply?

findepi: Quoting https://en.wikipedia.org/wiki/Gregorian_calendar:

_Thursday 4 October 1582 was followed by Friday 15 October 1582_

>  It sounds like Java changes it in October 1581, doesn't it? 

I think Java's java.util.Calendar and java.util.Date date calculations follow the Oct 4 → Oct 15 shift, do they not?

dongjoon-hyun: @findepi 
We had better be specific to Java version because Apache Orc still supports Java 7. I'm not aware of any change in Java 7. Do you have any reference saying Java 7 changes it's default calendar or time-related stuff? IIRC, everything starts at Java 8.
- https://github.com/apache/orc/blame/master/README.md#L47

findepi:
cc @omalley

proto/orc_proto.proto
omalley: Java hasn't changed its behavior, we've just become more aware of it.

Does making the text "Written using the hybrid Julian/Gregorian calendar with a cutover point in October 1582" make the intent clear?

dongjoon-hyun:
Thank you, @omalley . +1 for @omalley 's suggestion.

PR #539, opened 2020-08-13
Change calendar field to unused index

dain:
The calendar field is currently assigned to field 11, but this was previously used for stripeStatisticsLength and then removed with https://github.com/apache/orc/commit/548050271e5869bc6a5b4c8b735927d6d0721062#diff-f1e26f33ad210dfde6ab7123c7feb835L364

### DIFF
--- a/proto/orc_proto.proto
+++ b/proto/orc_proto.proto
@@ -368,7 +368,7 @@ message Footer {
 
   // information about the encryption in this file
   optional Encryption encryption = 10;
-  optional CalendarKind calendar = 11;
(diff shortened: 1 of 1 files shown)


omalley:
Apologies for reusing the field number, but the old usage was never released while the new one has been. To change the field number now would cause compatibility problems.

Did Presto produce files with the old attribute?

dain:
No it didn't effect me.  I though you folks might not have noticed this (and if so we could fix it before a release went out).  Good to hear that this wasn't released.  Thanks!

PR #540, opened 2020-08-22
ORC-661. Use LocalDate and day of epoch instead of java's Date for column statistics

omalley:
This PR adds 4 new methods to the DateColumnStatistics.

* getMinimumLocalDate -> LocalDate
* getMinimumDayOfEpoch -> int

and the matching getMaximum functions. The current functions are deprecated. The DateColumnStatisticsImpl moves the minimum and maximum values to int instead of Integer, so they can be modified. The initial values are Integer.MAX_VALUE and MIN_VALUE.

The PPD code is changed to use LocalDate instead of Date.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -17,6 +17,13 @@
  */
 package org.apache.orc.impl;
 
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.chrono.ChronoLocalDate;
+import java.time.format.DateTimeFormatter;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.common.type.HiveDecimal;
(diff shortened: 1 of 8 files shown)


omalley:
Ugh, that is unfortunate. Thank you for the catch, @dongjoon-hyun !

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -801,12 +805,17 @@ private static Comparable getBaseObjectForComparison(PredicateLeaf.Type type,
...
         }
       case DATE:
-        if (obj instanceof Date) {
+        if (obj instanceof LocalDate) {
           return obj;
+        } else if (obj instanceof java.sql.Date) {
+          return ((java.sql.Date) obj).toLocalDate();
+        } else if (obj instanceof Date) {
+          return LocalDateTime.ofInstant(((Date) obj).toInstant(),
+              ZoneOffset.UTC).toLocalDate();
```
dongjoon-hyun, line 814: So, this is the fix for the following?
```
ORC PPD evaluation for Date type uses java.sql.Date for min/max comparison causing incorrect results.
Date.compareTo uses millis offset which can return incorrect results depending on the timezone. 
Running the testcase in HIVE-19726 passed in Los Angeles but failed in Paris
as Date.compareTo return 0 for Los Angeles but returned -1 for Paris.
```

omalley:
@dongjoon-hyun let me write some more unit tests to make sure things are working correctly, but yes this should fix the timezone sensitive Date problem.

dongjoon-hyun:
Sure~ Thank you, @omalley .

omalley:
Ok, I made a few changes as well as updating the tests.
1. I moved from LocalDate to ChronoLocalDate, which is agnostic to the Chronology (eg. Hybrid vs Proleptic).
2. The LocalDate or HybridDate returned from get(Minimum/Maximum)LocalDate depend on the calendar that is selected.
3. We handle the conversion from ChronoLocalDate to string explicitly so that don't get the chronology.

We also need an update to storage-api to allow ChronoLocalDate in SARG predicates.

PR #542, opened 2020-08-31
ORC-623 Fix evaluation of SArgs over all null columns for ints and doubles.

omalley:
This fixes evaluation of predicate push down filters for columns of ints and doubles with all null values.

### DIFF
--- a/java/core/src/test/org/apache/orc/TestVectorOrcFile.java
+++ b/java/core/src/test/org/apache/orc/TestVectorOrcFile.java
@@ -4062,6 +4062,137 @@ public class TestVectorOrcFile {
     assertEquals(0, batch.size);
   }
 
+  /**
+   * Test predicate pushdown on nulls, with different combinations of
+   * values and nulls.
+   */
+  @Test
+  public void testPredicatePushdownAllNulls() throws Exception {
+    TypeDescription schema = createInnerSchema();
+    try (Writer writer = OrcFile.createWriter(testFilePath,
+        OrcFile.writerOptions(conf).setSchema(schema).rowIndexStride(1024).version(fileFormat))) {
+      VectorizedRowBatch batch = schema.createRowBatch();
+      batch.size = 1024;
+
+      // write 1024 rows of (null, "val")
+      batch.cols[0].noNulls = false;
+      batch.cols[0].isNull[0] = true;
+      batch.cols[0].isRepeating = true;
+      batch.cols[1].isRepeating = true;
+      ((BytesColumnVector) batch.cols[1]).setVal(0, "val".getBytes(StandardCharsets.UTF_8));
+      writer.addRowBatch(batch);
+
+      // write 1024 rows of (1, null)
+      batch.cols[0].isNull[0] = false;
+      ((LongColumnVector) batch.cols[0]).vector[0] = 123;
+      batch.cols[1].noNulls = false;
+      batch.cols[1].isNull[0] = true;
+      writer.addRowBatch(batch);
+
+      // write 1024 rows of (null, null)
+      batch.cols[0].isNull[0] = true;
+      writer.addRowBatch(batch);
+    }
+
+    try (Reader reader = OrcFile.createReader(testFilePath,
+        OrcFile.readerOptions(conf).filesystem(fs))) {
+      assertEquals(3072, reader.getNumberOfRows());
+      VectorizedRowBatch batch = reader.getSchema().createRowBatch();
+
+      // int1 is not null
+      SearchArgument sarg =
+          SearchArgumentFactory.newBuilder()
+              .startNot()
+              .isNull("int1", PredicateLeaf.Type.LONG)
+              .end()
+              .build();
+      // should find one row group
+      try (RecordReader rows = reader.rows(reader.options().searchArgument(sarg, new String[]{}))) {
+        rows.nextBatch(batch);
+        assertEquals(1024, batch.size);
+        assertEquals(true, batch.cols[0].isRepeating);
+        assertEquals(123, ((LongColumnVector) batch.cols[0]).vector[0]);
+        assertEquals(false, rows.nextBatch(batch));
+      }
+
+      // string1 is not null
+      sarg = SearchArgumentFactory.newBuilder()
+          .startNot()
+          .isNull("string1", PredicateLeaf.Type.STRING)
+          .end()
+          .build();
+      // should find one row group
+      try (RecordReader rows = reader.rows(reader.options().searchArgument(sarg, new String[]{}))) {
+        rows.nextBatch(batch);
+        assertEquals(1024, batch.size);
+        assertEquals(true, batch.cols[1].isRepeating);
+        assertEquals("val", ((BytesColumnVector) batch.cols[1]).toString(0));
+        assertEquals(false, rows.nextBatch(batch));
+      }
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4061,6 +4061,115 @@ public void testPredicatePushdownWithNan() throws Exception {
     assertEquals(0, batch.size);
   }
 
+  @Test
+  public void testPredicatePushdownIssue1() throws Exception {
```
shardulm94, line 4065: Nit: Make method name more descriptive?

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4061,6 +4061,115 @@ public void testPredicatePushdownWithNan() throws Exception {
...
+          Integer.toHexString(10*i).getBytes(StandardCharsets.UTF_8));
+    }
+    writer.addRowBatch(batch);
+    writer.close();
+
+    Reader reader = OrcFile.createReader(testFilePath,
+        OrcFile.readerOptions(conf).filesystem(fs));
+    assertEquals(3500, reader.getNumberOfRows());
+
+    // all values for int1 are null, so not(isNull(int1)) should always be false and not rows should be read
```
shardulm94, line 4091: Typo `not rows should be read` -> `no rows should be read`

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4061,6 +4061,115 @@ public void testPredicatePushdownWithNan() throws Exception {
...
+        .range(0L, Long.MAX_VALUE)
+        .searchArgument(sarg, new String[]{}));
+    batch = reader.getSchema().createRowBatch(3500);
+
+    rows.nextBatch(batch);
+    assertEquals(0, batch.size);
+  }
+
+  @Test
+  public void testPredicatePushdownIssue2() throws Exception {
```
shardulm94, line 4108: Nit: Make method name more descriptive?

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4061,6 +4061,115 @@ public void testPredicatePushdownWithNan() throws Exception {
...
+      }
+    }
+    writer.addRowBatch(batch);
+    writer.close();
+    Reader reader = OrcFile.createReader(testFilePath,
+        OrcFile.readerOptions(conf).filesystem(fs));
+    assertEquals(3500, reader.getNumberOfRows());
+
+    // values for int1 are either null or 1, so not(in(int1, 1)) should always be true because null
+    // occurs every 2nd row and hence all row groups should be returned
```
shardulm94, line 4141: This should now change from `hence all row groups should be returned` to `hence no row groups should be returned` as my initial understanding when I wrote the comment was wrong

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4061,6 +4061,115 @@ public void testPredicatePushdownWithNan() throws Exception {
...
+    RecordReader rows = reader.rows(reader.options()
+        .range(0L, Long.MAX_VALUE)
+        .searchArgument(sarg, new String[]{}));
+    batch = reader.getSchema().createRowBatch(2500);
+
+    rows.nextBatch(batch);
+    assertEquals(0, batch.size);
+
+    // values for string1 are either null or "val", so not(in(string1, "val")) should always be true because null
+    // occurs every 2nd row and hence all row groups should be returned
```
shardulm94, line 4157: This should now change from `hence all row groups should be returned` to `hence no row groups should be returned` as my initial understanding when I wrote the comment was wrong

java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
```diff
@@ -403,16 +406,18 @@ public void testGetMin() throws Exception {
 
   private static OrcProto.ColumnStatistics createIntStats(Long min,
                                                           Long max) {
+    OrcProto.ColumnStatistics.Builder result =
+        OrcProto.ColumnStatistics.newBuilder();
     OrcProto.IntegerStatistics.Builder intStats =
         OrcProto.IntegerStatistics.newBuilder();
     if (min != null) {
       intStats.setMinimum(min);
+      result.setNumberOfValues(1);
```
shardulm94, line 415: I am not sure why this is needed? Also if min != max wouldn't we have > 1 values?
PSA: I am not familiar with the Orc codebase, especially the protobuf parts. So may be totally missing the point.

omalley: Ok, the fix was about handling ColumnStatistics when there were no non-null values. So the added branch checks the number of values in the statistics. In real files, the only way to have a minimum or maximum is to have had at least one value, so it works.

In the sarg tests though, we were creating artificial ColumnStatistics that had minimum and maximum without any values. They broke with the new code. The fix was to make the artificial ColumnStatistics have a non-zero number of values.

omalley:
Ok, I just made some additional changes to test more variants. @shardulm94 let me know what you think.

shardulm94:
LGTM! 👍

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4061,6 +4061,137 @@ public void testPredicatePushdownWithNan() throws Exception {
...
+
+      // write 1024 rows of (null, "val")
+      batch.cols[0].noNulls = false;
+      batch.cols[0].isNull[0] = true;
+      batch.cols[0].isRepeating = true;
+      batch.cols[1].isRepeating = true;
+      ((BytesColumnVector) batch.cols[1]).setVal(0, "val".getBytes(StandardCharsets.UTF_8));
+      writer.addRowBatch(batch);
+
+      // write 1024 rows of (1, null)
```
wgtmac, line 4084: (123, null)

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4061,6 +4061,137 @@ public void testPredicatePushdownWithNan() throws Exception {
...
+        writer.addRowBatch(batch);
+      }
+    }
+
+    try (Reader reader = OrcFile.createReader(testFilePath,
+                          OrcFile.readerOptions(conf).filesystem(fs))) {
+      assertEquals(3*1024, reader.getNumberOfRows());
+      VectorizedRowBatch batch = reader.getSchema().createRowBatch();
+
+      // int1 is not null -- should select 0 of the row groups
```
wgtmac, line 4171: should be consistent with sarg, same for line 4183

wgtmac (approved):

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
omalley: Thanks for the catch, @wgtmac !

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
omalley: oops.

PR #545, opened 2020-09-20
ORC-667: Positional mapping for nested struct types should not applied by default

dongjoon-hyun:

[ORC-644](https://github.com/apache/orc/commit/bad383e9074b1597af50e85e77e7b6ce79147e39) (Support positional mapping for nested types) introduced a correctness regression on nested struct types. By default, `orc.force.positional.evolution` is false. However, currently it's applied for nested struct fields.

[ORC-644](https://github.com/apache/orc/commit/bad383e9074b1597af50e85e77e7b6ce79147e39) changes like the following. However, `buildConversion` is a recursive function and `positionalLevels` becomes negative in the nested case.
```java
buildConversion(fileChildren.get(i),
-  readerChildren.get(i), 0);
+  readerChildren.get(i), positionalLevels - 1);
```

Pass the newly added test case.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
+++ b/java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
@@ -513,7 +513,7 @@ public class SchemaEvolution {
             isOnlyImplicitConversion = false;
           }
 
-          if (positionalLevels == 0) {
+          if (positionalLevels <= 0) {
             List<String> readerFieldNames = readerType.getFieldNames();
             List<String> fileFieldNames = fileType.getFieldNames();
 
--- a/java/core/src/test/org/apache/orc/impl/TestSchemaEvolution.java
+++ b/java/core/src/test/org/apache/orc/impl/TestSchemaEvolution.java
@@ -1737,6 +1737,20 @@ public class TestSchemaEvolution {
   }
 
   @Test
+  public void testStructInArrayWithoutPositionalEvolution() throws IOException {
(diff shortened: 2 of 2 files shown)


dongjoon-hyun:
cc @ArvinZheng

java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
```diff
@@ -513,7 +513,7 @@ void buildConversion(TypeDescription fileType,
             isOnlyImplicitConversion = false;
           }
 
-          if (positionalLevels == 0) {
+          if (positionalLevels <= 0) {
```
dongjoon-hyun, line 516: This can be negative because of ORC-667.

ArvinZheng: because of ORC-644

ArvinZheng:
LGTM

ArvinZheng (approved):

java/core/src/java/org/apache/orc/impl/SchemaEvolution.java
dongjoon-hyun: Oh. Thanks! It's fixed.

dongjoon-hyun:
Thank you for review, @ArvinZheng .

omalley (approved):
+1

dongjoon-hyun:
Thank you, @omalley !

PR #548, opened 2020-10-05
ORC-670. RecordReaderImpl.findColumns should respect orc.schema.evolution.case.sensitive

dongjoon-hyun:

Apache ORC 1.6.0 ~ 1.6.5 cannot handle predicate pushdown case-insensitively.

This is a performance regression from Apache ORC 1.5.x. Without this patch, ORC 1.6.x cannot pass SPARK-32622 (case sensitivity in predicate pushdown) unit test cases.

This makes Apache ORC pass Apache Spark's `sql/core` module unit tests by fixing the last three failures. The remaining failures are in all `sql/hive` module due to another reasons.

Pass the newly added unit test case.

### DIFF
--- a/java/core/src/java/org/apache/orc/TypeDescription.java
+++ b/java/core/src/java/org/apache/orc/TypeDescription.java
@@ -818,8 +818,12 @@ public class TypeDescription
    * @return the subtype
    */
   public TypeDescription findSubtype(String columnName) {
+    return findSubtype(columnName, true);
+  }
+
+  public TypeDescription findSubtype(String columnName, Boolean isSchemaEvolutionCaseAware) {
     ParserUtils.StringPosition source = new ParserUtils.StringPosition(columnName);
-    TypeDescription result = ParserUtils.findSubtype(this, source);
+    TypeDescription result = ParserUtils.findSubtype(this, source, isSchemaEvolutionCaseAware);
     if (source.hasCharactersLeft()) {
       throw new IllegalArgumentException("Remaining text in parsing field name "
           + source);
--- a/java/core/src/java/org/apache/orc/impl/ParserUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/ParserUtils.java
@@ -246,6 +246,12 @@ public class ParserUtils {
 
   public static TypeDescription findSubtype(TypeDescription schema,
                                             ParserUtils.StringPosition source) {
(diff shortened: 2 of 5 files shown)


java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
```diff
@@ -112,6 +112,16 @@ public void testFindColumn() throws Exception {
     assertEquals(3, RecordReaderImpl.findColumns(evo, "e"));
   }
 
+  @Test
+  public void testFindColumnCaseInsensitively() throws Exception {
```
dongjoon-hyun, line 116: This passed in `branch-1.5` without any change.

dongjoon-hyun:
Could you review this PR please, @omalley ?

java/core/src/java/org/apache/orc/TypeDescription.java
```diff
@@ -818,8 +818,12 @@ public TypeDescription findSubtype(int goal) {
    * @return the subtype
    */
   public TypeDescription findSubtype(String columnName) {
+    return findSubtype(columnName, true);
+  }
+
+  public TypeDescription findSubtype(String columnName, Boolean isSchemaEvolutionCaseAware) {
```
omalley, line 824: This should be boolean.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -246,6 +246,12 @@ public static TypeDescription parseType(ParserUtils.StringPosition source) {
 
   public static TypeDescription findSubtype(TypeDescription schema,
                                             ParserUtils.StringPosition source) {
+    return findSubtype(schema, source, true);
+  }
+
+  public static TypeDescription findSubtype(TypeDescription schema,
+                                            ParserUtils.StringPosition source,
+                                            Boolean isSchemaEvolutionCaseAware) {
```
omalley, line 254: This should be boolean instead of Boolean.

omalley (changes requested):
This looks good, Dongjoon. It would be nice to replace the Boolean types with boolean to avoid the boxing/unboxing costs.

omalley:
Ok, I just made the signature change and committed it.

dongjoon-hyun:
Thank you so much, @omalley !

PR #549, opened 2020-10-05
ORC-671. Add OrcTail.getStripeStatistics back

dongjoon-hyun:

This PR aims to add `OrcTail.getStripeStatistics` back to mitigate backward compatibility partially. After this PR, the migration becomes much easier because changing `new OrcTail` is easier.
- If the existing app creates `OrcTail` via `extractFileTail` or `buildEmptyTail`, it will recover.
- For the other cases, the users need to migrate to `Reader.getStripeStatistics` or new `OrcTail` constructor.

Apache ORC 1.6.0 ~ 1.6.5 removes `getStripeStatistics` during Proleptic time support. We had better keep the backward compatibility as a wrapper for the new API.
```
[info]   java.lang.RuntimeException: ORC split generation failed with exception: java.lang.NoSuchMethodError: org.apache.orc.impl.OrcTail.getStripeStatistics()Ljava/util/List;
[info]   at org.apache.hadoop.hive.ql.io.orc.OrcInputFormat.generateSplitsInfo(OrcInputFormat.java:1731)
[info]   at org.apache.hadoop.hive.ql.io.orc.OrcInputFormat.getSplits(OrcInputFormat.java:1817)
```

This is a recovery of the old API.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/OrcTail.java
+++ b/java/core/src/java/org/apache/orc/impl/OrcTail.java
@@ -18,17 +18,24 @@ package org.apache.orc.impl;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
 import java.util.List;
 
 import org.apache.orc.CompressionKind;
 import org.apache.orc.OrcFile;
 import org.apache.orc.OrcProto;
 import org.apache.orc.OrcUtils;
+import org.apache.orc.Reader;
 import org.apache.orc.StripeInformation;
+import org.apache.orc.StripeStatistics;
 import org.apache.orc.TypeDescription;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


dongjoon-hyun:
This is still `Draft` because this reveals another missing APIs.

omalley:
Oh, I missed that you wanted to add more. I added a test case. :)

omalley:
I guess let's use a different PR for the follow up, but you can use the same jira.

omalley:
I guess let's use a different PR for the follow up, but you can use the same jira.

dongjoon-hyun:
Thank you, @omalley . Yes. I will use different JIRA for the remaining issue because the remaining issue is about another API.

dongjoon-hyun:
Also, thank you for the test case and I tested in `branch-1.6` with that test case and backported to `branch-1.6`, too.
- https://github.com/apache/orc/commit/d78cc39a9299b62bc8a5d8f5c3fac9201e03cb8b

PR #552, opened 2020-10-15
ORC-673: PPD LTE Point equality comparison is wrong when RG MIN==MAX

pgaref:

FIx PPD Less than Equals comparison for min==max corner case

Better PPD evaluation

Extra JUnit test


dongjoon-hyun:
+1, LGTM. Merged to branch-1.6.

PR #564, opened 2020-11-05
ORC-274: Remove columnNames field from Reader.Options

pgaref:
*  Remove the columnNames field from Reader.Options
*  Add searchArgument(SearchArgument sarg) as default method for SARG construction
*  Deprecate searchArgument(SearchArgument sarg, String[] columnNames)

Codebase cleaning and simplification with API consistency

Extended existing tests

### DIFF
--- a/java/core/src/java/org/apache/orc/Reader.java
+++ b/java/core/src/java/org/apache/orc/Reader.java
@@ -186,7 +186,6 @@ public interface Reader extends Closeable {
     private long length = Long.MAX_VALUE;
     private int positionalEvolutionLevel;
     private SearchArgument sarg = null;
-    private String[] columnNames = null;
     private Boolean useZeroCopy = null;
     private Boolean skipCorruptRecords = null;
     private TypeDescription schema = null;
@@ -272,14 +271,24 @@ public interface Reader extends Closeable {
     }
 
     /**
+     *  @deprecated columnNames are no needed for setting search argument. Method exists for backward compatibility.
+     *
      * Set search argument for predicate push down.
      * @param sarg the search argument
      * @param columnNames the column names for
      * @return this
      */
     public Options searchArgument(SearchArgument sarg, String[] columnNames) {
+      return searchArgument(sarg);
+    }
+
+    /**
+     * Set search argument for predicate push down.
+     * @param sarg the search argument
+     * @return this
+     */
+    public Options searchArgument(SearchArgument sarg) {
       this.sarg = sarg;
-      this.columnNames = columnNames;
       return this;
     }
 
@@ -390,8 +399,15 @@ public interface Reader extends Closeable {
       return preFilterColumns;
     }
 
+    /**
+     * Get the column-name mapping of search argument leaves.
+     * @return column names of predicate leaves or NULL when search argument not set
+     */
     public String[] getColumnNames() {
-      return columnNames;
+      if (this.sarg != null) {
+        return sarg.getLeaves().stream().map(l -> l.getColumnName()).toArray(String[]::new);
+      }
+      return null;
     }
 
     public long getMaxOffset() {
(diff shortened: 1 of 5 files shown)


pgaref:
@dongjoon-hyun can you please take a look?

dongjoon-hyun (changes requested):
Thank you for making a contribution, @pgaref . However, can we do this after we make Apache Hive and Apache Spark use Apache ORC 1.6?

```java
public String[] getColumnNames() {
  return columnNames;
}
```

Historically, Apache ORC removes many APIs without considering the downstream customers. Currently, Apache Hive and Apache Spark cannot use ORC 1.6 due to that.

pgaref:
Hey @dongjoon-hyun -- sure  API backward compatibility is definitely something we should keep an eye on -- I was personally trying to upgrade to 1.6 on HIVErecently and bumped into too many API incompatibilities.
However, we could also use schema/schemaEvolution and SARGS to derive columnNames (without keeping the extra field).
Would something like the above make sense?

dongjoon-hyun:
Ya. Sure, but my suggestion is we can do that after we finish Hive/Spark migration on 1.6 first~ Let's not increase the API compatibility gap any more.

pgaref:
> Ya. Sure, but my suggestion is we can do that after we finish Hive/Spark migration on 1.6 first~ Let's not increase the API compatibility gap any more.

Sure, sounds like a plan :)

dongjoon-hyun:
I'll convert this PR as `Draft` status to prevent any accidental merge for a while.

dongjoon-hyun:
Sorry for the delay here, @pgaref . I switch to from `Draft` to `Normal`. If this PR passes all three project CIs (Hive/Iceberg/Spark), we are good to go~

pgaref:
> Sorry for the delay here, @pgaref . I switch to from `Draft` to `Normal`. If this PR passes all three project CIs (Hive/Iceberg/Spark), we are good to go~

Thanks @dongjoon-hyun  for reviving this ! 
Just updated the PR to address the getColumnNames incompatibility -- if we wanted to use schema instead of predicate leaves we would have to move the method to the ReaderImpl class thus kept it simple here.

java/core/src/java/org/apache/orc/Reader.java
```diff
@@ -390,8 +399,15 @@ public SearchArgument getSearchArgument() {
...
     }
 
+    /**
+     * Get the search argument leaves' column-name mapping.
+     * @return an array of column names, one for every predicate leave, or NULL when search argument not set
+     */
     public String[] getColumnNames() {
-      return columnNames;
+      if (this.sarg != null) {
+        return sarg.getLeaves().stream().map(l -> l.getColumnName()).toArray(String[]::new);
```
dongjoon-hyun, line 408: Since this logic doesn't look consistent with the old usage, I have a few questions.
- What if the user-given `columeNames` has a different order from the leaves?
- What if the user-given `columnNames` are different case-sensitively?
- What if the user-given `columnNames` has more columns or less columns?

cc @omalley , too.

pgaref: Hey @dongjoon-hyun these are valid concerns, in general columnNames var is not used anymore (even on 1.5 branch).
The actual mapping to internal colIdx is happening as part of mapSargColumnsToOrcInternalColIdx which in turn calls RecordReaderImpl.findColumns where:

* Depending on the schema isSchemaEvolutionCaseAware var -- the SArgs columnNames can be case-insensitive or not
* If there is not reader schema-file type mapping the colName will be -1 (no Sargs applied for that column)
* If the column is not found in the Reader schema at all, it will throw an exception

In that sense, its the caller's responsibility to use the right Reader schema, with correct column names, define if needed the case-sensitivity (as part of schema.isSchemaEvolutionCaseAware) and handle any missing schema type issues.

PR #565, opened 2020-11-05
ORC-684: [C++] Make Floating point NaN check more strict

pgaref:
ORC-636 added an extra check (part of RecordReaderImp) disabling PPD when DoubleColumnStats min, max or sum is a non-Finite number.

However, the check is unnecessarily broad – the way stats are updated, using just sum for the check would suffice.

Make NAN check for DoubleStats (used for PPD) more strict

TestPredicateLeaf.cc


dongjoon-hyun:
Thank you, @pgaref . Is this a performance issue?

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Merged to master

PR #454, opened 2019-11-24
ORC-503: Introduce Maven wrapper

yuokada:
https://issues.apache.org/jira/browse/ORC-503

PR #566, opened 2020-11-06
ORC-503. Add Maven Wrapper

dongjoon-hyun:

This PR aims to add `Maven Wrapper`.

This will help ORC build and testing at the environment without Maven installation.

Manual.
```
cd java
./mvnw clean
```

This closes #454

### DIFF
--- a/java/mvnw
+++ b/java/mvnw
@@ -1,0 +1,102 @@
+#!/usr/bin/env bash
+
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Determine the current working directory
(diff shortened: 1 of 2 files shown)


dongjoon-hyun:
cc @omalley , @wgtmac , @pgaref

java/mvnw
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env bash
```
dongjoon-hyun, line 1: This file is a simplified version of Apache Spark mvn wrapper.
- https://github.com/apache/spark/blob/master/build/mvn

pgaref:
Hey @dongjoon-hyun  thanks for looking at this!
Wrapper script LGTM --  however I believe we still have to replace the MVN build/test calls with MVNW right?
For instance:

- https://github.com/apache/orc/blob/master/.github/workflows/master.yml#L46
- https://github.com/apache/orc/blob/master/java/CMakeLists.txt#L49

And as we are only now going to depend on MVNW maybe even remove maven plugin installation from Dockerfiles completely?

- https://github.com/apache/orc/blob/master/docker/debian10/Dockerfile#L33

Another thing we have to check is if the  MAVEN_OPTS are actually propagated to the wrapper (which I believe they should):
https://github.com/apache/orc/blob/master/.github/workflows/master.yml#L22

dongjoon-hyun:
Thank you so much for review, @pgaref . 

For the following, we don't need to use this in our CI or testing tool. This is for new environment.
> however I believe we still have to replace the MVN build/test calls with MVNW right?

For the following, that is not a scope of this PR. If some Docker images has a problem with old Maven versions, we had better migrate it step by step.
> And as we are only now going to depend on MVNW maybe even remove maven plugin installation from Dockerfiles completely?

The Apache Spark script already handles it like the following and uses it in Apache Spark GitHub Action job.
> Another thing we have to check is if the MAVEN_OPTS are actually propagated to the wrapper (which I believe they should): 
```bash
# Set any `mvn` options if not already present
export MAVEN_OPTS=${MAVEN_OPTS:-"$_COMPILE_JVM_OPTS"}
```

dongjoon-hyun:
Merged to master!

PR #569, opened 2020-11-09
ORC-672: FIX ORC type conversion within arrays where array.length >1024

pgaref:

Orc type conversion is throwing ArrayIndexOutOfBoundsException when using ColumnVectors larger than 1024.
The issue is originating from the ConvertTreeReaderFactory where we convert types always using the default CV length instead of the specific batch size.

This PR takes into account batchSize when converting types and adds tests for all possible type conversions (with large batchSize) as part of TestConvertTreeReaderFactory.

TestConvertTreeReaderFactory


dongjoon-hyun (approved):
+1, LGTM (Pending CIs)

dongjoon-hyun:
Merged to branch-1.5. Thanks!

PR #572, opened 2020-11-26
ORC-688: Allow CHAR to be promoted to STRING

pgaref:
Enable String conversion from varchar. 
Currently createStringConvertTreeReader in ConvertTreeReaderFactory is missing that opportunity

Modify createStringConvertTreeReader to handle char, varchar and string as part of the same group.

TestSchemaEvolution.testCharToStringEvolution


dongjoon-hyun (approved):
+1, LGTM. Thank you, @pgaref .

PR #576, opened 2020-12-04
ORC-693: Update credential according to INFRA setup

dongjoon-hyun:

This PR updates the credentials according to the direction of https://issues.apache.org/jira/browse/INFRA-21150 .

To publish snapshot.

N/A (The snapshot publish job is triggered at commit only).


dongjoon-hyun:
Could you review this, @pgaref ? This will work.

PR #583, opened 2020-12-15
ORC-699. Minor improvements to the scan tool that make the exception messages optional.

omalley:

- Change the stripe id to be 0 based, which seems more consistent.
- Change the end of file stripe id to be the number of stripes rather than -1.
- Limit the recovery row to the start of the next stripe.
- Add --verbose that prints exceptions.
- Fix the --help to give more information

Tested by hand on the example files.


dongjoon-hyun (approved):
Thank you, @omalley . I checked `-v` option manually.

PR #586, opened 2020-12-22
ORC-703 : Fix RLE encoding bug on large negative integer.

chaoyli:

ORC has use RLE to encoding/decoding integer.
Four types are comprised of the RLE encoding/decoding algorithm.
Short Repeat : used for short repeating integer sequences.
Direct : used for integer sequences whose values have a relatively constant bit width.
Patched Base : used for integer sequences whose bit widths varies a lot.
Delta : used for monotonically increasing or decreasing sequences.

This bug occurs in **Patched Base** Type for large negative number.
In patched base, [we use 3 bits to store base value ](https://orc.apache.org/specification/ORCv2/)width that is encoded using 1 to 8 bytes.
If the base value is actually 8 bytes in length, the value for base width should be 7.
Currently, this value can go up to 8 what can result in inconsistent data as part of the encoding procedure.
In extreme cases, the encoding/decoding process can even be cored dump referring to an illegal address.

Pass the newly added UT.

### DIFF
--- a/c++/src/RleEncoderV2.cc
+++ b/c++/src/RleEncoderV2.cc
@@ -577,9 +577,10 @@ void RleEncoderV2::writePatchedBasedValues(EncodingOption& option) {
     // find the number of bytes required for base and shift it by 5 bits
     // to accommodate patch width. The additional bit is used to store the sign
     // of the base value.
-    uint32_t baseWidth = findClosestNumBits(option.min) + 1;
-    if (baseWidth > 64) {
-        baseWidth = 64;
+    uint32_t baseWidth = findClosestNumBits(option.min);
+    if (baseWidth < 64) {
+        // store the sign bit when baseWidth is less than 64
+        baseWidth++;
     } 
     const uint32_t baseBytes = baseWidth % 8 == 0 ? baseWidth / 8 : (baseWidth / 8) + 1;
     const uint32_t bb = (baseBytes - 1) << 5;
--- a/c++/src/RleEncoderV2.cc
+++ b/c++/src/RleEncoderV2.cc
@@ -577,11 +577,9 @@ void RleEncoderV2::writePatchedBasedValues(EncodingOption& option) {
     // find the number of bytes required for base and shift it by 5 bits
     // to accommodate patch width. The additional bit is used to store the sign
     // of the base value.
-    uint32_t baseWidth = findClosestNumBits(option.min);
-    if (baseWidth < 64) {
-        // store the sign bit when baseWidth is less than 64
-        baseWidth++;
-    } 
+    uint32_t baseBitsNum = findClosestNumBits(option.min);
+    // if Negative, sign bit is already accounted for
+    uint32_t baseWidth = isNegative ? baseBitsNum : baseBitsNum + 1;
     const uint32_t baseBytes = baseWidth % 8 == 0 ? baseWidth / 8 : (baseWidth / 8) + 1;
     const uint32_t bb = (baseBytes - 1) << 5;
 
--- a/c++/src/RleEncoderV2.cc
+++ b/c++/src/RleEncoderV2.cc
@@ -577,7 +577,10 @@ void RleEncoderV2::writePatchedBasedValues(EncodingOption& option) {
     // find the number of bytes required for base and shift it by 5 bits
     // to accommodate patch width. The additional bit is used to store the sign
     // of the base value.
-    const uint32_t baseWidth = findClosestNumBits(option.min) + 1;
+    uint32_t baseWidth = findClosestNumBits(option.min) + 1;
+    if (baseWidth > 64) {
+        baseWidth = 64;
+    } 
     const uint32_t baseBytes = baseWidth % 8 == 0 ? baseWidth / 8 : (baseWidth / 8) + 1;
     const uint32_t bb = (baseBytes - 1) << 5;
 
--- a/c++/test/TestRleEncoder.cc
+++ b/c++/test/TestRleEncoder.cc
@@ -23,6 +23,7 @@
 
 #include "wrap/orc-proto-wrapper.hh"
 #include "wrap/gtest-wrapper.h"
+#include <iostream>
 
 #ifdef __clang__
   DIAGNOSTIC_IGNORE("-Wmissing-variable-declarations")
@@ -393,7 +394,7 @@ namespace orc {
     data[511] = std::numeric_limits<int64_t>::max();
 
     // Invoke the encoder.
-    const bool isSigned = false;
+    const bool isSigned = true;
     MemoryOutputStream memStream(DEFAULT_MEM_STREAM_SIZE);
 
     std::unique_ptr<RleEncoder> encoder = getEncoder(RleVersion_2, memStream, isSigned);
@@ -416,7 +417,7 @@ namespace orc {
     data[511] = std::numeric_limits<int64_t>::max();
 
     // Invoke the encoder.
-    const bool isSigned = false;
+    const bool isSigned = true;
     MemoryOutputStream memStream(DEFAULT_MEM_STREAM_SIZE);
 
     std::unique_ptr<RleEncoder> encoder = getEncoder(RleVersion_2, memStream, isSigned);
@@ -438,7 +439,7 @@ namespace orc {
                                 33333L };
 
     // Invoke the encoder.
-    const bool isSigned = false;
+    const bool isSigned = true;
     MemoryOutputStream memStream(DEFAULT_MEM_STREAM_SIZE);
 
     std::unique_ptr<RleEncoder> encoder = getEncoder(RleVersion_2, memStream, isSigned);
@@ -462,7 +463,7 @@ namespace orc {
                                 std::numeric_limits<int64_t>::max() };
 
     // Invoke the encoder.
-    const bool isSigned = false;
+    const bool isSigned = true;
     MemoryOutputStream memStream(DEFAULT_MEM_STREAM_SIZE);
 
     std::unique_ptr<RleEncoder> encoder = getEncoder(RleVersion_2, memStream, isSigned);
--- a/c++/test/TestRleEncoder.cc
+++ b/c++/test/TestRleEncoder.cc
@@ -283,7 +283,7 @@ namespace orc {
   // This case is used to testing encoding large negative integer.
   // The minimum is -84742859065569280, it's a 57 bit width integer 
   // and 8 bytes is used to encoding it.
-  TEST_P(RleTest, RleV2_Patched_large_negative_number) {
+  TEST_P(RleTest, RleV2_Patched_large_negative_integer) {
     // write data
     const int numValues = 30;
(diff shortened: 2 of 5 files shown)


chaoyli:
The following data can be used to reproduced the problem.
```
-17887939293638656
-15605417571528704
-15605417571528704
-13322895849418752
-13322895849418752
-84742859065569280
-15605417571528704
-13322895849418752
-13322895849418752
-15605417571528704
-13322895849418752
-13322895849418752
-15605417571528704
-15605417571528704
-13322895849418752
-13322895849418752
-15605417571528704
-15605417571528704
-13322895849418752
-13322895849418752
-11040374127308800
-15605417571528704
-13322895849418752
-13322895849418752
-15605417571528704
-15605417571528704
-13322895849418752
-13322895849418752
-15605417571528704
-13322895849418752
```

dongjoon-hyun:
Thank you, @chaoyli !

c++/src/RleEncoderV2.cc
```diff
@@ -577,7 +577,10 @@ void RleEncoderV2::writePatchedBasedValues(EncodingOption& option) {
     // find the number of bytes required for base and shift it by 5 bits
     // to accommodate patch width. The additional bit is used to store the sign
     // of the base value.
-    const uint32_t baseWidth = findClosestNumBits(option.min) + 1;
+    uint32_t baseWidth = findClosestNumBits(option.min) + 1;
```
pgaref, line 580: Seems like getClosestFixedBits method can indeed return up to 64 bits and adding thee sign bit here would exceed 8 bytes -- would it make sense to add the logic as part of the underlying method?

An alternative would be to skip the sign bit addition when the width is already 64 bits

c++/test/TestRleEncoder.cc
```diff
@@ -280,6 +280,33 @@ namespace orc {
     decodeAndVerify(RleVersion_2, memStream, data, numValues, nullptr, isSigned);
   }
 
+  // This case is used to testing encoding large negative integer.
```
pgaref, line 283: Is there a case that this would be an issue for positive numbers as well?

Can we also add some checks for the expected encoded values?

java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
```diff
@@ -286,7 +286,10 @@ private void writePatchedBaseValues() throws IOException {
     // find the number of bytes required for base and shift it by 5 bits
     // to accommodate patch width. The additional bit is used to store the sign
     // of the base value.
-    final int baseWidth = utils.findClosestNumBits(min) + 1;
+    int baseWidth = utils.findClosestNumBits(min) + 1;
```
pgaref, line 289: Same comment applies here (as the C++ version) -- I would try to avoid the extra if statement.

Can we please add some java tests as well?

pgaref (changes requested):
@chaoyli  Thanks for the PR and the useful details! Please find some comments below

dongjoon-hyun:
Also, cc @omalley

c++/src/RleEncoderV2.cc
chaoyli: Change getClosestFixedBits() logic will affect another code paths.
Because the sign is only needed in PatchBase encoding.
So change the PatchBase encoding will be better.

chaoyli: I will change it to skip the sign bit addition when the width is already 64 bits

java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
chaoyli: OK, I add this java tests

c++/test/TestRleEncoder.cc
chaoyli: I have test large positive number, but found it will not use PatchedBase encoding, so it's OK.

java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
chaoyli: OK, I add java test into the pull request.

dongjoon-hyun:
Could you review this once more, @pgaref ?

pgaref:
> Could you review this once more, @pgaref ?

Sure, wanted to check the RLE path more thoroughly before approving the fix -- will do that within the day.

pgaref:
Apologies for the slight delay but this got more complicated than I expected.
While testing the newly added [Java test](https://github.com/apache/orc/pull/586/files#diff-8f851a9e2ba9c8fd9f357a9df6482df5c264fdfdbb0e7625e9b2fe71ca9366a7R1226) I realised that the values were encoded using DIRECT encoding instead of the expected PATCHED_BASE.

The reason for picking DIRECT encoding was that some of the values exceed the [BASE_VALUE_LIMIT](https://github.com/apache/orc/blob/f2201f396cdd0c4e5b15589ec499e477235a553c/java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java#L537) that was introduced as part of ORC-616 (setting an upper limit to 56 bits numbers).

More precisely, **RunLengthIntegerWriterV2** checks if Math.abs(min) -- where min is the min input number of the sequence -- is greater than or equal to (1 << 56) bits -- to make sure the value of baseBytes is never above 8.  Seems that this check was never introduced to the Cpp side of things.

That said, I would recommend to port the above logic to the Cpp library to keep things consistent.
@chaoyli  please let me know if you want to make the change and again thanks for the patience :)

pgaref:
Btw, seems like the ORC-616 FIX is actually missing PATCH encoding opportunities with the LIMIT check (and using the most wasteful DIRECT instead) https://github.com/apache/orc/pull/601/files#diff-c76000edca5c4182015a83c604c01b032d4a97e31c0050c0e3ce0ccd195f284fR536

In reality, the only thing we need to check is if the number to be encoded is negative (already including the sign bit) and avoid accounting for the extra sign bit. https://github.com/apache/orc/pull/601/files#diff-c76000edca5c4182015a83c604c01b032d4a97e31c0050c0e3ce0ccd195f284fR291

Created a proof of concept for the java side of things as part of #601  -- please let me know what you think

chaoyli:
@pgaref Thank you to check the RLE path. I have read the patch code for [ORC-616](https://issues.apache.org/jira/browse/ORC-616)

pgaref:
> @pgaref Thank you to check the RLE path. I have read the patch code for [ORC-616](https://issues.apache.org/jira/browse/ORC-616). I found it use DIRECT code instead. The solution in #586 may be better for encoding. Can I revert the [ORC-616](https://issues.apache.org/jira/browse/ORC-616) and change it to my solution? Could you agree it? If you agree it, I will write the code to update the pull request.

Hey @chaoyli -- definitely agree that ORC-616 is not ideal in terms of encoding. 
Can you also check #601 and my [comment above?](https://github.com/apache/orc/pull/586#issuecomment-753592692) before updating the PR -- I believe that checking for negative numbers is sufficient here.

Thanks for working on this :)

chaoyli:
@pgaref OK, I will check it. I have a question to consult. Although I have not construct a case about large positive integer, you can ensure that large position integer will be encoded as PatchedBase?

pgaref:
> @pgaref OK, I will check it. I have a question to consult. Although I have not construct a case about large positive integer, you can ensure that large position integer will be encoded as PatchedBase?

[testPatchedBaseMax4](https://github.com/apache/orc/blob/60c9e9757ffb766edaaaeb7bed9c565a0316612d/java/core/src/test/org/apache/orc/TestNewIntegerEncoding.java#L1168) should already take care of that but sure we can add a more specific test too -- lets make sure these tests are replicated on the Cpp side as well

chaoyli:
@pgaref OK, I will check the code and copy the testPatchedBaseMax4 to Cpp side. I will submit the following pull request in the next few days. It will be trouble you to review at that time.

chaoyli:
@pgaref Sorry, I have delayed this patch for days. I introduced the PatchedBase unit test to Cpp side. Thank you to review it.

pgaref:
> @pgaref Sorry, I have delayed this patch for days. I introduced the PatchedBase unit test to Cpp side. Thank you to review it.

Hey @chaoyli  -- np at all, we all have our own distractions. 
PR is on the right path, can you please check the failing tests on the Cpp side?
I would also keep the **testBaseValueLimit** on the Java side even thought you have a similar test case -- I would just rename to something like testMaxPatchBaseValue.

Thanks

chaoyli:
> np at all, we all have our own distractions.

OK

chaoyli:
@pgaref I have test RleV2_Patched_max1 many times in linux, It's all OK. But in MSVC, it'a failed. I copy the data from MVSC to  reproduce the bug, it's also OK. I have a question, in MSVC, the alignBitpacking in TestRleEncoder.cc will be changed and make the case failed. It's may be late,  sorry, I have many jobs in the previous two months.

pgaref:
> @pgaref I have test RleV2_Patched_max1 many times in linux, It's all OK. But in MSVC, it'a failed. I copy the data from MVSC to reproduce the bug, it's also OK. I have a question, in MSVC, the alignBitpacking in TestRleEncoder.cc will be changed and make the case failed. It's may be late, sorry, I have many jobs in the previous two months.

Hey @chaoyli  np at all -- I know @omalley had some comments about this (as part of #601) maybe he can share more here.

chaoyli:
@omalley can you share some message here?

java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
```diff
@@ -530,7 +531,7 @@ private void determineEncoding() {
       // fallback to DIRECT encoding.
       // The decision to use patched base was based on zigzag values, but the
       // actual patching is done on base reduced literals.
-      if ((brBits100p - brBits95p) != 0 && Math.abs(min) < BASE_VALUE_LIMIT) {
+      if ((brBits100p - brBits95p) != 0) {
```
dirtysalt, line 534: Thanks. Look like this is the correct way to resolve the problem instead of this workaround. w

java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
```diff
@@ -286,7 +285,9 @@ private void writePatchedBaseValues() throws IOException {
     // find the number of bytes required for base and shift it by 5 bits
     // to accommodate patch width. The additional bit is used to store the sign
     // of the base value.
-    final int baseWidth = utils.findClosestNumBits(min) + 1;
+    int baseBitsNum = utils.findClosestNumBits(min);
```
dirtysalt, line 288: The root cause here is because we add 1 here, so `baseWidth` ranges from 2 to 65,  and `baseBytes` ranges from 0 to 8 which does not fit into 3 bits.

PR #593, opened 2020-12-28
ORC-708: FIX Snapshot publishing

pgaref:
Our snapshot pub seems to be broken currently. This PR investigates the issue

FIX snapshot publishing

Manual

### DIFF
--- a/.github/workflows/publish_snapshot.yml
+++ b/.github/workflows/publish_snapshot.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   publish-snapshot:
-    if: github.repository == 'apache/orc'
+    if: ${{ github.repository == 'apache/orc' }}
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@master
--- a/.github/workflows/publish_snapshot.yml
+++ b/.github/workflows/publish_snapshot.yml
@@ -17,7 +17,7 @@ jobs:
         java-version: 8
 
     - name: Release Maven package
-      uses: samuelmeuli/action-maven-publish@v1
+      uses: samuelmeuli/action-maven-publish@v1.4.0
       with:
         directory: java
         server_id: apache.snapshots.https
--- a/.github/workflows/publish_snapshot.yml
+++ b/.github/workflows/publish_snapshot.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   publish-snapshot:
-    if: ${{ github.repository == 'apache/orc' }}
+    if: github.repository == 'apache/orc'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@master


pgaref:
Hey @dongjoon-hyun any idea how we can actually test this before landing on the branch? :)

dongjoon-hyun (changes requested):
Are you sure that your claim `snapshot` publishing is broken by ORC-704?
> ORC-704 broke snapshot publishing

I also double-check the result. Please see [Apache ORC GitHub Action](https://github.com/apache/orc/actions?query=workflow%3A%22Publish+Snapshot%22).
Screen Shot 2020-12-28 at 5 24 14 PM

The root cause analysis is wrong. The failure is due to the following policy change. For me, it looks like Apache repository policy change.

> samuelmeuli/action-maven-publish@v1 is not allowed to be used in apache/orc. Actions in this workflow must be: created by GitHub, within a repository owned by apache or match the following: apache/*, adoptopenjdk/*, gradle/wrapper-validation-action.

.github/workflows/publish_snapshot.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   publish-snapshot:
-    if: github.repository == 'apache/orc'
+    if: ${{ github.repository == 'apache/orc' }}
```
dongjoon-hyun, line 10: I know that the original code has no problem technically. 
I can give you the working reference. 

- https://github.com/apache/spark/blob/master/.github/workflows/publish_snapshot.yml#L9
- https://github.com/apache/spark/actions?query=workflow%3A%22Publish+Snapshot%22

So, please revise your PR description and revert this code change from this PR. You may try to file a new JIRA for this style fix.

.github/workflows/publish_snapshot.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         java-version: 8
 
     - name: Release Maven package
-      uses: samuelmeuli/action-maven-publish@v1
+      uses: samuelmeuli/action-maven-publish@v1.4.0
```
dongjoon-hyun, line 20: According to the error message, I'm not sure this will work here. Do you have any reference why this fix the current error?
```
samuelmeuli/action-maven-publish@v1 is not allowed to be used in apache/orc.
Actions in this workflow must be: created by GitHub, verified in the GitHub Marketplace,
within a repository owned by apache or match the following:
adoptopenjdk/*, apache/*, gradle/wrapper-validation-action.
```

pgaref: > Are you sure that your claim snapshot publishing is broken by ORC-704?
You are right, 704 is not related -- just happened that all commits after that failed to publish a snapshot.

Regarding the error message: the confusion comes from the fact that the repo is actually in the apache/* form (thus assumed it was a string/expression evaluation issue). 

>  Do you have any reference why this fix the current error?

No ref here, just wanted to check if this helps with the policy issue above (from the message I assume we shouldn't be seeing the issue and it could be a bug).
Please let me know If there are other things we could try to fix this, you definitely have more experience GH actions :)

dongjoon-hyun:
If you don't mind, I'm going to try `maven` command directly instead of the `samuelmeuli/action-maven-publish@v1`.

.github/workflows/publish_snapshot.yml
dongjoon-hyun: Ya, +1 for that.

pgaref:
> If you don't mind, I'm going to try `maven` command directly instead of the `samuelmeuli/action-maven-publish@v1`.

Sure, totally fine by me

dongjoon-hyun:
Thank you, @pgaref .
I made a PR, https://github.com/apache/orc/pull/593 .

PR #598, opened 2021-01-02
ORC-708: Use maven command directly

dongjoon-hyun:

This PR aims to use `maven deploy` directly.

Apache repository doesn't allow 3rd party GitHub Action script.

Tested locally

This closes #593

### DIFF
--- a/.github/workflows/publish_snapshot.yml
+++ b/.github/workflows/publish_snapshot.yml
@@ -4,14 +4,11 @@ on:
   push:
     branches:
     - master
-  pull_request:
-    branches:
-    - master
 
 jobs:
   publish-snapshot:
     if: github.repository == 'apache/orc'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
According to the GitHub Action log, the secret cannot be used on the PR request.

dongjoon-hyun:
First of all, the command is tested locally with my credential.
- https://repository.apache.org/content/repositories/snapshots/org/apache/orc/orc-core/1.7.0-SNAPSHOT/orc-core-1.7.0-20210102.225408-13.jar

According to the test result at GitHub Action (at the first three commits), we cannot test with the Apache secret. It makes sense to me because that can be a security flaw.

So, we need to land this patch first.

pgaref (approved):
LGTM -- pretty clean and elegant solution! Thanks @dongjoon-hyun !

dongjoon-hyun:
Thank you so much, @pgaref .

It is recovered now.
- https://github.com/apache/orc/actions?query=workflow%3A%22Publish+Snapshot%22

PR #599, opened 2021-01-03
ORC-708: Use maven command directly

pgaref:

This PR aims to use `maven deploy` directly.

Apache repository doesn't allow 3rd party GitHub Action script.

Tested locally.

### DIFF
--- a/.github/workflows/publish_snapshot.yml
+++ b/.github/workflows/publish_snapshot.yml
@@ -16,10 +16,11 @@ jobs:
       with:
         java-version: 8
 
-    - name: Release Maven package
-      uses: samuelmeuli/action-maven-publish@v1
-      with:
-        directory: java
-        server_id: apache.snapshots.https
-        nexus_username: ${{ secrets.NEXUS_USER }}
-        nexus_password: ${{ secrets.NEXUS_PW }}
+    - name: Publish snapshot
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Ditto. Any explanation about this, @pgaref ?
Why do you reset the branch and handle like this?

pgaref:
Hey @dongjoon-hyun  -- while merging #599 an extra dummy commit was created in the branch.
I did try to clean this up by force pushing and created a mess as I had two more branches locally tracking apache/orc.

I did manage to recover everything but caused all this mess. So sorry about this.
At this point, I am even thinking that force pushing should be even disabled at repo level to avoid this

dongjoon-hyun:
@pgaref . It's recoverable, but here are some tips for you.
1. First of all, when you make a PR branch, please make a branch from Apache branch, not from outdated your local branch to avoid this kind of issue from the beginning.
2. Second, please don't merge your PR by yourself. Usually, +1 from other committer is the requirement although you can still merge it and get approval after merging.
3. Third, please stop do `force-pushing` at any reason. That should be the last resort.

dongjoon-hyun:
Please note that `force-push` breaks the downstream repositories and enforces the downstream also need to do force-push accordingly.

PR #601, opened 2021-01-03
ORC-703 : Fix RLE encoding bug on large negative integer.

pgaref:


omalley:
This reintroduced an old bug, which has an inadequate test. I’m working on a fix.

pgaref:
Hey @omalley  -- this PR was part of the discussion on #586 which is still on going and was mostly targeting the Cpp side of things -- we can hopefully use it to add adequate Tests with a proper FIX on both sides

PR #634, opened 2021-01-25
ORC-741: Schema Evolution missing column not handled in the presence of filters

pavibhai:
The following changes are proposed during the resolution of columns:
* If a column is not found in the read schema, then fail with an `IllegalArgumentException`
* If a column is not found in the file, then do not fail but instead treat as `NullReader`

In the absence of the patch, when the filter columns referred to missing columns in the file we were incorrectly getting an `IllegalArgumentException` instead of handling it as a missing column in the file.

Unit tests were added to verify this behavior

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -104,17 +104,10 @@ public class RecordReaderImpl implements RecordReader {
    */
   static int findColumns(SchemaEvolution evolution,
                          String columnName) {
-    try {
-      TypeDescription readerColumn = evolution.getReaderBaseSchema().findSubtype(
-          columnName, evolution.isSchemaEvolutionCaseAware);
-      TypeDescription fileColumn = evolution.getFileType(readerColumn);
-      return fileColumn == null ? -1 : fileColumn.getId();
-    } catch (IllegalArgumentException e) {
-      if (LOG.isDebugEnabled()){
-        LOG.debug("{}", e.getMessage());
-      }
-      return -1;
-    }
+    TypeDescription readerColumn = evolution.getReaderBaseSchema().findSubtype(
+        columnName, evolution.isSchemaEvolutionCaseAware);
+    TypeDescription fileColumn = evolution.getFileType(readerColumn);
+    return fileColumn == null ? -1 : fileColumn.getId();
   }
 
   /**
@@ -231,11 +224,9 @@ public class RecordReaderImpl implements RecordReader {
     if (options.getPreFilterColumnNames() != null) {
       for (String colName : options.getPreFilterColumnNames()) {
         int expandColId = findColumns(evolution, colName);
+        // If the column is not present in the file then this can be ignored from read.
         if (expandColId != -1) {
           filterColIds.add(expandColId);
-        } else {
-          throw new IllegalArgumentException("Filter could not find column with name: " +
-              colName + " on " + evolution.getReaderBaseSchema());
         }
       }
       LOG.info("Filter Columns: " + filterColIds);
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -100,14 +100,21 @@ public class RecordReaderImpl implements RecordReader {
    *
    * @param evolution the mapping from reader to file schema
    * @param columnName  the fully qualified column name to look for
-   * @return the file column number or -1 if the column wasn't found
+   * @return the file column number or -1 if the column wasn't found in the file schema
+   * @throws IllegalArgumentException if the column was not found in the reader schema
    */
   static int findColumns(SchemaEvolution evolution,
                          String columnName) {
-    TypeDescription readerColumn = evolution.getReaderBaseSchema().findSubtype(
-        columnName, evolution.isSchemaEvolutionCaseAware);
-    TypeDescription fileColumn = evolution.getFileType(readerColumn);
-    return fileColumn == null ? -1 : fileColumn.getId();
+    try {
+      TypeDescription readerColumn = evolution.getReaderBaseSchema().findSubtype(
+          columnName, evolution.isSchemaEvolutionCaseAware);
+      TypeDescription fileColumn = evolution.getFileType(readerColumn);
+      return fileColumn == null ? -1 : fileColumn.getId();
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("Filter could not find column with name: " +
+                                         columnName + " on " + evolution.getReaderBaseSchema(),
+                                         e);
+    }
   }
 
   /**
--- a/java/mapreduce/src/test/org/apache/orc/mapred/TestOrcFileEvolution.java
+++ b/java/mapreduce/src/test/org/apache/orc/mapred/TestOrcFileEvolution.java
@@ -235,17 +235,18 @@ public class TestOrcFileEvolution {
   @Test
   public void testPreHive4243CheckEqual() {
     // Expect success on equal schemas
-    checkEvolution("struct<_col0:int,_col1:string>",
+    checkEvolutionPosn("struct<_col0:int,_col1:string>",
                    "struct<_col0:int,_col1:string>",
                    struct(1, "foo"),
-                   struct(1, "foo", null), false, addSarg, false);
+                   struct(1, "foo", null),
+                   false, addSarg, false);
   }
 
   @Test
   public void testPreHive4243Check() {
     // Expect exception on strict compatibility check
     thrown.expectMessage("HIVE-4243");
-    checkEvolution("struct<_col0:int,_col1:string>",
+    checkEvolutionPosn("struct<_col0:int,_col1:string>",
                    "struct<_col0:int,_col1:string,_col2:double>",
                    struct(1, "foo"),
                    struct(1, "foo", null), false, addSarg, false);
@@ -253,17 +254,18 @@ public class TestOrcFileEvolution {
 
   @Test
   public void testPreHive4243AddColumn() {
-    checkEvolution("struct<_col0:int,_col1:string>",
+    checkEvolutionPosn("struct<_col0:int,_col1:string>",
                    "struct<_col0:int,_col1:string,_col2:double>",
                    struct(1, "foo"),
-                   struct(1, "foo", null), true, addSarg, false);
+                   struct(1, "foo", null),
+                   true, addSarg, false);
   }
 
   @Test
   public void testPreHive4243AddColumnMiddle() {
     // Expect exception on type mismatch
     thrown.expect(SchemaEvolution.IllegalEvolutionException.class);
-    checkEvolution("struct<_col0:int,_col1:double>",
+    checkEvolutionPosn("struct<_col0:int,_col1:double>",
                    "struct<_col0:int,_col1:date,_col2:double>",
                    struct(1, 1.0),
                    null, true, addSarg, false);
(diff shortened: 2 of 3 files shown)


java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -104,17 +104,10 @@
...
-    } catch (IllegalArgumentException e) {
-      if (LOG.isDebugEnabled()){
-        LOG.debug("{}", e.getMessage());
-      }
-      return -1;
-    }
+    TypeDescription readerColumn = evolution.getReaderBaseSchema().findSubtype(
+        columnName, evolution.isSchemaEvolutionCaseAware);
+    TypeDescription fileColumn = evolution.getFileType(readerColumn);
+    return fileColumn == null ? -1 : fileColumn.getId();
```
dongjoon-hyun, line 110: This PR passed without this change. Shall we have a test coverage for this change?
And, could you update the function document by adding 
```java
@throws IllegalArgumentException ....(description)
```

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -231,11 +224,9 @@ protected RecordReaderImpl(ReaderImpl fileReader,
     if (options.getPreFilterColumnNames() != null) {
       for (String colName : options.getPreFilterColumnNames()) {
         int expandColId = findColumns(evolution, colName);
+        // If the column is not present in the file then this can be ignored from read.
         if (expandColId != -1) {
           filterColIds.add(expandColId);
-        } else {
-          throw new IllegalArgumentException("Filter could not find column with name: " +
-              colName + " on " + evolution.getReaderBaseSchema());
```
dongjoon-hyun, line 238: Could you confirm that what is the new error message? I understand that the thrown exception is the same, `IllegalArgumentException`, but I'm not sure about the information on the error message.

java/mapreduce/src/test/org/apache/orc/mapred/TestOrcFileEvolution.java
```diff
@@ -312,6 +314,26 @@ public void testPositional2() {
         false, addSarg, true);
   }
 
+  private void checkEvolutionPosn(String writerType, String readerType,
```
dongjoon-hyun, line 317: This function seems to be duplicated mostly with `checkEvolution`. Is the only difference is the column name, `_col0` and `a`?

java/mapreduce/src/test/org/apache/orc/mapred/TestOrcFileEvolution.java
```diff
@@ -235,35 +235,37 @@ public void testListEvolution() {
...
 
   @Test
   public void testPreHive4243AddColumn() {
-    checkEvolution("struct<_col0:int,_col1:string>",
+    checkEvolutionPosn("struct<_col0:int,_col1:string>",
                    "struct<_col0:int,_col1:string,_col2:double>",
                    struct(1, "foo"),
-                   struct(1, "foo", null), true, addSarg, false);
+                   struct(1, "foo", null),
+                   true, addSarg, false);
```
dongjoon-hyun, line 261: Let's keep the original form to reduce the patch size.

java/mapreduce/src/test/org/apache/orc/mapred/TestOrcFileEvolution.java
```diff
@@ -235,35 +235,37 @@ public void testListEvolution() {
   @Test
   public void testPreHive4243CheckEqual() {
     // Expect success on equal schemas
-    checkEvolution("struct<_col0:int,_col1:string>",
+    checkEvolutionPosn("struct<_col0:int,_col1:string>",
                    "struct<_col0:int,_col1:string>",
                    struct(1, "foo"),
-                   struct(1, "foo", null), false, addSarg, false);
+                   struct(1, "foo", null),
+                   false, addSarg, false);
```
dongjoon-hyun, line 242: Let's keep the original form to reduce the patch size.

java/mapreduce/src/test/org/apache/orc/mapred/TestOrcFileEvolution.java
dongjoon-hyun: Ya. I also don't have a better option. +1 for adding this.

dongjoon-hyun (commented):
Thank you for making a PR, @pavibhai . I left a few minor comments. Could you check them?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -104,17 +104,10 @@
    */
   static int findColumns(SchemaEvolution evolution,
```
dongjoon-hyun, line 105: For this one, we hit many issues at Apache ORC 1.6.x due to the breaking API and behavior change.
Although this is an implementation internal class, Hive and Spark's hive module assumes old behavior.
So, if we need this change, I'd like to recommend to have a configuration to support both behavior.

pavibhai:
> Thank you for making a PR, @pavibhai . I left a few minor comments. Could you check them?

Thanks for the feedback @dongjoon-hyun will go through the same and update.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pavibhai: The behavior change we did to this API is as follows:
* If the column is present is missing in readSchema then throw an exception of missing column
* If the column is missing in file but present in readSchema then we see this as missing column in file handled via NullReader

Wanted to confirm that the ask is to not throw an exception when Hive/Spark use a column name that is missing in the readSchema. I might be missing the scenario where this is the case, I was under the impression that column references should always be valid within the readSchema.

pgaref: I believe it is safe to assume that the column should always be part of the readerSchema -- if thats not the case and there is a column/schema missmatch the caller (Spark/Hive) will be responsible for handling the RuntimeException.

I like this approach because its cleaner (and distinguishes between the two cases) so +1 for that -- but lets first make sure we add the behavior change as part of the method doc.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pgaref: +1 on the above -- lets add a new test or extend testSchemaEvolutionMissingColumn -- where the "missing" column is not part of schema evolution and assert for expected exception

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -231,11 +224,9 @@ protected RecordReaderImpl(ReaderImpl fileReader,
     if (options.getPreFilterColumnNames() != null) {
       for (String colName : options.getPreFilterColumnNames()) {
         int expandColId = findColumns(evolution, colName);
+        // If the column is not present in the file then this can be ignored from read.
         if (expandColId != -1) {
           filterColIds.add(expandColId);
-        } else {
```
pgaref, line 236: The intention of the original code here was to notify the consumer that the filter can not be applied  and could lead to wrong results (keep in mind that row-filtering is not best-effort as row-group skipping and the consumer expects only the rows that actually pass the filter).
This was also a side-effect of the existing code that had no distinction between a column missing from the schema, and schema-fileReader missmatch as you very well mentioned above.

However, (in the former case) we can now end up with pretty generic exceptions from findSubtype that do not reveal the root cause:
**IllegalArgumentException("Field " + first + " not found in " + current.toString())**

I believe the best thing to do here (even if its a bit ugly) is capture the Exception and rethrow with an appropriate message.

pgaref (changes requested):
Thanks for the PR @pavibhai ! Please check some comments inline

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pavibhai: Added the test

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pavibhai: Yeah, will add this wrapping into the find method as suggested by @pgaref

java/mapreduce/src/test/org/apache/orc/mapred/TestOrcFileEvolution.java
pavibhai: This was primarily to fix the errors in some of the tests that were using positional names

java/mapreduce/src/test/org/apache/orc/mapred/TestOrcFileEvolution.java
pavibhai: Done

java/mapreduce/src/test/org/apache/orc/mapred/TestOrcFileEvolution.java
pavibhai: Done

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pavibhai: @dongjoon-hyun My guess is that there are some unit tests that need fixing. I had to do the same here in ORC that is the reason for that new `checkEvolutionPosn` method in the test class.

@pgaref Updated the javadoc for the method.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pavibhai: Agreed. Wrapped the exception in the method as suggested.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -110,10 +111,9 @@ static int findColumns(SchemaEvolution evolution,
       TypeDescription fileColumn = evolution.getFileType(readerColumn);
       return fileColumn == null ? -1 : fileColumn.getId();
     } catch (IllegalArgumentException e) {
-      if (LOG.isDebugEnabled()){
-        LOG.debug("{}", e.getMessage());
-      }
-      return -1;
+      throw new IllegalArgumentException("Filter could not find column with name: " +
```
pgaref, line 114: Style is a bit off here?

java/mapreduce/src/test/org/apache/orc/mapred/TestOrcFileEvolution.java
```diff
@@ -232,10 +233,23 @@ public void testListEvolution() {
                    addSarg);
   }
 
+  @Test
+  public void testMissingColumnFromReaderSchema() {
+    // Expect failure if the column is missing from the reader schema, as column a that is added
+    // by the SArg is missing from the reader schema
+    Assert.assertThrows("Field a not found in", IllegalArgumentException.class,
+                        () -> checkEvolution("struct<b:int,c:string>",
```
pgaref, line 241: style?

pgaref:
> @pgaref @dongjoon-hyun
> Thanks for your review comments. I have addressed them, please check and let me know if something is amiss.

Thanks for the prompt reply @pavibhai  -- Latest PR is almost there, please check the two style comments above

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
omalley: @pgaref what are you asking for?

pgaref: Looked like the indentation was off here (on my IDE at least) but just realised it was just split in 2 lines instead of 3 -- ignore the comment please

pgaref (approved):

omalley (approved):

dongjoon-hyun (commented):
+1, late LGTM. Thank you all.

PR #637, opened 2021-01-27
ORC-747: Abstract Dictionary interface and refactoring

autumnust:
This PR contains the very first part to make the implementation of dictionary used for encoding pluggable, by introducing an interface `Dictionary`. The subsequent PR will add HashTable-based Dictionary implementation and compare with the existing RB-tree based dictionary. 

For reviewers: 
- Please comment if reflection-based initialization will be preferred or not since I didn't found much of that within the code base. 
- The `Dictionary` class isn't added with generic type declaration (e.g. we could use a generic type `T` to replace `Text`). Please let me know if that's necessary to add. 

Since this is mostly a refactoring, it is just passing all existing unit tests.

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -105,9 +105,11 @@ public enum OrcConf {
           "(default 10000 rows) else dictionary check will happen before\n" +
           "writing first stripe. In both cases, the decision to use\n" +
           "dictionary or not will be retained thereafter."),
-  DICTIONARY_FACTORY_CLASS_KEY("orc.dictionary.class.key", null,
-      "org.apache.orc.impl.StringRedBlackTree.StringRBTreeFactory",
-      "Key to specify the factory class in charge of generating dictionary."),
+  DICTIONARY_IMPL("orc.dictionary.implementation", null,
+      "rbtree",
+      "the implementation for the dictionary used for string-type column encoding.\n" +
+          "The legit value for this key could be 'rbtree' or 'hash' as it currently supports\n"
+          + " rbtree-based or hashTable-based implementation."),
   BLOOM_FILTER_COLUMNS("orc.bloom.filter.columns", "orc.bloom.filter.columns",
       "", "List of columns to create bloom filters for when writing."),
   BLOOM_FILTER_WRITE_VERSION("orc.bloom.filter.write.version",
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -105,11 +105,12 @@ public enum OrcConf {
           "(default 10000 rows) else dictionary check will happen before\n" +
           "writing first stripe. In both cases, the decision to use\n" +
           "dictionary or not will be retained thereafter."),
-  DICTIONARY_IMPL("orc.dictionary.implementation", null,
+  DICTIONARY_IMPL("orc.dictionary.implementation", "orc.dictionary.implementation",
       "rbtree",
       "the implementation for the dictionary used for string-type column encoding.\n" +
-          "The legit value for this key could be 'rbtree' or 'hash' as it currently supports\n"
-          + " rbtree-based or hashTable-based implementation."),
+          "The choices are:\n"
+          + " rbtree - use red-black tree as the implementation for the dictionary.\n"
+          + " hash (yet to be implemented) - use hash table as the implementation for the dictionary."),
   BLOOM_FILTER_COLUMNS("orc.bloom.filter.columns", "orc.bloom.filter.columns",
       "", "List of columns to create bloom filters for when writing."),
   BLOOM_FILTER_WRITE_VERSION("orc.bloom.filter.write.version",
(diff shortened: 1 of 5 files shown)


dongjoon-hyun:
Thank you for making a PR, @autumnust . It seems that many failure. Could you test it locally first?
```
Error:  Tests run: 1146, Failures: 0, Errors: 48, Skipped: 1
```

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -37,7 +37,7 @@
   ROW_INDEX_STRIDE("orc.row.index.stride",
       "hive.exec.orc.default.row.index.stride", 10000,
       "Define the default ORC index stride in number of rows. (Stride is the\n"+
-          " number of rows n index entry represents.)"),
+          " number of rows an index entry represents.)"),
```
dongjoon-hyun, line 40: Let's avoid touching irrelevant part. You can fix this as a separate PR.

dongjoon-hyun (commented):
This PR is partial for ORC-50. So, I create a new JIRA for this PR. After we finish all implementation, we can resolve ORC-50 as an umbrella JIRA.

dongjoon-hyun:
Gentle ping, @autumnust .

java/core/src/java/org/apache/orc/OrcConf.java
omalley: I don't think we need a separate jira/pr for this. I'll commit it separately.

autumnust:
> Gentle ping, @autumnust .

Sorry I got preempted by some other thing last week. I will get back to fixing the failure in Github Action ASAP. Thanks @dongjoon-hyun  for your review in first round.

java/core/src/java/org/apache/orc/impl/Dictionary.java
```diff
@@ -0,0 +1,99 @@
...
+  void visit(Visitor visitor) throws IOException;
+
+  void clear();
+
+  /**
+   * Given the position index, return the original string before being encoded.
+   */
+  void getText(Text result, int originalPosition);
+
+  int add(String value);
```
omalley, line 50: We shouldn't need the add(String).

java/core/src/java/org/apache/orc/impl/Dictionary.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.hadoop.io.Text;
+
+
+/**
+ * Interface to define the dictionary used for encoding value in columns of specific types like string, char, varchar, etc.
+ */
+public interface Dictionary {
+  int INITIAL_DICTIONARY_SIZE = 4096;
+
+  interface DictionaryFactory {
```
omalley, line 34: I don't think I'd bother defining a factory interface.

java/core/src/java/org/apache/orc/impl/writer/StringBaseTreeWriter.java
```diff
@@ -64,6 +64,15 @@
                        WriterEncryptionVariant encryption,
                        WriterContext writer) throws IOException {
     super(schema, encryption, writer);
+    Configuration conf = writer.getConfiguration();
+
```
omalley, line 68: I'd make a function that looks at the configuration value and builds the hash or red-black dictionary directly. I can't see a need to make the dictionary mechanism pluggable.

omalley (changes requested):

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -105,6 +105,11 @@
           "(default 10000 rows) else dictionary check will happen before\n" +
           "writing first stripe. In both cases, the decision to use\n" +
           "dictionary or not will be retained thereafter."),
+  DICTIONARY_IMPL("orc.dictionary.implementation", null,
```
pgaref, line 108: hash option missing?

java/core/src/java/org/apache/orc/impl/writer/StringBaseTreeWriter.java
```diff
@@ -55,15 +59,32 @@
   // If the number of keys in a dictionary is greater than this fraction of
   //the total number of non-null rows, turn off dictionary encoding
   private final double dictionaryKeySizeThreshold;
+  protected Dictionary dictionary;
   protected boolean useDictionaryEncoding = true;
   private boolean isDirectV2 = true;
   private boolean doneDictionaryCheck;
   private final boolean strideDictionaryCheck;
 
+  static Dictionary createDict(Configuration conf) {
```
pgaref, line 68: make private?

java/core/src/java/org/apache/orc/OrcConf.java
autumnust: the value for this config is being processed `createDict`.

java/core/src/java/org/apache/orc/impl/writer/StringBaseTreeWriter.java
autumnust: addressed.

java/core/src/java/org/apache/orc/OrcConf.java
pgaref: I guess what I meant was that we should explicitly list the choices (similar to BLOOM_FILTER_WRITE_VERSION):
* rbTree
* hash (not yet implemented

Moreover, I would set the hiveConfName same as the attribute -- I believe null means that it does not apply to Hive

autumnust: Thanks, updated the description in a way similar to `BLOOM_FILTER_WRITE_VERSION`.

PR #641, opened 2021-02-11
ORC-508 remove hadoop dependency

omalley:
This change makes Hadoop an optional dependence for ORC reader/writers.

It creates a new package org.apache.orc.core that contains a new version of OrcFile, OrcConf, etc. that don't use Hadoop classes.

The old classes stay in their old place and should be compatible. The tests required relatively little change.

I also included a new version of the tools that does not include Hadoop. It is naturally much smaller (3.6mb vs 40mb).

I compiled Hive master branch with the patch to find any places where I could minimize the API changes. Most of the changes are things that changed in ORC 1.6 to master, but they aren't much. https://github.com/omalley/hive/commit/12c1e6416990b7387b7bb823a62468e5a49f24d6 

Remaining question:
* We still need Writable and WritableComparable from hadoop. Those interfaces are tiny and haven't changed in more than a decade. I currently put them into the nonhadoop tool directory, but they should probably be somewhere else. Thoughts?


omalley:
Other notes/questions:
* There is a new implementation of HadoopShims that avoids using Hadoop.
* For classes where I needed a second implementation that doesn't use Hadoop, I used *Core.
* Do we actually want to commit the nonhadoop module? It is a complete duplicate of tools. It is however, a nice demonstration that you can read and write ORC files without hadoop on the classpath at all.
* If we do keep nonhadoop, it currently duplicates the sample data from tools, which should probably be fixed.

omalley:
Clearly we also need to update a fair amount of documentation also.

dongjoon-hyun (commented):
Wow. It's a massive change. Could you make CIs happy? It seems to be failing now for some reasons.

dongjoon-hyun:
Hi, @omalley . Is there something I can help you with this PR?

PR #642, opened 2021-02-17
ORC-749 Add checkstyle:check to analyze profile.

omalley:
This change makes the analyze profile also run checkstyle:check. It pulls the manual rat and checkstyle checks into cmake's build.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -41,8 +41,5 @@ jobs:
         mkdir -p ~/.m2
         mkdir build
         cd build
-        cmake ..
+        cmake -DANALYZE_JAVA=ON ..
         make package test-out
-        cd ../java
-        mvn apache-rat:check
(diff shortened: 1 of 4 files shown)


dongjoon-hyun (commented):
Thank you, @omalley . Is this blocked by ORC-750 ?

pgaref:
> Thank you, @omalley . Is this blocked by [ORC-750](https://issues.apache.org/jira/browse/ORC-750) ?

These look like findbugs issues -- the interesting fact is that they only show up with java > 8

omalley:
By turning on the analyze profile in the CI build, now we get find bugs again in CI.

Java 11 causes a new set of findbugs false alarms, so those need to be suppressed.

I'll follow up with ORC-750 that enables the analyze profile for the benchmark too.

dongjoon-hyun (approved):
+1, LGTM.

PR #645, opened 2021-02-20
ORC-753: [C++] Clear stream state of ColumnReader after seek

wgtmac:
C++ ORC ColumnReaders of some types have cached buffer state to issue next I/O. This is fine if we read streams continuously. However, if we try to skip some row groups and seek to a random row group, the cached buffer state is wrong and may lead to undefined behavior. This patch fixes it by clearing all stream states in the ColumnReader.

This fix is a prerequisite to introduce PPD to C++ ORC reader in that the read pattern requires seeking to any row group.

This is a trivial fix and just make sure it doesn't break any unit test case.

### DIFF
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -534,6 +534,9 @@ namespace orc {
     std::unordered_map<uint64_t, PositionProvider>& positions) {
     ColumnReader::seekToRowGroup(positions);
     inputStream->seek(positions.at(columnId));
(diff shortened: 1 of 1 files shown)


wgtmac:
@dongjoon-hyun I have split this minor fix from the PPD patch. Please review it when you have time. Thanks!

xndai (approved):
LGTM

dongjoon-hyun:
+1, late LGTM. Thank you, @wgtmac .

PR #647, opened 2021-03-01
ORC-755: Added OrcFilterContext that accomplishes the following:

pavibhai:
* Change the filter from Consumer<VectorizedRowBatch> to Consumer<OrcFilterContext>
* OrcFilterContext has findVector method that allows cached determination of ColumnVectors from name
* OrcFilterContext offers utility methods for null determination on the vector branch
* Refactors ParserUtils to introduce TypeWalker and VectorWalker to centralize the walk logic
* Tests have been modified to support the new interface

* Change the filter from Consumer<VectorizedRowBatch> to Consumer<OrcFilterContext>
* Add the following convenience methods
  * OrcFilterContext has findVector method that allows cached determination of ColumnVectors from name
  * OrcFilterContext offers utility methods for null determination on the vector branch
* Refactors ParserUtils to introduce TypeWalker and VectorWalker to centralize the walk logic

The changes are needed to allow easy access to ColumnVectors by name as part of [ORC-744]

Unit tests were added to the project.

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcFilterContext.java
+++ b/java/core/src/java/org/apache/orc/OrcFilterContext.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.orc;
+
+import org.apache.hadoop.hive.ql.exec.vector.ColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.ListColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.MapColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
+import org.apache.hadoop.hive.ql.io.filter.MutableFilterContext;
+
+/**
+ * This defines the input for any filter operation. This is an extension of
+ * [[{@link VectorizedRowBatch}]] with schema.
+ * <p>
+ * This offers a convenience method of finding the column vector from a given column name
+ * that the filters can invoke to get access to the column vector.
+ */
+public interface OrcFilterContext extends MutableFilterContext {
+  /**
+   * Retrieves the column vector that matches the specified name. Allows support for nested struct
+   * references e.g. order.date where data is a field in a struct called order.
+   *
+   * @param name The column name whose vector should be retrieved
+   * @return The column vector
+   * @throws IllegalArgumentException if the field is not found or if the nested field is not part
+   *                                  of a struct
+   */
+  ColumnVector[] findColumnVector(String name);
+
+  /**
+   * Utility method for determining if the leaf vector of the branch can be treated as having
+   * noNulls.
+   * This method navigates from the top to the leaf and checks if we have nulls anywhere in the
+   * branch as compared to checking just the leaf vector.
+   *
+   * @param vectorBranch The input vector branch from the root to the leaf
+   * @return true if the entire branch satisfies noNull else false
+   */
+  static boolean noNulls(ColumnVector[] vectorBranch) {
+    for (ColumnVector v : vectorBranch) {
+      if (!v.noNulls) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
+   * Utility method for determining if a particular row element in the vector branch is null.
+   * This method navigates from the top to the leaf and checks if we have nulls anywhere in the
+   * branch as compared to checking just the leaf vector.
+   *
+   * @param vectorBranch The input vector branch from the root to the leaf
+   * @param idx          The row index being tested
+   * @return true if the entire branch is not null for the idx otherwise false
+   * @throws IllegalArgumentException If a multivalued vector such as List or Map is encountered in
+   *                                  the branch.
+   */
+  static boolean isNull(ColumnVector[] vectorBranch, int idx) throws IllegalArgumentException {
+    for (ColumnVector v : vectorBranch) {
+      if (v instanceof ListColumnVector || v instanceof MapColumnVector) {
+        throw new IllegalArgumentException(String.format(
+          "Found vector: %s in branch. List and Map vectors are not supported in isNull "
+          + "determination", v));
+      }
+      if (!v.noNulls && v.isNull[idx]) {
+        return true;
+      }
+    }
+    return false;
+  }
+}
--- a/java/core/src/java/org/apache/orc/OrcFilterContext.java
+++ b/java/core/src/java/org/apache/orc/OrcFilterContext.java
@@ -34,10 +34,12 @@ import org.apache.hadoop.hive.ql.io.filter.MutableFilterContext;
 public interface OrcFilterContext extends MutableFilterContext {
   /**
    * Retrieves the column vector that matches the specified name. Allows support for nested struct
-   * references e.g. order.date where data is a field in a struct called order.
+   * references e.g. order.date where date is a field in a struct called order.
    *
    * @param name The column name whose vector should be retrieved
-   * @return The column vector
+   * @return The column vectors from the root to the column name. The array levels match the name
+   * levels with Array[0] referring to the top level, followed by the subsequent levels. For
+   * example of order.date Array[0] refers to order and Array[1] refers to date
    * @throws IllegalArgumentException if the field is not found or if the nested field is not part
    *                                  of a struct
    */
@@ -79,7 +81,7 @@ public interface OrcFilterContext extends MutableFilterContext {
           "Found vector: %s in branch. List and Map vectors are not supported in isNull "
           + "determination", v));
       }
-      if (!v.noNulls && v.isNull[idx]) {
+      if (!v.noNulls && (v.isRepeating || v.isNull[idx])) {
         return true;
       }
     }
--- a/java/core/src/java/org/apache/orc/OrcFilterContext.java
+++ b/java/core/src/java/org/apache/orc/OrcFilterContext.java
@@ -81,7 +81,8 @@ public interface OrcFilterContext extends MutableFilterContext {
           "Found vector: %s in branch. List and Map vectors are not supported in isNull "
           + "determination", v));
       }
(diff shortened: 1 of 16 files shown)


pavibhai:
@omalley Thanks for your suggestions on the findColumnVector method, I think I incorporated all of them here
@pgaref Thanks for your suggestions on reusing the findSubType method, I have included that into this PR.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -244,80 +250,172 @@ public static TypeDescription parseType(ParserUtils.StringPosition source) {
 
   private static final Pattern INTEGER_PATTERN = Pattern.compile("^[0-9]+$");
 
+  public static ColumnVector[] findBranchVectors(TypeDescription schema,
```
omalley, line 253: I think this would be better named findColumnVectors.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -244,80 +250,172 @@ public static TypeDescription parseType(ParserUtils.StringPosition source) {
...
     return findSubtype(schema, source, true);
   }
 
-  public static TypeDescription findSubtype(TypeDescription schema,
-                                            ParserUtils.StringPosition source,
-                                            boolean isSchemaEvolutionCaseAware) {
-    List<String> names = ParserUtils.splitName(source);
-    if (names.size() == 1 && INTEGER_PATTERN.matcher(names.get(0)).matches()) {
-      return schema.findSubtype(Integer.parseInt(names.get(0)));
+  private static final ThreadLocal<TypeWalker> typeWalker = ThreadLocal.withInitial(TypeWalker::new);
```
omalley, line 265: I don't think we need the ThreadLocals. This code is only called at setup.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -244,80 +250,172 @@ public static TypeDescription parseType(ParserUtils.StringPosition source) {
...
-            for (int i = 0; i < current.getFieldNames().size(); i++) {
-              if (current.getFieldNames().get(i).equalsIgnoreCase(first)) {
-                posn = i;
-                break;
+
+    protected void handleLevel(int posn) {
+      this.current = current.getChildren().get(posn);
+    }
+
+    private void findInternal() {
```
omalley, line 280: I think I'd prefer to leave the findInternal as a static method that takes a visitor. I've posted a branch with such a change on my github. https://github.com/omalley/orc/tree/orc-755

omalley (changes requested):
Pavan, thank you for breaking up the large patch. This helps a lot.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
pavibhai: agreed

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
```diff
@@ -0,0 +1,202 @@
...
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TestOrcFilterContext {
+  private final TypeDescription schema = TypeDescription.createStruct()
```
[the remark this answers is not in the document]
pavibhai: will do

java/core/src/java/org/apache/orc/impl/ParserUtils.java
pavibhai: Will check and revert, thanks for the proposal

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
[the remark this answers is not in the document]
omalley: You don't need to change it, I just noticed that you were writing a lot of code using the builders that we used before we had the type parser. :)

java/core/src/java/org/apache/orc/impl/OrcFilterContextImpl.java
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * This defines the input for any filter operation. This is an extension of
+ * [[{@link VectorizedRowBatch}]] with schema.
+ * <p>
+ * This offers a convenience method of finding the column vector from a given column name
+ * that the filters can invoke to get access to the column vector.
+ */
+public class OrcFilterContextImpl implements OrcFilterContext {
+  VectorizedRowBatch batch = null;
```
pgaref, line 38: batch access level?

java/core/src/java/org/apache/orc/OrcFilterContext.java
```diff
@@ -0,0 +1,88 @@
...
+ * This defines the input for any filter operation. This is an extension of
+ * [[{@link VectorizedRowBatch}]] with schema.
+ * <p>
+ * This offers a convenience method of finding the column vector from a given column name
+ * that the filters can invoke to get access to the column vector.
+ */
+public interface OrcFilterContext extends MutableFilterContext {
+  /**
+   * Retrieves the column vector that matches the specified name. Allows support for nested struct
+   * references e.g. order.date where data is a field in a struct called order.
```
pgaref, line 37: typo: data

java/core/src/java/org/apache/orc/OrcFilterContext.java
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * @param name The column name whose vector should be retrieved
+   * @return The column vector
+   * @throws IllegalArgumentException if the field is not found or if the nested field is not part
+   *                                  of a struct
+   */
+  ColumnVector[] findColumnVector(String name);
```
pgaref, line 44: Clearly mention that this returns an arrays of Vectors -- e.g., order with return a StructColumn on array[0] and the field columns on the remaining slots.

pgaref (commented):
Thanks for the PR @pavibhai this looks much better now. Left some, mostly minor, comments

java/core/src/java/org/apache/orc/OrcFilterContext.java
pavibhai: Added

java/core/src/java/org/apache/orc/OrcFilterContext.java
pavibhai: Fixed

java/core/src/java/org/apache/orc/impl/OrcFilterContextImpl.java
pavibhai: Left as Package Local for use in Tests. Is that what you were referring to?

java/core/src/java/org/apache/orc/impl/ParserUtils.java
pavibhai: I have incorporated the suggestion with some tweaks.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
pavibhai: Make sense, removed this. Given that ColumnVectors are cached, this should not happen after the initial setup.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -249,75 +255,210 @@ public static TypeDescription findSubtype(TypeDescription schema,
...
+    }
+  }
+
+  static class ColumnFinder implements TypeVisitor {
+    private ColumnVector[] top;
+    private ColumnVector current = null;
+    private final ColumnVector[] result;
+    private int resultIdx = 0;
+
+    ColumnFinder(TypeDescription schema, VectorizedRowBatch batch, List<String> levels) {
```
omalley, line 412: Instead of passing in the List of strings, I think I'd pass in the levels as an int.

omalley (changes requested):

java/core/src/java/org/apache/orc/impl/ParserUtils.java
pavibhai: done

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -249,75 +255,210 @@ public static TypeDescription findSubtype(TypeDescription schema,
...
         case UNION: {
           try {
-            int posn = Integer.parseInt(first);
+            posn = Integer.parseInt(first);
             if (posn < 0 || posn >= current.getChildren().size()) {
               throw new NumberFormatException("off end of union");
             }
-            current = current.getChildren().get(posn);
           } catch (NumberFormatException e) {
             throw new IllegalArgumentException("Field " + first +
```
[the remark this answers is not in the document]
pavibhai: Currently the code passes the exception context to the IllegalArgumentException, that is the reason this was not simplified as posn = -1

java/core/src/java/org/apache/orc/OrcFilterContext.java
```diff
@@ -0,0 +1,90 @@
...
+   *                                  the branch.
+   */
+  static boolean isNull(ColumnVector[] vectorBranch, int idx) throws IllegalArgumentException {
+    for (ColumnVector v : vectorBranch) {
+      if (v instanceof ListColumnVector || v instanceof MapColumnVector) {
+        throw new IllegalArgumentException(String.format(
+          "Found vector: %s in branch. List and Map vectors are not supported in isNull "
+          + "determination", v));
+      }
+      if (!v.noNulls && (v.isRepeating || v.isNull[idx])) {
```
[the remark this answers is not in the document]
pavibhai: Thanks for clarifying that v.noNull = false does not mean that we have at least one null. I have made the requested change.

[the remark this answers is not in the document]
pgaref: I believe it does as we populate it from the present stream -- the problem is here you are checking for a particular row across CVs so if the flag is false you need to check across them.

java/core/src/java/org/apache/orc/impl/OrcFilterContextImpl.java
pgaref: Yes exactly, I saw you added a getter for testing thats perfectly fine

java/core/src/java/org/apache/orc/OrcFilterContext.java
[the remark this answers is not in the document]
pavibhai: I am a little confused here. I was checking at each level, we had a loop that started at the root level and navigates down by checking if the row is null at each level, the first level when we find a null we return back as having null.

My approach was to say that if `v.noNulls == false` and `v.isRepeating == true` then the row is null with the assumption that at this level it means every element is a null. If we have already hit null at this level we return and do not check the lower levels.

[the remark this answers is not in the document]
pgaref: I believe you are right, if we know noNulls == false and isRepeating==true I dont think we need to check isNull[0].

omalley (approved):

PR #652, opened 2021-03-12
ORC-758: Avoid seeking and decompressing of compressed stream

pavibhai:
If the seek on a compressed stream is already satisfied by the current decompression then avoid seeking and decompressing the same block again.

Seeks performed within the same compression block avoid the costs of decompression.

Modified an existing test to verify that the compressed buffer changes only when the seek is needed and not in other cases.
Additionally all of the existing tests pass.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/InStream.java
+++ b/java/core/src/java/org/apache/orc/impl/InStream.java
@@ -397,13 +397,14 @@ public abstract class InStream extends InputStream {
     }
   }
 
-  private static class CompressedStream extends InStream {
+  static class CompressedStream extends InStream {
     private final int bufferSize;
     private ByteBuffer uncompressed;
     private final CompressionCodec codec;
     protected ByteBuffer compressed;
     protected DiskRangeList currentRange;
     private boolean isUncompressedOriginal;
+    protected long currentCompressedStart;
 
     /**
      * Create the stream without resetting the input stream.
@@ -471,6 +472,7 @@ public abstract class InStream extends InputStream {
     }
 
     private void readHeader() throws IOException {
+      currentCompressedStart = compressed.position();
       int b0 = readHeaderByte();
       int b1 = readHeaderByte();
       int b2 = readHeaderByte();
@@ -488,11 +490,14 @@ public abstract class InStream extends InputStream {
         isUncompressedOriginal = true;
       } else {
         if (isUncompressedOriginal) {
+          // Since the previous chunk was uncompressed, allocate the buffer and set original false
           allocateForUncompressed(bufferSize, slice.isDirect());
           isUncompressedOriginal = false;
         } else if (uncompressed == null) {
+          // If the buffer was not allocated then allocate the same
           allocateForUncompressed(bufferSize, slice.isDirect());
         } else {
+          // Since the buffer is already allocated just clear the same
           uncompressed.clear();
         }
         codec.decompress(slice, uncompressed);
@@ -551,15 +556,20 @@ public abstract class InStream extends InputStream {
 
     @Override
     public void seek(PositionProvider index) throws IOException {
-      seek(index.getNext());
+      boolean seeked = seek(index.getNext());
       long uncompressedBytes = index.getNext();
-      if (uncompressedBytes != 0) {
-        readHeader();
-        uncompressed.position(uncompressed.position() +
-                              (int) uncompressedBytes);
-      } else if (uncompressed != null) {
-        // mark the uncompressed buffer as done
-        uncompressed.position(uncompressed.limit());
+      if (!seeked && uncompressed != null) {
+        // Only reposition uncompressed
+        uncompressed.position((int) uncompressedBytes);
+      } else {
+        if (uncompressedBytes != 0) {
+          readHeader();
+          uncompressed.position(uncompressed.position() +
+                                (int) uncompressedBytes);
+        } else if (uncompressed != null) {
+          // mark the uncompressed buffer as done
+          uncompressed.position(uncompressed.limit());
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -551,15 +556,20 @@ public void changeIv(Consumer<byte[]> modifier) {
...
-        uncompressed.position(uncompressed.position() +
-                              (int) uncompressedBytes);
-      } else if (uncompressed != null) {
-        // mark the uncompressed buffer as done
-        uncompressed.position(uncompressed.limit());
+      if (!seeked && uncompressed != null) {
+        // Only reposition uncompressed
+        uncompressed.position((int) uncompressedBytes);
+      } else {
+        if (uncompressedBytes != 0) {
```
pgaref, line 565: add some explanation here as well?

java/core/src/test/org/apache/orc/impl/TestInStream.java
```diff
@@ -434,8 +437,30 @@ public void testCompressed() throws Exception {
...
+        assertEquals(c, ((InStream.CompressedStream) in).compressed);
+      } else {
+        seeksPerformed += 1;
+        in.seek(positions[i]);
+        assertNotEquals(c, ((InStream.CompressedStream) in).compressed);
+      }
+      positions[i].reset();
+      assertEquals(i & 0xff, in.read());
+    }
+    assertEquals(1, seeksPerformed);
```
pgaref, line 459: Why are we assuming 1 seek here? do we have 2 chunks total?
Shall we move this to its own test?

pgaref (commented):
Hey @pavibhai thanks for pinging me, changes look good overall, left some comments.
Left me know what you think

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -397,13 +397,14 @@ public String toString() {
     }
   }
 
-  private static class CompressedStream extends InStream {
+  static class CompressedStream extends InStream {
```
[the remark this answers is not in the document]
pavibhai: I agree will make public that is clearer and consistent. Only the EncryptedCompressedStream remains as private and not public after this change.

java/core/src/java/org/apache/orc/impl/InStream.java
pavibhai: Added

java/core/src/test/org/apache/orc/impl/TestInStream.java
pavibhai: Makes sense let me add a separate test for this

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -397,13 +397,14 @@ public String toString() {
...
     private final int bufferSize;
     private ByteBuffer uncompressed;
     private final CompressionCodec codec;
     protected ByteBuffer compressed;
     protected DiskRangeList currentRange;
     private boolean isUncompressedOriginal;
+    protected long currentCompressedStart;
```
omalley, line 407: This should be initialized to -1 so that we don't have someone seek to 0 and get incorrect data.

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -551,15 +556,20 @@ public void changeIv(Consumer<byte[]> modifier) {
...
+      boolean seeked = seek(index.getNext());
       long uncompressedBytes = index.getNext();
-      if (uncompressedBytes != 0) {
-        readHeader();
-        uncompressed.position(uncompressed.position() +
-                              (int) uncompressedBytes);
-      } else if (uncompressed != null) {
-        // mark the uncompressed buffer as done
-        uncompressed.position(uncompressed.limit());
+      if (!seeked && uncompressed != null) {
```
omalley, line 561: Can we get into the case of !seeked && uncompressed == null? I suspect not, but this code would probably be doing the wrong thing by ignoring the case.

java/core/src/test/org/apache/orc/impl/TestInStream.java
```diff
@@ -19,6 +19,9 @@
 package org.apache.orc.impl;
```
omalley, line 19: It would be nice to have a test case for blocks that used the original bytes (header.isOriginal).

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -471,6 +472,7 @@ private int readHeaderByte() {
     }
 
     private void readHeader() throws IOException {
+      currentCompressedStart = compressed.position();
```
omalley, line 475: I think this needs to be this.position.

omalley (changes requested):

pgaref:
Hey @pavibhai can you please check the remaining comments?

java/core/src/test/org/apache/orc/impl/TestInStream.java
pavibhai: Added the test case

java/core/src/java/org/apache/orc/impl/InStream.java
pavibhai: Fixed

java/core/src/java/org/apache/orc/impl/InStream.java
pavibhai: This is a good catch @omalley fixed this and added a test to verify that the seek skip takes place even when the DiskRange does not perfectly align with the stream

java/core/src/java/org/apache/orc/impl/InStream.java
pavibhai: @omalley !seeked would mean that previously a readHeader has taken place which would result in the uncompressed being allocated, unless when someone seeks to -1. I will add an assert that this is unexpected, if you have any other suggestion please let me know.

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -558,9 +558,14 @@ public void changeIv(Consumer<byte[]> modifier) {
...
       boolean seeked = seek(index.getNext());
       long uncompressedBytes = index.getNext();
-      if (!seeked && uncompressed != null) {
-        // Only reposition uncompressed
-        uncompressed.position((int) uncompressedBytes);
+      if (!seeked) {
+        if (uncompressed != null) {
+          // Only reposition uncompressed
+          uncompressed.position((int) uncompressedBytes);
+        } else {
```
omalley, line 565: I'd drop the empty else block, but the comment is good.

pavibhai: Sure, dropped the empty block

omalley (approved):

PR #653, opened 2021-03-12
ORC-759: StructBatchReader should always skip processing on the rootReader

pavibhai:
StructBatchReader assumes rootReader has no nulls and skips rootReader processing in the following methods:
* seek
* skip
* startStripe

This behavior will be consistent with the behavior of nextBatch which already makes this assumption.

Make the behavior consistent on all the methods.

Regression test, none of the existing tests have failed.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/reader/tree/PrimitiveBatchReader.java
+++ b/java/core/src/java/org/apache/orc/impl/reader/tree/PrimitiveBatchReader.java
@@ -15,7 +15,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.apache.orc.impl.reader.tree;
 
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
--- a/java/core/src/java/org/apache/orc/impl/reader/tree/PrimitiveBatchReader.java
+++ b/java/core/src/java/org/apache/orc/impl/reader/tree/PrimitiveBatchReader.java
@@ -15,10 +15,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.apache.orc.impl.reader.tree;
 
-import java.io.IOException;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
+import org.apache.orc.impl.PositionProvider;
+import org.apache.orc.impl.reader.StripePlanner;
+
+import java.io.IOException;
 
 public class PrimitiveBatchReader extends BatchReader {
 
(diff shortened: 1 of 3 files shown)


java/core/src/java/org/apache/orc/impl/reader/tree/StructBatchReader.java
```diff
@@ -20,11 +20,18 @@
...
 import org.apache.orc.impl.TreeReaderFactory;
+import org.apache.orc.impl.reader.StripePlanner;
 
 import java.io.IOException;
 import java.util.Set;
 
+/**
+ * Handles the Struct rootType for batch handling. The handling assumes that the root
+ * {@link org.apache.orc.impl.TreeReaderFactory.StructTreeReader} has no nulls, this is required as
+ * the {@link VectorizedRowBatch} does not represent the root Struct as a vector.
```
pgaref, line 33: Nice, I would rephrase to something like: ..has no nulls. Root Struct vector is not represented as part of the final  the {@link  org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch}.

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -621,9 +631,20 @@ private ByteBuffer slice(int chunkLength) throws IOException {
...
+     * Seek to the desired chunk based on the input position.
+     *
+     * @param desired position in the compressed stream
+     * @return Indicates whether a seek was performed or not
+     * @throws IOException when seeking outside the stream bounds
+     */
+    boolean seek(long desired) throws IOException {
       if (desired == 0 && bytes == null) {
-        return;
+        return false;
```
pgaref, line 643: Lets remove ORC-758 leftovers from the PR

java/core/src/java/org/apache/orc/impl/reader/tree/StructBatchReader.java
```diff
@@ -76,4 +83,31 @@ public void nextBatch(VectorizedRowBatch batch, int batchSize) throws IOExceptio
       }
     }
   }
+
+  @Override
+  public void startStripe(StripePlanner planner) throws IOException {
```
pgaref, line 88: Makes sense, maybe we should make these methods abstract on BatchReader instead?
Then we would only keep setVectorColumnCount and resetBatch methods with implementation that apply both for Primitive and Struct reading.

java/core/src/test/org/apache/orc/impl/TestInStream.java
```diff
@@ -434,8 +437,30 @@ public void testCompressed() throws Exception {
       int x = in.read();
       assertEquals(i & 0xff, x);
     }
+
+    // Forward seek test
```
pgaref, line 441: another ORC-758 leftover

pgaref (commented):
Almost there, left some comments @pavibhai

java/core/src/java/org/apache/orc/impl/reader/tree/StructBatchReader.java
pavibhai: Changed

java/core/src/java/org/apache/orc/impl/reader/tree/StructBatchReader.java
pavibhai: Good idea. Pushed the implementation into PrimitiveBatchReader.

Now only the methods `setVectorColumnCount` and `resetBatch` methods are implemented in BatchReader.

java/core/src/java/org/apache/orc/impl/InStream.java
pavibhai: Removed ORC-758 references

pgaref (approved):
Hey @pavibhai I am +1 on latest changes -- can merge after addressing the minor comment above :)

java/core/src/java/org/apache/orc/impl/reader/tree/PrimitiveBatchReader.java
```diff
@@ -15,10 +15,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
```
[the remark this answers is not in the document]
pavibhai: Sure removed

pgaref:
Thanks @pavibhai ! This is now merged!

dongjoon-hyun:
Thank you, @pavibhai and @pgaref .

PR #657, opened 2021-03-18
ORC-765: Added build option to compile libraries with position independent code

bgivertz:

I have added a BUILD_POSITION_INDEPENDENT_LIB option to the CMake file which is turned OFF by default. However, if a user needs to compile the static libraries with PIC, they are able to specify so by turning the option ON. This will effect liborc and all of its external dependencies. If turned on, the option will set the CMAKE_POSITION_INDEPENDENT_CODE flag.

When building the ORC c++ library, it will compile itself and its third party dependencies as static libraries. However, on linux, it is not possible to link these libraries to a shared object unless they are compiled with position independent code.

It is difficult to add tests in order to ensure that libraries have been compiled with position independent code. However, I have used the feature and confirmed that when the option is turned on, the libraries successfully link with a shared object and when turned off, they do not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,10 @@ option(INSTALL_VENDORED_LIBS
 option(STOP_BUILD_ON_WARNING
     "Fail the build on C++ warnings"
     ON)
+  
+option(BUILD_POSITION_INDEPENDENT_LIB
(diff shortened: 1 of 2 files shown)


omalley (approved):
LGTM +1

PR #661, opened 2021-03-22
ORC-763: [C++] Fix ORC timestamp inconsistencies with Java ORC

wgtmac:
The Java ORC reader takes one second off from timestamp values with negative second and nanosecond greater than 999999. The C++ ORC reader/writer should use same logic to be consistent.

WIthout this fix, we may see data inconsistency of timestamp values from different ORC readers.

It can be tested manually by C++/Java ORC tools.

### DIFF
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -374,7 +374,7 @@ namespace orc {
         }
         int64_t writerTime = secsBuffer[i] + epochOffset;
         secsBuffer[i] = writerTimezone.convertToUTC(writerTime);
-        if (secsBuffer[i] < 0 && nanoBuffer[i] != 0) {
+        if (secsBuffer[i] < 0 && nanoBuffer[i] > 999999) {
           secsBuffer[i] -= 1;
         }
       }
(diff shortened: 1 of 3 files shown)


pgaref (commented):
Thanks for this @wgtmac  ! 
Looks to be consistent with https://github.com/apache/orc/blob/master/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java#L1227-L1229

Shall we also add a test with the TS around 1970 as per ticket to check behaviour?

dongjoon-hyun (approved):
Thank you, @wgtmac !
+1, LGTM (and +1 for @pgaref 's comment)

dongjoon-hyun:
Thank you for adding the test case.

pgaref (approved):

dongjoon-hyun:
Thank you, @dongjoon-hyun and @pgaref .
Merged to master.

dongjoon-hyun:
This is backported to branch-1.6, too.

pgaref:
Thanks for the fix @wgtmac ! 
I was thinking about the backport too, thanks @dongjoon-hyun !

PR #668, opened 2021-03-24
ORC-742: Added Lazy IO for non filter columns

pavibhai:
    * Identify columns in the presence of a filter into LEAD and FOLLOW columns
            * LEAD columns are read first
            * FOLLOW columns are read only if the filter selects an output
    * RecordReaderImpl.nextBatch performs read until a batch has value or the file is exhaused instead of returning empty batches as was the case previously
    * IO of FOLLOW columns happens at the level of RowGroup
    * In the presence of filters batches respected row group boundaries

The code changes allow for a lazy evaluation of FOLLOW columns, which in the case of reads with minimal hits gives substantial savings both of IO and CPU.

This patch includes Unit tests that verify the IO savings accomplished as a result of this change.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -100,7 +100,7 @@ public class RecordReaderImpl implements RecordReader {
   private final DataReader dataReader;
   private final int maxDiskRangeChunkLimit;
   private final StripePlanner planner;
-  // identifies the type of read, ALL(read everything), LEADS(read only the filter columns)
+  // identifies the type of read, ALL(read everything), LEADERS(read only the filter columns)
   private final TypeReader.ReadPhase startReadPhase;
   // identifies that follow columns bytes must be read
   private boolean needsFollowColumnsRead;
@@ -131,7 +131,17 @@ public class RecordReaderImpl implements RecordReader {
     }
   }
 
-  static TypeDescription findCommonColumnType(SchemaEvolution evolution, String columnName) {
+  /**
+   * Given a column name such as 'a.b.c', this method returns the column 'a.b.c' if present in the
+   * file. In case 'a.b.c' is not found in file then it tries to look for 'a.b', then 'a'. If none
+   * are present then it shall return null.
+   *
+   * @param evolution the mapping from reader to file schema
+   * @param columnName the fully qualified column name to look for
+   * @return the file column type or null in case none of the branch columns are present in the file
+   * @throws IllegalArgumentException if the column was not found in the reader schema
+   */
+  static TypeDescription findMostCommonColumn(SchemaEvolution evolution, String columnName) {
     try {
       TypeDescription readerColumn = evolution.getReaderBaseSchema().findSubtype(
         columnName, evolution.isSchemaEvolutionCaseAware);
@@ -273,7 +283,7 @@ public class RecordReaderImpl implements RecordReader {
           // Add -1 to filter columns so that the NullTreeReader is invoked during the LEADERS phase
           filterColIds.add(-1);
           // Determine the common parent and include these
-          expandCol = findCommonColumnType(evolution, colName);
+          expandCol = findMostCommonColumn(evolution, colName);
         }
         while(expandCol != null && expandCol.getId() != -1) {
           // classify the column and the parent branch as LEAD
@@ -1317,21 +1327,23 @@ public class RecordReaderImpl implements RecordReader {
   }
 
   /**
-   * This method plans the follow columns to position them based on the row idx determined by
-   * filtering the lead columns
+   * This method prepares the non-filter column readers for next batch. This involves the following
+   * 1. Determine position
+   * 2. Perform IO if required
+   * 3. Position the non-filter readers
    *
-   * Given that this is just repositioning of the follow columns, this method shall never have to
+   * This method is repositioning the non-filter columns and as such this method shall never have to
    * deal with navigating the stripe forward or skipping row groups, all of this should have already
-   * taken place based on the leading columns.
+   * taken place based on the filter columns.
    * @param toFollowRow The rowIdx identifies the required row position within the stripe for
    *                    follow read
    * @param fromFollowRow Indicates the current position of the follow read, exclusive
-   * @return the read level for reading followers, this will be FOLLOWERS_AND_PARENTS in case of a
-   * seek otherwise will be FOLLOWERS
+   * @return the read phase for reading non-filter columns, this shall be FOLLOWERS_AND_PARENTS in
+   * case of a seek otherwise will be FOLLOWERS
    */
   private TypeReader.ReadPhase prepareFollowReaders(long toFollowRow, long fromFollowRow)
     throws IOException {
-    // 1. Determine the row group that this matches in the stripe
+    // 1. Determine the required row group and skip rows needed from the RG start
     int needRG = computeRGIdx(toFollowRow);
     // The current row is not yet read so we -1 to compute the previously read row group
     int readRG = computeRGIdx(fromFollowRow - 1);
@@ -1346,38 +1358,40 @@ public class RecordReaderImpl implements RecordReader {
       skipRows = toFollowRow - (needRG * rowIndexStride);
     }
 
-    // 2. Plan the row group idx for the following columns, only if the rowgroup has changed,
-    // otherwise we have already read the required bytes
+    // 2. Plan the row group idx for the non-filter columns if this has not already taken place
     if (needsFollowColumnsRead) {
       needsFollowColumnsRead = false;
       planner.readFollowData(indexes, includedRowGroups, needRG, false);
       reader.startStripe(planner, TypeReader.ReadPhase.FOLLOWERS);
     }
 
-    // 3. Seek to the required row group and skip the rows as needed
+    // 3. Position the non-filter readers to the required RG and skipRows
     TypeReader.ReadPhase result = TypeReader.ReadPhase.FOLLOWERS;
     if (needRG != readRG || toFollowRow < fromFollowRow) {
-      // seek both the lead parents and follow. This will position the parents present vector to
-      // facilitate the subsequent skip on the leader readers which will determine the number of
-      // non null values to skip
+      // When having to change a row group or in case of back navigation, seek both the filter
+      // parents and non-filter. This will re-position the parents present vector. This is needed
+      // to determine the number of non-null values to skip on the non-filter columns.
       seekToRowEntry(reader, needRG, TypeReader.ReadPhase.FOLLOWERS_AND_PARENTS);
+      // skip rows on both the filter parents and non-filter as both have been positioned in the
+      // previous step
       reader.skipRows(skipRows, TypeReader.ReadPhase.FOLLOWERS_AND_PARENTS);
       result = TypeReader.ReadPhase.FOLLOWERS_AND_PARENTS;
     } else if (skipRows > 0) {
-      // in case we are only skipping within the row group, reposition the lead parents back to the
-      // position of the follow and move both the lead parents and follow forward, this will compute
-      // the correct non null skips on follow children
+      // in case we are only skipping within the row group, position the filter parents back to the
+      // position of the follow. This is required to determine the non-null values to skip on the
+      // non-filter columns.
       seekToRowEntry(reader, readRG, TypeReader.ReadPhase.LEADER_PARENTS);
       reader.skipRows(fromFollowRow - (readRG * rowIndexStride), TypeReader.ReadPhase.LEADER_PARENTS);
+      // Move both the filter parents and non-filter forward, this will compute the correct
+      // non-null skips on follow children
       reader.skipRows(skipRows, TypeReader.ReadPhase.FOLLOWERS_AND_PARENTS);
       result = TypeReader.ReadPhase.FOLLOWERS_AND_PARENTS;
     }
     // Identifies the read level that should be performed for the read
-    // FOLLOWERS_WITH_PARENTS indicates repositioning identifying both FOLLOW children and parents
-    // in scope
-    // FOLLOWERS indicates read only of the FOLLOW level without the parents, which is used during
-    // contiguous read as the parent information is captured into the RowBatch that is used as part
-    // of TreeReader.nextVector
+    // FOLLOWERS_WITH_PARENTS indicates repositioning identifying both non-filter and filter parents
+    // FOLLOWERS indicates read only of the non-filter level without the parents, which is used during
+    // contiguous read. During a contiguous read no skips are needed and the non-null information of
+    // the parent is available in the column vector for use during non-filter read
     return result;
   }
 
@@ -1518,6 +1532,7 @@ public class RecordReaderImpl implements RecordReader {
       readStripe();
     }
     if (rowIndexColsToRead == null) {
+      // Read the row indexes only if they were not already read as part of readStripe()
       readCurrentStripeRowIndex();
     }
 
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -153,7 +153,7 @@ public class RecordReaderImpl implements RecordReader {
         } else {
           return fileColumn;
         }
-      } while(readerColumn != null);
+      } while (readerColumn != null);
       return null;
     } catch (IllegalArgumentException e) {
       throw new IllegalArgumentException("Filter could not find column with name: " +
@@ -285,7 +285,7 @@ public class RecordReaderImpl implements RecordReader {
           // Determine the common parent and include these
           expandCol = findMostCommonColumn(evolution, colName);
         }
-        while(expandCol != null && expandCol.getId() != -1) {
+        while (expandCol != null && expandCol.getId() != -1) {
           // classify the column and the parent branch as LEAD
           filterColIds.add(expandCol.getId());
           rowIndexCols[expandCol.getId()] = true;
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -131,6 +131,27 @@ public class RecordReaderImpl implements RecordReader {
     }
   }
 
+  static TypeDescription findCommonColumnType(SchemaEvolution evolution, String columnName) {
+    try {
+      TypeDescription readerColumn = evolution.getReaderBaseSchema().findSubtype(
+        columnName, evolution.isSchemaEvolutionCaseAware);
+      TypeDescription fileColumn;
+      do {
+        fileColumn = evolution.getFileType(readerColumn);
+        if (fileColumn == null) {
+          readerColumn = readerColumn.getParent();
+        } else {
+          return fileColumn;
+        }
+      } while(readerColumn != null);
+      return null;
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException("Filter could not find column with name: " +
+                                         columnName + " on " + evolution.getReaderBaseSchema(),
+                                         e);
+    }
+  }
+
   /**
    * Find the mapping from predicate leaves to columns.
    * @param sargLeaves the search argument that we need to map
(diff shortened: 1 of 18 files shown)


pavibhai:
@pgaref @omalley 
I have not broken this PR down any further so as to keep a merge functionally complete as compared to just structural changes of adding a read level for example.

If it helps with the review I am happy to break the PR into multiple commits. Please let me know.

pavibhai:
@pgaref Following up on this to bring to your attention. If you get sometime please go through the same and provide your feedback.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -230,15 +247,22 @@ protected RecordReaderImpl(ReaderImpl fileReader,
     SortedSet<Integer> filterColIds = new TreeSet<>();
     if (options.getPreFilterColumnNames() != null) {
       for (String colName : options.getPreFilterColumnNames()) {
-        int expandColId = findColumns(evolution, colName);
+        TypeDescription expandCol = findColumnType(evolution, colName);
         // If the column is not present in the file then this can be ignored from read.
-        if (expandColId != -1) {
-          filterColIds.add(expandColId);
+        while(expandCol != null && expandCol.getId() != -1) {
```
pgaref, line 252: missing space

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -958,19 +981,21 @@ public SargApplier(SearchArgument sarg,
       filterColumns = mapSargColumnsToOrcInternalColIdx(sargLeaves,
                                                         evolution);
       this.rowIndexStride = rowIndexStride;
+      this.evolution = evolution;
+      exceptionCount = new long[sargLeaves.size()];
+      this.useUTCTimestamp = useUTCTimestamp;
+    }
+
+    public void setRowIndexCols(boolean[] rowIndexCols) {
```
pgaref, line 989: Is this change related? Shall we move it a separate ticket if thats the case?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -1302,25 +1427,33 @@ private int findStripe(long rowNumber) {
...
         Arrays.fill(included, true);
       }
       copy.parseStripe(stripes.get(stripeIndex), included);
-      return copy.readRowIndex(sargColumns, null);
+      return copy.readRowIndex(readCols, null);
     }
   }
 
-  private void seekToRowEntry(BatchReader reader, int rowEntry)
+  private void seekToRowEntry(BatchReader reader, int rowEntry, EnumSet<TypeReader.ReadLevel> readLevel)
```
pgaref, line 1456: We already have readLevel as part of our RecordReaderIml instance, any reason we also pass it as param here?

java/mapreduce/src/java/org/apache/orc/mapreduce/OrcMapreduceRecordReader.java
```diff
@@ -96,16 +99,22 @@ public boolean nextKeyValue() throws IOException, InterruptedException {
...
     }
+    int rowIdx = batch.selectedInUse ? batch.selected[rowInBatch] : rowInBatch;
     if (schema.getCategory() == TypeDescription.Category.STRUCT) {
       OrcStruct result = (OrcStruct) row;
       List<TypeDescription> children = schema.getChildren();
       int numberOfChildren = children.size();
       for(int i=0; i < numberOfChildren; ++i) {
-        result.setFieldValue(i, OrcMapredRecordReader.nextValue(batch.cols[i], rowInBatch,
-            children.get(i), result.getFieldValue(i)));
+        TypeDescription child = children.get(i);
```
pgaref, line 108: unrelated change?

java/mapreduce/src/java/org/apache/orc/mapred/OrcMapredRecordReader.java
```diff
@@ -98,16 +101,22 @@ public boolean next(NullWritable key, V value) throws IOException {
...
     }
+    int rowIdx = batch.selectedInUse ? batch.selected[rowInBatch] : rowInBatch;
     if (schema.getCategory() == TypeDescription.Category.STRUCT) {
       OrcStruct result = (OrcStruct) value;
       List<TypeDescription> children = schema.getChildren();
       int numberOfChildren = children.size();
       for(int i=0; i < numberOfChildren; ++i) {
-        result.setFieldValue(i, nextValue(batch.cols[i], rowInBatch,
-            children.get(i), result.getFieldValue(i)));
+        TypeDescription child = children.get(i);
```
pgaref, line 110: unrelated change?

java/core/src/test/org/apache/orc/impl/TestSchemaEvolution.java
```diff
@@ -1732,10 +1732,10 @@ public void testTypeConversion() throws IOException {
         dataReader, OrcFile.WriterVersion.ORC_14, true, Integer.MAX_VALUE);
     boolean[] columns = new boolean[]{true, true, true};
     planner.parseStripe(dataReader.getStripe(0), columns)
-        .readData(null, null, false);
-    reader.startStripe(planner);
+        .readData(null, null, false, TypeReader.ReadLevel.ALL);
+    reader.startStripe(planner, TypeReader.ReadLevel.ALL);
     VectorizedRowBatch batch = readType.createRowBatch();
-    reader.nextBatch(batch, 10);
+    reader.nextBatch(batch, 10, TypeReader.ReadLevel.ALL);
```
pgaref, line 1738: Use overloaded method for all and create a separate test for partial?

java/core/src/java/org/apache/orc/impl/reader/StripePlanner.java
```diff
@@ -144,10 +165,23 @@ public StripePlanner parseStripe(StripeInformation stripe,
    * @return the buffers that were read
    */
   public BufferChunkList readData(OrcIndex index,
-                                boolean[] rowGroupInclude,
-                                boolean forceDirect) throws IOException {
+                                  boolean[] rowGroupInclude,
```
pgaref, line 168: Update doc here? how readLevel is changing the behaviour of the method?

java/core/src/java/org/apache/orc/impl/reader/tree/TypeReader.java
```diff
@@ -24,22 +25,51 @@
...
+   * @param readLevel the requested read level
+   * @return true if allowed by read level or if it is a LEAD_PARENT otherwise false
+   */
+  static boolean allowChild(TypeReader reader, EnumSet<ReadLevel> readLevel) {
+    return readLevel.contains(reader.getReadLevel())
+           || reader.getReadLevel() == ReadLevel.LEAD_PARENT;
+  }
+
+  enum ReadLevel {
+    LEAD_CHILD,    // Read only the elementary filter columns
```
pgaref, line 65: Do we really need to distinguish between LEAD_CHILD and LEAD_PARENT? 
Why not just LEAD and FOLLOW?

pgaref (changes requested):
Hey @pavibhai thanks for pinging me! 
This is much cleaner than before but still hard to parse as it contains many critical changes.

Left some preliminary comments but my general advice would be to have one final split of this where we just introduce the enums and prepare the Reader for the actual planning/reading changes -- for instance I am sure we can cut down on the places where we pass ReadLevel around, and introduce some overloaded method where necessary.

In the final PR we will then only need to introduce the logic of partialPlan and follow Data Read.

Please let me know what you think

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pavibhai: will add

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pavibhai: This is still needed as we do two types of seeks within RecordReader one for just the filter columns and the other that performs only the non-filter columns when the RecordReader readLevel says filter columns only.

java/mapreduce/src/java/org/apache/orc/mapreduce/OrcMapreduceRecordReader.java
pavibhai: Sorry, yes this should have been a separate bug fix patch that MapRed reader does not respect inclusions when setting the values.

Will pull this out.

java/mapreduce/src/java/org/apache/orc/mapred/OrcMapredRecordReader.java
pavibhai: Sorry, yes this should have been a separate bug fix patch that MapRed reader does not respect inclusions when setting the values.

Will pull this out.

java/core/src/test/org/apache/orc/impl/TestSchemaEvolution.java
pavibhai: To be clear you are asking can we add overloaded methods on the reader that if the level is not passed it will be treated as ALL. Please confirm.

I think that method will only be used from test, in all other cases the reader level is always supplied by the RecordReader.

java/core/src/java/org/apache/orc/impl/reader/StripePlanner.java
pavibhai: Will add

java/core/src/java/org/apache/orc/impl/reader/tree/TypeReader.java
pavibhai: The LEAD_PARENT has significance that it should still propagate activities on the child even if it does not satisfy the requested read level. That is why I chose to differentiate between PARENT and CHILD.

pavibhai:
> Left some preliminary comments but my general advice would be to have one final split of this where we just introduce the enums and prepare the Readers for the actual planning/reading changes -- for instance I am sure we can cut down on the places where we pass ReadLevel around, and introduce some overloaded method where necessary.
> 
> Then, in the final PR we will then only need to introduce the logic of partialPlan and follow Data Read which will be much cleaner.
> 
> Please let me know what you think :)

@pgaref Thanks for your feedback. Instead of separating PRs as the PR with just the level changes as it does not add any functionality, maybe alternatively shall I break the PR into multiple commits so as to differentiate these changes:
1. Introduce ReadLevel and add the parameters to the methods
2. Introduce the LazyIO based on the read level

Let me know what you think.

java/core/src/test/org/apache/orc/impl/TestSchemaEvolution.java
pgaref: If its only used by the test then lets leave it as

java/core/src/java/org/apache/orc/impl/reader/tree/TypeReader.java
pgaref: Not sure I follow, in general every column is a child of the root struct in ORc right?
 Can you share an example please where we need the distinction?

pgaref:
> > Left some preliminary comments but my general advice would be to have one final split of this where we just introduce the enums and prepare the Readers for the actual planning/reading changes -- for instance I am sure we can cut down on the places where we pass ReadLevel around, and introduce some overloaded method where necessary.
> > Then, in the final PR we will then only need to introduce the logic of partialPlan and follow Data Read which will be much cleaner.
> > Please let me know what you think :)
> 
> @pgaref Thanks for your feedback. Instead of separating PRs as the PR with just the level changes does not add any functionality, maybe alternatively shall I break the PR into multiple commits so as to differentiate these changes:
> 
> 1. Introduce ReadLevel and add the parameters to the methods
> 2. Introduce the LazyIO based on the read level
> 
> Let me know what you think.

Hey @pavibhai that could also help, see what works better for you :)

java/core/src/java/org/apache/orc/impl/reader/tree/TypeReader.java
pavibhai: Sure.

Consider the following schema and classification
* f1: Long
* f2: String
* f3: Struct
  * a: String
  * b: Long

In the above schema if the filter was based on f2 and f3.a then the classification will be as follows:
* f1: Long: FOLLOW
* f2: String: LEAD_CHILD
* f3: Struct: LEAD_PARENT
  * a: String: LEAD_CHILD
  * b: Long: FOLLOW

The difference of LEAD_PARENT comes into play from the standpoint of seeks. Where a seek for follow involves the seek of the LEAD_PARENTs as well to determine the non null rows in the reader.

This is documented in [prepareFollowReaders](https://github.com/apache/orc/blob/cc48b40e245f717ffc79c170f94f750da4fccf32/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java#L1305)

This will come into play when let us say the lead read has already happened to say Row 8123 while the previous follow read has happened to say 1000. Now we want to skip to rows 8123 on the follow readers, this would require the determination of number of non null values on the parent, in our example for f3.b.

This is achieved in the following steps:
1. Seek LEAD_PARENT to follow position i.e. 1000
2. Seek LEAD_PARENT + FOLLOW to desired position 8123, this will ensure that non nulls is determined correctly and the follows are seeked.

LEAD_CHILD are unaffected and remain at 1000 through the above process, at the end all of them are on 1000. Hope this is clear. If this is better handled via a call, I am open for it.

pgaref:
Hey @pavibhai  thanks updating this! Any remaining comments to address? Is this ready for a final review?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pavibhai: It is related. I see that the rowIndexCols are not used within the SArgApplier. Additionally now the RowIndexCols is decided both by the SArgs and the Filters so I wanted to pull this out of the SArgApplier into RecordReaderImpl where it is used.

pavibhai:
> Hey @pavibhai thanks updating this! Any remaining comments to address? Is this ready for a final review?

@pgaref thanks for following up. I think most of the changes are complete. The key item that I have not yet done is splitting the PR. 

I have had some discussions with @omalley and based on that the key change you will see in the PR right now is the separation of ReadLevel into ReadCategory and ReadPhase.

Hope that makes sense.

java/core/src/java/org/apache/orc/impl/reader/tree/TypeReader.java
pgaref: Thanks for the detailed explanation!

java/core/src/java/org/apache/orc/impl/reader/tree/TypeReader.java
```diff
@@ -24,22 +25,70 @@
...
+  void skipRows(long rows, ReadPhase readPhase) throws IOException;
 
   void nextVector(ColumnVector previous,
                   boolean[] isNull,
                   int batchSize,
-                  FilterContext filterContext) throws IOException;
+                  FilterContext filterContext,
+                  ReadPhase readPhase) throws IOException;
 
   int getColumnId();
```
pgaref, line 47: not sure we need getColumnId() anymore?

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -189,12 +192,34 @@ public boolean useProlepticGregorian() {
     public boolean fileUsedProlepticGregorian() {
       return fileUsedProlepticGregorian;
     }
+
+    @Override
+    public TypeReader.ReaderCategory getReaderCategory(int columnId) {
+      TypeReader.ReaderCategory result;
+      if (getColumnFilterCallback() == null || getColumnFilterIds().contains(columnId)) {
```
pgaref, line 199: Not sure why consider getColumnFilterCallback null case here? 
When there is not filter set we consider all ReaderCategories to be FILTER_PARENT or FILTER_CHILD?

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -2856,11 +2962,16 @@ public static TypeReader createTreeReader(TypeDescription readerType,
     }
   }
 
+  public static boolean isDecimalAsLong(OrcFile.Version version, int precision) {
```
pgaref, line 2965: This could be part of another PR tbh

java/core/src/java/org/apache/orc/impl/reader/StripePlanner.java
```diff
@@ -445,55 +509,83 @@ private static boolean hasSomeRowGroups(boolean[] includedRowGroups) {
...
     if (hasSomeRowGroups(includedRowGroups)) {
       InStream.StreamOptions compression = dataReader.getCompressionOptions();
       boolean isCompressed = compression.getCodec() != null;
       int bufferSize = compression.getBufferSize();
       OrcProto.RowIndex[] rowIndex = index.getRowGroupIndex();
       for (StreamInformation stream : dataStreams) {
-        if (RecordReaderUtils.isDictionary(stream.kind, encodings[stream.column])) {
-          addChunk(result, stream, stream.offset, stream.length);
+        if (readPhase == TypeReader.ReadPhase.ALL
+            ||(readPhase == TypeReader.ReadPhase.LEADERS
```
pgaref, line 523: nit. space after || 
nit. reduce new lines

java/core/src/java/org/apache/orc/impl/reader/tree/TypeReader.java
pavibhai: Sure. This was not added as part of this PR. I can raise a separate PR to remove this or include the removal here.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
pavibhai: Good catch. I missed cleaning up this condition, I already inverted in the previous commit that in the absence of filters we treat every column as NON_FILTER. Am removing `getColumnFilterCallback() == null` condition.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
pavibhai: I need this for the next PR conversion of SArg to Filter. I can remove this here and include it as part of that PR

pgaref: Sounds good! The SArgs PR will be smaller anyway so better to add there

java/core/src/java/org/apache/orc/impl/reader/tree/TypeReader.java
pgaref: Lets do that as part of the follow up PR -- this part of the code is not released yet (only part of main branch) so we have more flexibility

java/core/src/java/org/apache/orc/impl/reader/StripePlanner.java
pavibhai: > nit. space after ||
done

> nit. lets keep conditions on a single line
brought the subcondition into a single line not the entire condition as it extends the line boundary significantly

> nit. reduce new lines (e.g., processStream call)
Fixed the processStream call however I am unclear on how to handle parameters. I have generally been following chompdown if long, I see both variants in the code chomp or just wrap.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
pavibhai: reverted this change

java/core/src/java/org/apache/orc/impl/reader/StripePlanner.java
pgaref: Looks good now, thanks!

pgaref (approved):
Thanks for addressing all the comments @pavibhai ! 
+1 on the latest changes pending tests

pavibhai:
> This is now merged to main! Thanks @pavibhai for all the hard work!

Thanks @pgaref

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -23,6 +23,7 @@
 import java.time.ZoneOffset;
 import java.time.chrono.ChronoLocalDate;
 import java.time.format.DateTimeFormatter;
+import org.apache.commons.lang.ArrayUtils;
```
dongjoon-hyun, line 26: Although ORC-734 removed `commons-lang` usage from Apache ORC in favor of `lang3`, this seems to bring it back. I'll make a PR to remove it.

dongjoon-hyun: Here is the PR.
- https://github.com/apache/orc/pull/811

PR #669, opened 2021-03-24
ORC-773: BinaryColumnWriter does not update Bloom filter

noirello:
Add missing Bloom filter update, when new item is added to a binary column.

Even when the Bloom filter is enabled for a binary column, the filter won't be updated. 

Copied the logic from the string column in testBloomFilter.

### DIFF
--- a/c++/src/ColumnWriter.cc
+++ b/c++/src/ColumnWriter.cc
@@ -1683,6 +1683,9 @@ namespace orc {
       if (!notNull || notNull[i]) {
         directDataStream->write(data[i], unsignedLength);
 
+        if (enableBloomFilter) {
+          bloomFilter->addBytes(data[i], length[i]);
+        }
         binStats->update(unsignedLength);
         ++count;
       }
--- a/c++/test/TestWriter.cc
+++ b/c++/test/TestWriter.cc
@@ -1666,13 +1666,13 @@ namespace orc {
       .setMemoryPool(getDefaultPool())
       .setRowIndexStride(10000)
       .setFileVersion(fileVersion)
-      .setColumnsUseBloomFilter({1, 2});
+      .setColumnsUseBloomFilter({1, 2, 3});
 
     // write 65535 rows of data
     MemoryOutputStream memStream(DEFAULT_MEM_STREAM_SIZE);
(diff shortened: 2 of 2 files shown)


c++/src/ColumnWriter.cc
```diff
@@ -1683,6 +1683,9 @@ namespace orc {
       if (!notNull || notNull[i]) {
         directDataStream->write(data[i], unsignedLength);
 
+        if (enableBloomFilter) {
```
pgaref, line 1686: +1 This looks to be on par with the ColumnWriters above

pgaref (commented):
+1 from my side, @wgtmac please let me know what you think!

wgtmac:
Thanks for pinning me @pgaref !   The change is pretty clean. Thanks @noirello for contributing!

I haven't implemented this in the C++ BinaryColumnWriter because the created BloomFilter is not used on the reader side: https://github.com/apache/orc/blob/master/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java#L741 . I am not sure if there is any valid case where literals of binary type appear in the predicate. If that is valid, we should also include the fix on the reader side.

noirello:
Oh I see. I didn't check the predicate implementation. I just only saw that the [Java counterpart](https://github.com/apache/orc/blob/master/java/core/src/java/org/apache/orc/impl/writer/BinaryTreeWriter.java#L95) updates the filter, and I thought it could be problematic initialising but not updating it on the C++ side.

But now I checked Trino as another reference, where it seems neither the reader expects Bloom filter for a binary column (just like the implementations here), nor the writer creates Bloom filter for binary statistics.

I guess there aren't many valid use cases for it. Should I close this with the related issue?

wgtmac (approved):
@noirello Thanks for confirming! I am OK to accept this PR to make it consistent with the Java side and it is pretty neat and clean. @pgaref  @dongjoon-hyun What do you think?

dongjoon-hyun (approved):
Thank you for pinging me, @wgtmac .
+1, LGTM, too.

dongjoon-hyun:
Thank you for your contribution, @noirello .

PR #674, opened 2021-03-29
ORC-777: Make the vectorized row batch size configurable in MR record readers and writers

sv2000:

The changes in this PR create overloaded constructors in MapreduceRecordReader/Writer classes that accept the rowbatch size as an argument. In order to use the overloaded constructor, the caller is expected to override the getRecordWriter() method (or getRecordReader()), as exemplified in the TestMapreduceOrcOutputFormat class. 

The change is needed to provide greater configurability of the ORC MR readers/writers.
Current behavior of OrcInput(Output)Format is to create an ORC reader/writer with the default row batch size of 1024. This may need to be overridden in certain scenarios to prevent OOMs during reads/writes.

Added unit tests in TestOrcOutputFormat and TestMapreduceOrcOutputFormat, where we extend the OrcOutputFormat class to override the getRecordWriter() method. The change to override the row batch size config was also tested by running a MR job that used the snapshot version of the orc-mapreduce library.

### DIFF
--- a/java/mapreduce/src/java/org/apache/orc/mapred/OrcMapredRecordReader.java
+++ b/java/mapreduce/src/java/org/apache/orc/mapred/OrcMapredRecordReader.java
@@ -69,12 +69,13 @@ public class OrcMapredRecordReader<V extends WritableComparable>
   }
 
   protected OrcMapredRecordReader(Reader fileReader,
-      Reader.Options options) throws IOException {
+                                  Reader.Options options) throws IOException {
     this(fileReader, options, VectorizedRowBatch.DEFAULT_SIZE);
   }
 
   protected OrcMapredRecordReader(Reader fileReader,
-                                  Reader.Options options, int rowBatchSize) throws IOException {
+                                  Reader.Options options,
+                                  int rowBatchSize) throws IOException {
     this.batchReader = fileReader.rows(options);
     if (options.getSchema() == null) {
       schema = fileReader.getSchema();
--- a/java/mapreduce/src/java/org/apache/orc/mapred/OrcMapredRecordReader.java
+++ b/java/mapreduce/src/java/org/apache/orc/mapred/OrcMapredRecordReader.java
@@ -45,6 +45,7 @@ import org.apache.hadoop.io.NullWritable;
 import org.apache.hadoop.io.ShortWritable;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.io.WritableComparable;
+import org.apache.orc.OrcConf;
 import org.apache.orc.Reader;
 import org.apache.orc.RecordReader;
 import org.apache.orc.TypeDescription;
@@ -69,14 +70,19 @@ public class OrcMapredRecordReader<V extends WritableComparable>
   }
 
   protected OrcMapredRecordReader(Reader fileReader,
-                                  Reader.Options options) throws IOException {
+      Reader.Options options) throws IOException {
+    this(fileReader, options, (int) OrcConf.ROW_BATCH_SIZE.getDefaultValue());
+  }
+
+  protected OrcMapredRecordReader(Reader fileReader,
+                                  Reader.Options options, int rowBatchSize) throws IOException {
     this.batchReader = fileReader.rows(options);
     if (options.getSchema() == null) {
       schema = fileReader.getSchema();
     } else {
       schema = options.getSchema();
     }
-    this.batch = schema.createRowBatch();
+    this.batch = schema.createRowBatch(rowBatchSize);
     rowInBatch = 0;
   }
 
--- a/java/mapreduce/src/java/org/apache/orc/mapred/OrcMapredRecordReader.java
+++ b/java/mapreduce/src/java/org/apache/orc/mapred/OrcMapredRecordReader.java
@@ -45,7 +45,6 @@ import org.apache.hadoop.io.NullWritable;
 import org.apache.hadoop.io.ShortWritable;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.io.WritableComparable;
-import org.apache.orc.OrcConf;
 import org.apache.orc.Reader;
 import org.apache.orc.RecordReader;
 import org.apache.orc.TypeDescription;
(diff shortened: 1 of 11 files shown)


java/mapreduce/src/java/org/apache/orc/mapreduce/OrcMapreduceRecordReader.java
```diff
@@ -53,16 +53,20 @@ public OrcMapreduceRecordReader(RecordReader reader,
     rowInBatch = 0;
     this.row = (V) OrcStruct.createValue(schema);
   }
+  public OrcMapreduceRecordReader(Reader fileReader,
+      Reader.Options options) throws IOException {
+    this(fileReader, options, VectorizedRowBatch.DEFAULT_SIZE);
```
dongjoon-hyun, line 58: Is there a reason why we use `DEFAULT_SIZE`, a fixed value?

dongjoon-hyun (commented):
Thank you for making a contribution.

sv2000:
@dongjoon-hyun Thanks for your comment! With regard to "Is there a reason why we use DEFAULT_SIZE, a fixed value?", I use the default value to maintain backward compatibility when orc.row.batch.size is not configured.

pgaref (commented):
Not sure if an extra conf is the right way to go here.

- ```orc.row.batch.size``` is pretty generic even though it is only used for MR
-  So far it was the callers responsibility to set the appropriate batch size so I am wondering if the same should happen for MR -- using an overloaded constructor or even JobConf?

sv2000:
@pgaref IIUC, what you are proposing is to have an overloaded constructor in OrcMapreduceRecordReader/Writer that accepts rowBatchSize as an argument (like we do in this PR), but leave it to the caller to override the getRecordWriter() method, which in turn calls the overloaded constructor. This way we don't need to introduce a separate config. LMK if I am missing anything.

pgaref:
> @pgaref IIUC, what you are proposing is to have an overloaded constructor in OrcMapreduceRecordReader/Writer that accepts rowBatchSize as an argument (like we do in this PR), but leave it to the caller to override the getRecordWriter() method, which in turn calls the overloaded constructor. This way we don't need to introduce a separate config. LMK if I am missing anything.

Hey @sv2000  -- yes you are on point. Just want to avoid having an extra configuration here as it could confuse users. 
If we can introduce a new constructor for Reader/Writer and release the change (on 1.6 or 1.7) it could sort out the issue.
Thoughts?

sv2000:
> > @pgaref IIUC, what you are proposing is to have an overloaded constructor in OrcMapreduceRecordReader/Writer that accepts rowBatchSize as an argument (like we do in this PR), but leave it to the caller to override the getRecordWriter() method, which in turn calls the overloaded constructor. This way we don't need to introduce a separate config. LMK if I am missing anything.
> 
> Hey @sv2000 -- yes you are on point. Just want to avoid having an extra configuration here as it could confuse users.
> If we can introduce a new constructor for Reader/Writer and release the change (on 1.6 or 1.7) it could sort out the issue.
> Thoughts?

@pgaref I am ok with the proposed change. One potential downside though would be that every user that needs to configure the row batch size will have to override the getRecordWriter() method. But may be worth the cost to pay to avoid the confusion of introducing a new config.

dongjoon-hyun:
If we reach an agreement, shall we move forward in that way, @sv2000 ?

sv2000:
@pgaref @dongjoon-hyun Addressed comments on this PR. Can you help take a look?

java/mapreduce/src/java/org/apache/orc/mapred/OrcMapredRecordReader.java
```diff
@@ -69,14 +69,19 @@ public OrcMapredRecordReader(RecordReader reader,
   }
 
   protected OrcMapredRecordReader(Reader fileReader,
-                                  Reader.Options options) throws IOException {
+      Reader.Options options) throws IOException {
```
dongjoon-hyun, line 72: Could you keep the original style here? This looks inconsistent with the other existing one at line 64 and with your new code at line 77.

java/mapreduce/src/java/org/apache/orc/mapreduce/OrcMapreduceRecordReader.java
```diff
@@ -53,16 +53,20 @@ public OrcMapreduceRecordReader(RecordReader reader,
     rowInBatch = 0;
     this.row = (V) OrcStruct.createValue(schema);
   }
```
dongjoon-hyun, line 55: Please add one empty line between constructors.

java/mapreduce/src/java/org/apache/orc/mapreduce/OrcMapreduceRecordReader.java
```diff
@@ -53,16 +53,20 @@ public OrcMapreduceRecordReader(RecordReader reader,
...
+  public OrcMapreduceRecordReader(Reader fileReader,
+      Reader.Options options) throws IOException {
```
dongjoon-hyun, line 57: Shall we match the indentation with line 49?

java/mapreduce/src/java/org/apache/orc/mapreduce/OrcMapreduceRecordReader.java
```diff
@@ -53,16 +53,20 @@ public OrcMapreduceRecordReader(RecordReader reader,
...
+    this(fileReader, options, VectorizedRowBatch.DEFAULT_SIZE);
+  }
 
   public OrcMapreduceRecordReader(Reader fileReader,
-                                  Reader.Options options) throws IOException {
+                                  Reader.Options options, int rowBatchSize) throws IOException {
```
dongjoon-hyun, line 62: Let's put `rowBatchSize` in a new line like the following.
```java
Reader fileReader,
Reader.Options options,
int rowBatchSize) throws IOException {
```

dongjoon-hyun:
To @sv2000 . I left a few minor comments about coding styling.
To @pgaref . Could you confirm that all of your concerns are addressed, please?

sv2000:
Thanks @dongjoon-hyun. I addressed your comments. Can you help take a look?

pgaref:
Hey @sv2000  thanks for making the changes -- thats exactly what I had in mind!
Let update PRs description and some of the Mapreduce Tests to use the new constructors -- will merge after that.

dongjoon-hyun (approved):
Thanks, @sv2000 . +1 from my side. I'll leave this to @pgaref according to his last comments.

sv2000:
@pgaref Added unit tests to use the overloaded constructors and updated the PR description. LMK if the changes look ok.

pgaref (approved):
LGTM

PR #675, opened 2021-03-30
ORC-666: [C++] Add timestamp with local timezone

coderex2522:
The patch is tested through 
a. build type from string
b. write timestamp_instant column
c. read/write timestamp column with different timezone.
all test case already pass.

### DIFF
--- a/c++/test/TestWriter.cc
+++ b/c++/test/TestWriter.cc
@@ -706,11 +706,10 @@ namespace orc {
 
   void testWriteTimestampWithTimezone(FileVersion fileVersion,
                                       const char* writerTimezone,
+                                      int64_t writerTime,
                                       const char* readerTimezone,
-                                      const std::string& tsStr,
-                                      int isDst = 0) {
-    char* tzBk = getenv("TZ");  // backup TZ env
-
+                                      int64_t readerTime,
+                                      const std::string& tsStr) {
     MemoryOutputStream memStream(DEFAULT_MEM_STREAM_SIZE);
     MemoryPool* pool = getDefaultPool();
     std::unique_ptr<Type> type(Type::buildTypeFromString("struct<col1:timestamp>"));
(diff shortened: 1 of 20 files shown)


dongjoon-hyun:
cc @omalley

c++/test/TestWriter.cc
```diff
@@ -699,6 +705,172 @@ namespace orc {
...
+                                                  &memStream,
+                                                  fileVersion,
+                                                  0,
+                                                  writerTimezone);
+    auto batch = writer->createRowBatch(rowCount);
+    auto& structBatch = dynamic_cast<StructVectorBatch&>(*batch);
+    auto& tsBatch = dynamic_cast<TimestampVectorBatch&>(*structBatch.fields[0]);
+
+    // write timestamp in the writer timezone
+#ifdef _WIN32
```
wgtmac, line 737: I'd rather not expose platform specific logic and put it into Adaptor.hh.in and Adaptor.cc. See https://github.com/apache/orc/blob/master/c%2B%2B/src/Adaptor.cc#L46 for example.

wgtmac (commented):
Thanks for the contribution @coderex2522! Basically this patch does two things: 1) support setting writer/reader timezone to remove the hack which hardcodes timezone to GMT. 2) support a new TIMESTAMP_INSTANT type. Because they are orthogonal, it is better to split them into two separate patches. Additionally, please make sure the build passes on Windows as well.

coderex2522:
> Thanks for the contribution @coderex2522! Basically this patch does two things: 1) support setting writer/reader timezone to remove the hack which hardcodes timezone to GMT. 2) support a new TIMESTAMP_INSTANT type. Because they are orthogonal, it is better to split them into two separate patches. Additionally, please make sure the build passes on Windows as well.

OK, the first thing has already move into ORC-784 pull request.

PR #676, opened 2021-03-31
MINOR: Fix RLEv1 encoding examples in ORCv0/v1/v2 docs

dirtysalt:
Change the literals-encoding example to [2,3,6,7,11]. Either [2,3,4,7,11] or first 5 prime numbers [2,3,5,7,11] has run, and can not be used as a literals-encoding example. 

------

The first 5 prime numbers are [2,3,5,7,11], and there is a run. The encoded data is
- 2 [0xff, 0x02]
- 3,5,7 (delta=2,length=3,base=3) [0x00, 0x02, 0x03]
- 11 [0xff, 0xb]

And if we use example [2,3,4,7,11] as the spec written, there is also a run .The encoded data is 
- 2,3,4 (delta=1, length=3, base=2) [0x00, 0x01, 0x02]
- 7,11  [0xfe, 0x7, 0xb]

------
And I have verified the encoded data with the following C++ code. It used RLEEncoderv1 to encode [2,3,5,7,11] and dump data to a local file called `rle.data`. 

```c++

int main() {
    auto pool = getDefaultPool();
    auto outputStream = writeLocalFile("rle.data");
    auto buf = new BufferedOutputStream(*pool, outputStream.get(), 1024, 32);
    std::unique_ptr<BufferedOutputStream> pbuf(buf);
    auto encoder = new RleEncoderV1(std::move(pbuf), false);
    encoder->write(2);
    encoder->write(3);
    encoder->write(5);
    encoder->write(7);
    encoder->write(11);
    encoder->flush();
    return 0;
}
```

And then I use `od -A n -t x1 rle.data` to see data in hex format. The result is following

```
 ff 02 00 02 03 ff 0b
```


wgtmac:
Thanks @dirtysalt for the good catch! The illustration in the documentation is wrong as you have noticed. But I'd like to remove the word "prime" from that section as it aims to illustrate rule of LITERAL (not RUN) in the RLEv1. The correction may also consider ORCv0.md and ORCv2.md.

.gitignore
```diff
@@ -9,3 +9,4 @@ target
 *.iws
 .idea
 .DS_Store
+.vscode/
```
dongjoon-hyun, line 12: Hi, @dirtysalt . This is orthogonal change from the proposed document fix. Apache ORC recommends that each commit should have one theme. Could you revert this change from this PR?

temp/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.data
+*.exe
```
dongjoon-hyun, line 2: Why do you add `temp` directory, @dirtysalt ? Could you revert this `temp` directory?

.gitignore
dirtysalt: Yes. Sorry, I don't expect this PR  will trace my master branch commit automatically. Reverted already.

temp/.gitignore
dirtysalt: Same issue, sorry.

dongjoon-hyun:
Thank you for updating, @dirtysalt . Could you revise the PR description accordingly together? For example, `first 5 prime numbers`?

dongjoon-hyun:
For PR title, I updated it according to the AS-IS PR content.

dirtysalt:
Yes. Updated. I add a short description of PR at the head. I want to keep the old description because it provides some detailed explanation why not to use those examples. Not sure it's ok or not. Thanks for your review and suggestion.

dongjoon-hyun (approved):
+1, LGTM. Thank you, @dirtysalt .
Merged to master/1.6

PR #679, opened 2021-04-06
ORC-780: Add LZ4 Compression to the C++ Writer

noirello:

A new BlockCompressionStream class was implemented, that can be used for LZ4 compression kind.

The C++ Writer doesn't support LZ4 compression. This PR aims to fix that.

The existing compression tests were copied and changed to use LZ4 compression,

### DIFF
--- a/c++/src/Compression.cc
+++ b/c++/src/Compression.cc
@@ -36,6 +36,10 @@
 #define ZSTD_CLEVEL_DEFAULT 3
 #endif
 
+#ifndef LZ4_ACCELERATION_MAX
+#define LZ4_ACCELERATION_MAX 65537
+#endif
+
 namespace orc {
 
   class CompressionStreamBase: public BufferedOutputStream {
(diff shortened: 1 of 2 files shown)


c++/src/Compression.cc
```diff
@@ -897,6 +901,70 @@ DIAGNOSTIC_POP
...
+                                            static_cast<int>(compressorBuffer.size()),
+                                            level);
+    if (result == 0) {
+      throw std::runtime_error("Error during block compression using lz4.");
+    }
+    return static_cast<uint64_t>(result);
+  }
+
+  void Lz4CompressionSteam::init() {
+    state = static_cast<LZ4_stream_t*>(malloc(static_cast<size_t>(LZ4_sizeofState())));
```
wgtmac, line 957: why not use LZ4_createStream() to create one and free it via LZ4_freeStream() ?

c++/src/Compression.cc
```diff
@@ -1064,9 +1132,15 @@ DIAGNOSTIC_PUSH
         (new ZSTDCompressionStream(
           outStream, level, bufferCapacity, compressionBlockSize, pool));
     }
+    case CompressionKind_LZ4: {
+      int level = (strategy == CompressionStrategy_SPEED) ?
+              LZ4_ACCELERATION_MAX : 1;
```
wgtmac, line 1137: replace magic number 1 with macro ACCELERATION_DEFAULT defined in lz4.h

c++/src/Compression.cc
```diff
@@ -36,6 +36,10 @@
 #define ZSTD_CLEVEL_DEFAULT 3
 #endif
 
+#ifndef LZ4_ACCELERATION_MAX
+#define LZ4_ACCELERATION_MAX 65537
```
wgtmac, line 40: Why set to 65537? Will it overflows to be a negative number as the acceleration parameter used in lz4 is of type int.

wgtmac (commented):
Thanks for the patch @noirello . Just left some comments. Let me know what's your thoughts. Thanks!

c++/src/Compression.cc
noirello: I just overlooked it. Somehow the `void*` cast through me off a little. Fixed it.

c++/src/Compression.cc
noirello: These macros are defined in `lz4.c` instead of the header file for some reason. That's why I needed to redefined them here. Their values are looked up from the docs/[sources](https://github.com/lz4/lz4/blob/bdc9d3b0c10cbf7e1ff40fc6e27dad5f693a00e7/lib/lz4.c#L47-L57).

wgtmac (approved):
LGTM +1

dongjoon-hyun (approved):

dongjoon-hyun:
+1, LGTM.

dongjoon-hyun:
Merged to master. Thank you!

dongjoon-hyun:
I updated the JIRA for Apache ORC 1.7.0

PR #682, opened 2021-04-12
ORC-781: [C++] Make type annotations available from C++

boroknagyz:
ORC-522 added support for type annotations, but only to the Java ORC
library. This patch adds support for type annotations for the C++
ORC library.

Both reads and writes are supported. orc-metadata now also prints out
the type attriburtes.

Testing
 * added unit tests in TestAttributes.cc
 * added unit test to TestFileMetadata.cc


boroknagyz:
The Java test failure seems unrelated to my change. @wgtmac, could you please take a look at this change?

wgtmac:
> The Java test failure seems unrelated to my change. @wgtmac, could you please take a look at this change?

Thanks for pinging me. I will take a look shortly.

c++/src/TypeImpl.cc
```diff
@@ -121,6 +121,32 @@ namespace orc {
...
+  std::vector<std::string> TypeImpl::getAttributeKeys() const {
+    std::vector<std::string> ret;
+    ret.reserve(attributes.size());
+    for (auto& attribute : attributes) {
+      ret.push_back(attribute.first);
+    }
+    return ret;
+  }
+
+  std::string TypeImpl::getAttributeValue(const std::string& key) const {
```
wgtmac, line 144: I'd prefer to add a new interface`bool hasAttributeKey(const std::string& key) const` to check existence and throw here if the key does not exist.

c++/include/orc/Type.hh
```diff
@@ -58,6 +58,11 @@ namespace orc {
     virtual uint64_t getMaximumLength() const = 0;
     virtual uint64_t getPrecision() const = 0;
     virtual uint64_t getScale() const = 0;
+    virtual Type& setAttribute(const std::string& key,
```
wgtmac, line 61: What's the benefit to make return type as Type& for setAttribute() and removeAttribute() ? Is it better to make it void?

c++/src/TypeImpl.cc
boroknagyz: I've added hasAttributeKey() to Type, and throw std::range_error in removeAttribute() and getAttributeValue() when key is not present

c++/include/orc/Type.hh
boroknagyz: This allows method chaining. I've followed the Java implementation:
https://github.com/apache/orc/blob/f381f2dd78e6b305b8a4e0b07124034de72ef1a1/java/core/src/java/org/apache/orc/TypeDescription.java#L242-L252
But I can switch to void if you prefer that.

wgtmac (approved):
+1 LGTM

boroknagyz:
Thanks for the review @wgtmac! Could you please merge this PR?

PR #683, opened 2021-04-13
ORC-784:[C++] Read/Write timestamp column with the specified timezone option

coderex2522:
this patch is tested by the ut of writeTimestampWithTimezone.

### DIFF
--- a/c++/src/Writer.cc
+++ b/c++/src/Writer.cc
@@ -41,6 +41,7 @@ namespace orc {
     std::set<uint64_t> columnsUseBloomFilter;
     double bloomFilterFalsePositiveProb;
     BloomFilterVersion bloomFilterVersion;
+    std::string timezone;
 
     WriterOptionsPrivate() :
                             fileVersion(FileVersion::v_0_12()) { // default to Hive_0_12
@@ -56,6 +57,7 @@ namespace orc {
       enableIndex = true;
       bloomFilterFalsePositiveProb = 0.05;
       bloomFilterVersion = UTF8;
+      timezone = "GMT";
     }
   };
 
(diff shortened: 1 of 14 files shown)


coderex2522:
Hey @wgtmac, the UT of this patch don't support on windows system. There are several problems here 
1) Windows don't support acquiring the structure timezone.
2) The timezones of Windows cannot match exactly linux system. 
3) the function calls about timezone in Windows need to acquire process token.

c++/src/Writer.cc
```diff
@@ -56,6 +57,7 @@ namespace orc {
       enableIndex = true;
       bloomFilterFalsePositiveProb = 0.05;
       bloomFilterVersion = UTF8;
+      timezone = "GMT";
```
wgtmac, line 60: The Java ORC writer uses local time zone by default. However, the C++ ORC writer uses GMT by default in this [fix](https://github.com/apache/orc/commit/8ea3592f6aeddb5e6930798e4ff95065c6177e00). I hard-coded it to GMT at that time to work around the insufficient support of timezone conversion on the C++ side. I'd suggest to leave some comments to explain this history.

c++/test/TestWriter.cc
```diff
@@ -699,6 +704,98 @@ namespace orc {
     }
   }
 
+#ifndef _MSC_VER
```
wgtmac, line 707: leave some comments to explain why you skip windows here.

c++/src/Writer.cc
coderex2522: Done

coderex2522 (commented):
@wgtmac I added some comments, please take a look.

c++/test/TestWriter.cc
coderex2522: Done.

coderex2522 (commented):
done

wgtmac (approved):
LGTM +1

PR #688, opened 2021-04-20
ORC-666: [C++] support a new TIMESTAMP_INSTANT type

coderex2522:
How was this patch tested?
1. Read and write tests for new TIMESTAMP_INSTANT type columns in unit test WriteTest.writeTimestampInstant
2. build type from the string containing the new TIMESTAMP_INSTANT type in unit test TestType.buildTypeFromString

### DIFF
--- a/c++/src/TypeImpl.cc
+++ b/c++/src/TypeImpl.cc
@@ -169,6 +169,8 @@ namespace orc {
       return "binary";
     case TIMESTAMP:
       return "timestamp";
+    case TIMESTAMP_INSTANT:
+      return "timestamp_with_local_time_zone";
     case LIST:
       return "array<" + (subTypes[0] ? subTypes[0]->toString() : "void") + ">";
     case MAP:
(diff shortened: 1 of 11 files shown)


c++/src/TypeImpl.cc
```diff
@@ -609,6 +614,8 @@ namespace orc {
       return std::unique_ptr<Type>(new TypeImpl(BINARY));
     } else if (category == "timestamp") {
       return std::unique_ptr<Type>(new TypeImpl(TIMESTAMP));
+    } else if (category == "timestamp_with_local_time_zone") {
```
wgtmac, line 617: To be consistent with the Java side, the type name literal should remove the underscores: [commit](https://github.com/omalley/orc/blob/8289f7d2640833bba46be88941d1013eb2083023/java/core/src/java/org/apache/orc/TypeDescription.java#L124)

wgtmac (approved):
+1 LGTM

c++/src/TypeImpl.cc
coderex2522: Done, the type name has been modified to be consistent with the Java side.

PR #690, opened 2021-04-22
fix orc-scan tool & more tools installed

dirtysalt:
1. `buf` pointer is not forwarded after reading a chunk
2.  some tools are not published.

### DIFF
--- a/c++/src/OrcHdfsFile.cc
+++ b/c++/src/OrcHdfsFile.cc
@@ -142,18 +142,20 @@ namespace orc {
         throw ParseError("Buffer is null");
       }
 
+      char* buf_ptr = reinterpret_cast<char*>(buf);
       hdfs::Status status;
       size_t total_bytes_read = 0;
       size_t last_bytes_read = 0;
 
       do {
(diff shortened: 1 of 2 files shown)


c++/src/OrcHdfsFile.cc
```diff
@@ -142,18 +142,20 @@ namespace orc {
         throw ParseError("Buffer is null");
       }
 
+      char* buf_ptr = (char*)buf;
```
wgtmac, line 145: please use reinterpret_cast

wgtmac (commented):
Please make sure all test cases pass. Thanks! @dirtysalt

c++/src/OrcHdfsFile.cc
dirtysalt: @wgtmac  sorry. updated. all test cases are passed.

wgtmac (approved):
+1 LGTM. Thanks @dirtysalt !

dongjoon-hyun:
Thank you, @dirtysalt and @wgtmac !

PR #691, opened 2021-04-26
ORC-790: TIMESTAMP_INSTANT should be primitive

boroknagyz:
This patch makes TIMESTAMP_INSTANT a primitive type.

### DIFF
--- a/java/core/src/java/org/apache/orc/TypeDescription.java
+++ b/java/core/src/java/org/apache/orc/TypeDescription.java
@@ -116,7 +116,7 @@ public class TypeDescription
     MAP("map", false),
     STRUCT("struct", false),
     UNION("uniontype", false),
-    TIMESTAMP_INSTANT("timestamp with local time zone", false);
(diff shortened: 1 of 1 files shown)


java/core/src/java/org/apache/orc/TypeDescription.java
```diff
@@ -116,7 +116,7 @@ public int compareTo(TypeDescription other) {
     MAP("map", false),
     STRUCT("struct", false),
     UNION("uniontype", false),
-    TIMESTAMP_INSTANT("timestamp with local time zone", false);
+    TIMESTAMP_INSTANT("timestamp with local time zone", true);
```
dongjoon-hyun, line 119: cc @omalley since this is ORC-189.

dongjoon-hyun: Also, cc @prasanthj , @wgtmac , @williamhyun

dongjoon-hyun:
Thank you, @boroknagyz and @omalley !

PR #694, opened 2021-05-05
ORC-793: Deprecate the incorrect getInt method and add a new setInt within OrcConf

autumnust:

There was two duplicated `getInt` method existed in `OrcConf` but no `setInt`. From the existing implementation it is obvious that the method `getInt` is a typo and it should `setInt` instead.

This patch is rather straightforward, running through existing unit tests.

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -244,7 +244,7 @@ public enum OrcConf {
     return getInt(null, conf);
   }
 
-  public void getInt(Configuration conf, int value) {
+  public void setInt(Configuration conf, int value) {
     conf.setInt(attribute, value);
   }
 
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -244,6 +244,15 @@ public enum OrcConf {
     return getInt(null, conf);
   }
 
+  /**
+   * @deprecated Use {@link #getInt(Configuration)} instead. This method was
+   * incorrectly added and shouldn't be used anymore.
+   */
+  @Deprecated
+  public void getInt(Configuration conf, int value) {
+    // noop
+  }
+
   public void setInt(Configuration conf, int value) {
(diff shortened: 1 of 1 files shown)


java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -244,7 +244,7 @@ public int getInt(Configuration conf) {
     return getInt(null, conf);
   }
 
-  public void getInt(Configuration conf, int value) {
+  public void setInt(Configuration conf, int value) {
```
dongjoon-hyun, line 247: Wow. Nice catching.

dongjoon-hyun (changes requested):
Thank you so much, @autumnust . 

Although this is a correct fix, Apache ORC follows semantic versioning rule. We cannot remove a public method like this at Apache ORC 1.7.x.

Instead, could you add a new function `public void setInt(Configuration conf, int value)` and add a deprecated annotation and description to the wrong old function, `public void getInt()`, please?

autumnust:
> Thank you so much, @autumnust .
> 
> Although this is a correct fix and `void getInt` doesn't make senses, Apache ORC follows [semantic versioning](https://semver.org) rule. We cannot remove a public method like this at Apache ORC 1.7.x.
> 
> Instead, could you add a new function `public void setInt(Configuration conf, int value)` and add a deprecated annotation and description to the wrong old function, `public void getInt()`, please? And, the PR title and description should be revised accordingly.
> 
> cc @omalley , @pgaref , @wgtmac , @williamhyun

thanks for suggestions. @dongjoon-hyun  Do you think it is necessary to make the original incorrect `getInt` a noop?

java/core/src/java/org/apache/orc/OrcConf.java
pgaref: Good catch indeed!

dongjoon-hyun (approved):
+1, LGTM. Thank you so much, @autumnust .
Merged to master/1.6/1.5.

dongjoon-hyun:
Also, thank you, @pgaref !

PR #695, opened 2021-05-05
ORC-614: [C++] Implement efficient seek() in decompression streams

wgtmac:

The current implementation of ZlibDecompressionStream::seek and
BlockDecompressionStream::seek resets the state of the decompressor
and the underlying file reader and throws away their buffers.

This commit introduces two optimizations which rely on reusing
the buffers that still contain useful data, and therefore reducing
the time spent reading/uncompressing the buffers again.

The first case is when the seeked position is already read
and decompressed into the output stream.

The second case is when the seeked position is already read from
the input stream, but has not been decompressed yet, ie. it's
not in the output stream.

Tests:
- Run the ORC tests, and the Impala tests working on ORC tables.
- The regression that #476 would cause is not present anymore.

### DIFF
--- a/c++/src/Compression.cc
+++ b/c++/src/Compression.cc
@@ -355,8 +355,8 @@ DIAGNOSTIC_PUSH
 
     // The starting and current position of the buffer for the uncompressed
     // data. It either points to the data buffer or the underlying input stream.
-    const char* outputBufferStart;
-    const char* outputBuffer;
+    const char *outputBufferStart;
+    const char *outputBuffer;
     // The original (ie. the overall) and the actual length of the uncompressed
     // data.
     size_t uncompressedBufferLength;
@@ -367,7 +367,7 @@ DIAGNOSTIC_PUSH
     size_t remainingLength;
 
     // the last buffer returned from the input
-    const char* inputBufferStart;
+    const char *inputBufferStart;
     const char *inputBuffer;
     const char *inputBufferEnd;
 
@@ -541,7 +541,7 @@ DIAGNOSTIC_PUSH
    */
   void DecompressionStream::seek(PositionProvider& position) {
     size_t seekedPosition = position.current();
-    // Case 3.: the seeked position is the one that is currently buffered and
+    // Case 1: the seeked position is the one that is currently buffered and
     // decompressed. Here we only need to set the output buffer's pointer to the
     // seeked position. Note that after the headerPosition comes the 3 bytes of
     // the header.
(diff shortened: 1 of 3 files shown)


dongjoon-hyun (commented):
Thank you for the fix. BTW, is this applicable to `branch-1.6` too, @wgtmac ?

wgtmac:
> Thank you for the fix. BTW, is this applicable to `branch-1.6` too, @wgtmac ?

@dongjoon-hyun No, the `main` branch has an innegligible refactoring of the DecompressionStream which does not exist in the `branch-1.6`. So I suppose this fix can only work here.

dongjoon-hyun:
Got it. Thank you for the explanation.

c++/src/Compression.cc
```diff
@@ -503,16 +532,47 @@ DIAGNOSTIC_PUSH
...
+   * 1. The seeked position is already read and decompressed into
+   *    the output stream.
+   * 2. It is already read from the input stream, but has not been
+   *    decompressed yet, ie. it's not in the output stream.
+   * 3. It is not read yet from the inputstream.
+   */
   void DecompressionStream::seek(PositionProvider& position) {
-    // clear state to force seek to read from the right position
+    size_t seekedPosition = position.current();
+    // Case 3.: the seeked position is the one that is currently buffered and
```
dongjoon-hyun, line 544: It seems that this is `1` at line 536. Shall we match the number by changing this `Case 3.:` -> `Case 1:`?

c++/src/Compression.cc
```diff
@@ -503,16 +532,47 @@ DIAGNOSTIC_PUSH
...
+    if (seekedPosition < static_cast<uint64_t>(input->ByteCount()) &&
+        seekedPosition >= inputBufferStartPosition) {
+      // Case 2.: The input is buffered, but not yet decompressed. No need to
+      // force re-reading the inputBuffer, we just have to move it to the
+      // seeked position.
+      position.next(); // Skip the input level position.
+      inputBuffer
+        = inputBufferStart + (seekedPosition - inputBufferStartPosition);
+    } else {
+      // Case 1.: The seeked position is not in the input buffer, here we are
```
dongjoon-hyun, line 570: Ditto. This looks like `3. It is not read yet from the inputstream.`. 
Switching from `Case 1.:` to `Case 3:` will be better. WDYT?

c++/src/Compression.cc
```diff
@@ -353,19 +353,29 @@ DIAGNOSTIC_PUSH
...
+    size_t uncompressedBufferLength;
     size_t outputBufferLength;
-    // the size of the current chunk
+
+    // The remaining size of the current chunk that is not yet consumed
+    // ie. decompressed or returned in output if state==DECOMPRESS_ORIGINAL
     size_t remainingLength;
 
     // the last buffer returned from the input
+    const char* inputBufferStart;
```
dongjoon-hyun, line 370: nit. Shall we match with the coding style with the existing line 371 and 372?
```cpp
- const char* inputBufferStart;
+ const char *inputBufferStart;
```

dongjoon-hyun (commented):
I left a few comments. BTW, do you think it's possible to have a simple test case for this?

Also, cc @pgaref

c++/src/Compression.cc
wgtmac: done

c++/src/Compression.cc
wgtmac: done

c++/src/Compression.cc
wgtmac: done

dongjoon-hyun (approved):
Thank you for updating, @wgtmac . +1, LGTM.

dongjoon-hyun:
Merged to main for Apache ORC 1.7.0.

PR #698, opened 2021-05-14
ORC-797: Allow writers to get the stripe information.

omalley:

It adds a new method to the Writer that lets the user get information about the stripes of the file.

Currently Iceberg has to reopen the file after writing the file and we'd like to short circuit that.

Added a unit test.


dongjoon-hyun (approved):
+1, LGTM. Thank you, @omalley !

PR #701, opened 2021-05-18
ORC-800: Add `key` and `value` nullness check in `MapVectorBatch`

czxrrr:
MapVectorBatch::toString will encounter segment fault if map.value is selected while map.key is not selected.

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -160,6 +160,7 @@ namespace orc {
       std::fill(selectedColumns.begin(), selectedColumns.end(), true);
     }
     selectParents(selectedColumns, *contents->schema.get());
+    selectMaps(selectedColumns, *contents->schema.get());
     selectedColumns[0] = true; // column 0 is selected by default
   }
 
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -100,6 +100,20 @@ namespace orc {
   }
 
   /**
+   * Recurses over a type tree and selects the map.key and map.value of every selected map type.
+   */
+  void ColumnSelector::selectMaps(std::vector<bool>& selectedColumns, const Type& type) {
+    size_t id = static_cast<size_t>(type.getColumnId());
+    for(uint64_t c=0; c < type.getSubtypeCount(); ++c) {
+      if (type.getKind() == MAP && selectedColumns[id]) {
+        selectChildren(selectedColumns, *type.getSubtype(c));
+      } else {
+        selectMaps(selectedColumns, *type.getSubtype(c));
+      }
+    }
+  }
+
+  /**
    * Recurses over a type tree and build two maps
    * map<TypeName, TypeId>, map<TypeId, Type>
    */
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -100,20 +100,6 @@ namespace orc {
   }
 
   /**
-   * Recurses over a type tree and selects the map.key and map.value of every selected map type.
-   */
-  void ColumnSelector::selectMaps(std::vector<bool>& selectedColumns, const Type& type) {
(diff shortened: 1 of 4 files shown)


wgtmac (commented):
Could you please add an UT to reproduce this?

dongjoon-hyun:
+1 for @wgtmac 's comment.

czxrrr:
I Added a testcase. Without this commit, toString() function call will segment fault.

czxrrr:
Testcase was just attached.

dongjoon-hyun:
Thank you, @czxrrr

c++/src/Reader.cc
```diff
@@ -99,6 +99,20 @@ namespace orc {
...
   }
 
+  /**
+   * Recurses over a type tree and selects the map.key and map.value of every selected map type.
+   */
+  void ColumnSelector::selectMaps(std::vector<bool>& selectedColumns, const Type& type) {
+    size_t id = static_cast<size_t>(type.getColumnId());
+    for(uint64_t c=0; c < type.getSubtypeCount(); ++c) {
+      if (type.getKind() == MAP && selectedColumns[id]) {
+        selectChildren(selectedColumns, *type.getSubtype(c));
```
wgtmac, line 109: This implementation means all sub types of the MAP KEY or MAP VALUE are selected recursively. For example, if the map value type is a struct type which contains 100 sub types, we need to read all of them even the SQL query does not request so.

c++/src/Reader.hh
```diff
@@ -96,6 +96,9 @@ namespace orc {
     // For each child of type, select it if one of its children
     // is selected.
     bool selectParents(std::vector<bool>& selectedColumns, const Type& type);
+
+    // For selected map type, select its children, key and value
+    void selectMaps(std::vector<bool>& selectedColumns, const Type& type);
```
wgtmac, line 101: Does LIST type has same issue?

tools/test/TestMatch.cc
```diff
@@ -1085,6 +1085,27 @@ TEST(TestMatch, selectColumns) {
...
+      if (i>=19 && i<=23)
+        EXPECT_TRUE(c[i]);
+      else
+        EXPECT_TRUE(!c[i]);
+    }
+    batch = rowReader->createRowBatch(1);
+    std::ostringstream expectedMapSchema;
+    expectedMapSchema << "Struct vector <0 of 1; Map vector <Byte vector <0 of 1>, "
+        << "Struct vector <0 of 1; Long vector <0 of 1>; Byte vector <0 of 1>; > with 0 of 1>; >";
+    EXPECT_EQ(expectedMapSchema.str(), batch->toString());
```
wgtmac, line 1107: I think it is valid to read only map key or map value column, or even a subset of key or value column if itself is a complex type. And you reported the bug from toString() function. Should we only fix this function to handle the case where its sub vector is null?

c++/src/Reader.hh
czxrrr: List won't have such issues, because list only has one child. Parent and Child have to be both selected or both unselected.

tools/test/TestMatch.cc
czxrrr: The bug does not only exist in ReaderImpl::toString() and it exists in ReaderImpl::getMemoryUse() ALSO.
Once key or value missing, even rootBatch->getMemoryUse() will fail. 
In my opinion, If people really wanna querying keys or values independently, I think struct of array of keys and array of values is more suitable, they do not need map.

czxrrr: But letting people query part of map would be more general way to solve it. Let me try fix it in this way.

wgtmac (approved):
LGTM +1

PR #702, opened 2021-05-24
ORC-801: Clean up Logging

belugabehr:
Clean up some logging statements with the project. Remove instances of log-and-throw anti-pattern, remove superfluous logging guards, use anchors in debug statements instead of concatenation.

Trivial performance increases, decrease in lines of code (readability), log-and-throw makes troubleshooting more difficult because the same error may be logged multiple times.

Existing unit tests, but should be no change in functionality.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -141,10 +141,8 @@ public class RecordReaderImpl implements RecordReader {
     OrcFile.WriterVersion writerVersion = fileReader.getWriterVersion();
     SchemaEvolution evolution;
     if (options.getSchema() == null) {
-      if (LOG.isInfoEnabled()) {
-        LOG.info("Reader schema not provided -- using file schema " +
-            fileReader.getSchema());
-      }
+      LOG.info("Reader schema not provided -- using file schema " +
+          fileReader.getSchema());
       evolution = new SchemaEvolution(fileReader.getSchema(), null, options);
     } else {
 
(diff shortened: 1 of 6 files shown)


dongjoon-hyun (changes requested):
Hi, @belugabehr . Thank you for making a PR.
Could you fill the PR template?

belugabehr:
@dongjoon-hyun PR template completed.  Thanks!

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -1030,13 +1026,11 @@ public SargApplier(SearchArgument sarg,
...
-                        " Skipping ORC PPD." +
-                        " Stats: " + stats +
-                        " Predicate: " + predicate;
-                    LOG.warn(reason, e);
-                  }
+                  final String reason = e.getClass().getSimpleName() + " when evaluating predicate." +
+                      " Skipping ORC PPD." +
+                      " Stats: " + stats +
+                      " Predicate: " + predicate;
+                  LOG.warn(reason, e);
```
dongjoon-hyun, line 1033: Shall we revert this change?

belugabehr: Hello. This change is intentional.  There is never a reason to wrap a `warn` level message in logging guards: `isWarnEnabled`.  The answer will always be "true"

dongjoon-hyun: Ah, got it. It makes sense.

dongjoon-hyun (approved):
+1, LGTM. Thank you, @belugabehr .
Merged to master for Apache ORC 1.7.0.

PR #703, opened 2021-05-24
ORC-802: Make Maven Minimum 3.6.0

belugabehr:

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ The subdirectories are:
 ### Building
 
 * Install java 1.8 or higher
-* Install maven 3 or higher
+* Install maven 3.6.0 or higher
 * Install cmake
 
 To build a release version with debug information:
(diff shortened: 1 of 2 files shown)


dongjoon-hyun (changes requested):
Hi, @belugabehr .
Apache ORC community provides `Maven Wrapper`, `mvnw`. Could you try that?

belugabehr:
@dongjoon-hyun Thanks for the update.

I have verified that the script does indeed build the project, but I've never seen such a setup before (and it's not documented in the README).  What is the value of doing it this way?  It is not expected at all.

If this is absolutely the preferred way, I can update the docs to mention this. 

Thanks!

dongjoon-hyun:
Thanks. +1 for switching to a doc-only updating PR.

dongjoon-hyun:
I'll close this for now. Please feel free to reopen this when you are ready.

PR #705, opened 2021-05-26
ORC-804: MaskDescriptionImpl should use List instead of Set

dongjoon-hyun:

This PR aims to use `List` instead of `Set` in `MaskDescriptionImpl`.

If there are multiple encrypted columns with the same types for a single key, only the first column is masked.

Pass the CIs with the newly added test cases.


dongjoon-hyun:
Hi, @omalley , @pgaref , @wgtmac , @williamhyun .
Could you review this?

wgtmac (approved):
It looks pretty straightforward to me. +1

dongjoon-hyun:
Thank you so much, @wgtmac ! Merged to master/branch-1.6.

PR #712, opened 2021-06-04
ORC-810: FIX expected Output Stats when using Hash Dictionary

pgaref:
ORC-757 introduced HT as the default string dictionary implementation.
testStripeLevelStats need to be updated and existing tests with ORC files written with RBTree dict should use the same dict to avoid size inconsistencies

Fix tests upstream

Passed tests locally


pgaref:
cc @autumnust

dongjoon-hyun (approved):
+1, LGTM.

BTW, we had better have both test coverage in the end.

PR #282, opened 2018-06-09
ORC-377: [c++] Add SnappyCompressionStream

czxrrr:
Currently ORC C++ only support writing with Zlib compression
This branch aims to add support for Snappy compression

PR #713, opened 2021-06-05
ORC-377: [c++] Add SnappyCompressionStream

noirello:

A new BlockCompressionStream class was implemented for Snappy compression.

The C++ Writer doesn't support Snappy compression. This PR aims to fix that.

The existing compression tests were copied and changed to use Snappy compression.

PR #282 aims for the very same thing, but it's not complete and inactive for a while now. The BlockCompressionStream has been introduced since it was opened, thus it was easier for me to start over from the current state and open a new pull request.
I hope it's not a problem that I've stepped in to add the functionality.

### DIFF
--- a/c++/src/Compression.cc
+++ b/c++/src/Compression.cc
@@ -1031,6 +1031,47 @@ DIAGNOSTIC_POP
   }
 
   /**
+   * Snappy block compression
+   */
+  class SnappyCompressionStream: public BlockCompressionStream {
+  public:
+    SnappyCompressionStream(OutputStream * outStream,
+                        int compressionLevel,
+                        uint64_t capacity,
+                        uint64_t blockSize,
+                        MemoryPool& pool)
+                        : BlockCompressionStream(outStream,
+                                                 compressionLevel,
+                                                 capacity,
+                                                 blockSize,
+                                                 pool) {
+    }
+
+    virtual std::string getName() const override {
+      return "SnappyCompressionStream";
+    }
+    
+    virtual ~SnappyCompressionStream() override {
+    }
+
+  protected:
+    virtual uint64_t doBlockCompression() override;
+
+    virtual uint64_t estimateMaxCompressionSize() override {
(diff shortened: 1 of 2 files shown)


c++/src/Compression.cc
```diff
@@ -1030,6 +1030,47 @@ DIAGNOSTIC_POP
...
+                                                 blockSize,
+                                                 pool) {
+    }
+
+    virtual std::string getName() const override {
+      return "SnappyCompressionStream";
+    }
+    
+    virtual ~SnappyCompressionStream() override {
+    }
```
wgtmac, line 1055: // PASS

c++/src/Compression.cc
```diff
@@ -1030,6 +1030,47 @@ DIAGNOSTIC_POP
...
+
+  protected:
+    virtual uint64_t doBlockCompression() override;
+
+    virtual uint64_t estimateMaxCompressionSize() override {
+      return static_cast<uint64_t>(snappy::MaxCompressedLength(static_cast<size_t>(bufferSize)));
```
wgtmac, line 1061: break this long line into two

c++/src/Compression.cc
```diff
@@ -1204,7 +1245,12 @@ DIAGNOSTIC_PUSH
         (new Lz4CompressionSteam(
           outStream, level, bufferCapacity, compressionBlockSize, pool));
     }
-    case CompressionKind_SNAPPY:
+    case CompressionKind_SNAPPY: {
+      int level = 0;
```
wgtmac, line 1249: Does snappy support different levels? If true, we should assign different levels according to CompressionStrategy

noirello: There's no mention about it in the header file, and I couldn't find anything elsewhere, that would indicate it does.

wgtmac (approved):
+1 LGTM. Thanks @noirello !

dongjoon-hyun:
Thank you. I updated the JIRA for Apache ORC 1.7.0.

PR #731, opened 2021-06-30
ORC-826: Do Not Use Collection Contains/Get

belugabehr:

Discovered multiple places in the code where a Collection is accessed with a 'contains' and a 'get'.  However, this is superfluous as a 'get' will return a `null` value if the Collection doesn't contain the item.  This can be utilized to save multiple trips into the same map.

Small performance, less code

No change in behavior so existing unit test suffice.


williamhyun (approved):
+1 LGTM

PR #732, opened 2021-07-01
ORC-827: Utilize Array copyOf

belugabehr:
Utilize JDK Array copyOf

Less code, small performance boost perhaps

Existing unit tests as functionality has not changed.

### DIFF
--- a/java/core/src/java/org/apache/orc/InMemoryKeystore.java
+++ b/java/core/src/java/org/apache/orc/InMemoryKeystore.java
@@ -276,7 +276,16 @@ public class InMemoryKeystore implements KeyProvider {
       algorithm = EncryptionAlgorithm.AES_CTR_128;
     }
 
-    final byte[] buffer = Arrays.copyOf(masterKey, algorithm.keyLength());
+    final byte[] buffer = new byte[algorithm.keyLength()];
+    if (algorithm.keyLength() > masterKey.length) {
+
+      System.arraycopy(masterKey, 0, buffer, 0, masterKey.length);
+      /* fill with zeros */
+      Arrays.fill(buffer, masterKey.length, buffer.length - 1, (byte) 0);
+
+    } else {
+      System.arraycopy(masterKey, 0, buffer, 0, algorithm.keyLength());
+    }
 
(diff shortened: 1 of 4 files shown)


java/core/src/java/org/apache/orc/InMemoryKeystore.java
```diff
@@ -278,16 +276,7 @@ public InMemoryKeystore addKey(String keyName, int version,
...
-    final byte[] buffer = new byte[algorithm.keyLength()];
-    if (algorithm.keyLength() > masterKey.length) {
-
-      System.arraycopy(masterKey, 0, buffer, 0, masterKey.length);
-      /* fill with zeros */
-      Arrays.fill(buffer, masterKey.length, buffer.length - 1, (byte) 0);
-
-    } else {
-      System.arraycopy(masterKey, 0, buffer, 0, algorithm.keyLength());
-    }
```
belugabehr, line 290: As I understand this code, it's a bit superfluous.  There is no need to "fill with zeros" as the array will, by default in Java, pre-populated with zero value.

And this:

> For any indices that are valid in the copy but not the original, the copy will contain (byte)0. Such indices will exist if and only if the specified length is greater than that of the original array.

https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#copyOf(byte[],%20int)

dongjoon-hyun:
cc @omalley

dongjoon-hyun (commented):
BTW, @belugabehr . Although you don't have your patch in Apache Orc 1.6.9, could you participate Apache ORC 1.6.9 RC0 vote?
- https://lists.apache.org/thread.html/rab78879c2ff44b0357d49b592974545d5e649852c53a696510e19bdd%40%3Cdev.orc.apache.org%3E

belugabehr:
@dongjoon-hyun Are you able to assist in moving this in?

Thanks!

belugabehr:
@dongjoon-hyun This one has been getting a bit stale.  What additional details can I provide?

Thanks!!

dongjoon-hyun (approved):
+1, LGTM (Pending CIs)

dongjoon-hyun:

PR #734, opened 2021-07-02
ORC-829: Optimize Serialization percentileBits

belugabehr:

Optimize Serialization percentileBits method

Speed and simplicity

Existing unit tests, no functionality changes.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/SerializationUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/SerializationUtils.java
@@ -34,6 +34,7 @@ import java.io.OutputStream;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.sql.Date;
+import java.util.Arrays;
 import java.util.TimeZone;
 
 public final class SerializationUtils {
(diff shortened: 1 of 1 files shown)


java/core/src/java/org/apache/orc/impl/SerializationUtils.java
```diff
@@ -287,27 +293,24 @@ public long zigzagDecode(long val) {
...
-                            double p) {
+  public int percentileBits(long[] data, int offset, int length, double p) {
     if ((p > 1.0) || (p <= 0.0)) {
       return -1;
     }
 
-    // histogram that store the encoded bit requirement for each values.
-    // maximum number of bits that can encoded is 32 (refer FixedBitSizes)
-    int[] hist = new int[32];
+    Arrays.fill(this.histBuffer, 0);
```
pgaref, line 301: Any reason we removed the comment here?

pgaref (commented):
Hey @belugabehr  changes LGTM -- just wondering if we can/should incorporate the Perf tests you did when discovered this? 

Thanks

belugabehr:
> Hey @belugabehr changes LGTM -- just wondering if we can/should incorporate the Perf tests you did when discovered this?

@pgaref Hey, thanks for taking a look.

Nothing special here.  I am just slowly working through the `core-benchmark` module.  I've examined the write side a bit (see: #734 #735 #736) and hope to find time next week to look at the read side.

To be clear, I think the biggest savings of this PR comes from using `Long.numberOfLeadingZeros` instead of implementing a loop that can loop as many as 64-times per value written.  The rest of the stuff I changed based on manual inspection of the code.  I did not test each piece individually, but as I mentioned, after making these changes this method didn't even register in the performance capture.

PR #736, opened 2021-07-02
ORC-831: Do Not Copy String When Flushing Dictionary

belugabehr:
The code to flush first loads (copies) the data into a Text object and then writes it to the direct stream. Instead, pass the directStream to the dictionary and have it write directly to the stream instead of an intermediate Text object.

Performance.

No change in behavior so using existing unit tests.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/Dictionary.java
+++ b/java/core/src/java/org/apache/orc/impl/Dictionary.java
@@ -44,6 +44,10 @@ public interface Dictionary {
 
   /**
    * Given the position index, return the original string before being encoded.
+   * The value of the Text in the Dictionary is copied into {@code result}.
+   *
+   * @param result the holder to copy the dictionary text into
+   * @param position the position where the key was added
    */
   void getText(Text result, int position);
 
--- a/java/core/src/java/org/apache/orc/impl/Dictionary.java
+++ b/java/core/src/java/org/apache/orc/impl/Dictionary.java
@@ -55,6 +55,8 @@ public interface Dictionary {
    * Given the position index, write the original string, before being encoded,
    * to the OutputStream.
    *
+   * @param out the output stream to which to write the data
+   * @param position the position where the key was originally added
    * @return the number of byte written to the stream
    * @throws IOException if an I/O error occurs
    */
--- a/java/core/src/java/org/apache/orc/impl/Dictionary.java
+++ b/java/core/src/java/org/apache/orc/impl/Dictionary.java
@@ -47,6 +47,15 @@ public interface Dictionary {
    */
   void getText(Text result, int position);
 
+  /**
+   * Given the position index, write the original string, before being encoded,
+   * to the OutputStream.
+   *
+   * @return the number of byte written to the stream
+   * @throws IOException if an I/O error occurs
+   */
+  int writeTo(OutputStream out, int position) throws IOException;
+
   int add(byte[] bytes, int offset, int length);
 
   int size();
--- a/java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
@@ -17,8 +17,10 @@
  */
 package org.apache.orc.impl;
 
-import org.apache.hadoop.io.Text;
+import java.io.IOException;
+import java.io.OutputStream;
 
+import org.apache.hadoop.io.Text;
 
 public class DictionaryUtils {
   private DictionaryUtils() {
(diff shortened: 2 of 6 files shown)


java/core/src/java/org/apache/orc/impl/Dictionary.java
```diff
@@ -47,6 +47,15 @@
    */
```
pgaref, line 47: Can we please mention here that this method performs a copy from a byte array to Text?

java/core/src/java/org/apache/orc/impl/Dictionary.java
```diff
@@ -47,6 +47,15 @@
...
   void getText(Text result, int position);
 
+  /**
+   * Given the position index, write the original string, before being encoded,
+   * to the OutputStream.
```
pgaref, line 52: Can we please mention here that this is directly writing to out without extra copies?

pgaref (approved):
Thanks again for the PR @belugabehr this looks great! I would only add a comment or two to clarify the expected copies

java/core/src/java/org/apache/orc/impl/Dictionary.java
belugabehr: @pgaref Thanks for checking this out too.

Since this is an API, I don't think there should be that level of details/constraints in regards to the implementations.  Anyone implementing this API can choose to do what they want. :)

java/core/src/java/org/apache/orc/impl/Dictionary.java
belugabehr: Donr.

java/core/src/java/org/apache/orc/impl/Dictionary.java
pgaref: Thats true, but the way we pass the text argument here this seems to be an one way option -- wound not mind adding the explanation on getText HT/Tree implementations though

pgaref (approved):

java/core/src/java/org/apache/orc/impl/Dictionary.java
```diff
@@ -44,9 +44,22 @@
...
+   * @param result the holder to copy the dictionary text into
+   * @param position the position where the key was added
    */
   void getText(Text result, int position);
 
+  /**
+   * Given the position index, write the original string, before being encoded,
+   * to the OutputStream.
+   *
+   * @return the number of byte written to the stream
```
dongjoon-hyun, line 58: Shall we add `@param` for `out` and `position`?

java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
```diff
@@ -222,10 +222,10 @@ public void setText(Text result, int offset, int length) {
    * @param out the stream to write to
    * @param offset the first offset to write
    * @param length the number of bytes to write
-   * @throws IOException
+   * @throws IOException if an I/O error occurs
```
dongjoon-hyun, line 225: @belugabehr . Could you revert this file change from this PR?
- First of all, for the line 227 and 228, this is irrelevant. They are only style changes.
- Second, do we need to add this additional comment, `if an I/O error occurs`? This is a Java default semantic. It doesn't add any new information and ORC users don't think this function throws IOException always.

dongjoon-hyun (changes requested):

java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
```diff
@@ -42,4 +44,29 @@ public static void getTextInternal(Text result, int position, DynamicIntArray ke
...
   }
+
+  /**
+   * Write a UTF8 string from the byteArray, using the offset in index-array,
+   * into an OutputStream
+   *
+   * @param out the output stream
+   * @param position position in the keyOffsets
+   * @param keyOffsets starting offset of the key (in byte) in the byte array
+   * @param byteArray storing raw bytes of all key seen in dictionary
```
dongjoon-hyun, line 55: `key` -> `keys`?

dongjoon-hyun:
Sorry for adding last-minute comments after approval. Please consider the above comments.

belugabehr:
> Sorry for adding last-minute comments after approval.

No worries at all. Thanks for being engaged and helping me get these PRs reviewed.  You are very generous with your time.

I'll address your comments.

java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
belugabehr: Sure.  Just trying to be thorough and document everything, but I agree that this change does not add any value for this particular PR.

java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
belugabehr: Thanks.

java/core/src/java/org/apache/orc/impl/Dictionary.java
belugabehr: Done. Thanks.

dongjoon-hyun (approved):
+1, LGTM. Merged to main.

dongjoon-hyun:
Thank you, @belugabehr and @pgaref

java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
```diff
@@ -42,4 +44,29 @@ public static void getTextInternal(Text result, int position, DynamicIntArray ke
...
+  /**
+   * Write a UTF8 string from the byteArray, using the offset in index-array,
+   * into an OutputStream
+   *
+   * @param out the output stream
+   * @param position position in the keyOffsets
+   * @param keyOffsets starting offset of the key (in byte) in the byte array
+   * @param byteArray storing raw bytes of all keys seen in dictionary
+   * @return the number of bytes written to the output stream
+   * @throw IOException if an I/O error occurs
```
dongjoon-hyun, line 57: Unfortunately, this breaks our `Javadoc` generation. Here is the fix.
- https://github.com/apache/orc/pull/867

PR #738, opened 2021-07-07
ORC-833: Calculate nextVector Batch Size Once

belugabehr:

The number of iterations of the loops in RunLengthIntegerReaderV2 nextVector look at two distinct, but known, values. Instead of check each value in each loop, pre-compute the batch size.

Code clarity, and if the optimizer doesn't already do it automagically, saves a few branches.

No changes in functionality.  Uses existing unit tests.


dongjoon-hyun (approved):
+1, LGTM. Thank you, @belugabehr .
I updated an indentation at one line.

dongjoon-hyun:
Merged to main.

PR #739, opened 2021-07-07
ORC-838: Simplify `compareTo/equals/putBuffer` of ByteBufferAllocatorPool

belugabehr:
Simplify ByteBufferAllocatorPool Key class. Cache hash value for Key as it's computed more than once: putBuffer to insert into Tree and getBuffer to remove it from tree.

Less code. Performance.

No changes to functionality. Use existing unit tests.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
@@ -570,7 +570,7 @@ public class RecordReaderUtils {
       // If our key is not unique on the first try, try again
       do {
         key = new Key(buffer.capacity(), currentGeneration++);
-      while (tree.putIfAbsent(key, buffer) != null);
+      } while (tree.putIfAbsent(key, buffer) != null);
     }
   }
 }
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
@@ -504,13 +504,10 @@ public class RecordReaderUtils {
     private static final class Key implements Comparable<Key> {
       private final int capacity;
       private final long insertionGeneration;
-      private final int hash;
 
       Key(int capacity, long insertionGeneration) {
         this.capacity = capacity;
         this.insertionGeneration = insertionGeneration;
-        this.hash = new HashCodeBuilder().append(capacity).append(insertionGeneration)
-            .toHashCode();
       }
 
       @Override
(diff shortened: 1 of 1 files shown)


java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -504,38 +504,33 @@ static void readDiskRanges(FSDataInputStream file,
     private static final class Key implements Comparable<Key> {
       private final int capacity;
       private final long insertionGeneration;
+      private final int hash;
 
-      Key(int capacity, long insertionGeneration) {
+      Key (int capacity, long insertionGeneration) {
```
kbendick, line 509: Nit: Is this extra space expected?

belugabehr: Nope.  Not sure why I even did that.  Thanks for pointing it out.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -504,38 +504,33 @@ static void readDiskRanges(FSDataInputStream file,
     private static final class Key implements Comparable<Key> {
       private final int capacity;
       private final long insertionGeneration;
+      private final int hash;
 
       Key(int capacity, long insertionGeneration) {
         this.capacity = capacity;
         this.insertionGeneration = insertionGeneration;
+        this.hash = new HashCodeBuilder().append(capacity).append(insertionGeneration)
```
guiyanakuang, line 512: I found that the equals and hashCode methods of the Key are not used at all, because the TreeMap implementation logic only use of compareTo. It is redundant to put the hashCode calculation statement in the constructor, and in fact there is no need to override them.

guiyanakuang: I also agree with leave equals and hashCode. Just putting the hashcode calculation in the constructor seems redundant in the current implementation. 

ByteBufferAllocatorPool needs to support multiple calls to getBuffer with the same length before the buffer is returned for reuse. I think there may not be a suitable data structure other than TreeMap.

guiyanakuang: sorry, I didn't notice revert hash cache.

dongjoon-hyun:
Thank you for narrowing down the scoping by reverting.

dongjoon-hyun (approved):
+1, LGTM.
Merged to main for Apache ORC 1.8.0.

dongjoon-hyun:
Thank you, @belugabehr , @kbendick , @guiyanakuang .

PR #740, opened 2021-07-07
ORC-836: StringGroupFromDoubleTreeReader Use Double toString

belugabehr:
When converting from Double to String, use the Double class methods directly. Lower overhead of encoding by using ASCII.

Performance.

No change in functionality.  Use existing unit tests.


belugabehr:
@dongjoon-hyun This is a trivial one.  Can you please take a look? Thanks!

dongjoon-hyun (commented):
Thank you for pinging me, @belugabehr .
- The title looks wrong to me: `Use Long toString`. Do you want to say `Use Double.toString`?
- Could you check if we have more instances like this?

belugabehr:
@dongjoon-hyun Thanks!!

You are correct, I meant to say "Double" instead of ~Long~

I changed the commit message.

All other instances correctly use the `toString` instead of `valueOf`, this was the only one.

dongjoon-hyun (approved):
+1, LGTM. Thank you for confirming, @belugabehr .
Merged to main.

PR #741, opened 2021-07-08
ORC-835: Cache TRUE/FALSE Bytes in StringGroupFromBooleanTreeReader

belugabehr:
Cache byte value for TRUE and FALSE strings.

Performance.

No change in functionality. Uses existing unit tests.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
@@ -1068,6 +1068,8 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
   }
 
   public static class StringGroupFromBooleanTreeReader extends StringGroupFromAnyIntegerTreeReader {
+    private static final TRUE_BYTES = "TRUE".getBytes(StandardCharsets.US_ASCII);
+    private static final FALSE_BYTES = "FALSE".getBytes(StandardCharsets.US_ASCII);
 
     StringGroupFromBooleanTreeReader(int columnId, TypeDescription fileType,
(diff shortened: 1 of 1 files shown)


java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -1068,6 +1068,8 @@ public void nextVector(ColumnVector previousVector,
   }
 
   public static class StringGroupFromBooleanTreeReader extends StringGroupFromAnyIntegerTreeReader {
+    private static final byte[] TRUE_BYTES = "TRUE".getBytes(StandardCharsets.US_ASCII);
+    private static final byte[] FALSE_BYTES = "FALSE".getBytes(StandardCharsets.US_ASCII);
```
kbendick, line 1072: In my opinion, it would be better to continue to use the UTF_8 charset to reduce the impact of this change.


Is there a benefit to not using it? My understanding is that the final result should be more or less equivalent because the letters are part of ASCII but I'm not an expert on UTF8 codepoints.

belugabehr: The two will produce the same output since all numeric values (0-9) are ASCII values.  There is slightly less overhead with using ASCII directly as opposed to using UTF8 which will discover that all values are ASCII.

kbendick (approved):

belugabehr:
@kbendick OK to merge? Thanks!

dongjoon-hyun (approved):
+1, LGTM. Thank you, @belugabehr and @kbendick .
Merged to main for Apache ORC 1.8.0

PR #746, opened 2021-07-12
ORC-843: Enforce Non-Null Fields in TypeDescription

belugabehr:
Enforce fields in `TypeDescription` not be `null`.

Makes code less error-prone and removes the need to remember to check for `null` especially since it's not documented.

### DIFF
--- a/java/core/src/java/org/apache/orc/TypeDescription.java
+++ b/java/core/src/java/org/apache/orc/TypeDescription.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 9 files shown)


java/core/src/java/org/apache/orc/TypeDescription.java
```diff
@@ -50,6 +51,25 @@
...
+  private int precision = DEFAULT_PRECISION;
+  private int scale = DEFAULT_SCALE;
+
+  public TypeDescription(final Category category) {
+    this.category = Objects.requireNonNull(category);
+    this.children =
+        (category.isPrimitive) ? Collections.emptyList() : new ArrayList<>();
+    this.fieldNames = (category != Category.STRUCT) ? Collections.emptyList()
+        : new ArrayList<>();
+  }
```
dongjoon-hyun, line 71: Please don't relocate the position.

dongjoon-hyun (changes requested):
Hi, @belugabehr . Could you minimize your PR?

PR #748, opened 2021-07-12
ORC-842: Remove newKey from StringHashTableDictionary

belugabehr:
When adding keys to `StringHashTableDictionary`, do not copy data into intermediate container.

Performance.

No changes to functionality. Use existing unit tests. Existing unit test was updated to use new internal mechanism for testing.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
+++ b/java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
@@ -23,7 +23,6 @@ import java.io.OutputStream;
 import java.util.Arrays;
 
 import org.apache.hadoop.io.Text;
-import org.apache.hadoop.io.WritableComparator;
 
 
 /**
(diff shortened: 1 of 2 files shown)


belugabehr:
Under the covers, Text#hashcode uses `WritableComparator.hashBytes`.  If we remove the dependency on Hadoop, this will have to change, but preserves current functionality for now.

java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
```diff
@@ -173,10 +170,18 @@ public int size() {
...
-    return Math.floorMod(text.hashCode(), capacity);
+    return getIndex(text.getBytes(), 0, text.getLength());
+  }
+
+  /**
+   * Compute the hash value and find the corresponding index. Uses same
+   * implementation as {@code Text#hashCode()}.
+   */
+  int getIndex(final byte[] bytes, final int offset, final int length) {
+    return Math.floorMod(WritableComparator.hashBytes(bytes, offset, length),
```
kbendick, line 183: Now that there is an implicit dependency on the hash function of `Test#hashCode` and this class, is there any way to enforce that a change in the hash function would reflect here as well?

pgaref: Thats exactly what I had in mind with my comment above..

belugabehr: I am happy to sub this out with non-hadoop hashcode if you would like.  Like I said, I came up with this only to preserve the previous behavior.

autumnust:
@pgaref  Thanks for pinging. Changes LGTM, and +1 for non-hadoop hashcode method.

belugabehr:
Changed to no longer use Hadoop for hash code implementation.

belugabehr:
Passes locally but looks like something is failing with a build download:

```none
[  6%] Performing download step (download, verify and extract) for 'protobuf_ep'
CMake Error at protobuf_ep-stamp/protobuf_ep-download-RELWITHDEBINFO.cmake:49 (message):
```

pgaref:
Thank you all! This is now merged to main

PR #751, opened 2021-07-15
ORC-848: Recycle Internal Buffer in StringHashTableDictionary

belugabehr:
When clearing a StringHashTableDictionary, re-use the existing internal buffer.

Nit performance.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
+++ b/java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
@@ -75,18 +75,19 @@ public class StringHashTableDictionary implements Dictionary {
     this.capacity = initialCapacity;
     this.loadFactor = loadFactor;
     this.keyOffsets = new DynamicIntArray(initialCapacity);
-    initHashBuckets(initialCapacity);
+    initHashBuckets(false);
     this.threshold = (int)Math.min(initialCapacity * loadFactor, MAX_ARRAY_SIZE + 1);
   }
 
-  private void initHashBuckets(int capacity) {
-    DynamicIntArray[] buckets = new DynamicIntArray[capacity];
-    for (int i = 0; i < capacity; i++) {
+  private void initHashBuckets(final boolean reinit) {
+    final DynamicIntArray[] newBuckets =
+        (reinit) ? this.hashBuckets : new DynamicIntArray[this.capacity];
+    for (int i = 0; i < this.capacity; i++) {
       // We don't need large bucket: If we have more than a handful of collisions,
       // then the table is too small or the function isn't good.
-      buckets[i] = createBucket();
+      newBuckets[i] = createBucket();
     }
-    hashBuckets = buckets;
+    hashBuckets = newBuckets;
   }
 
   private DynamicIntArray createBucket() {
@@ -112,7 +113,7 @@ public class StringHashTableDictionary implements Dictionary {
   public void clear() {
     byteArray.clear();
     keyOffsets.clear();
-    initHashBuckets(this.capacity);
+    initHashBuckets(true);
   }
 
   @Override


java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
```diff
@@ -75,18 +75,19 @@ public StringHashTableDictionary(int initialCapacity, float loadFactor) {
...
-    DynamicIntArray[] buckets = new DynamicIntArray[capacity];
-    for (int i = 0; i < capacity; i++) {
+  private void initHashBuckets(final boolean reinit) {
+    final DynamicIntArray[] newBuckets =
+        (reinit) ? this.hashBuckets : new DynamicIntArray[this.capacity];
+    for (int i = 0; i < this.capacity; i++) {
       // We don't need large bucket: If we have more than a handful of collisions,
       // then the table is too small or the function isn't good.
-      buckets[i] = createBucket();
+      newBuckets[i] = createBucket();
```
wgtmac, line 88: The DynamicIntArray-typed elements are re-created anyway. Just curious if we can see obvious gain.

java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
```diff
@@ -75,18 +75,19 @@ public StringHashTableDictionary(int initialCapacity, float loadFactor) {
...
+    initHashBuckets(false);
     this.threshold = (int)Math.min(initialCapacity * loadFactor, MAX_ARRAY_SIZE + 1);
   }
 
-  private void initHashBuckets(int capacity) {
-    DynamicIntArray[] buckets = new DynamicIntArray[capacity];
-    for (int i = 0; i < capacity; i++) {
+  private void initHashBuckets(final boolean reinit) {
+    final DynamicIntArray[] newBuckets =
+        (reinit) ? this.hashBuckets : new DynamicIntArray[this.capacity];
```
kbendick, line 84: Do we need to call some kind of `clear` or null out the current elements to ensure they're able to be GC'd as soon as possible? I've noticed that this new `hash` dictionary uses more memory than the `rbtree` dictionary in some tests I'm working on.

java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
belugabehr: Hello. Thanks for the review. The purpose of this patch is to preserve the `buckets` array since this method simply repopulates the contents. There's no reason to throw away a 4KB array to simply replace it with a new one .

java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
belugabehr: As soon as each element in `buckets` is overwritten with a new bucket, it becomes available for GC.  There is no value in writing a null value and the immediately replacing it with a new bucket.

java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
```diff
@@ -75,18 +75,19 @@ public StringHashTableDictionary(int initialCapacity, float loadFactor) {
...
     this.keyOffsets = new DynamicIntArray(initialCapacity);
-    initHashBuckets(initialCapacity);
+    initHashBuckets(false);
     this.threshold = (int)Math.min(initialCapacity * loadFactor, MAX_ARRAY_SIZE + 1);
   }
 
-  private void initHashBuckets(int capacity) {
-    DynamicIntArray[] buckets = new DynamicIntArray[capacity];
-    for (int i = 0; i < capacity; i++) {
+  private void initHashBuckets(final boolean reinit) {
```
pgaref, line 82: I would even remove the parameter of this method -- we just need to ensure that hashBuckets is initialized right?
So we can just check if its non-null (assuming the capacity remains the same) otherwise create the new buckets

belugabehr: Ha. Yes. You're absolutely correct.  I will change this to simplify.

pgaref (approved):
Thanks for updating the PR @belugabehr -- latest patch LGTM

PR #735, opened 2021-07-02
ORC-830: Do Not Copy String When Adding to StringHashTableDictionary

belugabehr:
### What changes were proposed in this pull request?
When there is a collision adding a value into a StringHashTableDictionary, a temp Text object is created and then each value in the byte array is copied into the temp Text until a match is found (or worst-case scenario, a match is not found and every value is loaded).

Instead of loading (copying) the values, just compare directly against the byte array without copying the data into a intermediate (temp) buffer.

### Why are the changes needed?
Performance.  StringTreeWriter#writeBatch() consumes some number the cycles, much of which is spent in getText().  By Changing this getText() implementation (removing the copy), the number of cycles consumed is decreased measurably.

### How was this patch tested?
No functionality change, utilized existing unit tests.

PR #754, opened 2021-07-19
ORC-852: Allow DynamicByteArray to return a ByteBuffer

belugabehr:
Allow `DyanmicByteArray` and `Dictionary` to return a ByteBuffer of its content.

Performance.

No change to functionality. Existing unit tests used.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/Dictionary.java
+++ b/java/core/src/java/org/apache/orc/impl/Dictionary.java
@@ -22,6 +22,7 @@ import org.apache.hadoop.io.Text;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.ByteBuffer;
 
 
 /**
@@ -51,6 +52,8 @@ public interface Dictionary {
    * @param position the position where the key was added
    */
   void getText(Text result, int position);
+
+  ByteBuffer getText(int position);
 
   /**
    * Given the position index, write the original string, before being encoded,
--- a/java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
@@ -47,6 +47,15 @@ public class DictionaryUtils {
     byteArray.setText(result, offset, length);
   }
 
+  /**
+   * Return a {@code ByteBuffer} containing the data at a certain offset within a
+   * {@code DynamicByteArray}.
+   *
+   * @param position position in the keyOffsets
+   * @param keyOffsets starting offset of the key (in byte) in the byte array
+   * @param byteArray storing raw bytes of all keys seen in dictionary
+   * @return the number of bytes written to the output stream
+   */
   public static ByteBuffer getTextInternal(int position, DynamicIntArray keyOffsets,
       DynamicByteArray byteArray) {
     final int offset = keyOffsets.get(position);
--- a/java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
@@ -21,6 +21,7 @@ import org.apache.hadoop.io.Text;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.ByteBuffer;
 
 public class DictionaryUtils {
   private DictionaryUtils() {
@@ -44,6 +45,17 @@ public class DictionaryUtils {
       length = keyOffsets.get(position + 1) - offset;
     }
     byteArray.setText(result, offset, length);
+  }
+
+  public static ByteBuffer getTextInternal(int position, DynamicIntArray keyOffsets, DynamicByteArray byteArray) {
+    final int offset = keyOffsets.get(position);
+    final int length;
+    if (position + 1 == keyOffsets.size()) {
+      length = byteArray.size() - offset;
+    } else {
+      length = keyOffsets.get(position + 1) - offset;
+    }
+    return byteArray.get(offset, length);
   }
 
   /**
--- a/java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
@@ -47,7 +47,8 @@ public class DictionaryUtils {
     byteArray.setText(result, offset, length);
   }
 
-  public static ByteBuffer getTextInternal(int position, DynamicIntArray keyOffsets, DynamicByteArray byteArray) {
+  public static ByteBuffer getTextInternal(int position, DynamicIntArray keyOffsets,
+      DynamicByteArray byteArray) {
     final int offset = keyOffsets.get(position);
     final int length;
     if (position + 1 == keyOffsets.size()) {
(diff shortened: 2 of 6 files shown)


java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
```diff
@@ -294,6 +295,16 @@ public void setByteBuffer(ByteBuffer result, int offset, int length) {
     return result;
   }
 
+  public ByteBuffer get(int offset, int length) {
+    final int currentChunk = offset / chunkSize;
+    final int currentOffset = offset % chunkSize;
+    final int currentLength = Math.min(length, chunkSize - currentOffset);
+    if (currentLength == length) {
+      return ByteBuffer.wrap(data[currentChunk], currentOffset, length);
```
belugabehr, line 303: This is the "win" right here.  If the data requested is contiguous within a single data chunk, then a view to the data can be returned without needing to copy into a buffer.

belugabehr:
Using ByteBuffer here will allow us to remove the `Text` version of this method in a subsequent release and move the project closer to removing Hadoop as a dependency (ORC-508)

pgaref:
@belugabehr can we resolve conflicts here please?

belugabehr:
@pgaref Will do, however, I think it would be better to get #735, #748 in first since they remove some instances of `Text` and then I can re-work this one to convert the remaining pieces to `ByteBuffer`

java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java
```diff
@@ -120,6 +122,11 @@ public void getText(Text result, int positionInKeyOffset) {
     DictionaryUtils.getTextInternal(result, positionInKeyOffset, this.keyOffsets, this.byteArray);
   }
 
+  @Override
+  public ByteBuffer getText(int positionInKeyOffset) {
+    return DictionaryUtils.getTextInternal(positionInKeyOffset, this.keyOffsets, this.byteArray);
+  }
```
kbendick, line 128: Is it safe to be exporting this ByteBuffer? If it is reused / overwritten in some way, will that affect any of the existing code.

Specifically, I'm wondering about the now returned `ByteBuffer` from `setByteBuffer`, as the data can be returned without any copying.

belugabehr: Hey @kbendick,

> Is it safe to be exporting this ByteBuffer? If it is reused / overwritten in some way, will that affect any of the existing code.

I guess everyone has a different scale of what "safe" is.  If your concern is regarding the callers changing values in the returned ByteBuffer affected the data stored in the Dictionary, then yes, that is a possibility right now.  There are two ways to handle that:

1. Document this behavior and assume the caller will handle the data responsibly
2. Return a read-only `ByteBuffer` which will prevent callers from modifying the data

I'm not sure I understand the second part of your question.

belugabehr: On the second part, if you are referring to the new value returned by `setByteBuffer`, there's no functionality change there.  I am simply returning the passed in value for convenience:

```java
ByteBuffer bb = ByteBuffer.allocate(16);
dictionary.setByteBuffer(bb, 0, 16);
System.out.println(bb);

// v.s.

ByteBuffer bb = ByteBuffer.allocate(16);
System.out.println(dictionary.setByteBuffer(bb, 0, 16));
```

kbendick: That was indeed what I was referring to. That answers my question. Thank you @belugabehr!

belugabehr:
@pgaref @kbendick Got the other items merged into the project, so this one is now ready.  Thanks!

belugabehr:
@pgaref @kbendick Care to take a look over once more? Thank you!

java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
```diff
@@ -46,6 +47,18 @@ public static void getTextInternal(Text result, int position,
     byteArray.setText(result, offset, length);
   }
 
+  public static ByteBuffer getTextInternal(int position, DynamicIntArray keyOffsets,
```
pgaref, line 50: Shall we name this explicitly getTexBuffertInternal ?

java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
```diff
@@ -46,6 +47,18 @@ public static void getTextInternal(Text result, int position,
...
+      DynamicByteArray byteArray) {
+    final int offset = keyOffsets.get(position);
+    final int length;
+    if (position + 1 == keyOffsets.size()) {
+      length = byteArray.size() - offset;
+    } else {
+      length = keyOffsets.get(position + 1) - offset;
+    }
```
pgaref, line 58: Length calculation is repeated at least 5 times already -- lets move this to a helper function

java/core/src/java/org/apache/orc/impl/Dictionary.java
```diff
@@ -52,6 +53,8 @@
    */
   void getText(Text result, int position);
 
+  ByteBuffer getText(int position);
```
pgaref, line 56: maybe name getTextBuffer instead?

pgaref (commented):
Hey @belugabehr  -- left some comments here, let me know what you think

java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
dongjoon-hyun, line 50: Could you add a function description please, @belugabehr ?

java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
dongjoon-hyun: I agree with @pgaref 's perspective. Could you elaborate a little bit more about your refactoring idea? Is that better?
>  I actually think this DictionaryUtils class can probably be removed and roll the functionality into StringHashTable directly.

dongjoon-hyun (approved):
+1, LGTM with one minor comment and one question. Thank you, @belugabehr .

java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
belugabehr: Done!

belugabehr:
@dongjoon-hyun I made the change regarding JavaDocs.  Please review.

In regards to my comment:

> I actually think this DictionaryUtils class can probably be removed and roll the functionality into StringHashTable directly.

Looking into it a bit more, I think the best thing to do is to create a new data structure that stores the `DynamicByteArray` and its offsets (`DynamicIntArray`).  It should govern access and resizing the two as needed.  That would remove `DictionaryUtils`, make the methods more unit test-able and remove a bunch of logic from the `StringHashTableDictionary` class that deals with maintenance (like growing the arrays).  Alas, I still believe this should be accomplished as a separate issue/task.

dongjoon-hyun:
Thank you, @belugabehr .
Merged to main.

PR #755, opened 2021-07-20
ORC-853: Optimize writeDouble Implementation

belugabehr:
Use the optimized Apache Avro implementation of writing a `double` to a stream.  Remove extra method which supports this functionality; condense it into a single method.

Performance. Had a measurable impact on the Driver performance harness for writing `double` values.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/SerializationUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/SerializationUtils.java
@@ -173,7 +173,6 @@ public final class SerializationUtils {
     output.write(writeBuffer, 0, 8);
   }
 
-
   /**
    * Write the arbitrarily sized signed BigInteger in vint format.
    *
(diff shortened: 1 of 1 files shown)


pgaref (approved):
Code rewrite LGTM -- @belugabehr is the benefit consistent across runs?
Let's also point to the AVRO path with comment please.

belugabehr:
@pgaref Thanks for the review.  I'll add additional comments.

Yes. I only posted the results of one run, but it was representative of the greater series of tests.  Anyway, the reason I even thought to do this was because I have worked a bit with Avro in the pass and this strange implementation really stuck with me.  Some time ago, I found it super interesting and I tried different ways to out-perform it, but I was never able to.  Now, TBH there is some amount of YMMV here, as this may be sensitive to CPU architecture, but the person that implemented this in the first place found it was optimal, and I could only confirm that with my personal setup, so there is some merit to it.

pgaref:
Merged, Thanks @belugabehr !

PR #756, opened 2021-07-20
ORC-854: Optimize ReadFully for Full Reads

belugabehr:
Optimize ORC's readFully method for a condition where the buffer is always fully read.

Better performance reading Float/Long/Double values.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/SerializationUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/SerializationUtils.java
@@ -138,22 +138,19 @@ public final class SerializationUtils {
         + ((long) (readBuffer[7] & 0xff) << 56));
   }
 
-  private void readFully(final InputStream in, final byte[] buffer, final int len)
+  private void readFully(final InputStream in, final byte[] buffer, int len)
       throws IOException {
-    final int firstRead = in.read(buffer, 0, len);
-    if (firstRead != len) {
-      if (firstRead < 0) {
+    int offset = 0;
+    for (;;) {
+      final int n = in.read(buffer, offset, len);
+      if (n == len) {
+        return;
+      }
+      if (n < 0) {
         throw new EOFException("Read past EOF for " + in);
       }
-      int remaining = len - firstRead;
-      while (remaining > 0) {
-        final int location = len - remaining;
-        final int count = in.read(buffer, location, remaining);
-        if (count < 0) {
-          throw new EOFException("Read past EOF for " + in);
-        }
-        remaining -= count;
-      }
+      offset += n;
+      len -= n;
     }
   }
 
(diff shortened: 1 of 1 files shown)


pgaref (approved):
Thanks @belugabehr  ! Changes LGTM

kbendick (commented):
Are we sure that the buffers / `InputStreams` passed to `readFully` do not contain data after their usage? And that they have not been read previously such thatream the entire contents of the `InputStream` truly represents all of the data for the one usage of that buffer?

I know by name it sounds like it, but my experience is that sometimes names can be misleading. As an example, perhaps the user meant read fully the data from offset to offset + length.

Also, since we've been adding a lot of smaller optimizations recently (which can indeed have a big effect in some cases), would it be a good idea to establish something like JMH benchmarks or something to ensure that the optimizations make sense? I know JMH isn't perfect, and many of the optimization PRs have been visually apparent to be optimizing, but throwing that out there to see what others think. 🙂

belugabehr:
> Are we sure that the buffers / InputStreams passed to readFully do not contain data after their usage?

In this context, the "readFully" refers to the the buffer that is passed in, not that the `InputStream` itself is read fully.  That is, the buffer is fully populated (unless EOF is seen first).  The typical `read` function of `InputStream` can return any number of bytes, so you have to be able to handle that, and continuously loop to extract enough data to fully fill the buffer.

belugabehr:
> Also, since we've been adding a lot of smaller optimizations recently (which can indeed have a big effect in some cases), would it be a good idea to establish something like JMH benchmarks or something to ensure that the optimizations make sense? I know JMH isn't perfect, and many of the optimization PRs have been visually apparent to be optimizing, but throwing that out there to see what others think.

I think something like ORC should definitely have that, and looking at the code, there might even be some JMH tests in there, but we would need additional documentation on how to do that.

For now, I am using the performance benchmarks currently available, using simple timers and/or examining under a profiler.  I've come up with quite a few ideas for PRs, but not all of them have shown much perf improvement and I don't bother submitting them. :)

kbendick:
> > Are we sure that the buffers / InputStreams passed to readFully do not contain data after their usage?
> 
> In this context, the "readFully" refers to the the buffer that is passed in, not that the `InputStream` itself is read fully. That is, the buffer is fully populated (unless EOF is seen first). The typical `read` function of `InputStream` can return any number of bytes, so you have to be able to handle that, and continuously loop to extract enough data to fully fill the buffer.

Ah I misspoke. I was indeed referring to the passed in buffers. That's what I get for reviewing code really late at night.

Looking at it with fresher eyes, it does in fact look like there is no behavioral change between the two. My actual concern was that the buffers passed in were not necessarily intended to be read from the beggining / their current location / mark (and hence the offset parameter). That was my actual concern. But seeing where the code had to be changed, it doesn't appear as though anywhere it's used the offset is anything but zero, so this is functionally the same.

Thanks @belugabehr!

kbendick (approved):
LGTM.

belugabehr:
@kbendick @pgaref Good to merge? :)

wgtmac:
> @kbendick @pgaref Good to merge? :)

I have merged it. Thanks @belugabehr for contribution and @kbendick @pgaref for review!

PR #773, opened 2021-07-26
ORC-867: Upgrade hive-storage-api to 2.8.1

dongjoon-hyun:

This PR aims to upgrade `hive-storage-api` from 2.7.2 to 2.8.1.

This will bring the latest improvement and bug fixes.

Pass the CIs.

### DIFF
--- a/java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
+++ b/java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
@@ -311,6 +311,9 @@ public class TestRecordReaderImpl {
     }
 
     @Override
+    public int getId() { return -1; }
+
+    @Override
     public List<Object> getLiteralList() {
       return null;
     }
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -72,7 +72,7 @@
     <min.hadoop.version>2.2.0</min.hadoop.version>
     <hadoop.version>2.7.3</hadoop.version>
     <tools.hadoop.version>2.10.1</tools.hadoop.version>
-    <storage-api.version>2.7.2</storage-api.version>
+    <storage-api.version>2.8.1</storage-api.version>
     <zookeeper.version>3.7.0</zookeeper.version>
     <maven.version>3.6.3</maven.version>
     <maven-assembly-plugin.version>3.3.0</maven-assembly-plugin.version>


dongjoon-hyun:
cc @williamhyun and @pavibhai

kbendick (approved):
This is great. Thank you @dongjoon-hyun!

Is there anything we can change / remove from the build (pom) file? I took a look, and it didn't seem like it as the only explicit dependency is on the commons-lang3, but felt worth bringing up just in case.

dongjoon-hyun:
Thank you for review and approval, @kbendick . 
Since `commons-lang3` is already in `core` module, there is nothing to change for this.
- https://github.com/apache/orc/blob/main/java/core/pom.xml#L48

dongjoon-hyun:
Oh..  `guava` dependency seems to be promoted from `test` dependency to `compile` dependency.
Let me check my PR to `Draft` first.

java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
```diff
@@ -310,6 +310,9 @@ public Object getLiteral() {
       return null;
     }
 
+    @Override
```
pavibhai, line 313: Looks like we will do another release of StorageAPI. I was wondering if we should push this as a default?

dongjoon-hyun: Yes, we need 2.8.1. This PR will stay here and will retarget to 2.8.1, @pavibhai .

BTW, what do you mean by the following?
> we should push this as a default?

pavibhai: I was wondering if we should add a default implementation to this new method into PredicateLeaf. But not required if this is the only failure we are seeing.

dongjoon-hyun:
For this one, we will retry when `hive-storage-api` 2.8.1 is ready.
-

omalley (approved):
LGTM, other than the pom needs to switch to storage-api 2.8.1. Build & unit tests run fine.

dongjoon-hyun:
Thank you, @omalley . I also saw that 2.8.1 is published already.
- https://repo1.maven.org/maven2/org/apache/hive/hive-storage-api/2.8.1/

The pom is updated and I rebased this PR.

dongjoon-hyun:
If CIs pass, I'll merge this to main/branch-1.7.

dongjoon-hyun:
All Java test passed with GitHub Action (Java8/11/16/17-ea).

PR #787, opened 2021-07-29
ORC-880: Reduce Visibility of DictionaryUtils

belugabehr:
Reduce visibility of class `DictionaryUtils`.

Prevent a situation where an external user has a dependency on this utility class and therefore the internals cannot change.

No changes to functionality.  Use existing unit tests.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/DictionaryUtils.java
@@ -23,7 +23,6 @@ import java.io.OutputStream;
 import org.apache.hadoop.io.Text;
 
 class DictionaryUtils {
-
   private DictionaryUtils() {
     // Utility class does nothing in constructor
   }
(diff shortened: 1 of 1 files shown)


dongjoon-hyun (changes requested):
+1 for the intention, but -1 for this minor PR change because it's not worth of doing this.

In the same `impl` directory, we have many `public` Utils classes.
```
$ ls *Utils.java
CryptoUtils.java        DictionaryUtils.java    ParserUtils.java        SerializationUtils.java Utf8Utils.java
DateUtils.java          OrcAcidUtils.java       RecordReaderUtils.java  TypeUtils.java
```

belugabehr:
@dongjoon-hyun Would you rather I abandon this effort altogether or apply this same change to all the `util` classes?

dongjoon-hyun:
We cannot hide all of them because it's an unnecessarily breaking change. At 1.7.0, we already recovered many things to help the downstreams' migration. Although they are designed to be internal, someone use them. So, I'd rather close ORC-880 as 'Won't Do'.

BTW, I really agree with your architectural design mind and your intention here. Sorry for not accepting your efforts, @belugabehr .

dongjoon-hyun:
Thank you for closing this, @belugabehr .

PR #803, opened 2021-08-02
ORC-897: Optimization loop termination condition

guiyanakuang:

```java
private static boolean sameKeys(EncryptionKey[] first,
                                  EncryptionKey[] next) {
    if (first.length != next.length) {
      return false;
    }
    for(int k = 0; k < first.length && k < next.length; ++k) {
      if (!first[k].getKeyName().equals(next[k].getKeyName()) ||
          first[k].getKeyVersion() != next[k].getKeyVersion() ||
          first[k].getAlgorithm() != next[k].getAlgorithm()) {
        return false;
      }
    }
    return true;
  }
```
There are five similar codes.
The length of `first` and `next` are the same, otherwise the loop is unreachable.
k <next.length is redundant conditions.
I think they can be removed.

Optimization loop termination condition in readerIsCompatible method

Pass the CIs.

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcFile.java
+++ b/java/core/src/java/org/apache/orc/OrcFile.java
@@ -1039,7 +1039,7 @@ public class OrcFile {
     if (first.length != next.length) {
       return false;
     }
-    for(int k = 0; k < first.length && k < next.length; ++k) {
+    for(int k = 0; k < first.length; ++k) {
       if (!first[k].getKeyName().equals(next[k].getKeyName()) ||
           first[k].getKeyVersion() != next[k].getKeyVersion() ||
           first[k].getAlgorithm() != next[k].getAlgorithm()) {
@@ -1054,7 +1054,7 @@ public class OrcFile {
     if (first.length != next.length) {
       return false;
     }
-    for(int k = 0; k < first.length && k < next.length; ++k) {
+    for(int k = 0; k < first.length; ++k) {
       if (!first[k].getName().equals(next[k].getName())) {
         return false;
       }
@@ -1063,7 +1063,7 @@ public class OrcFile {
       if (firstParam.length != nextParam.length) {
         return false;
       }
-      for(int p=0; p < firstParam.length && p < nextParam.length; ++p) {
+      for(int p=0; p < firstParam.length; ++p) {
         if (!firstParam[p].equals(nextParam[p])) {
           return false;
         }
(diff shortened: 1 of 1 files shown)


dongjoon-hyun (commented):
Thank you for making a PR @guiyanakuang .

Could you elaborate why this is redundant in the PR description? For example, `if (first.length != next.length) {` ensures that they are the same?
> Remove redundant conditions

Are these all instances like this?

pgaref:
> Thank you for making a PR @guiyanakuang .
> 
> Could you elaborate why this is redundant in the PR description?
> 
> > Remove redundant conditions
> 
> Are these all instances like this?

Guess we already make sure   __first.length == next.length__ 
https://github.com/apache/orc/pull/803/files#diff-4c43c10da88510010bb511ba6e6892bb5ea6cc29be1b0b9d97c3bbaae5d382e2L1039

dongjoon-hyun:
Yep, @pgaref . I noticed that and wanted to have it in the PR description. :)

pgaref (approved):
+1 tests pending

dongjoon-hyun:
Before merging, it would be great if we make it sure that these are all instances like this (at least in this file or module), @pgaref .

dongjoon-hyun (approved):
Thank you for updating the PR description, @guiyanakuang .
+1, LGTM.

pgaref:
> Before merging, it would be great if we make it sure that these are all instances like this (at least in this file or module), @pgaref .


Sure, all seem to be using firstRoots/nextRoots or first/next length -- happy with the changes

guiyanakuang:
Thank you so much for review and approval, @dongjoon-hyun !

guiyanakuang:
Thank you, @pgaref !

dongjoon-hyun:
BTW, @pgaref . When you click the squash button, GitHub shows you the commit title and body input boxes.

The content of those input boxes came from the contributor's commit. So, you need to copy and past the PR title and PR description into it manually.

pgaref:
> BTW, @pgaref . When you click the squash button, GitHub shows you the commit title and body input boxes.
> 
> The content of those input boxes came from the contributor's commit. So, you need to copy and past the PR title and PR description into it manually. Then, you can make a correct Apache ORC commit instead of using the user's commit message.
> 
> Screen Shot 2021-08-02 at 9 24 09 AM

Right, in this case I wanted to avoid pasting the code sample above. You are right thought we should keep it consistent.

Cheers

PR #810, opened 2021-08-03
ORC-904: Use Map for userMetadata in ReaderImpl

guiyanakuang:

ReaderImpl.java
```java
  @Override
  public ByteBuffer getMetadataValue(String key) {
    for(OrcProto.UserMetadataItem item: userMetadata) {
      if (item.hasName() && item.getName().equals(key)) {
        return item.getValue().asReadOnlyByteBuffer();
      }
    }
    throw new IllegalArgumentException("Can't find user metadata " + key);
  }

  @Override
  public boolean hasMetadataValue(String key) {
    for(OrcProto.UserMetadataItem item: userMetadata) {
      if (item.hasName() && item.getName().equals(key)) {
        return true;
      }
    }
    return false;
  }
```
I think the data structure of userMetadata can be modified to map,  avoid traversal

There may be performance problems if there are more userMetadata keys

Pass the CIs.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ReaderImpl.java
@@ -81,7 +81,7 @@ public class ReaderImpl implements Reader {
   private final int metadataSize;
   protected final List<OrcProto.Type> types;
   private final TypeDescription schema;
-  private final Map<String, ByteString> userMetadata;
+  private final UserMetadata userMetadata;
   private final List<OrcProto.ColumnStatistics> fileStats;
   private final List<StripeInformation> stripes;
   protected final int rowIndexStride;
@@ -206,6 +206,45 @@ public class ReaderImpl implements Reader {
     }
   }
 
+  public static class UserMetadata {
+    private final List<OrcProto.UserMetadataItem> innerUserMetadata;
+
+    private Map<String, ByteString> metadataCache;
+
+    public UserMetadata(List<OrcProto.UserMetadataItem> userMetadata) {
+      this.innerUserMetadata = userMetadata;
+    }
+
+    public List<String> getMetadataKeys() {
+      List<String> result = new ArrayList<>();
+      for(OrcProto.UserMetadataItem item: innerUserMetadata) {
+        result.add(item.getName());
+      }
+      return result;
+    }
+
+    private void lazyInitCache() {
+      if (metadataCache == null) {
+        metadataCache = new TreeMap<>();
+        for(OrcProto.UserMetadataItem item: innerUserMetadata) {
+          metadataCache.putIfAbsent(item.getName(), item.getValue());
+        }
+      }
+    }
+
+    public ByteBuffer getMetadataValue(String key) {
+      lazyInitCache();
+      return metadataCache.computeIfAbsent(key, s -> {
+        throw new IllegalArgumentException("Can't find user metadata " + key);
+      }).asReadOnlyByteBuffer();
+    }
+
+    public boolean hasMetadataValue(String key) {
+      lazyInitCache();
+      return metadataCache.containsKey(key);
+    }
+  }
+
   @Override
   public long getNumberOfRows() {
     return numberOfRows;
@@ -213,19 +252,17 @@ public class ReaderImpl implements Reader {
 
   @Override
   public List<String> getMetadataKeys() {
-    return new ArrayList<>(userMetadata.keySet());
+    return userMetadata.getMetadataKeys();
   }
 
   @Override
   public ByteBuffer getMetadataValue(String key) {
-    return userMetadata.computeIfAbsent(key, s -> {
-      throw new IllegalArgumentException("Can't find user metadata " + key);
-    }).asReadOnlyByteBuffer();
+    return userMetadata.getMetadataValue(key);
   }
 
   @Override
   public boolean hasMetadataValue(String key) {
-    return userMetadata.containsKey(key);
+    return userMetadata.hasMetadataValue(key);
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


dongjoon-hyun (changes requested):
Thank you for making a PR, @guiyanakuang . Do you think we can have a benchmark result for your claim?

> There may be performance problems if there are more userMetadata keys

cc @pgaref , @williamhyun

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
```diff
@@ -569,7 +561,13 @@ public ReaderImpl(Path path, OrcFile.ReaderOptions options) throws IOException {
       this.rowIndexStride = tail.getFooter().getRowIndexStride();
       this.contentLength = tail.getFooter().getContentLength();
       this.numberOfRows = tail.getFooter().getNumberOfRows();
-      this.userMetadata = tail.getFooter().getMetadataList();
+      this.userMetadata = tail.getFooter().getMetadataList()
+              .stream().collect(Collectors.toMap(
```
dongjoon-hyun, line 565: If you don't mind, can we indent with `4-space` instead of `8-space`?

dongjoon-hyun (commented):
@guiyanakuang . Technically, this looks reasonable because Java `WriterImpl` ensures that there is no duplicated keys.
```java
private final Map<String, ByteString> userMetadata = new TreeMap<>();
```

However, it seems that `ORC Foot` spec itself has no assumption that the user metadata items have unique keys. I'm wondering if we have some other places to put this unique key assumption. Otherwise, this PR might not be safe with another ORC writers.
```proto
message Footer {
  optional uint64 headerLength = 1;
  optional uint64 contentLength = 2;
  repeated StripeInformation stripes = 3;
  repeated Type types = 4;
  repeated UserMetadataItem metadata = 5;
```

cc @omalley , @pgaref , @wgtmac , @williamhyun

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
omalley: The java stream apis are very slow. It would perform better if you used a loop. This is especially important, because now the set is build preemptively, even if the user doesn't need the metadata values.

omalley (changes requested):

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
guiyanakuang: > If you don't mind, can we indent with `4-space` instead of `8-space`?
I will fix

guiyanakuang: > The java stream apis are very slow. It would perform better if you used a loop. This is especially important, because now the set is build preemptively, even if the user doesn't need the metadata values.

How about lazy initialization of the set when the getMetadataValue and hasMetadataValue methods need to be called

guiyanakuang: > Thank you for making a PR, @guiyanakuang . Do you think we can have a benchmark result for your claim?
> 
> > There may be performance problems if there are more userMetadata keys
> 
> cc @pgaref , @williamhyun

Sorry for the late reply due to different time zones. I made a guess based on the time complexity of the algorithm and have not yet benchmarked it. getMetadataValue is used in compatibility checking and metadata merging, which would have an O(n^2) time complexity if used in a loop.

guiyanakuang:
> @guiyanakuang . Technically, this looks reasonable because Java `WriterImpl` ensures that there is no duplicated keys at least Java layer(writer and reader)
> 
> ```java
> private final Map<String, ByteString> userMetadata = new TreeMap<>();
> ```
> 
> However, it seems that `ORC Foot` spec itself has no assumption that the user metadata items have unique keys. I'm wondering if we already have some other places to put this unique key assumption. Otherwise, this PR might not be safe with another ORC writers.
> 
> ```proto
> message Footer {
>   optional uint64 headerLength = 1;
>   optional uint64 contentLength = 2;
>   repeated StripeInformation stripes = 3;
>   repeated Type types = 4;
>   repeated UserMetadataItem metadata = 5;
> ```
> 
> cc @omalley , @pgaref , @wgtmac , @williamhyun

@dongjoon-hyun Additional commits solve the compatibility problem

```java
    private void lazyInitCache() {
      if (metadataCache == null) {
        metadataCache = new TreeMap<>();
        for(OrcProto.UserMetadataItem item: innerUserMetadata) {
          metadataCache.putIfAbsent(item.getName(), item.getValue());
        }
      }
    }
```
`metadataCache.putIfAbsent(item.getName(), item.getValue());` , the same key will only get the first value. Keeping the same logic as the loop traversal.
`getMetadataKeys` is still the same as before and can get non-unique keys.

dongjoon-hyun:
Thank you for updating, @guiyanakuang . Now, do we have both structure? It looks like a regression in memory perspective.

```
  public static class UserMetadata {
     private final List<OrcProto.UserMetadataItem> innerUserMetadata;

     private Map<String, ByteString> metadataCache;
```

guiyanakuang:
The metadataCache is only initialised when the metadata is needed for compatibility checking or merging. In fact the reader is quickly cleaned up and only survives in the merge context.

guiyanakuang:
I think I misunderstood, the user is also allowed to actively call methods to get the metadata, so that there are two data structures. Perhaps an internal method could be added to speed up the file merge, separate from the user getting the metadata.
Sorry, I've been reading and studying the orc source code for an upcoming project. Some of the suggestions may be superficial.

dongjoon-hyun:
How many users metadata do you have, @guiyanakuang ?

guiyanakuang:
> How many users metadata do you have, @guiyanakuang ?

Usually around 10.The largest have thousands of this order of magnitude. It may be abusing this feature, but so far there are no performance problems. This pr I just think from reading the source code that the time complexity can be optimised. I think I'll close this pr  and optimize it when I have performance issues in production.

dongjoon-hyun:
I agree with you, @guiyanakuang .

Although you revised your PR, I still suspect that some reader might expect the duplicated keys together (not the first one or the last one). Let's revisit this when the performance issue become severe.

dongjoon-hyun:
Anyway, thank you for your contribution, @guiyanakuang ! 
I'll promptly review your upcoming PRs.

guiyanakuang:
[ORC Specification v1](https://orc.apache.org/specification/ORCv1/)
image

PR #607, opened 2021-01-05
ORC-603. Upgrade `tools` to use Hadoop 2.10.1

dongjoon-hyun:
### What changes were proposed in this pull request?

This PR aims to upgrade `tools` module to use Apache Hadoop 2.10.1.

### Why are the changes needed?

This will make `orc-tools` CLI to use the latest and stable Hadoop 2.10.1 version.

### How was this patch tested?

Pass the CIs and manually check the dependency change with `mvn dependency:tree`

```
[INFO] ----------------------< org.apache.orc:orc-tools >----------------------
[INFO] Building ORC Tools 1.7.0-SNAPSHOT                                  [5/6]
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- maven-dependency-plugin:2.10:tree (default-cli) @ orc-tools ---
[INFO] org.apache.orc:orc-tools:jar:1.7.0-SNAPSHOT
[INFO] +- org.apache.orc:orc-core:jar:1.7.0-SNAPSHOT:compile
[INFO] |  +- org.apache.orc:orc-shims:jar:1.7.0-SNAPSHOT:compile
[INFO] |  +- com.google.protobuf:protobuf-java:jar:2.5.0:compile
[INFO] |  +- io.airlift:aircompressor:jar:0.16:compile
[INFO] |  +- javax.xml.bind:jaxb-api:jar:2.2.11:compile
[INFO] |  +- org.jetbrains:annotations:jar:17.0.0:compile
[INFO] |  \- org.threeten:threeten-extra:jar:1.5.0:compile
[INFO] +- com.google.code.gson:gson:jar:2.2.4:compile
[INFO] +- com.opencsv:opencsv:jar:3.9:compile
[INFO] +- commons-cli:commons-cli:jar:1.3.1:compile
[INFO] +- org.apache.commons:commons-lang3:jar:3.11:compile
[INFO] +- org.apache.hadoop:hadoop-common:jar:2.10.1:compile
[INFO] |  +- org.apache.hadoop:hadoop-annotations:jar:2.10.1:compile
[INFO] |  +- org.apache.commons:commons-math3:jar:3.1.1:compile
[INFO] |  +- xmlenc:xmlenc:jar:0.52:compile
[INFO] |  +- org.apache.httpcomponents:httpclient:jar:4.5.2:compile
[INFO] |  |  \- org.apache.httpcomponents:httpcore:jar:4.4.4:compile
[...]

PR #816, opened 2021-08-05
ORC-910: Enforce maven-dependency-plugin

williamhyun:
This PR aims to enforce maven-dependency-plugin. 

This will prevent unintended dependency changes. 

Pass the CIs and check the CI log.

### DIFF
--- a/java/bench/pom.xml
+++ b/java/bench/pom.xml
@@ -468,6 +468,15 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-enforcer-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <configuration>
+          <ignoredDependencies>
+            <ignoredDependency>*:*</ignoredDependency>
+          </ignoredDependencies>
+        </configuration>
+      </plugin>
     </plugins>
     <pluginManagement>
       <plugins>
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -307,6 +307,22 @@
           </configuration>
         </plugin>
         <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-dependency-plugin</artifactId>
+          <version>3.1.2</version>
+          <executions>
+            <execution>
+              <phase>package</phase>
+              <goals>
+                <goal>analyze-only</goal>
+              </goals>
+            </execution>
+          </executions>
+          <configuration>
+            <failOnWarning>true</failOnWarning>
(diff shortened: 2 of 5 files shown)


java/bench/pom.xml
```diff
@@ -468,6 +468,15 @@
...
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <configuration>
+          <ignoredDependencies>
+            <ignoredDependency>*:*</ignoredDependency>
+          </ignoredDependencies>
+        </configuration>
+      </plugin>
```
williamhyun, line 479: `benchmark` module is not published to the Maven Central.

java/pom.xml
```diff
@@ -306,6 +306,22 @@
             <failOnViolation>true</failOnViolation>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-dependency-plugin</artifactId>
+          <version>3.1.2</version>
```
williamhyun, line 312: 3.2.x has a bug. So, we should use the latest 3.1.x which is 3.1.2.

java/pom.xml
```diff
@@ -306,6 +306,22 @@
...
+          <executions>
+            <execution>
+              <phase>package</phase>
+              <goals>
+                <goal>analyze-only</goal>
+              </goals>
+            </execution>
+          </executions>
+          <configuration>
+            <failOnWarning>true</failOnWarning>
```
williamhyun, line 322: This will enforce this plugin in `analyze` profile.

java/tools/pom.xml
```diff
@@ -62,7 +66,7 @@
     </dependency>
     <dependency>
       <groupId>org.apache.hadoop</groupId>
-      <artifactId>hadoop-hdfs</artifactId>
+      <artifactId>hadoop-hdfs-client</artifactId>
```
williamhyun, line 69: After ORC-603 (https://github.com/apache/orc/pull/607), we use `hadoop-hdfs-client`.

williamhyun:
cc: @dongjoon-hyun @pgaref

dongjoon-hyun (approved):
Thank you, @williamhyun . It looks nice to me.

java/tools/pom.xml
dongjoon-hyun: Thank you for catching this, @williamhyun .

dongjoon-hyun:
I verified this by checking the GitHub Action log.
```
[INFO] --- maven-dependency-plugin:3.1.2:analyze-only (default) @ orc-shims ---
[INFO] No dependency problems found
```

PR #852, opened 2021-08-11
ORC-940: Use Hadoop 3.3.1 in bench module

dongjoon-hyun:

This PR aims to use Hadoop 3.3.1 instead of 2.7.3 in benchmark module.

Both Apache Hive 3.1.2 and Apache Spark 3.1.2 depends on Apache Hadoop 3.

Pass the CIs.


dongjoon-hyun:
cc @williamhyun .

BTW, do you think we can add hive bench and spark bench integration test, too?

williamhyun (approved):
+1 LGTM

PR #854, opened 2021-08-12
ORC-942: Remove javax.xml.bind:jaxb-api dependency

dongjoon-hyun:

This PR aims to remove `javax.xml.bind:jaxb-api` dependency.

This was added via ORC-411 to support Java 10+ at Apache ORC 1.4.5.

Since we only use `DatatypeConverter.printHexBinary` API, we can use the similar logic of `commons-codec` `Hex.encodeHexString`.

Pass the CIs with Java8/11/16/17-ea.

Closes #846

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/mask/SHA256MaskFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/mask/SHA256MaskFactory.java
@@ -22,7 +22,6 @@ import org.apache.hadoop.hive.ql.exec.vector.ColumnVector;
 import org.apache.orc.DataMask;
 import org.apache.orc.TypeDescription;
 
-import javax.xml.bind.DatatypeConverter;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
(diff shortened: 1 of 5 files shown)


dongjoon-hyun:
The original code is `Eclipse Distribution License v1.0`. Could you give me some advice, @omalley ?

dongjoon-hyun:
I realized that the license issue is serious. So, I switched the logic to Apache `commons-codec'`s one.

java/core/src/java/org/apache/orc/impl/mask/SHA256MaskFactory.java
```diff
@@ -73,6 +72,20 @@
     }
   }
 
+  private static final char[] DIGITS = {
+      '0', '1', '2', '3', '4', '5', '6', '7',
+      '8', '9', 'A', 'B', 'C', 'D', 'E', 'F'
+  };
+
+  public static String printHexBinary(byte[] data) {
```
dongjoon-hyun, line 80: https://github.com/apache/commons-codec/blob/b7ba05ad6b99ae9d5f93cd18e936b3e8908202fb/src/main/java/org/apache/commons/codec/binary/Hex.java#L287

williamhyun (approved):
+1 LGTM

dongjoon-hyun:
Thank you so much, @williamhyun !

PR #861, opened 2021-08-14
MINOR: Add more exclusion patterns at Spark bench uber jar building

williamhyun:
This PR aims to add more exclusion patterns at Spark bench uber jar building. 

Since ORC-940, it seems that we have more warnings.
This will suppress more new warnings.

Check GHA build logs.


dongjoon-hyun (approved):
+1, LGTM.
Thank you for doing this follow-up for ORC-940.

PR #862, opened 2021-08-15
ORC-949: Add CustomImportOrder checkstyle rule

williamhyun:
This PR aims to add the `CustomImportOrder` rule

To enforce coding style.

Pass the CIs.


dongjoon-hyun (approved):
+1, LGTM although I quickly checked it. Thank you for adding this, @williamhyun .

Although this is a little huge change, I'm fine if we can keep our codebase consistently from now by adding automatic checking.

PR #866, opened 2021-08-15
ORC-953: Add `since` tag to `org.apache.orc.Writer` interface

dongjoon-hyun:

This PR aims to add `since` tag to `org.apache.orc.Writer` interface.

To help ORC developers to check the API availability easily by improving code and Javadoc.

Manual.

### DIFF
--- a/java/core/src/java/org/apache/orc/Writer.java
+++ b/java/core/src/java/org/apache/orc/Writer.java
@@ -27,6 +27,7 @@ import java.util.List;
 
 /**
  * The interface for writing ORC files.
+ * @since 1.1.0
  */
 public interface Writer extends Closeable {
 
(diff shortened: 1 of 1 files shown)


java/core/src/java/org/apache/orc/Writer.java
```diff
@@ -150,6 +162,7 @@ void appendStripe(byte[] stripe, int offset, int length,
    *
    * @return stripe information
    * @throws IOException
+   * @since 1.6.0
```
williamhyun, line 165: Looks like this is from 1.6.8, could you fix this?

dongjoon-hyun: Oh. My bad. Thank you!

PR #869, opened 2021-08-16
ORC-946: Unified json library

guiyanakuang:

The java project depends on several json libraries. Replacing jackson/jettison with Gson 

```
use jackson-core in orc-benchmarks-core
org.apache.orc.bench.core.convert.json.JsonWriter

use jettison in tools
org.apache.orc.tools.KeyTool
org.apache.orc.tools.JsonFileDump
org.apache.orc.tools.PrintData
```

gson vs. jettison has some inconsistent

1.  the scope of character escaping is inconsistent; 
    jettison escapes '/'
    gson does not modify
2. Inconsistent floating point writing; 
    jettison removes trailing zeros and decimal points where possible. 
    gson leaves it as is.
3. prettyPrint;
    jettison's prettyPrint is faulty, '[' and '{' are often linked together without a line break. 
    gson prettyPrint is fine.

These three behaviours are not forward compatible in this pr.

I don't think prettyPrint needs to be made compatible. The other two I would like to hear your opinion on.

Reduce dependencies and use a unified json lib.

Pass the CIs.

### DIFF
--- a/java/tools/src/test/resources/orc-file-dump.json
+++ b/java/tools/src/test/resources/orc-file-dump.json
@@ -28,7 +28,7 @@
       }
     }
   },
-  "calendar": "Julian\/Gregorian",
+  "calendar": "Julian/Gregorian",
   "stripeStatistics": [
     {
       "stripeNumber": 1,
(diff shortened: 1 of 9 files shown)


dongjoon-hyun:

java/tools/src/test/resources/orc-file-dump.json
```diff
@@ -1323,46 +1331,52 @@
...
+          }
         }
-      }]
+      ]
     }
   ],
   "fileLength": 272513,
   "paddingLength": 0,
-  "paddingRatio": 0,
+  "paddingRatio": 0.0,
```
dongjoon-hyun, line 1380: I agree that new floating point string looks better.

java/tools/src/test/resources/orc-file-dump.json
```diff
@@ -28,7 +28,7 @@
       }
     }
   },
-  "calendar": "Julian\/Gregorian",
+  "calendar": "Julian/Gregorian",
```
dongjoon-hyun, line 31: This one looks good too.

dongjoon-hyun (approved):

guiyanakuang:
Thanks to dongjoon-hyun for the review and approval!

PR #873, opened 2021-08-18
ORC-811 Benchmarks and documentation for ORC-742 and ORC-743

pavibhai:
Includes Benchmark tests and Documentation for ORC-742 and ORC-743

Benchmarks show the benefits of adding LazyIO during reads and the documentation explains the design of the same.

There is no new functionality added here so the regression tests were run and the benchmark tests were completed to verify the improvements.

### DIFF
--- a/java/bench/core/pom.xml
+++ b/java/bench/core/pom.xml
@@ -96,12 +96,28 @@
       <artifactId>jmh-core</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-generator-annprocess</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-log4j12</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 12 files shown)


dongjoon-hyun (commented):
Thank you, @pavibhai .

dongjoon-hyun (commented):
@pavibhai . Can we move some of the tests from `bench` module to normal modules? It's a little weird for `bench` module to have a test coverage because this is not executed by default.

pavibhai:
> @pavibhai . Can we move some of the tests from `bench` module to normal modules? It's a little weird for `bench` module to have a test coverage because this is not executed by default.

@dongjoon-hyun These tests have been added specifically to verify the benchmark code with unit tests. The other modules already have this test coverage, so these tests should belong here.

dongjoon-hyun:
Then, could you spin off them as a new test PR, @pavibhai ?

pavibhai:
> Then, could you spin off them as a new test PR, @pavibhai ?

Sorry I am confused. These are tests I am adding for the benchmarks that are being added as part of this PR. Why do you want to separate the tests into another PR?

omalley:
I think it is better practice to include the tests with the PR that adds the code. Since the tests are for the benchmark code, it makes sense to include them here.

dongjoon-hyun:
Got it if you think so, @omalley .

pavibhai:
@omalley @pgaref the benchmark and documentation PR is ready for review. Please take a look at it when you get a chance.

PR #877, opened 2021-08-23
ORC-959: [C++] Fix crash of SargsApplier in finding nested field names

stiga-huang:

This patch fixes a crash in the C++ reader when SearchArgument is created on nested columns of collection types (e.g. array/map). Only STRUCT type has field names for subtypes. So we should check the current node type before checking any field names. JIRA description contains details on how to reproduce the crash.

Same as the above.

Added unit tests in c++/test/TestSargsApplier.cc to reproduce the crash and verify the fix on ARRAY and MAP types.

### DIFF
--- a/c++/src/sargs/SargsApplier.cc
+++ b/c++/src/sargs/SargsApplier.cc
@@ -25,7 +25,8 @@ namespace orc {
   uint64_t SargsApplier::findColumn(const Type& type,
                                     const std::string& colName) {
     for (uint64_t i = 0; i != type.getSubtypeCount(); ++i) {
-      if (type.getFieldName(i) == colName) {
+      // Only STRUCT type has field names
+      if (type.getKind() == STRUCT && type.getFieldName(i) == colName) {
         return type.getSubtype(i)->getColumnId();
       } else {
         uint64_t ret = findColumn(*type.getSubtype(i), colName);
--- a/c++/src/sargs/SargsApplier.hh
+++ b/c++/src/sargs/SargsApplier.hh
@@ -82,6 +82,8 @@ namespace orc {
 
   private:
     friend class TestSargsApplier_findColumnTest_Test;
+    friend class TestSargsApplier_findArrayColumnTest_Test;
+    friend class TestSargsApplier_findMapColumnTest_Test;
     static uint64_t findColumn(const Type& type, const std::string& colName);
 
   private:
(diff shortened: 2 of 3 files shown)


dongjoon-hyun (commented):
Thank you for making a PR too, @stiga-huang .

dongjoon-hyun:
cc @wgtmac and @pgaref

c++/src/sargs/SargsApplier.cc
```diff
@@ -25,7 +25,8 @@ namespace orc {
   uint64_t SargsApplier::findColumn(const Type& type,
                                     const std::string& colName) {
     for (uint64_t i = 0; i != type.getSubtypeCount(); ++i) {
-      if (type.getFieldName(i) == colName) {
+      // Only STRUCT type has field names
+      if (type.getKind() == STRUCT && type.getFieldName(i) == colName) {
```
dongjoon-hyun, line 29: The directly previous code seems to be ORC-40 at 1.7.0.
Does 1.6.x has the same issue, @stiga-huang ?

stiga-huang: No, predicate push down in the C++ reader is supported since 1.7.0. ORC-1.6.x doesn't have these codes.

dongjoon-hyun: Great. Thank you for the confirmation.

dongjoon-hyun (commented):
BTW, can we have a test coverage for case-sensitivity? For example, what is the contract of `findColumn` when the `colName` is different case-sensitively like `col1` and `Col1`?

dongjoon-hyun (approved):
+1, LGTM.

I verified the unit test provides the coverage correctly.
```
Test project /Users/dongjoon/APACHE/orc-merge/build
    Start 1: orc-test
1/8 Test #1: orc-test .........................***Exception: SegFault  1.28 sec
ORC version: 1.8.0-SNAPSHOT
```

For case-sensitivity test coverage, I filed ORC-962 and we can handle that separately.

dongjoon-hyun:
Merged to main/1.7. Thank you, @stiga-huang .

wgtmac:
Thanks @stiga-huang for a quick fix!

stiga-huang:
Thank @dongjoon-hyun for the quick review!

PR #878, opened 2021-08-23
ORC-963: Build benchmark module always for integration testing

williamhyun:

This PR aims to build the `benchmark` module always for integration testing. 

Since we always run the integration tests, we need to build the `benchmark` module. Otherwise, those tests will fail when we run manually without any configurations.

Manually.


dongjoon-hyun (approved):
+1, LGTM. Thank you, @williamhyun .

PR #879, opened 2021-08-23
ORC-960: [C++] Support creating SearchArguments using column ids

stiga-huang:

Currently in the C++ reader, SearchArgumentBuilder only provides interfaces for creating SearchArguments using field column names. Column names can be ambiguous if there are nested struct columns using the same names. Array item column or map key/value columns even don't have names.

This patch addes corresponding interfaces to create SearchArguments using column ids. Refactors some codes using templates to avoid duplicated codes.

As described above and in the JIRA description, we need creating SearchArguments using column ids.

Added end-to-end tests in c++/test/TestPredicatePushdown.cc.

### DIFF
--- a/c++/include/orc/sargs/SearchArgument.hh
+++ b/c++/include/orc/sargs/SearchArgument.hh
@@ -82,7 +82,7 @@ namespace orc {
 
     /**
      * Add a less than leaf to the current item on the stack.
-     * @param column the name of the column
+     * @param column the field name of the column
      * @param type the type of the expression
      * @param literal the literal
      * @return this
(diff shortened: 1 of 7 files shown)


dongjoon-hyun (commented):
Thank you for making a PR, @stiga-huang .

dongjoon-hyun:
cc @wgtmac

wgtmac (approved):
Overall LGTM. It is possible that user mixes use of fieldName and columnId in a single SearchArgumentBuilder which is weird but valid.

dongjoon-hyun:
Thank you, @wgtmac . 
According to the affected version of ORC-960, @stiga-huang seems to create ORC-960 for Apache ORC 1.7.0.
Although Apache ORC 1.7 has been in the feature freeze status, do you think we need to land this in branch-1.7?

stiga-huang:
Impala will wait for this feature. It'd be good if the coming ORC 1.7.0 release can have it.

wgtmac:
> Thank you, @wgtmac .
> According to the affected version of [ORC-960](https://issues.apache.org/jira/browse/ORC-960), @stiga-huang seems to create [ORC-960](https://issues.apache.org/jira/browse/ORC-960) for Apache ORC 1.7.0.
> Although Apache ORC 1.7 has been in the feature freeze status, do you think we need to land this in branch-1.7?

This change is limited to C++ only. I am OK if landing it into branch-1.7 will not delay the release process.

dongjoon-hyun:
Got it, @stiga-huang and @wgtmac .

dongjoon-hyun:
Merged to main/branch-1.7 for Apache ORC 1.7.

dongjoon-hyun:
BTW, the current release blocker is ORC-811 (LazyIO of non-filter columns benchmark) and Apache Iceberg Integration Testing.
- https://github.com/dongjoon-hyun/iceberg/pull/7

PR #884, opened 2021-08-27
ORC-973: [C++] Support providing IN-predicate literals using vector

stiga-huang:

Add two more interfaces for creating IN-predicate literals using vector. This provide more flexibility that doesn't requires callers know the list size at compilation time.

Described as above.

Add unit test in c++/test/TestPredicateLeaf.cc


dongjoon-hyun:
Merged to main/1.7.

PR #885, opened 2021-08-27
ORC-974: Remove Text Key from StringRedBlackTree

belugabehr:
Remove the dependency on Hadoop's Text class from RedBlackTree.  This has the added benefit of skipping the step where the incoming bytes are copied into the Text class.

Allows for removing dependency on Hadoop later; improve performance by removing copy.

No functionality change. Using existing unit tests.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RedBlackTree.java
+++ b/java/core/src/java/org/apache/orc/impl/RedBlackTree.java
@@ -147,8 +147,8 @@ abstract class RedBlackTree {
    * @param greatGrandparent Grandparent's parent
    * @return Does parent also need to be checked and/or fixed?
    */
-    private boolean add(int node, boolean fromLeft, int parent, int grandparent, int greatGrandparent,
-            byte[] bytes, int offset, int length) {
+    private boolean add(int node, boolean fromLeft, int parent, int grandparent,
+            int greatGrandparent, byte[] bytes, int offset, int length) {
     if (node == NULL) {
       if (root == NULL) {
         lastAdd = insert(NULL, NULL, false);
@@ -271,9 +271,10 @@ abstract class RedBlackTree {
 
   /**
    * Add the new key to the tree.
- * @param length 
- * @param offset 
- * @param bytes 
+   *
+   * @param bytes
+   * @param offset
+   * @param length
    * @return true if the element is a new one.
    */
   protected boolean doAdd(byte[] bytes, int offset, int length) {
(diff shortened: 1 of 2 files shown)


java/core/src/java/org/apache/orc/impl/RedBlackTree.java
```diff
@@ -147,8 +147,8 @@ private void setRight(int position, int right) {
    * @param greatGrandparent Grandparent's parent
    * @return Does parent also need to be checked and/or fixed?
    */
-  private boolean add(int node, boolean fromLeft, int parent,
-                      int grandparent, int greatGrandparent) {
+    private boolean add(int node, boolean fromLeft, int parent, int grandparent,
```
dongjoon-hyun, line 150: indentation?

java/core/src/java/org/apache/orc/impl/RedBlackTree.java
```diff
@@ -271,10 +271,14 @@ private boolean add(int node, boolean fromLeft, int parent,
...
    * Add the new key to the tree.
+   *
+   * @param bytes
+   * @param offset
+   * @param length
    * @return true if the element is a new one.
    */
-  protected boolean add() {
-    add(root, false, NULL, NULL, NULL);
+  protected boolean doAdd(byte[] bytes, int offset, int length) {
```
dongjoon-hyun, line 280: Is this renamed to avoid conflicts?

java/core/src/java/org/apache/orc/impl/StringRedBlackTree.java
```diff
@@ -29,49 +30,42 @@
...
-    return addNewKey();
+    return add(value.getBytes(), 0, value.getLength());
   }
 
   @Override
   public int add(byte[] bytes, int offset, int length) {
-    newKey.set(bytes, offset, length);
-    return addNewKey();
+    // if the newKey is actually new, add it to our byteArray and store the offset & length
+    if (doAdd(bytes, offset, length)) {
```
guiyanakuang, line 53: > Is this renamed to avoid conflicts?

@dongjoon-hyun, should be here to avoid recursive calls.

dongjoon-hyun: Ya, that was my question.

belugabehr: Yes.  There is already a `public` method with the same name/signature.

java/core/src/java/org/apache/orc/impl/RedBlackTree.java
belugabehr: Yes.  There is already a `public` method with the same name/signature.

java/core/src/java/org/apache/orc/impl/RedBlackTree.java
dongjoon-hyun: Gentle ping, @belugabehr .

PR #886, opened 2021-08-28
ORC-975: Avoid double counting closestFixedBits

guiyanakuang:

```
310  int idx = encodeBitWidth(findClosestNumBits(data[i]));
```

```
public int encodeBitWidth(int n) {
    n = getClosestFixedBits(n);
    .....
}
```

```
public int findClosestNumBits(long value) {
  final int numBits = 64 - Long.numberOfLeadingZeros(value);
  return getClosestFixedBits(numBits);
}
```

getClosestFixedBits is called twice.
Extracts the new method findNumBits from findClosestNumBits. Avoid double counting closestFixedBits.

Avoid redundant calls.

Pass the CIs.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/SerializationUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/SerializationUtils.java
@@ -270,8 +270,16 @@ public final class SerializationUtils {
    * @return bits required to store value
    */
   public int findClosestNumBits(long value) {
-    final int numBits = 64 - Long.numberOfLeadingZeros(value);
-    return getClosestFixedBits(numBits);
+    return getClosestFixedBits(findNumBits(value));
+  }
+
+  /**
+   * Count the number of bits left after subtracting leading zeros
+   * @param value
+   * @return bits left after subtracting leading zeros
+   */
+  public int findNumBits(long value) {
+    return 64 - Long.numberOfLeadingZeros(value);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


java/core/src/java/org/apache/orc/impl/SerializationUtils.java
```diff
@@ -270,8 +270,16 @@ public static BigInteger readBigInteger(InputStream input) throws IOException {
...
-    return getClosestFixedBits(numBits);
+    return getClosestFixedBits(findNumBits(value));
+  }
+
+  /**
+   * Count the number of bits left after subtracting leading zeros
+   * @param value
+   * @return bits left after subtracting leading zeros
+   */
+  public int findNumBits(long value) {
```
dongjoon-hyun, line 281: Do you have a reference for this naming, @guiyanakuang ? Can we have a more specific name?

guiyanakuang: I've looked up some references and I still don't know what the name of the complement code to remove the high zeros is. How about `validNumBits`.

dongjoon-hyun: Ya, it's tricky. Then, shall we make this `private`? In that case, we don't need to care much about naming.

guiyanakuang: Yes, I should not expose this internal method.  Fix in 6fa50b2

dongjoon-hyun (approved):
Merged to main/1.7.

dongjoon-hyun:
This is not backported to branch-1.7 because there is a conflict.

PR #887, opened 2021-08-28
ORC-976: Optimize compute zigzagliterals

guiyanakuang:

```java
  private void computeZigZagLiterals() {
    // populate zigzag encoded literals
    long zzEncVal = 0;
    for (int i = 0; i < numLiterals; i++) {
      if (signed) {
        zzEncVal = utils.zigzagEncode(literals[i]);
      } else {
        zzEncVal = literals[i];
      }
      zigzagLiterals[i] = zzEncVal;
    }
  }
```

Avoid conditional judgments in loops. 
The unsigned case can use literals instead of zigzagLiterals. 

optimize write performance.

Pass the CIs.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
+++ b/java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
@@ -414,7 +414,10 @@ public class RunLengthIntegerWriterV2 implements IntegerWriter {
 
     // we need to compute zigzag values for DIRECT encoding if we decide to
     // break early for delta overflows or for shorter runs
-    computeZigZagLiterals();
+    // only signed numbers need to compute zigzag values
+    if (signed) {
+      computeZigZagLiterals();
+    }
 
     long[] currentZigzagLiterals = signed ? zigzagLiterals : literals;
     zzBits100p = utils.percentileBits(currentZigzagLiterals, 0, numLiterals, 1.0);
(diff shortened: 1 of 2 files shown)


dongjoon-hyun (changes requested):
Thank you, @guiyanakuang . In this case, we had batter improve C++ code together for consistency. Could you update it in this PR?

https://github.com/apache/orc/blob/main/c++/src/RleEncoderV2.cc#L170-L180

cc @wgtmac

java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
```diff
@@ -545,14 +547,10 @@ private void determineEncoding() {
 
   private void computeZigZagLiterals() {
     // populate zigzag encoded literals
-    long zzEncVal = 0;
-    for (int i = 0; i < numLiterals; i++) {
-      if (signed) {
-        zzEncVal = utils.zigzagEncode(literals[i]);
-      } else {
-        zzEncVal = literals[i];
+    if (signed) {
```
dongjoon-hyun, line 550: Shall we remove `if` statement and add an assertion at the first line of this function?

dongjoon-hyun:
cc @pgaref and @williamhyun , too.

java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
guiyanakuang: I quite agree. fix in 7de084b

guiyanakuang:
> Thank you, @guiyanakuang . In this case, we had batter improve C++ code together for consistency. Could you update it in this PR?
> 
> https://github.com/apache/orc/blob/main/c++/src/RleEncoderV2.cc#L170-L180
> 
> cc @wgtmac

I think I can, although I don't use c++ much

dongjoon-hyun:
Thank you for updating and considering C++ change together, @guiyanakuang .

guiyanakuang:
Update C++ improve, please help review code. I don't use C++ much. @dongjoon-hyun @wgtmac

dongjoon-hyun (approved):
+1, LGTM from my side. Thank you so much for updating, @guiyanakuang .
I'll wait for the other committers' review.

dongjoon-hyun:
Also, cc @omalley , too

wgtmac (approved):
LGTM +1. Thanks @guiyanakuang and @dongjoon-hyun !

dongjoon-hyun:
Thank you, @guiyanakuang and @wgtmac !
Merged to main/1.7.

PR #891, opened 2021-09-01
ORC-978: Fix NPE in TestFlinkOrcReaderWriter

guiyanakuang:

This PR aims to fix NPE in TestFlinkOrcReaderWriter.

```java
@Override
  public void close() throws IOException {
    if (!isClosed) {
      try {
        if (batch.size > 0) {
          writer.addRowBatch(batch);
          batch.reset();
        }
      } finally {
        writer.close();
        this.isClosed = true;
      }
    }
  }
```

writer.addRowBatch(batch);

```java
    .......
      checkMemory();
    } catch (Throwable t) {
      try {
        close();
      } catch (Throwable ignore) {
        // ignore
      }
      if (t instanceof IOException) {
        throw (IOException) t;
      } else {
        throw new IOException("Problem adding row to " + path, t);
      }
    }
```
addRowBatch method throws java.lang.OutOfMemoryError causing writerImpl to close twice in TestFlinkOrcReaderWriter case.
The first close already set the rawWriter to null, so the second time throw a NullPointerException.

I added status variables to ensure that close will only do the necessary action the first time.

Fix NPE in TestFlinkOrcReaderWriter.

Add UT for this bug.

### DIFF
--- a/java/core/src/test/org/apache/orc/impl/TestWriterImpl.java
+++ b/java/core/src/test/org/apache/orc/impl/TestWriterImpl.java
@@ -118,4 +118,12 @@ public class TestWriterImpl {
     Reader r = OrcFile.createReader(testFilePath, OrcFile.readerOptions(conf));
     assertEquals(r.getStripes(), w.getStripes());
   }
+
+  @Test
+  public void testCloseIsIdempotent() throws IOException {
+    conf.set(OrcConf.OVERWRITE_OUTPUT_FILE.getAttribute(), "true");
+    Writer w = OrcFile.createWriter(testFilePath, OrcFile.writerOptions(conf).setSchema(schema));
+    w.close();
+    w.close();
+  }
 }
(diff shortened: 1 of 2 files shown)


dongjoon-hyun:
Thank you so much for working on this, @guiyanakuang .

cc @pgaref , @wgtmac , @williamhyun , @omalley

java/core/src/test/org/apache/orc/impl/TestWriterImpl.java
```diff
@@ -118,4 +118,12 @@ public void testStripes() throws Exception {
     Reader r = OrcFile.createReader(testFilePath, OrcFile.readerOptions(conf));
     assertEquals(r.getStripes(), w.getStripes());
   }
+
+  @Test
+  public void testCloseIsIdempotent() throws IOException {
```
dongjoon-hyun, line 123: Thank you for adding a test case for this.

dongjoon-hyun (approved):
+1, LGTM.
Merged to main/1.7.

PR #893, opened 2021-09-02
ORC-980: Filter processing respects the case-sensitivity flag

pavibhai:
The filter processing incorrectly ignores the case-sensitivity flag for schema supplied by the reader. This fixes this bug and uses the flag in determining field and vector information.

In the absence of this the read incorrectly fails by performing a case-sensitive match even when case-insensitive match was requested by the reader.

Unit Tests were added to verify the failure and subsequently the fix.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -283,6 +283,7 @@ public class RecordReaderImpl implements RecordReader {
     Consumer<OrcFilterContext> filterCallBack = null;
     BatchFilter filter = FilterFactory.createBatchFilter(options,
                                                          evolution.getReaderBaseSchema(),
+                                                         evolution.isSchemaEvolutionCaseAware,
                                                          fileReader.getFileVersion(),
                                                          false);
     if (filter != null) {
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -283,7 +283,7 @@ public class RecordReaderImpl implements RecordReader {
     Consumer<OrcFilterContext> filterCallBack = null;
     BatchFilter filter = FilterFactory.createBatchFilter(options,
                                                          evolution.getReaderBaseSchema(),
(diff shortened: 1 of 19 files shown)


dongjoon-hyun:
cc @omalley

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -283,6 +283,7 @@ protected RecordReaderImpl(ReaderImpl fileReader,
     Consumer<OrcFilterContext> filterCallBack = null;
     BatchFilter filter = FilterFactory.createBatchFilter(options,
                                                          evolution.getReaderBaseSchema(),
+                                                         evolution.isSchemaEvolutionCaseAware,
```
dongjoon-hyun, line 286: For consistency, could you use the getter function?

java/core/src/test/org/apache/orc/TestRowFilteringIOSkip.java
```diff
@@ -570,6 +570,29 @@ public void schemaEvolutionLong2StringColumn() throws IOException {
     assertEquals(1, rowCount);
   }
 
+  @Test
+  public void readCaseInsensitive() throws IOException {
```
dongjoon-hyun, line 574: Do we have a case-sensitive test coverage?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pavibhai: Good point, changed.

java/core/src/test/org/apache/orc/TestRowFilteringIOSkip.java
pavibhai: The default value is true for case-sensitivity so all the tests are the other tests are case-sensitive tests. I added an explicit failure test with case-sensitivity when the name is not found.

dongjoon-hyun: Thank you so much!

dongjoon-hyun (approved):
+1, LGTM. Thank you, @pavibhai .
Merged to main/1.7.

PR #897, opened 2021-09-04
ORC-715: Add MapReduce test case

guiyanakuang:

This pr aims to recover a test coverage which is removed at
https://github.com/apache/orc/commit/db62ea5730797b046a3b7d5d1a3b026b08921209

Refer Hadoop example and previous mapreduce tests.
https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-jobclient/src/test/java/org/apache/hadoop/mapreduce/TestMapReduce.java
https://github.com/apache/orc/pull/612/files#diff-15999a33f815f28f3251acabae2a16dd2f1114d8c73e6fab26c207c08884ce68

BTW This pr bypasses the bug in older versions of hadoop-common by configuring `mapreduce.job.output.key.comparator.class`.  Fix [ORC-964 ](https://issues.apache.org/jira/projects/ORC/issues/ORC-964?filter=allopenissues) issue.

Recover a test coverage

Pass the CIs.

### DIFF
--- a/java/mapreduce/src/test/org/apache/orc/mapreduce/TestMrUnit.java
+++ b/java/mapreduce/src/test/org/apache/orc/mapreduce/TestMrUnit.java
@@ -21,13 +21,12 @@ package org.apache.orc.mapreduce;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.hive.ql.exec.vector.BytesColumnVector;
-import org.apache.hadoop.hive.ql.exec.vector.LongColumnVector;
-import org.apache.hadoop.hive.ql.exec.vector.StructColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
 import org.apache.hadoop.io.DataInputBuffer;
+import org.apache.hadoop.io.IntWritable;
 import org.apache.hadoop.io.NullWritable;
 import org.apache.hadoop.io.RawComparator;
+import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapred.JobConf;
 import org.apache.hadoop.mapred.JobConfigurable;
 import org.apache.hadoop.mapreduce.Job;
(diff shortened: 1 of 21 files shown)


dongjoon-hyun (commented):
Thank you for making this, @guiyanakuang .
Mostly this PR looks good. The only difference I noticed is that this is testing vectorized code path instead.  Is this change intentional?

java/mapreduce/src/test/org/apache/orc/mapreduce/TestMrUnit.java
```diff
@@ -0,0 +1,253 @@
...
+      .fromString("struct<one:struct<x:int,y:int>,two:struct<z:string>>");
+
+  private static final TypeDescription OUT_SCHEMA = TypeDescription
+      .fromString("struct<first:struct<x:int,y:int>,second:struct<z:string>>");
+
+  static {
+    OrcConf.MAPRED_SHUFFLE_KEY_SCHEMA.setString(CONF, "struct<x:int,y:int>");
+    OrcConf.MAPRED_SHUFFLE_VALUE_SCHEMA.setString(CONF, "struct<z:string>");
+    OrcConf.MAPRED_OUTPUT_SCHEMA.setString(CONF, OUT_SCHEMA.toString());
+    CONF.set("mapreduce.job.output.key.comparator.class", OrcKeyComparator.class.getName());
```
dongjoon-hyun, line 74: Thank you for adding this. Could you add a comment referring ORC-964?

guiyanakuang:
> Thank you for making this, @guiyanakuang .
> Mostly this PR looks good. The only difference I noticed is that this is testing vectorized code path instead. Is this change intentional?

This is not intentional. I'll change pr later to use OrcMapreduceRecordWriter and OrcMapreduceRecordReader instead.

java/mapreduce/src/test/org/apache/orc/mapreduce/TestMrUnit.java
guiyanakuang: OK 👌

dongjoon-hyun (approved):

cc @williamhyun because this will be a part of ORC-716 .

dongjoon-hyun:
Merged to main/1.7.

PR #902, opened 2021-09-07
ORC-985: Change default back to rbtree

omalley:
The new hash table implementation in ORC-757 results in significantly larger ORC files compared to ORC 1.6. Changing the default back to rbtree.

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -107,7 +107,7 @@ public enum OrcConf {
           "writing first stripe. In both cases, the decision to use\n" +
           "dictionary or not will be retained thereafter."),
   DICTIONARY_IMPL("orc.dictionary.implementation", "orc.dictionary.implementation",
-      "hash",
+      "rbtree",
       "the implementation for the dictionary used for string-type column encoding.\n" +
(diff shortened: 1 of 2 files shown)


dongjoon-hyun (approved):
+1, LGTM. Thank you, @omalley .
Merged to main/1.7 to unblock Apache ORC 1.7.0 release.

dongjoon-hyun:
cc @autumnust , @pgaref , @wgtmac , @williamhyun

wgtmac:
@omalley Do we have a brief summary of the size comparison?

dongjoon-hyun:
According to the JIRA description, it was a 5% regression.
> Running some tests, I noticed a 5% regression in file sizes with branch 1.7 compared to 1.6. I need to track this down.

wgtmac:
> According to the JIRA description, it was a 5% regression.
> 
> > Running some tests, I noticed a 5% regression in file sizes with branch 1.7 compared to 1.6. I need to track this down.

Thanks for the reply! In our environment we uses hash-based dictionary by default. In some tests I can only see 1%~2% gain after switching to sort-based dictionary.

dongjoon-hyun:
Got it. Thank you for sharing, @wgtmac .

PR #904, opened 2021-09-09
ORC-984: Save the software version that wrote each ORC file.

omalley:

I add a string to the file footer that records the version that wrote the file. We already had recorded the implementation that wrote the file as Footer.writer. I added a method that combines these two fields in the reader to produce a user facing string that describes the software version.

I also add a field to the meta data tool to show the version that wrote the file.

Because of that change and the fact that the files change size based on whether the ORC version is a snapshot or not, I had to extend the tests for TestFileDump to allow some slop for the size and to ignore the file version.

It passes the unit tests after I updated the tools tests.


omalley:
The software version in the API and tool shows up as "ORC Java 1.8.0-SNAPSHOT". When run from IntelliJ the unit tests can't find the software version, so it comes in "ORC Java unknown".

omalley:
I haven't updated the C++ reader, writer, and tool yet.

java/core/src/java/org/apache/orc/Reader.java
```diff
@@ -158,6 +158,16 @@
    */
   OrcFile.WriterVersion getWriterVersion();
 
+  /**
+   * Get the implementation and version of the software that wrote the file.
+   * It defaults to "ORC Java" for old files. For current files, we include the
+   * version also.
+   * @since 1.5.13
```
dongjoon-hyun, line 165: Thank you for adding `since` info, @omalley .
BTW, do we need to backport this new feature to `branch-1.5`?

dongjoon-hyun:
BTW, according to GitHub Action jobs, all Java11+ seems to complain for some reasons.

guiyanakuang:
> BTW, according to GitHub Action jobs, all Java11+ seems to complain for some reasons.

This should be a bug in findbugs. Java 11 compiles try-with-resources to generate bytecode differently than java8. Findbugs is considered RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE from bytecode analysis. 
https://github.com/spotbugs/spotbugs/issues/259

dongjoon-hyun (approved):

java/core/src/java/org/apache/orc/Reader.java
omalley: That was my intention.

omalley:
Ok, the changes for C++ were relatively straightforward. I did update the C++ with the new writer versions, such as the important ORC-14.

I updated the orc_split_elim_new.orc so that it was written by the 1.8.0-SNAPSHOT. I also added orc_split_elim_cpp.orc that is the same file written by c++. The Java and C++ tools show the files as having being written by the corresponding versions.

omalley:
BTW, to generate the new c++ ORC file, I wrote a little utility that takes an ORC file and rewrites it using c++. It is in https://github.com/omalley/orc/tree/cpp-convert .

tools/test/TestMatch.cc
```diff
@@ -163,6 +167,7 @@ namespace orc {
                                        "TestOrcFile.columnProjection.jsn.gz",
                                        "struct<int1:int,string1:string>",
                                        "0.12",
+				       "ORC Java",
```
dongjoon-hyun, line 170: Indentation seems to be broken due to `tab` character.

tools/test/TestMatch.cc
```diff
@@ -182,6 +187,7 @@ namespace orc {
                                        "string>>,map:map<string,struct<int1:"
                                        "int,string1:string>>>",
                                        "0.12",
+				       "ORC Java",
```
dongjoon-hyun, line 190: ditto

tools/test/TestMatch.cc
```diff
@@ -448,17 +472,34 @@ namespace orc {
...
                                        1,
                                        CompressionKind_ZLIB,
                                        262144,
                                        10000,
                                        std::map<std::string, std::string>()),
+                    OrcFileDescription("orc_split_elim_cpp.orc",
+                                       "orc_split_elim_cpp.jsn.gz",
+                                       ("struct<userid:bigint,string1:string,"
+                                        "subtype:double,"
+					"decimal1:decimal(16,6),"
```
dongjoon-hyun, line 487: ditto.

dongjoon-hyun:
So, everything is ready, right, @omalley ?

c++/src/Common.cc
```diff
@@ -58,12 +58,38 @@ namespace orc {
...
+        return "ORC-14";
     }
     std::stringstream buffer;
     buffer << "future - " << version;
     return buffer.str();
   }
 
+  std::string writerIdToString(uint32_t id) {
+    switch (id) {
+    case ORC_JAVA_WRITER:
```
dongjoon-hyun, line 75: This indentation seems to be different from the other `switch` statements in this file

dongjoon-hyun (approved):
+1, LGTM. Thank you, @omalley .
Merged to main/1.7/1.6/1.5 according to the PR content.

PR #907, opened 2021-09-14
ORC-992: Reached max repeat length, we can directly decide to use DELTA encoding

guiyanakuang:

Reached max repeat length, we can directly decide to use DELTA encoding.
The computeZigZagLiterals, zzBits100p, and determine isFixedDelta within the determineEncoding method are all redundant in the current case.

Similar practices.
RunLengthIntegerWriterV2.java 767-775
```java
          if (fixedRunLength >= MIN_REPEAT) {
            if (fixedRunLength <= MAX_SHORT_REPEAT_LENGTH) {
              encoding = EncodingType.SHORT_REPEAT;
            } else {
              encoding = EncodingType.DELTA;
              isFixedDelta = true;
            }
            writeValues();
          }
```

Optimization for reached max repeat length.

Pass the CIs.

### DIFF
--- a/c++/src/RleEncoderV2.cc
+++ b/c++/src/RleEncoderV2.cc
@@ -124,7 +124,8 @@ void RleEncoderV2::write(int64_t val) {
         }
 
         if (fixedRunLength == MAX_LITERAL_SIZE) {
-            determineEncoding(option);
+            option.encoding = DELTA;
+            option.isFixedDelta = true;
             writeValues(option);
         }
         return;
--- a/java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
+++ b/java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
@@ -755,7 +755,8 @@ public class RunLengthIntegerWriterV2 implements IntegerWriter {
 
           // if fixed runs reached max repeat length then write values
           if (fixedRunLength == MAX_SCOPE) {
-            determineEncoding();
+            encoding = EncodingType.DELTA;
+            isFixedDelta = true;
             writeValues();
           }
         } else {


dongjoon-hyun (commented):
Thank you, @guiyanakuang .

guiyanakuang:
@dongjoon-hyun @wgtmac Can you help review this, please?

c++/src/RleEncoderV2.cc
```diff
@@ -124,7 +124,8 @@ void RleEncoderV2::write(int64_t val) {
         }
 
         if (fixedRunLength == MAX_LITERAL_SIZE) {
-            determineEncoding(option);
+            option.encoding = DELTA;
+            option.isFixedDelta = true;
```
dongjoon-hyun, line 128: After this PR, the comments at line 134 will become obsolete.
> // case 2: variable delta run

java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
```diff
@@ -755,7 +755,8 @@ public void write(long val) throws IOException {
 
           // if fixed runs reached max repeat length then write values
           if (fixedRunLength == MAX_SCOPE) {
-            determineEncoding();
+            encoding = EncodingType.DELTA;
+            isFixedDelta = true;
```
dongjoon-hyun, line 759: Ditto. This PR makes line 763 a little obsolete.
> // variable delta run

dongjoon-hyun (commented):
Sorry for late review. BTW, do we need to update only `encoding` and `isFixedDelta`? What about other variables like `fixedDelta`?

guiyanakuang:
> Sorry for late review. BTW, do we need to update only `encoding` and `isFixedDelta`? What about other variables like `fixedDelta`?

`fixedDelta` initialized to 0, and `writeValues()` ends with a call to `clear()` to set the necessary variables to their default values. `fixedDelta=0` is exactly what we need. This is also the case with `bitsDeltaMax`.

c++/src/RleEncoderV2.cc
guiyanakuang: The return on line 131 has ended the handling of the fixed delta run case. I think the comment on line 134 is not obsolete.  The following is exactly how to handle variable delta run.

java/core/src/java/org/apache/orc/impl/RunLengthIntegerWriterV2.java
guiyanakuang: I think the `else` on line 762 separates the fixed delta run from the variable delta run. 763 This is followed by the variable delta run.

guiyanakuang: However, java and C++ have different styles of processing
```java
//Java
if (prevDelta == 0 && currentDelta == 0) {
  // fixed delta run
} else {
  // variable delta run
}
```

```C++
//c++
if (prevDelta == 0 && currentDelta == 0) {
  // fixed delta run
  return;
} 
// variable delta run
```

dongjoon-hyun (approved):
+1, LGTM. Sorry for the delay, @guiyanakuang .
Merged to master for Apache ORC 1.8.

PR #913, opened 2021-09-18
ORC-1005: Make that the java and C++ impl of determineEncoding in RLEV2 are consistent.

guiyanakuang:

1. Make that the java and C++ impl of determineEncoding in RLEV2 are consistent. 
2. Extracts the prepare for Direct or PatchedBase encoding into a separate function (prepareForDirectOrPatchedBase) to avoid code duplication.

The current Java determineEncoding impl is not optimal and may compute ZigzagLiterals and zzBits100p before deciding on DELTA encoding, which is redundant.

The C++ implementation is better, keeping Java consistent with C++.

Pass the CIs.

### DIFF
--- a/c++/src/RleEncoderV2.cc
+++ b/c++/src/RleEncoderV2.cc
@@ -276,6 +276,20 @@ void RleEncoderV2::preparePatchedBlob(EncodingOption& option) {
     }
 }
 
+/**
+ * Prepare for Direct or PatchedBase encoding
+ * compute zigZagLiterals and zzBits100p (Max number of encoding bits required)
+ * @return zigzagLiterals
+ */
+int64_t* RleEncoderV2::prepareForDirectOrPatchedBase(EncodingOption& option) {
+    if (isSigned) {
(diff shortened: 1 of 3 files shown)


c++/src/RleEncoderV2.cc
```diff
@@ -276,6 +276,20 @@ void RleEncoderV2::preparePatchedBlob(EncodingOption& option) {
     }
 }
 
+/**
+ * Prepare for Direct or PatchedBase encoding
+ * compute zigZagLiterals and zzBits100p (Max number of encoding bits required)
+ * @return zigzagLiterals
+ */
+int64_t* RleEncoderV2::prepareForDirectOrPatchedBase(EncodingOption& option) {
```
dongjoon-hyun, line 284: Thank you for this refactoring, @guiyanakuang .

dongjoon-hyun (approved):
+1, LGTM. Merged to master for Apache ORC 1.8.
cc @wgtmac , @pgaref , @williamhyun

PR #915, opened 2021-09-21
ORC-98: Add support for t-digests to ORC

guiyanakuang:

CollectionColumnStatisticsImpl / IntegerStatisticsImpl / DoubleStatisticsImpl / DateStatisticsImpl supports the recording of statistical information via TDigest. 

1. Easy access to histograms/quartiles. Get the distribution of the data.
2. As supporting information for your query. For example it can assist in filtering conditions and finding faster paths for pruning.

Adding relevant unit tests.

### Current Status
- [x] A java implementation to enhance ColumnStatistics with a plugin
- [x] A C++  implementation to enhance ColumnStatistics with a plugin
- [ ] Add a command to the tool to see the field's CustomStatistics name
- [ ] Add benchmarks (Complete 1/3)
- [ ] More unit tests
- [ ] More comments


java/core/pom.xml
```diff
@@ -75,6 +75,10 @@
       <groupId>org.threeten</groupId>
       <artifactId>threeten-extra</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.tdunning</groupId>
```
dongjoon-hyun, line 79: This reminds me the following Spark PR. In the community, it's a little hard to add this kind of dependency without a significant justification.
- https://github.com/apache/spark/pull/26197/files#diff-283ead138b37209525130359830c4c6eeafaf5c6639b0df82b770389d24002bfR134

dongjoon-hyun: Do we have some benchmark result for this enhancement?
> For example it can assist in filtering conditions and finding faster paths for pruning.

java/core/src/java/org/apache/orc/PhysicalWriter.java
```diff
@@ -77,6 +77,8 @@ void writeIndex(StreamName name,
   void writeBloomFilter(StreamName name,
                         OrcProto.BloomFilterIndex.Builder bloom) throws IOException;
 
+//  void writeTDigest(StreamName name, OrcProto) throws IOException;
+
```
dongjoon-hyun, line 81: Shall we remove this comments?

java/core/pom.xml
guiyanakuang: > This reminds me the following Spark PR. In the community, it's a little hard to add this kind of dependency without a significant justification.
> 
> * https://github.com/apache/spark/pull/26197/files#diff-283ead138b37209525130359830c4c6eeafaf5c6639b0df82b770389d24002bfR134

I understand the point and the fear of not being maintained. But the actual algorithm is short and easy to understand.

> https://github.com/addthis/stream-lib/blob/master/src/main/java/com/clearspring/analytics/stream/quantile/TDigest.java

Can I copy this file into the ORC project instead of the dependency?

guiyanakuang: > Do we have some benchmark result for this enhancement?
> 
> > For example it can assist in filtering conditions and finding faster paths for pruning.

@dongjoon-hyun, Currently pr is still in draft form. I will add benchmarking, unit testing and documentation. It might be a bit slow because I need to finish the work at hand first. 😃

java/core/src/java/org/apache/orc/PhysicalWriter.java
guiyanakuang: Fix in 9133711

java/core/pom.xml
wgtmac: I agree with @dongjoon-hyun that introducing a new dependency adds more complexity to this project. It looks like t-digest has been widely adopted by many projects and big name companies: https://www.sciencedirect.com/science/article/pii/S2665963820300403. It would be good if we can make it as a plugin instead of a direct dependency.

wgtmac: @guiyanakuang I am also curious of your preference of t-digest over DataSketches which is also an Apache project shipped with similar feature.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -423,6 +502,12 @@ public void merge(ColumnStatisticsImpl other) {
       if (!overflow) {
         intb.setSum(sum);
       }
+      if (tDigest != null && digestConf.getPersistence()) {
+        ByteBuffer byteBuffer = ByteBuffer.allocate(tDigest.byteSize());
+        tDigest.asBytes(byteBuffer);
+        byteBuffer.flip();
```
wgtmac, line 508: Why do we need a flip here?

java/core/src/java/org/apache/orc/DigestConf.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class DigestConf {
+
+  public static final DigestConf NO_CREATE_DIGEST = new DigestConf(false, 0);
+
+  public static final DigestConf DEFAULT_DIGEST = new DigestConf(true, 1000);
+
+  private final boolean createDigest;
+
+  private final double digestCompression;
```
wgtmac, line 30: What does it mean? Does t-digest supports compression natively (using standard compression algorithm) or simple serialization?

java/core/src/test/org/apache/orc/impl/TestDigest.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+    int row = batch.size++;
+    id.vector[row] = 1_000_000;
+    byte[] buffer = String.valueOf(1_000_000).getBytes(StandardCharsets.UTF_8);
+    name.setRef(row, buffer, 0, buffer.length);
+    writer.addRowBatch(batch);
+    writer.close();
+  }
+
+  private void read() throws IOException {
```
wgtmac, line 89: Should we encrypt the digest if column encryption is enabled to avoid data leak?

proto/orc_proto.proto
```diff
@@ -26,12 +26,14 @@ message IntegerStatistics  {
...
   optional sint64 maximum = 2;
   optional sint64 sum = 3;
+  optional bytes digest = 4;
 }
 
 message DoubleStatistics {
   optional double minimum = 1;
   optional double maximum = 2;
   optional double sum = 3;
+  optional bytes digest = 4;
```
wgtmac, line 36: Does t-digest has breaking compatibility of serialization among different versions? Additionally, dealing with serialized t-digests between Java and C++ is also something we should be aware of.

wgtmac (commented):
@guiyanakuang This work is interesting. However, we should be careful if it involves format change.

java/core/pom.xml
guiyanakuang: @wgtmac  I first learned about TDigest from this [issue](https://issues.apache.org/jira/projects/ORC/issues/ORC-98?filter=allopenissues), which inspired me to. So I don't actually have a preference, I see that DataSketches has implementations in different languages, and if it is well compatible with reading and writing serialisation in different languages, then I am in favour of using DataSketches. I need some time to learn it. :)

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
guiyanakuang: A statement after ： `intb.setDigest(ByteString.copyFrom(byteBuffer));`

```
  public static ByteString copyFrom(ByteBuffer bytes) {
    return copyFrom(bytes, bytes.remaining());
  }
```
ByteBuffer to ByteString, flipped to read the correct size.

java/core/src/test/org/apache/orc/impl/TestDigest.java
guiyanakuang: Current statistics already support encryption.
https://github.com/apache/orc/blob/cf720d7b2333618c652445299486cb9600bdbe4f/java/core/src/java/org/apache/orc/impl/ReaderImpl.java#L355-L382

proto/orc_proto.proto
guiyanakuang: Yes you are right, after researching DataSketches I will come back with my proposal.

guiyanakuang:
> @guiyanakuang This work is interesting. However, we should be careful if it involves format change.

@wgtmac Couldn't agree with you more. I am also submitting this draft in the hope of having a full discussion. Thanks for the review.

guiyanakuang:
After some testing and some thought, I have decided to modify this pr in the following way, and we will discuss any disagreements.

## Enhancing ColumnStatistics with a plugin approach
The data structures in either TDigest or datasketches can be specific implementations in the plugin. orc-core does not add any dependencies, test and benchmark modules add dependencies and specific implementations.

```proto
message Digest {
  optional string digestName = 1;
  optional bytes digestContent = 2;
}

message DoubleStatistics {
  optional double minimum = 1;
  optional double maximum = 2;
  optional double sum = 3;
  optional Digest digest = 4;
}
```
Both Java and C++ will use digestName to find specific plugin implementation. Failed to find degrades to a default empty implementation.

1. Does digest has breaking compatibility of serialization among different versions ?
Since Digest is defined as optional. Older versions will automatically ignore the Digest field when reading newer versions of files, I did some tests that looked good. This will also be added to the unit test.

2. How do I deal with the serialisation of digest between Java and C++ ?
As the enhancement is provided in the form of a plugin, if the user needs java to write C++ to read or otherwise. This requires a user implementation to ensure serialisation between languages. I thought we could add example based on datasketches (which has multiple language implementations).

Also, I think I'll add a command to the tool to see the field's digestName.

guiyanakuang:
In some environments the test has an overflow check that is not as expected. I haven't figured out why, but I believe this overflow detection code is incorrect.
https://github.com/apache/orc/blob/6da96bb8ceb64528d082974efed411c4c29f3408/c%2B%2B/src/Statistics.cc#L180-L190
A counter-example can easily be given
Assume `sum=1`, `update(std::numeric_limits<int64_t>::max(), 3);`
`value * repetitions + _stats.getSum()` is overflowed, but is still a positive number : 9223372036854775806
@dongjoon-hyun @wgtmac What do you think? :)

wgtmac:
> In some environments the test has an overflow check that is not as expected. I haven't figured out why, but I believe this overflow detection code is incorrect.
> https://github.com/apache/orc/blob/6da96bb8ceb64528d082974efed411c4c29f3408/c%2B%2B/src/Statistics.cc#L180-L190
> 
> 
> A counter-example can easily be given
> Assume `sum=1`, `update(std::numeric_limits<int64_t>::max(), 3);`
> `value * repetitions + _stats.getSum()` is overflowed, but is still a positive number : 9223372036854775806
> @dongjoon-hyun @wgtmac What do you think? :)

This piece of code was simply copied from the java side:
https://github.com/apache/orc/blob/main/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java#L362-L379

We need make sure value * repetitions does not overflow before checking the sum. : (

guiyanakuang:
@wgtmac You're right, we need to check the value * repetitions first, I've created the issue, added the c++ and java component tags and I'll try to fix it later.

wgtmac:
> Hi @dongjoon-hyun @wgtmac. After some testing and some thought, I have decided to modify this pr in the following way, and we will discuss any disagreements.
> 
> ## Enhancing ColumnStatistics with a plugin approach
> The data structures in either TDigest or datasketches can be specific implementations in the plugin. orc-core does not add any dependencies, test and benchmark modules add dependencies and specific implementations.
> 
> ```proto
> message Digest {
>   optional string digestName = 1;
>   optional bytes digestContent = 2;
> }
> 
> message DoubleStatistics {
>   optional double minimum = 1;
>   optional double maximum = 2;
>   optional double sum = 3;
>   optional Digest digest = 4;
> }
> ```
> 
> Both Java and C++ will use digestName to find specific plugin implementation. Failed to find degrades to a default empty implementation.
> 
> 1. Does digest has breaking compatibility of serialization among different versions ?
>    Since Digest is defined as optional. Older versions will automatically ignore the Digest field when reading newer versions of files, I did some tests that looked good. This will also be added to the unit test.
> 2. How do I deal with the serialisation of digest between Java and C++ ?
>    As the enhancement is provided in the form of a plugin, if the user needs java to write C++ to read or otherwise. This requires a user implementation to ensure serialisation between languages. I thought we could add example based on datasketches (which has multiple language implementations).
> 
> Also, I think I'll add a command to the tool to see the field's digestName.

So I suppose you want to use digestName as a hint to let readers know how to decode it? Should we write some useful information including type (digest, datasketches or sth else), version, binding (java, c++, etc.) so that the readers are able to decide if they can understand it? For the plugin approach on the java side, we may provide an individual module to help read/write digest.

guiyanakuang:
Agreed, I also think that using a string to mark the implementation feels unreliable. But the good thing is that the current implementation proves the feasibility. A new module that implements custom digest feels good.

dongjoon-hyun:
cc @omalley since this is ORC-98.

guiyanakuang:
I have completed a benchmark test (dba9a1a) using the current implementation for the time being. To show the benefits of custom statistics.
Added OptimizeFilterBenchmark. Tested the performance of the default query, and of the filter condition with the base percentage of filter values re-ordered by TDigest.

proportion: Ratio of cardinal number between columns
quota: Minimum cardinal number

```
Benchmark                            (proportion)  (quota)  Mode  Cnt     Score     Error  Units
OptimizeFilterBenchmark.noUseTDigest             2       10  avgt   20  1052.305 ±  10.632  us/op
OptimizeFilterBenchmark.noUseTDigest             2      100  avgt   20  1109.375 ±  10.162  us/op
OptimizeFilterBenchmark.noUseTDigest             2     1000  avgt   20  1173.790 ±  11.696  us/op
OptimizeFilterBenchmark.noUseTDigest             3       10  avgt   20  1056.139 ±   8.359  us/op
OptimizeFilterBenchmark.noUseTDigest             3      100  avgt   20  1154.665 ±   9.152  us/op
OptimizeFilterBenchmark.noUseTDigest             3     1000  avgt   20  1168.113 ±   9.115  us/op
OptimizeFilterBenchmark.useTDigest               2       10  avgt   20  1116.076 ±   6.330  us/op
OptimizeFilterBenchmark.useTDigest               2      100  avgt   20  1162.956 ±   9.865  us/op
OptimizeFilterBenchmark.useTDigest               2     1000  avgt   20  1220.028 ±  22.544  us/op
OptimizeFilterBenchmark.useTDigest               3       10  avgt   20  1114.617 ±  10.220  us/op
OptimizeFilterBenchmark.useTDigest               3      100  avgt   20  1219.488 ± 138.798  us/op
OptimizeFilterBenchmark.useTDigest               3     1000  avgt   20   651.001 ±  20.784  us/op
```

Tests show some performance loss in reading custom statistics structures, around 100 us/op. When there are sparse values in the filter conditions, reordering the filter conditions results in a larger performance gain, as this allows for early pruning.

dongjoon-hyun:
@guiyanakuang . You are an Apache ORC PMC member now. How do you want to conclude this efforts? Do you want to deliver it to Apache ORC 1.8.0 or 1.9.0?

guiyanakuang:
Sorry I've put this effort on hold. 

On the one hand I don't have an abundance of time at the moment, and as I'm not committed to open source full time, I can only use my spare time.

On the other hand, this is a lot more work than I thought it would be.
1. we need to refactor the statistics code to make it easier for users to define plugins to implement statistical records
2. we need to ensure read/write compatibility with the C++/java version

So I'd like to close this pr for now, but if anyone else is interested in completing this piece of work, or breaking it down into more detailed topics, I'd be happy to work on parts of it.

dongjoon-hyun:
I fully understand. Thank you for making a decision.

PR #916, opened 2021-09-25
ORC-1007: Fix a warning from the shade plugin

guiyanakuang:

Filtering module-info.class file during build of uber.jar
Fix the warning.
```
Warning:  Discovered module-info.class. Shading will break its strong encapsulation.
Warning:  Discovered module-info.class. Shading will break its strong encapsulation.
Warning:  Discovered module-info.class. Shading will break its strong encapsulation.
```

I don't think the module system makes sense for uber.jar, it's useful during development. The module-info.class of multiple modules will override each other during the build of uber.jar, and I think its strong encapsulation may have an effect on uber.jar.

Pass the CIs.

### DIFF
--- a/java/bench/spark/pom.xml
+++ b/java/bench/spark/pom.xml
@@ -178,6 +178,7 @@
                 <filter>
                   <artifact>*:*</artifact>
                   <excludes>
+                    <exclude>module-info.class</exclude>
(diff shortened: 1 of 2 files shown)


dongjoon-hyun (approved):

PR #921, opened 2021-09-29
ORC-1012: Support specifying columns in orc-scan

stiga-huang:

The PR adds an option to support specifying columns to be read in the orc-scan tool.

Currently the orc-scan tool will scan all columns. I find it useful to specifying columns when profiling read performance on specified data types.

Manually tested with the new option.
Added tests in tools/test/TestFileScan.cc.

### DIFF
--- a/tools/src/FileScan.cc
+++ b/tools/src/FileScan.cc
@@ -26,11 +26,12 @@
 #include <iostream>
 #include <string>
 
-void scanFile(std::ostream & out, const char* filename, uint64_t batchSize) {
+void scanFile(std::ostream & out, const char* filename, uint64_t batchSize,
+              const orc::RowReaderOptions& rowReaderOpts) {
   orc::ReaderOptions readerOpts;
   std::unique_ptr<orc::Reader> reader =
     orc::createReader(orc::readFile(filename), readerOpts);
-  std::unique_ptr<orc::RowReader> rowReader = reader->createRowReader();
+  std::unique_ptr<orc::RowReader> rowReader = reader->createRowReader(rowReaderOpts);
   std::unique_ptr<orc::ColumnVectorBatch> batch =
     rowReader->createRowBatch(batchSize);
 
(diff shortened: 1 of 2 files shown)


dongjoon-hyun (changes requested):
Thank you, @stiga-huang . BTW, do we support only ordinal column numbers?

Could you give explain how to select columns from complex types like nested columns like `struct`? It would be great if we can have a test coverage for complex types, too.

cc @wgtmac

wgtmac:
> It's great. Thank you, @stiga-huang . BTW, do we support only ordinal column numbers?
> 
> Could you give explain how to select columns from complex types like nested columns like `struct`? It would be great if we can have a test coverage for complex types, too.
> 
> cc @wgtmac

The RowReaderOptions support column pruning by fieldName (include(list<string>)), fieldId (include(list<int>)), and columnId (includeTypes(list<int>)). The last one supports pruning nested column in the struct. I think that is what @dongjoon-hyun has mentioned. Support nested pruning requires user to specify the exact columnId which is not easy if the schema is rather complex.

stiga-huang:
Currently the other tools (orc-contents and orc-memory) that support specifying columns use fieldId instead of columnId. So only top-level field ids are supported. The new option in orc-scan has the same meaning as these tools. I'd suggest adding another option, e.g. --columnIds=1,2,3, for specifying columnIds. Maybe in a separate PR. Do you think it's ok?

BTW, I slightly modify the test to include a top-level STRUCT column (fieldId=9 in examples/TestOrcFile.testSeek.orc).

wgtmac:
> Currently the other tools (orc-contents and orc-memory) that support specifying columns use fieldId instead of columnId. So only top-level field ids are supported. The new option in orc-scan has the same meaning as these tools. I'd suggest adding another option, e.g. --columnIds=1,2,3, for specifying columnIds. Maybe in a separate PR. Do you think it's ok?
> 
> BTW, I slightly modify the test to include a top-level STRUCT column (fieldId=9 in examples/TestOrcFile.testSeek.orc).

I am OK with the current PR (i.e. column selection by the field ids). The orc-scan tool is usually used to verify the integrity of an ORC file. We can add more complex column selection option to it later.

dongjoon-hyun (approved):
Thanks. I agree with you guys now, @stiga-huang and @wgtmac . Sorry for the delay.

dongjoon-hyun:
Since this is `tools`-only change, I backported this to branch-1.7.

PR #922, opened 2021-09-29
ORC-1013: Renaming the object of WriterContext in all TreeWriter's derived class

autumnust:
This is a simple change: Through reading the source code recently for doing another changes, I renamed the instance of `WriterContext` from `writer` into `writerContext`. The existing naming of the object seems misleading. The scope of this refactoring is small as the parameter is only used within the method itself. 

There's also a comment change in the `OrcFile` to reflect current status. 

For better readability. 

There's no function change so passing the existing unit tests should be sufficient.


java/core/src/java/org/apache/orc/OrcFile.java
```diff
@@ -691,7 +691,7 @@ public WriterOptions physicalWriter(PhysicalWriter writer) {
     }
 
     /**
-     * A package local option to set the memory manager.
+     * A public option to set the memory manager.
```
dongjoon-hyun, line 694: We usually don't mix like this. Especially, this minor change exists in an independent file. Please revert this. You can do this later in another PR.

java/core/src/java/org/apache/orc/impl/OutStream.java
```diff
@@ -399,4 +399,3 @@ public void suppress() {
     receiver.suppress();
   }
 }
-
```
dongjoon-hyun, line 402: Ditto. Please revert this from this PR.

dongjoon-hyun (changes requested):
If we want to change this, I prefer to use `context` instead of writerContext` because it's more consistent with the base class, `TreeWriterBase`.
```
  TreeWriterBase(TypeDescription schema,
                 WriterEncryptionVariant encryption,
                 WriterContext context) throws IOException {
```

dongjoon-hyun:
- To @autumnust . Thank you for making a PR. I agree with your intention.
- To @guiyanakuang . It's okay to approve this if you think it looks good to you. To become a committer, you had better practice to press `Approve` button with `LGTM`. :)

guiyanakuang:
@dongjoon-hyun Thanks for the reminder, I get it, the committer needs to take more responsibility.

java/core/src/java/org/apache/orc/impl/OutStream.java
autumnust: yes I think it is already updated from the newer commit as leaving last line of a file blank is actually required by the checkstyle.

java/core/src/java/org/apache/orc/OrcFile.java
autumnust: I raised another PR for this line: https://github.com/apache/orc/pull/923 
I can rebase against the upstream once that's merged in.

dongjoon-hyun: Why do you want to blocked by another PR, @autumnust ?

dongjoon-hyun:
@autumnust . I cannot merge this because of https://github.com/apache/orc/pull/922/files#diff-4c43c10da88510010bb511ba6e6892bb5ea6cc29be1b0b9d97c3bbaae5d382e2R694 . 

As you see here, we had better keep each PR independently next time. If you start to lock something like this, it degrades the community throughput because we are working asynchronously. :)

dongjoon-hyun:
The main branch is fixed. Could you rebase your PR, @autumnust ?

PR #924, opened 2021-09-30
ORC-1014. Add details when we get IOException from the file system

omalley:

It catches and wraps IOException in the data reader to give information about the range of bytes that the reader was trying to read.

To help debugging.

By reading files in S3 with a version of Hadoop that had HADOOP-16109.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
@@ -414,7 +414,15 @@ public class RecordReaderUtils {
     long offset = first.getOffset();
     int readSize = (int) (computeEnd(first, last) - offset);
     byte[] buffer = new byte[readSize];
-    file.readFully(offset, buffer, 0, buffer.length);
+    try {
+      file.readFully(offset, buffer, 0, buffer.length);
+    } catch(IOException e) {
+      throw new IOException(String.format("Failed while reading %s %d:%d",
+                                          file,
+                                          offset,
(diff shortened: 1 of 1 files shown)


java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -414,7 +414,15 @@ static void readRanges(FSDataInputStream file,
     long offset = first.getOffset();
     int readSize = (int) (computeEnd(first, last) - offset);
     byte[] buffer = new byte[readSize];
-    file.readFully(offset, buffer, 0, buffer.length);
+    try {
+      file.readFully(offset, buffer, 0, buffer.length);
+    } catch(IOException e) {
+      throw new IOException(String.format("Failed while reading %s %d:%d",
+                                          file,
```
autumnust, line 421: is `file.toString` only giving some object reference? I am assuming you are trying to print the path instead

omalley: Actually, especially on S3 it prints a bunch of other data, which is helpful. I haven't looked at DFSInputStream recently, does it just use the Object.toString?

omalley: Argh, yeah, I can see that it doesn't.

dongjoon-hyun: Could you address @autumnust 's comment?

PR #930, opened 2021-10-05
ORC-1019: Remove redundant jackson dependencies

guiyanakuang:

This PR aims to remove redundant jackson dependencies.
Unfortunately, ORC-946 forgot to remove the bench dependency on jackson. In fact, the bench module does not directly depend on jackson, only spark indirectly depends on the specified version of jackson. 

Minimising dependencies.

Pass the CIs.

Closes #928

### DIFF
--- a/java/bench/pom.xml
+++ b/java/bench/pom.xml
@@ -37,7 +37,6 @@
     <orc.version>${project.version}</orc.version>
     <parquet.version>1.12.0</parquet.version>
     <spark.version>3.1.2</spark.version>
-    <jackson.version>2.12.5</jackson.version>
     <hadoop.version>3.3.1</hadoop.version>
     <junit.version>5.8.0</junit.version>
   </properties>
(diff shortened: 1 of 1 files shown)


dongjoon-hyun (commented):
Thank you for making a PR, @guiyanakuang . BTW, this looks like a revert of ORC-807 (from @belugabehr )

So, is this the current status, @guiyanakuang ?
> In fact, the bench module does not directly depend on jackson, only spark indirectly depends on the specified version of jackson.

guiyanakuang:
> Thank you for making a PR, @guiyanakuang . BTW, this looks like a revert of [ORC-807](https://issues.apache.org/jira/browse/ORC-807) (from @belugabehr )
> 
> So, is this the current status, @guiyanakuang ?
> 
> > In fact, the bench module does not directly depend on jackson, only spark indirectly depends on the specified version of jackson.

Yes, because the code that depends on jackson has been rewritten with gson in ORC-946, which reminds me that we don't even need to declare Spark's dependency on jackson, there is no conflict.

I immediately made a new commit to remove the explicit dependency of spark on jackson.

dongjoon-hyun:
Ya, I agree with your updated opinion, @guiyanakuang . Then, shall we close this PR?

dongjoon-hyun:
Got it, @guiyanakuang !

dongjoon-hyun (approved):
+1, LGTM.

PR #936, opened 2021-10-12
ORC-1027: Allows the discovery of filters via the plugin interface.

pavibhai:
Allow injection of filters to be applied during read via a PluginService using the Java ServiceLoader.

The current integration uses SearchArguments this does allow easy integration but cannot be used to represent complex predicates e.g. predicates using a function, in these cases the predicate applicable is not optimized using LazyIO.

This allows a means to workaround this limitation by injecting user defined filters that are applied during read.

Regression Tests were successful
Additional Tests were added to test the Plugin functionality.

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -176,13 +176,13 @@ public enum OrcConf {
                         + "must have the filter\n"
                         + "reapplied to avoid using unset values in the unselected rows.\n"
                         + "If unsure please leave this as false."),
-  ALLOW_PLUGIN_FILTER("orc.filter.plugin.enable",
-                      "orc.filter.plugin.enable",
+  ALLOW_PLUGIN_FILTER("orc.filter.plugin",
+                      "orc.filter.plugin",
                       false,
                       "Enables the use of plugin filters during read. The plugin filters "
                       + "are discovered against the service "
                       + "org.apache.orc.filter.PluginFilterService, if multiple filters are "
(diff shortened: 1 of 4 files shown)


pavibhai:
> Thank you. This looks like a nice new feature for Apache ORC 1.8.

Thanks @dongjoon-hyun for your review comments. If ORC 1.8 is sometime away, would it be possible to consider the same into 1.7 as a patch release? If ORC 1.8 might get release in a couple of months then we don't need a 1.7 patch release.

dongjoon-hyun:
That's good questions.

>  If ORC 1.8 is sometime away, would it be possible to consider the same into 1.7 as a patch release?

Since Apache ORC follows the semantic versioning, it's difficult to bring this new features to 1.7.1.

> If ORC 1.8 might get release in a couple of months then we don't need a 1.7 patch release.

According to our release page, Apache ORC release cadence requires at least one year for next minor version.

- https://orc.apache.org/docs/releases.html
1.7.0: 2021-09-15 (Two years and 12 days)
1.6.0: 2019-09-03 (One year and 4 month)
1.5.0: 2018-05-14 (One year and 6 days from 1.4)
1.4.0: 2017-05-08

It means literally Apache ORC community needed more and more time to collect new features for new feature releases (new minor version like 1.8). In other words, we spent more efforts and time to maintain the old branches like we spent 40 months for 1.5.x.

It's a good candidate for discussion in our dev mailing list, @pavibhai . You can initiate the thread about your requirements and thoughts. What we need is building consensus and having a good release plan.

cc @omalley , @pgaref , @wgtmac , @williamhyun , @guiyanakuang

PR #941, opened 2021-10-19
ORC-1030: Java Tools may have missed OrcFile.MAGIC during file recovery

guiyanakuang:

This pr is aimed at fixing some bugs where the Java Tools recovery file command did not accurately find OrcFileMAGIC.
```java
        while (remaining > 0) {
          int toRead = (int) Math.min(DEFAULT_BLOCK_SIZE, remaining);
          byte[] data = new byte[toRead];
          long startPos = corruptFileLen - remaining;
          fdis.readFully(startPos, data, 0, toRead);

          // find all MAGIC string and see if the file is readable from there
          int index = 0;
          long nextFooterOffset;
          byte[] magicBytes = OrcFile.MAGIC.getBytes(StandardCharsets.UTF_8);
          while (index != -1) {
            index = indexOf(data, magicBytes, index + 1);
            if (index != -1) {
              nextFooterOffset = startPos + index + magicBytes.length + 1;
              if (isReadable(corruptPath, conf, nextFooterOffset)) {
                footerOffsets.add(nextFooterOffset);
              }
            }
          }

          System.err.println("Scanning for valid footers - startPos: " + startPos +
              " toRead: " + toRead + " remaining: " + remaining);
          remaining = remaining - toRead;
        }
```
Example:
First read:   data = [..........,'O', 'R']
Second read:  data = ['C', ...........]
The previous implementation only looked within data and did not consider OrcFile.MAGIC across two reads.

Fix bug: Java Tools may miss OrcFileMAGIC during file recovery.

Add UT.

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/FileDump.java
+++ b/java/tools/src/java/org/apache/orc/tools/FileDump.java
@@ -602,10 +602,8 @@ public final class FileDump {
 
       // Move side file to backup path
       Path sideFilePath = OrcAcidUtils.getSideFile(corruptPath);
-      if (fs.exists(sideFilePath)) {
-        Path backupSideFilePath = new Path(backupDataPath.getParent(), sideFilePath.getName());
-        moveFiles(fs, sideFilePath, backupSideFilePath);
-      }
+      Path backupSideFilePath = new Path(backupDataPath.getParent(), sideFilePath.getName());
+      moveFiles(fs, sideFilePath, backupSideFilePath);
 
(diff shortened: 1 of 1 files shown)


guiyanakuang:
> It would be great if we can have more specific PR title instead of `Fix some bugs`. If you need, you can split the PR into smaller ones.

It might be better to split it into two, I'll refine it later

dongjoon-hyun:
Thank you for updating. Is this ready for review, @guiyanakuang ?

PR #944, opened 2021-10-22
ORC-1020: [C++] Optimize RleDecoderV2::nextDirect base on bit sizes

stiga-huang:

This PR optimizes the C++ implementation of orc::RleDecoderV2::nextDirect. It provides different methods for different bit sizes, which significantly reduces the instructions and branches. Note that only bit sizes of 4,8,16,24,32,40,48,56,64 are optimized in this PR. bitSize 1 and 2 are skipped since they usually have short runs which won't benefit from loop unrolling. Other deprecated bit sizes, e.g. 3,5,6,7 are also skipped.

When the bit size is specified, we can better exhaust the buffer before calling readByte(). This also improves performance.

Perf tests show that orc::RleDecoderV2::nextDirect() dominant the time in reading random integers. There are more details in the JIRA description.
Performance improvement for scanning 1 billion unsigned numbers in different bit sizes:

bitSize | original time(s) | optimized time(s) | speedup
-- | -- | -- | --
4 | 5.93 | 3.25 | 1.8246153846
8 | 6.24 | 2.62 | 2.3816793893
16 | 7.24 | 2.59 | 2.7953667954
24 | 8.28 | 3.04 | 2.7236842105
32 | 9.45 | 2.9 | 3.2586206897
40 | 10.62 | 3.42 | 3.1052631579
48 | 11.67 | 3.56 | 3.2780898876
56 | 12.86 | 4.21 | 3.054631829
64 | 14.05 | 7.76 | 1.8105670103

Tested on Ubuntu16.04 with a 6 cores CPU (12 virtual cores) and 32GB RAM.
CPU: Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz

Add more tests in TestRleDecoder.cc

### DIFF
--- a/c++/src/RLEv2.hh
+++ b/c++/src/RLEv2.hh
@@ -161,56 +161,25 @@ private:
     bitSize = 0;
   }
 
-  unsigned char readByte() {
-  if (bufferStart == bufferEnd) {
-    int bufferLength;
-    const void* bufferPointer;
-    if (!inputStream->Next(&bufferPointer, &bufferLength)) {
-      throw ParseError("bad read in RleDecoderV2::readByte");
-    }
-    bufferStart = static_cast<const char*>(bufferPointer);
-    bufferEnd = bufferStart + bufferLength;
-  }
-
-  unsigned char result = static_cast<unsigned char>(*bufferStart++);
-  return result;
-}
+  unsigned char readByte();
 
   int64_t readLongBE(uint64_t bsz);
   int64_t readVslong();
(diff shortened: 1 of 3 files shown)


wgtmac (approved):
Sorry for the late reply. LGTM +1.

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Thank you, @stiga-huang , @yuruiz , @wgtmac !

PR #952, opened 2021-11-01
ORC-1004: Make orc writer support the selected vector

guiyanakuang:

This pr is aimed at making ORC Writer support write thed selected vector.

Added InternalVectorizedRowBatch and InternalColumnVector classes to encapsulate and replace VectorizedRowBatc and ColumnVector.
WholeVectorizedRowBatch/WholeColumnVector represents a batch/vector that does not use selected.
SelectedVectorizedRowBatch/SelectedColumnVector represents a batch/vector that does use selected.

During the write process we can use the same interface (`public int getValueOffset(int offset) `) to get the offsets, making the changes minimal and not doing anything redundant.

SelectedColumnVector

When writing to the encrypted column, a copy of the Selected Vector is made without modifying the maskData interface, resulting in an unused Selected Vector. Afterwards, the processing is the same as before.

Currently the ORC writer doesn't support the selected vector. This could cause clients that expect it to be supported to get trash rows in the output.

Add UT TestSelectedVector.java

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/SelectedColumnVector.java
+++ b/java/core/src/java/org/apache/orc/impl/SelectedColumnVector.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.orc.impl;
+
+import org.apache.hadoop.hive.ql.exec.vector.ColumnVector;
+
+import java.util.function.Function;
+
+public class SelectedColumnVector extends InternalColumnVector {
+
+  private final int[] selected;
+
+  public SelectedColumnVector(ColumnVector columnVector, int[] selected) {
+    super(columnVector);
+    this.selected = selected;
+  }
+
+  @Override
+  public void ensureSize(int size, boolean preserveData) {
+    throw new UnsupportedOperationException(
+        "The ensureSize operation is not supported for SelectedColumnVector");
+  }
+
+  @Override
+  public boolean isNull(int offset) {
+    return columnVector.isNull[getValueOffset(offset)];
+  }
+
+  @Override
+  public int getValueOffset(int offset) {
+    return selected[offset];
+  }
+
+  @Override
+  public Function<ColumnVector, InternalColumnVector> encapsulationFunction() {
+    return vector -> new SelectedColumnVector(vector, selected);
+  }
+
+}
--- a/java/core/src/java/org/apache/orc/impl/WriterImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/WriterImpl.java
@@ -683,20 +683,22 @@ public class WriterImpl implements WriterInternal, MemoryManager.Callback {
 
   @Override
   public void addRowBatch(VectorizedRowBatch batch) throws IOException {
+    InternalVectorizedRowBatch internalBatch = InternalVectorizedRowBatch.encapsulation(batch);
+    int batchSize = internalBatch.size();
     try {
       // If this is the first set of rows in this stripe, tell the tree writers
       // to prepare the stripe.
-      if (batch.size != 0 && rowsInStripe == 0) {
+      if (batchSize != 0 && rowsInStripe == 0) {
         treeWriter.prepareStripe(stripes.size() + 1);
       }
       if (buildIndex) {
         // Batch the writes up to the rowIndexStride so that we can get the
         // right size indexes.
         int posn = 0;
-        while (posn < batch.size) {
-          int chunkSize = Math.min(batch.size - posn,
+        while (posn < batchSize) {
+          int chunkSize = Math.min(batchSize - posn,
               rowIndexStride - rowsInIndex);
-          treeWriter.writeRootBatch(batch, posn, chunkSize);
+          treeWriter.writeRootBatch(internalBatch, posn, chunkSize);
           posn += chunkSize;
           rowsInIndex += chunkSize;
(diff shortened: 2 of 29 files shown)


dongjoon-hyun (commented):
Thank you for making a PR, @guiyanakuang . This looks like a big PR. :)

guiyanakuang:
> Thank you for making a PR, @guiyanakuang . This looks like a big PR. :)

Yes, although the design prioritizes minimizing changes, it still requires changing the writer for each column. This results in a significant number of modified rows. : )

wgtmac:
@guiyanakuang I am wondering whether the selected vector only applies to the root struct-typed column vector. And I am curious if this patch is a part of some other big work items?

guiyanakuang:
> @guiyanakuang I am wondering whether the selected vector only applies to the root struct-typed column vector. And I am curious if this patch is a part of some other big work items?

@wgtmac Yes, the selected marker is at the row level. It will only work on the root.  The purpose of this pr is to resolve [issue](https://issues.apache.org/jira/projects/ORC/issues/ORC-1004?filter=allopenissues). This feature helps me to write the results to the ORC file by filtering, which is useful for creating materialized views. It just happens to be useful for my current work, so I implemented it

wgtmac:
> > @guiyanakuang I am wondering whether the selected vector only applies to the root struct-typed column vector. And I am curious if this patch is a part of some other big work items?
> 
> @wgtmac Yes, the selected marker is at the row level. It will only work on the root. The purpose of this pr is to resolve [issue](https://issues.apache.org/jira/projects/ORC/issues/ORC-1004?filter=allopenissues). This feature helps me to write the results to the ORC file by filtering, which is useful for creating materialized views. It just happens to be useful for my current work, so I implemented it

As the selection vector only applies to the root struct type, is it possible to use null buffer of the root StructColumnVector (which is always not null for now)? We can ignore writing rows of null data there. This is much simpler but hacky.

guiyanakuang:
> > > @guiyanakuang I am wondering whether the selected vector only applies to the root struct-typed column vector. And I am curious if this patch is a part of some other big work items?
> > 
> > 
> > @wgtmac Yes, the selected marker is at the row level. It will only work on the root. The purpose of this pr is to resolve [issue](https://issues.apache.org/jira/projects/ORC/issues/ORC-1004?filter=allopenissues). This feature helps me to write the results to the ORC file by filtering, which is useful for creating materialized views. It just happens to be useful for my current work, so I implemented it
> 
> As the selection vector only applies to the root struct type, is it possible to use null buffer of the root StructColumnVector (which is always not null for now)? We can ignore writing rows of null data there. This is much simpler but hacky.

Sorry I didn't answer accurately, but in fact the root may also be a single column primitive type. Also the final VectorizedRowBatch obtained by filter query, the unselected rows may not be null. StructColumnVector seems to be written in such a way that it only appears when the sub-column type is struct.

java/core/src/java/org/apache/orc/impl/SelectedColumnVector.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  public boolean isNull(int offset) {
+    return columnVector.isNull[getValueOffset(offset)];
+  }
+
+  @Override
+  public int getValueOffset(int offset) {
+    return selected[offset];
```
wgtmac, line 47: What the main reason for using an int array as the selected vector?

java/core/src/java/org/apache/orc/impl/WriterImpl.java
```diff
@@ -683,20 +683,22 @@ public void addUserMetadata(String name, ByteBuffer value) {
...
         // Batch the writes up to the rowIndexStride so that we can get the
         // right size indexes.
         int posn = 0;
-        while (posn < batch.size) {
-          int chunkSize = Math.min(batch.size - posn,
+        while (posn < batchSize) {
+          int chunkSize = Math.min(batchSize - posn,
               rowIndexStride - rowsInIndex);
-          treeWriter.writeRootBatch(batch, posn, chunkSize);
+          treeWriter.writeRootBatch(internalBatch, posn, chunkSize);
```
wgtmac, line 701: If I understand correctly, the selected vector only applies to each row in the root VectorizedRowBatch. We can simply use a bool array in the root VectorizedRowBatch to indicate whether each row is selected or not. Then we can split the input VectorizedRowBatch into smaller pieces of continuously selected VectorizedRowBatchs and thus change posn & chunkSize accordingly.

java/core/src/java/org/apache/orc/impl/SelectedColumnVector.java
guiyanakuang: @wgtmac I am very sorry, I missed the message so late to reply

The reason is that I can very easily construct the InternalColumnVector from vectorizedRowBatch, selected is actually an array passed in by vectorizedRowBatch, so there is no need to provide additional space to store selection information.

SelectedVectorizedRowBatch.java

```java
  @Override
  public InternalColumnVector cols(int col) {
    return new SelectedColumnVector(vectorizedRowBatch.cols[col], vectorizedRowBatch.getSelected());
  }
```

java/core/src/java/org/apache/orc/impl/WriterImpl.java
guiyanakuang: Reusing Selected within a VectorizedRowBatch actually ensures consistent processing across types.
Let's look at a complex case provided in the test case 
```java
  public void testWriteComplexTypeUseSelectedVector() throws IOException {
    TypeDescription schema =
        TypeDescription.fromString("struct<a:map<int,uniontype<int,string>>," +
            "b:array<struct<c:int>>>");
    ....
 }    
```
For example, we have three rows of data, and the third row is selected after the filter. Before reading the third map, we need to determine the offsets of the key and value fields, since the first two maps may have multiple key-value pairs. Of course, if the key and value are also complex types, recursion may be required.

So processing only at the entry does not cover all cases and each type needs to be processed separately.

Imagine that the recursive processing eventually comes to the base type, which doesn't know what offset means. offset may be relative to the row, or it may be a value that has gone through multiple jumps.

wgtmac: I think the root treeWriter can deal with selected nested columns very well. Let me demonstrate my idea in this patch: https://github.com/wgtmac/orc/commit/42ee957be4113af62ccb59319d861fb9462a7c1e

guiyanakuang: I feel a sense of clarity,  @wgtmac thank you for the idea, I will use my test cases later to verify

guiyanakuang: The test passed perfectly. Very nice implementation. I'll close the pr. Wait for your pr and I can contribute test cases

wgtmac: Thanks for the verification! I will do it later.

PR #959, opened 2021-11-06
ORC-1047: Handle quoted field names during string schema parsing

noirello:
Improve parsing schema string with `Type::buildTypeFromString` to handle quoted field names and have stricter validations.

The current implementation cannot handle quoted field names and allows parsing string schemas that the Java implementation would reject (e.g. `struct<bigint>`, `map(boolean,float)`). It also cannot parse schema with `timestamp with local time zone` in the root.

Ran the existing test suites locally with the newly added tests for quoted field names and invalid schemas.

### DIFF
--- a/c++/src/TypeImpl.cc
+++ b/c++/src/TypeImpl.cc
@@ -650,7 +650,7 @@ namespace orc {
       if (!closed) {
         throw std::logic_error("Invalid field name. Unmatched quote");
       }
-      if (oss.tellp() == 0) {
+      if (oss.tellp() == std::streamoff(0)) {
         throw std::logic_error("Empty quoted field name.");
       }
       return std::make_pair(oss.str(), pos + 1);
--- a/c++/src/TypeImpl.cc
+++ b/c++/src/TypeImpl.cc
@@ -592,6 +592,45 @@ namespace orc {
     return return_value;
   }
 
+  std::pair<std::string, size_t> TypeImpl::parseName(const std::string &input,
+                                                     const size_t start,
+                                                     const size_t end) {
+    size_t pos = start;
+    if (input[pos] == '`') {
+      bool closed = false;
+      std::ostringstream oss;
+      while (pos < end) {
+        char ch = input[++pos];
+        if (ch == '`') {
+          if (pos < end && input[pos+1] == '`') {
+            ++pos;
+            oss.put('`');
+          } else {
+            closed = true;
+            break;
+          }
+        } else {
+          oss.put(ch);
+        }
+      }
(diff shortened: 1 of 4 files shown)


noirello:
Unfortunately, I have no idea why the new test case stuck during the AppVeyor build. I cannot build the exact version locally, but using a newer MSVC++ (2016) and gtest (1.10.0) works as expected,

guiyanakuang:
> Unfortunately, I have no idea why the new test case stuck during the AppVeyor build. I cannot build the exact version locally, but using a newer MSVC++ (2016) and gtest (1.10.0) works as expected,

You can log in to https://ci.appveyor.com/ with your github account and select your own fork ORC project so you can do active testing, the default is the main branch, which you can change in the settings

noirello:
I had to remove the field name with the non-ascii characters (`èœ`).
It caused a runtime error:

appveyor

It looks like, this has been improved in recent msvc versions.

dongjoon-hyun:
Thank you for updating, @noirello .

dongjoon-hyun:
cc @wgtmac

c++/src/TypeImpl.cc
```diff
@@ -218,7 +227,19 @@ namespace orc {
...
           result += ",";
         }
-        result += fieldNames[i];
+        if (isUnquotedFieldName(fieldNames[i])) {
+          result += fieldNames[i];
+        } else {
+          std::string name(fieldNames[i]);
+          size_t pos = 0;
+          while ((pos = name.find("`", pos)) != std::string::npos) {
+            name.replace(pos, 1, "``");
```
wgtmac, line 236: If we call toString() to a type with quote and get a string. Can we call buildTypeFromString() on that string to get back the original type object?

c++/src/TypeImpl.cc
```diff
@@ -678,79 +795,60 @@ namespace orc {
     } else if (category == "date") {
       return std::unique_ptr<Type>(new TypeImpl(DATE));
```
wgtmac, line 796: missing validatePrimitiveType before return

c++/src/TypeImpl.cc
wgtmac: Sorry for the late reply. I got the answer from your test case (TEST(TestType, quotedFieldNames)) already.

wgtmac (approved):
+1 LGTM

dongjoon-hyun:
Please feel free to merge this. :)

dongjoon-hyun:
Merged to `main` for Apache ORC 1.8.

PR #961, opened 2021-11-10
ORC-492: Avoid potential `ArrayIndexOutOfBoundsException` when getting WriterVersion

guiyanakuang:

This pr is aimed at avoiding potential ArrayIndexOutOfBoundsException possible when getting WriterVersion.

Fix bug.
Return the Future version when val is equal to versions.length to avoid throwing ArrayIndexOutOfBoundsException on line 248.

Pass the CIs.

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcFile.java
+++ b/java/core/src/java/org/apache/orc/OrcFile.java
@@ -242,7 +242,7 @@ public class OrcFile {
             val + " for writer " + writer);
       }
       WriterVersion[] versions = values[writer.id];
-      if (val < 0 || versions.length < val) {
+      if (val < 0 || versions.length <= val) {
         return FUTURE;
       }
       WriterVersion result = versions[val];


java/core/src/java/org/apache/orc/OrcFile.java
```diff
@@ -242,7 +242,7 @@ public static WriterVersion from(WriterImplementation writer, int val) {
             val + " for writer " + writer);
       }
       WriterVersion[] versions = values[writer.id];
-      if (val < 0 || versions.length < val) {
+      if (val < 0 || val >= versions.length) {
```
dongjoon-hyun, line 245: If you don't mind, could you keep the previous style like the following?
```java
- if (val < 0 || val >= versions.length) {
- if (val < 0 || versions.length <= val) {
```

dongjoon-hyun (approved):
+1, LGTM (with one minor style comment)

java/core/src/java/org/apache/orc/OrcFile.java
guiyanakuang: I adjusted the order on purpose, comparing variables on the same side looks better to understand, that's how I feel.
Of course I don't mind changing it back, but I'd like to mention your opinion.

guiyanakuang: 6adff23, keep the previous style.  : ）

dongjoon-hyun:
Thank you for updating. Merged to main.

dongjoon-hyun:
I backported this to branch-1.7.

PR #967, opened 2021-12-01
ORC-1053: Fix time zone offset precision when convert tool converts `LocalDateTime` to `Timestamp` is not consistent with the internal default precision of ORC

guiyanakuang:

```java
// use tool compute offset : 17762
int toolOffset = ((LocalDateTime) temporalAccessor).atZone(TimeZone.getTimeZone("America/New_York").toZoneId()).getOffset().getTotalSeconds();

// in orc internal compute offset: 18000
int orcInternalOffset = TimeZone.getTimeZone("America/New_York").getRawOffset() / 1000
```

This pr is designed to modify the implementation of the LocalDateTime to Timestamp conversion so that the time zone accuracy is consistent with the ORC internal accuracy during the conversion

Avoid inconsistencies between converted data and expected data from convert tools.

Add issue-specific unit test

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/convert/CsvReader.java
+++ b/java/tools/src/java/org/apache/orc/tools/convert/CsvReader.java
@@ -43,7 +43,6 @@ import java.sql.Timestamp;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
-import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.TemporalAccessor;
--- a/java/tools/src/java/org/apache/orc/tools/convert/CsvReader.java
+++ b/java/tools/src/java/org/apache/orc/tools/convert/CsvReader.java
@@ -303,8 +303,7 @@ public class CsvReader implements RecordReader {
           Timestamp timestamp = Timestamp.from(offsetDateTime.toInstant());
           vector.set(row, timestamp);
         } else if (temporalAccessor instanceof LocalDateTime) {
-          ZonedDateTime tz = ((LocalDateTime) temporalAccessor).atZone(ZoneId.systemDefault());
-          Timestamp timestamp = Timestamp.from(tz.toInstant());
+          Timestamp timestamp = Timestamp.valueOf((LocalDateTime) temporalAccessor);
           vector.set(row, timestamp);
         } else {
           column.noNulls = false;
@@ -371,5 +370,16 @@ public class CsvReader implements RecordReader {
       default:
         throw new IllegalArgumentException("Unhandled type " + schema);
     }
+  }
+
+  public static void main(String[] args) {
+    DateTimeFormatter timestampFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS zZ");
+    TemporalAccessor temporalAccessor =
+        timestampFormat.parseBest("0001-01-01 00:00:00.000 GMT+0000",
+            ZonedDateTime::from, OffsetDateTime::from, LocalDateTime::from);
+    System.out.println(temporalAccessor.getClass().getSimpleName());
+    ZonedDateTime zonedDateTime = (ZonedDateTime) temporalAccessor;
+    Timestamp timestamp = Timestamp.from(zonedDateTime.toInstant());
+    System.out.println(timestamp);
   }
 }
--- a/java/tools/src/test/org/apache/orc/tools/convert/TestConvert.java
+++ b/java/tools/src/test/org/apache/orc/tools/convert/TestConvert.java
@@ -89,7 +89,6 @@ public class TestConvert {
     VectorizedRowBatch batch = readSchema.createRowBatch();
     RecordReader rowIterator = reader.rows(reader.options().schema(readSchema));
     TimestampColumnVector tcv = (TimestampColumnVector) batch.cols[0];
-    rowIterator.nextBatch(batch);
 
     while (rowIterator.nextBatch(batch)) {
       for (int row = 0; row < batch.size; ++row) {
--- a/java/tools/src/test/org/apache/orc/tools/convert/TestConvert.java
+++ b/java/tools/src/test/org/apache/orc/tools/convert/TestConvert.java
@@ -29,16 +29,21 @@ import org.apache.orc.OrcFile;
 import org.apache.orc.Reader;
 import org.apache.orc.RecordReader;
 import org.apache.orc.TypeDescription;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
 import java.sql.Timestamp;
+import java.util.TimeZone;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class TestConvert {
+
+  public static final TimeZone DEFAULT_TIME_ZONE = TimeZone.getDefault();
 
   Path workDir = new Path(System.getProperty("test.tmp.dir"));
   Configuration conf;
@@ -52,6 +57,16 @@ public class TestConvert {
     fs.setWorkingDirectory(workDir);
     testFilePath = new Path("TestConvert.testConvert.orc");
     fs.delete(testFilePath, false);
+  }
+
+  @BeforeAll
+  public static void changeDefaultTimeZone() {
+    TimeZone.setDefault(TimeZone.getTimeZone("America/New_York"));
+  }
+
+  @AfterAll
+  public static void resetDefaultTimeZone() {
+    TimeZone.setDefault(DEFAULT_TIME_ZONE);
   }
 
   @Test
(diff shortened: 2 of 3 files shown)


guiyanakuang:
cc @dongjoon-hyun  @williamhyun  : )

dongjoon-hyun:
Thank you so much, @guiyanakuang !

java/tools/src/test/org/apache/orc/tools/convert/TestConvert.java
```diff
@@ -0,0 +1,89 @@
...
+  public void openFileSystem () throws Exception {
+    conf = new Configuration();
+    fs = FileSystem.getLocal(conf);
+    fs.setWorkingDirectory(workDir);
+    testFilePath = new Path("TestConvert.testConvert.orc");
+    fs.delete(testFilePath, false);
+  }
+
+  @Test
+  public void testConvertCustomTimestampFromCsv() throws IOException, ParseException {
```
dongjoon-hyun, line 58: Thank you for adding this.

java/tools/src/test/org/apache/orc/tools/convert/TestConvert.java
```diff
@@ -0,0 +1,89 @@
...
+    VectorizedRowBatch batch = readSchema.createRowBatch();
+    RecordReader rowIterator = reader.rows(reader.options().schema(readSchema));
+    TimestampColumnVector tcv = (TimestampColumnVector) batch.cols[0];
+    rowIterator.nextBatch(batch);
+
+    while (rowIterator.nextBatch(batch)) {
+      for (int row = 0; row < batch.size; ++row) {
+        Timestamp timestamp = Timestamp.valueOf(timeValues[row]);
+        assertEquals(timestamp.getTime(), tcv.time[row]);
+        assertEquals(timestamp.getNanos(), tcv.nanos[row]);
```
dongjoon-hyun, line 83: This seems to pass without your patch. Could you verify this again?

dongjoon-hyun:
Also, cc @wgtmac since C++ tool seems to have the same issue according to the JIRA.

java/tools/src/java/org/apache/orc/tools/convert/CsvReader.java
```diff
@@ -372,4 +370,15 @@ Converter buildConverter(IntWritable startOffset, TypeDescription schema) {
         throw new IllegalArgumentException("Unhandled type " + schema);
     }
   }
+
+  public static void main(String[] args) {
```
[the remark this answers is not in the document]
guiyanakuang: Very sorry for committing some validation code

java/tools/src/test/org/apache/orc/tools/convert/TestConvert.java
guiyanakuang: I verified before submitting that my local default timezone, and the America/New_York timezone on which jira uploaded the file, would not pass the test based on past code.

The reason for this is the inconsistent interpretation of the timestamp offfset accuracy when writing and reading.

guiyanakuang:
```java
    Timestamp timestamp1 = Timestamp.from(localDateTime.toInstant());
    Timestamp timestamp2 = Timestamp.valueOf(localDateTime);
```
Both methods use the local time zone to calculate the timestamp, but the offset precision of the time zone of the previous method is seconds (for New York is 17762 s), and the precision of the latter method is hours (for New York is 18000 s, which is 5 hours)
And the readings are interpreted by the latter precision (hour)

dongjoon-hyun: I ran this yesterday and today. It passed on my laptop without your patch.
```
$ git diff main --stat
 java/tools/src/test/org/apache/orc/tools/convert/TestConvert.java | 89 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 89 insertions(+)

$ mvn package -pl tools -Dtest=org.apache.orc.tools.convert.TestConvert
...
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.orc.tools.convert.TestConvert
Processing test.csv
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.337 s - in org.apache.orc.tools.convert.TestConvert
[INFO]
[INFO] Results:
[INFO]
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
...
```

dongjoon-hyun: The difference between you and me seems to be the timezones. You are in Asia timezone and I'm in America/Los Angeles timezone.

dongjoon-hyun: Can we make the test case to be independent from the tester's environment?

guiyanakuang: Maybe I should specify the time zone to test, but there are many methods inside ORC to get the local default time zone, so it's not good to cover it uniformly at the moment, I need to think.

guiyanakuang: @dongjoon-hyun I added two hooks using annotations to ensure that the default time zone during testing this class is New York, which feels good. Also I removed a redundant statement :sweat_smile:. 
You can try again when you have time, I'm sure the old code will not pass the test.

guiyanakuang:
> Also, cc @wgtmac since C++ tool seems to have the same issue according to the JIRA.

I have also debugged some of the C++ code. It looks like `strptime` can't convert values before 1970 on mac os.
https://github.com/apache/orc/blob/334bf1f2c605f38c7e75ec81d1dab93c31fc8459/tools/src/CSVFileImport.cc#L257

dongjoon-hyun (approved):
For C++ case, could you file a new JIRA for that?
ORC-1053 is enough for Java-only patch

dongjoon-hyun:
Merged to main/1.7.

guiyanakuang:
> +1, LGTM. Thank you, @guiyanakuang . For C++ case, could you file a new JIRA for that? [ORC-1053](https://issues.apache.org/jira/browse/ORC-1053) is enough for Java-only patch

Okay, I created the ORC-1055

dongjoon-hyun:
Thank you!

wgtmac:
> > Also, cc @wgtmac since C++ tool seems to have the same issue according to the JIRA.
> 
> I have also debugged some of the C++ code. It looks like `strptime` can't convert values before 1970 on mac os.
> 
> https://github.com/apache/orc/blob/334bf1f2c605f38c7e75ec81d1dab93c31fc8459/tools/src/CSVFileImport.cc#L257

Pre-1970 timestamp is a notorious issue. Not sure if these test cases will help: 
https://github.com/apache/orc/blob/main/c%2B%2B/test/TestWriter.cc#L621
https://github.com/apache/orc/blob/main/c%2B%2B/test/TestWriter.cc#L741

guiyanakuang:
@wgtmac , Unfortunately I think this is a new issue.
```c++
int main(int,char* argv[]){
  struct tm timeStruct;
  std::string name = "0001-01-01 00:00:00.000";
  char *left=strptime(name.c_str(), "%Y-%m-%d %H:%M:%S", &timeStruct);
  time_t _time = timegm(&timeStruct);
  char *tail;
  double d = strtod(left, &tail);
  long na;
  if (tail != left) {
    na = static_cast<long>(d * 1000000000.0);
  } else {
    na = 0;
  }
  std::cout << _time << " " << na << std::endl;
}
```
On my mac machine the output is "-1 0"
Using the online C++ runtime environment claimed on Linux, the output is "-62135596800 0"
Time format strings that appear to be older than 1970 get -1 on the mac.

> https://github.com/apache/orc/blob/main/c%2B%2B/test/TestWriter.cc#L621
> https://github.com/apache/orc/blob/main/c%2B%2B/test/TestWriter.cc#L741

It looks like the first test verifies that we can write time as a negative number.
The second one holds our custom time zone.  So I think this is a new issue.

Of course I verified the `converted_by_cpp.orc` uploaded by the user in ORC-1055, and it is clear that it is not converted on the mac, so there should be another issue that is not yet clear

wgtmac:
> @wgtmac , Unfortunately I think this is a new issue.
> 
> ```c++
> int main(int,char* argv[]){
>   struct tm timeStruct;
>   std::string name = "0001-01-01 00:00:00.000";
>   char *left=strptime(name.c_str(), "%Y-%m-%d %H:%M:%S", &timeStruct);
>   time_t _time = timegm(&timeStruct);
>   char *tail;
>   double d = strtod(left, &tail);
>   long na;
>   if (tail != left) {
>     na = static_cast<long>(d * 1000000000.0);
>   } else {
>     na = 0;
>   }
>   std::cout << _time << " " << na << std::endl;
> }
> ```
> 
> On my mac machine the output is "-1 0" Using the online C++ runtime environment claimed on Linux, the output is "-62135596800 0" Time format strings that appear to be older than 1970 get -1 on the mac.
> 
> > https://github.com/apache/orc/blob/main/c%2B%2B/test/TestWriter.cc#L621
> > https://github.com/apache/orc/blob/main/c%2B%2B/test/TestWriter.cc#L741
> 
> It looks like the first test verifies that we can write time as a negative number. The second one holds our custom time zone. So I think this is a new issue.
> 
> Of course I verified the `converted_by_cpp.orc` uploaded by the user in [ORC-1055](https://issues.apache.org/jira/browse/ORC-1055), and it is clear that it is not converted on the mac, so there should be another issue that is not yet clear

Thanks for the input. I will take a look later this week.

dongjoon-hyun:
Thank you so much, @guiyanakuang and @wgtmac .

PR #971, opened 2021-12-14
ORC-1060: Reduce memory usage when vectorized reading dictionary string encoding columns

expxiaoli:

In old code, when dictionary string encoding columns are read by vectorized reading, 2 copy of current stripe's dictionary data and 1 copy of next stripe's dictionary data are hold in memory when reading across different stripes. That could make vectorized reading's memory usage is larger than row reading. This patch fixes this issue, and only hold 1 copy of current stripe's dictionary data.

This patch logic has 3 parts:

1) Directly read data to primitive byte array, rather than using DynamicByteArray as intermediate variable. Using DynamicByteArray as intermediate variable causes 2 copy of current stripe's dictionary data are hold in memory.

2) Lazy read dictionary data until read current batch data. In previous code, RecordReaderImpl class's nextBatch method reads dictionary data of next stripe through advanceToNextRow method, then memory will hold two stripe's dictionary data. Through lazy read logic, only one stripe's dictionary data is hold in memory when reading across different stripes.

3) Before lazy read dictionary data from current stripe, remove batch data's reference to dictionary data from previous stripe. This could allow GC to clean previous stripe's dictionary data memory.

Reduce memory usage.

Pass the existing CIs.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
@@ -2190,7 +2190,7 @@ public class TreeReaderFactory {
       super.startStripe(planner, readPhase);
 
       StreamName name = new StreamName(columnId,
-              OrcProto.Stream.Kind.DICTIONARY_DATA);
+          OrcProto.Stream.Kind.DICTIONARY_DATA);
       dictionaryStream = planner.getStream(name);
       initDictionary = false;
 
@@ -2203,7 +2203,7 @@ public class TreeReaderFactory {
       // set up the row reader
       name = new StreamName(columnId, OrcProto.Stream.Kind.DATA);
       reader = createIntegerReader(encoding.getKind(),
-              planner.getStream(name), false, context);
+          planner.getStream(name), false, context);
     }
 
     private void readDictionaryLengthStream(InStream in, OrcProto.ColumnEncoding encoding)
--- a/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
@@ -2148,12 +2148,15 @@ public class TreeReaderFactory {
    */
   public static class StringDictionaryTreeReader extends TreeReader {
     private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
-    private DynamicByteArray dictionaryBuffer;
     private int[] dictionaryOffsets;
     protected IntegerReader reader;
+    private InStream lengthStream;
+    private InStream dictionaryStream;
+    private OrcProto.ColumnEncoding lengthEncoding;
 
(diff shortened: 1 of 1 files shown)


expxiaoli:
see background info here: https://issues.apache.org/jira/browse/ORC-1060

dongjoon-hyun:
Thank you for making a PR, @expxiaoli .

expxiaoli:
Here is perf test.
I create a orc table named src_table with map<string, string> column named mapping, which stripe's string dictionary could occupy 466M memory.
Then I run a spark query to read this column:
insert overwrite table res_table select mapping['tag_a'] from src_table;

With old orc lib, only if I set executor-memory to equal or larger than 2500M, the query could run successfully. Otherwise the query will fail with OOM exception. Here is perf result with MAT tool when I run query with 2500M executor-memory
内存占用情况_DynamicByteArray

With orc lib with this new patch, the query could run successfully when executor-memory is decreased to 1200M.

expxiaoli:
@pgaref @wgtmac Could you review and verify it? 
For reading string dictionary encoding column, this PR could reduce batch reading's memory usage to nearly row reading's memory usage, which could solve OOM issue in migration work from row reading to batch reading.

dongjoon-hyun (commented):
Is there any potential regression in terms of the speed?

expxiaoli:
@dongjoon-hyun In my perf test, speed is no regression for this patch. Here is scan time result for spark run query "insert overwrite table res_table select mapping['tag_a'] from src_table"

"scan time" is spark metric in DataSourceScanExec class's doExecuteColumnar method , which only metric time for scan operator

executor-memory | new ORC with this patch | old ORC
2500M                   | 37.2S                                 | 34.6S
2250M                   | 36.5S                                 | OOM
1100M                    | 30.2S                                 | OOM
1000M                    | OOM                                 | OOM

Besides, this patch removes memory allocation for DynamicByteArray as well as memory copy from DynamicByteArray to primitive byte array, and do NOT add other time consuming logic. I think there is NO potential regression for speed.

guiyanakuang:
Before
`InStream` ----> `DynamicByteArray(data[][])` -----> `byte[]`
After
`InStream` ----> `byte[]`

`DynamicByteArray` plays no other role in the context and seems completely redundant, I approve of this pr

dongjoon-hyun (approved):
+1, LGTM. Merged to master. Thank you, @expxiaoli and @expxiaoli .
Merry Christmas!

dongjoon-hyun:
@expxiaoli . I added you to the Apache ORC contributor group and assigned ORC-1060 to you.
Welcome to the Apache ORC community.

dongjoon-hyun:
I cherry-picked this to branch-1.7 for next Apache ORC 1.7.3 release.

expxiaoli:
Thanks @dongjoon-hyun

PR #973, opened 2021-12-20
[ORC-1062]Rename orc.schema.evolution.case.sensitive to orc.schema.evolution.case.aware

yikf:
Currently, OrcConf have a enum configuration to determine if the comparision of field names in schema evolution is case sensitive, However, the code block using this configuration thinks that this configuration should be orc.schema.evolution.case.aware

code simplification

Pass GitHub Action and jenkins

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -164,8 +164,8 @@ public enum OrcConf {
       "testing.  Setting this too low may negatively affect performance."),
   OVERWRITE_OUTPUT_FILE("orc.overwrite.output.file", "orc.overwrite.output.file", false,
     "A boolean flag to enable overwriting of the output file if it already exists.\n"),
-  IS_SCHEMA_EVOLUTION_CASE_SENSITIVE("orc.schema.evolution.case.sensitive",
-      "orc.schema.evolution.case.sensitive", true,
+  IS_SCHEMA_EVOLUTION_CASE_AWARE("orc.schema.evolution.case.aware",
+      "orc.schema.evolution.case.aware", true,
       "A boolean flag to determine if the comparision of field names " +
       "in schema evolution is case sensitive .\n"),
   ALLOW_SARG_TO_FILTER("orc.sarg.to.filter", "org.sarg.to.filter", false,
--- a/java/core/src/java/org/apache/orc/Reader.java
+++ b/java/core/src/java/org/apache/orc/Reader.java
@@ -229,7 +229,7 @@ public interface Reader extends Closeable {
     private Boolean tolerateMissingSchema = null;
     private boolean forcePositionalEvolution;
     private boolean isSchemaEvolutionCaseAware =
-        (boolean) OrcConf.IS_SCHEMA_EVOLUTION_CASE_SENSITIVE.getDefaultValue();
+        (boolean) OrcConf.IS_SCHEMA_EVOLUTION_CASE_AWARE.getDefaultValue();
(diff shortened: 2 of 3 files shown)


yikf:
cc @dongjoon-hyun Thanks for your review.

yikf:
@dongjoon-hyun  Also, the configuration default value is true, There are some compatibility problems

We have a case, In spark 2.4, the table can be queried normally even if the schema of the table and file is different in case
But in spark 3.1, we need to explicitly configure orc.schema.evolution.case.sensitive as false to query normally.(https://github.com/apache/orc/blob/aee81324507de1c5eb8e86bd7aa93a0f6b02bfd9/java/core/src/java/org/apache/orc/impl/SchemaEvolution.java#L514) this cause `fileIncluded ` not to include this column and cant read data

Wether we should set default value as false to maintain compatibility.

guiyanakuang:
@yikf, Thank you for making this pr, should we also modify this test
https://github.com/apache/orc/blob/aee81324507de1c5eb8e86bd7aa93a0f6b02bfd9/java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java#L119-L126

yikf:
> modify

Done

yikf:
> @yikf, Thank you for making this pr, should we also modify this test
> 
> https://github.com/apache/orc/blob/aee81324507de1c5eb8e86bd7aa93a0f6b02bfd9/java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java#L119-L126

Done

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -164,8 +164,8 @@
       "testing.  Setting this too low may negatively affect performance."),
   OVERWRITE_OUTPUT_FILE("orc.overwrite.output.file", "orc.overwrite.output.file", false,
     "A boolean flag to enable overwriting of the output file if it already exists.\n"),
-  IS_SCHEMA_EVOLUTION_CASE_SENSITIVE("orc.schema.evolution.case.sensitive",
-      "orc.schema.evolution.case.sensitive", true,
+  IS_SCHEMA_EVOLUTION_CASE_AWARE("orc.schema.evolution.case.aware",
+      "orc.schema.evolution.case.aware", true,
```
yikf, line 168: I'm not sure if this change will affect `hiveConf`

guiyanakuang: So would it be better to change `isSchemaEvolutionCaseAware` to `isSchemaEvolutionCaseSensitive`?

yikf: Yes, But there are many changes in this `isSchemaEvolutionCaseAware` to `isSchemaEvolutionCaseSensitive` and the default value also needs to be modified to be consistent with currently one.

yikf: BTW, Do we think the default value should be insensitive or aware, At present, it is aware

dongjoon-hyun (changes requested):
Sorry, but I'm -1 for this breaking change, @yikf .

yikf:
> Sorry, but I'm -1 for this breaking change, @yikf .

Shall we modify the semantics of the `orc.schema.evolution.case.sensitive`？The default value is true, This indicates that it is not sensitive, But in fact, it is regarded as sensitive when used

guiyanakuang:
> The default value is true, This indicates that it is not sensitive

@yikf, It looks like you have some misunderstanding of the source code, the default value is true which means it is sensitive.

PR #975, opened 2021-12-23
ORC-1055: [C++] Add the timezone option for the csv-import tool

coderex2522:
The pull request provides the csv-import tool with support for timezone settings

This is a new option to mitigate ORC-1055 situation.

The unit case is TestCSVFileImport.testTimezoneOption in TestCSVFileImport.cc


coderex2522:
@wgtmac Hi, please take a look!

coderex2522:
These timezones alias information is a reference to https://github.com/frohoff/jdk8u-dev-jdk/blob/master/src/share/classes/sun/util/calendar/ZoneInfoFile.java line:220-244

dongjoon-hyun (changes requested):
Thank you for making a PR, @coderex2522 .
- In general, we need a test case which failed before your PR. Do you think you can add a test coverage for that?
- This PR proposes to introduce `alias` which is borrowed from Java `oldMappings`. Does this PR handle both two JVM cases `sun.timezone.ids.oldmapping=true/false`?

wgtmac:
> * ed before your PR. Do

+1 for adding a test case. We can directly use the csv file from the JIRA.

After some investigation, I think the Java odlMappings may introduce new issues. For example, **CST** can either be Central Standard Time (America/Chicago) or China Standard Time (Asia/Shanghai).

dongjoon-hyun:
Thank you for updating, @coderex2522 .

coderex2522:
The PR title and descrition has been modified, @dongjoon-hyun please take a look again!

tools/src/CSVFileImport.cc
```diff
@@ -280,15 +280,19 @@ void usage() {
             << "                  [-s <size>] [--stripe=<size>]\n"
             << "                  [-c <size>] [--block=<size>]\n"
             << "                  [-b <size>] [--batch=<size>]\n"
+            << "                  [-t <string>] [--timezone=<string>]\n"
             << "                  <schema> <input> <output>\n"
             << "Import CSV file into an Orc file using the specified schema.\n"
+            << "The timezone can be viewed in the directory /usr/share/zoneinfo\n"
```
wgtmac, line 286: - It should be stated clearly that the timezone is writer timezone of timestamp types.
- You have specified it as a **required_argument** which conflicts with the comment here.

tools/test/TestCSVFileImport.cc
```diff
@@ -53,3 +53,27 @@ TEST (TestCSVFileImport, test10rows) {
   EXPECT_EQ(expected, output);
   EXPECT_EQ("", error);
 }
+
+TEST (TestCSVFileImport, testTimezoneOption) {
```
dongjoon-hyun, line 57: This test case succeeds without your patch, `CSVFileImport.cc`.

tools/test/TestCSVFileImport.cc
```diff
@@ -53,3 +53,27 @@ TEST (TestCSVFileImport, test10rows) {
...
+  // create an ORC file from importing the CSV file
+  const std::string pgm1 = findProgram("tools/src/csv-import");
+  const std::string csvFile =
+    findExample("TestCSVFileImport.testTimezoneOption.csv");
+  const std::string orcFile = "/tmp/test_csv_import_test_timezone_option.orc";
+  const std::string schema = "'struct<_a:timestamp>'";
+  std::string output;
+  std::string error;
+
+  std::string option = "--timezone=America/Los_Angeles";
```
dongjoon-hyun, line 67: Maybe, this test case always succeeds at `America/Los_Angeles` timezone? If then, can we choose less popular timezone?

tools/test/TestCSVFileImport.cc
coderex2522: I test the case again when I delete the line"options.setTimezoneName(timezoneName);" in CSVFileImport.cc. The case is failed. The error message show
"Expected: expected
      Which is: "{\"_a\": \"2021-12-26 16:00:00.0\"}\n"
To be equal to: output
      Which is: "{\"_a\": \"2021-12-27 00:00:00.0\"}\n"

tools/test/TestCSVFileImport.cc
coderex2522: If we don't set timezone for the WriterOptions, the writer will write the timestamp with the GMT timezone. And the orc-contents tool will scan the content in the GMT timezone.This case should not be successful?

wgtmac: We can add another test to use timezone as Europe/Paris or Asia/Shanghai.

dongjoon-hyun: @coderex2522 . Yes, the test case should fail without your patch and should pass with your patch. That's the purpose of verification of your code's contribution.

And, +1 for @wgtmac .

coderex2522: > We can add another test to use timezone as Europe/Paris or Asia/Shanghai.

Add the timezone 'Europe/Paris' test case done.

coderex2522:
@williamhyun Hi, can you help investigate this problem in continuous-integration/appveyor/pr? 
The error message shows "The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command."

guiyanakuang:
@coderex2522 It looks like ORC-1073 forgot to remove appveyor.yml in labeler.yml. Don't worry, it has nothing to do with your pr
https://github.com/apache/orc/blob/d84df2129adea01469871fc1ba6be5b758cbd7ac/.github/labeler.yml#L22

wgtmac (approved):
LGTM +1

dongjoon-hyun (approved):
+1, LGTM. Thank you, @coderex2522 , @wgtmac , @guiyanakuang .

dongjoon-hyun:
Merged to main/branch-1.7 for Apache ORC 1.7.3.

cc @williamhyun

PR #977, opened 2021-12-24
ORC-1064: Rework Attribute Map in TypeDescription class

belugabehr:

Simplify interactions with Map in TypeDescription class

Reduce code complexity

No functional changes.  Existing unit tests ran.

### DIFF
--- a/java/core/src/java/org/apache/orc/TypeDescription.java
+++ b/java/core/src/java/org/apache/orc/TypeDescription.java
@@ -26,9 +26,9 @@ import org.jetbrains.annotations.NotNull;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
 import java.util.regex.Pattern;
 
 /**
@@ -372,9 +372,7 @@ public class TypeDescription
         result.children.add(clone);
       }
     }
-    for (Map.Entry<String,String> pair: attributes.entrySet()) {
-      result.attributes.put(pair.getKey(), pair.getValue());
-    }
+    result.attributes.putAll(attributes);
     return result;
   }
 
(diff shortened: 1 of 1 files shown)


java/core/src/java/org/apache/orc/TypeDescription.java
```diff
@@ -372,9 +372,7 @@ public TypeDescription clone() {
         result.children.add(clone);
       }
     }
-    for (Map.Entry<String,String> pair: attributes.entrySet()) {
-      result.attributes.put(pair.getKey(), pair.getValue());
-    }
+    result.attributes.putAll(attributes);
```
belugabehr, line 375: No need to iterate like this.

java/core/src/java/org/apache/orc/TypeDescription.java
```diff
@@ -420,15 +418,9 @@ public boolean equals(Object other, boolean checkAttributes) {
     }
     if (checkAttributes) {
       // make sure the attributes are the same
-      List<String> attributeNames = getAttributeNames();
-      if (castOther.getAttributeNames().size() != attributeNames.size()) {
+      if (!this.attributes.equals(castOther.attributes)) {
```
belugabehr, line 421: The native `Map` equals method implements this already.

java/core/src/java/org/apache/orc/TypeDescription.java
```diff
@@ -420,15 +418,9 @@ public boolean equals(Object other, boolean checkAttributes) {
     }
     if (checkAttributes) {
       // make sure the attributes are the same
-      List<String> attributeNames = getAttributeNames();
```
belugabehr, line 423: Calling `getAttributeNames()` method here actually causes an ArrayList instantiation, and a sort of that list, for ever invocation.

java/core/src/java/org/apache/orc/TypeDescription.java
```diff
@@ -565,9 +557,7 @@ public int getScale() {
    * @return a list of sorted attribute names
    */
   public List<String> getAttributeNames() {
-    List<String> result = new ArrayList<>(attributes.keySet());
-    Collections.sort(result);
-    return result;
+    return new ArrayList<>(attributes.keySet());
```
belugabehr, line 560: Keep the `attributes` Map sorted so it doesn't have to be sorted every time this method is invoked.

PR #990, opened 2022-01-03
ORC-450: Add option to avoid materialization of array elements

rizaon:
ORC C++ library doesn't have a type id for the index field of a list
type. We have to select the type id of the whole array if we only want
to get the list indices, which causes unnecessary materialization on the
array elements. The offset stream is stored separately from the content
stream. We can materialize the list indices only.

This patch add the fourth option in ORC C++ library to select column
from ORC file, namely RowReaderOptions::includeTypesAndIntents. It
similar as RowReaderOptions::includeTypes, but with additional set of
ReadIntent for each type id. ListColumnReader can then refer to this
ReadIntent set to either read the list elements, read indices, or both.
ReadIntent_DATA is the default for all type id if the selection does not
specify any ReadIntent.

Adding read intent avoid introducing fake type id only to refer to the
list indices. Thus, expected type ids for an ORC file stay the same
after this patch.

This is needed to selectively avoid materialization of array elements.

Declare ReadIntent_POS in TestColumnReaderEncoded.testList and verify
that the resulting indices are correct.


c++/include/orc/Vector.hh
```diff
@@ -182,12 +182,20 @@ namespace orc {
...
     /**
      * The offset of the first element of each list.
      * The length of list i is offsets[i+1] - offsets[i].
      */
     DataBuffer<int64_t> offsets;
 
+    /**
+     * Position of each element in their respective array.
+     */
+    DataBuffer<int64_t> pos;
```
stiga-huang, line 197: `uint32_t` might be enough and `positions` might be better. BTW, I think clients can calculate this in realtime based on `offsets`. Maybe we don't need this field.

stiga-huang (changes requested):
The solution LGTM. Could you add a test for only reading array indices?

dongjoon-hyun (commented):
Thank you for making a PR, @rizaon .

c++/include/orc/Vector.hh
rizaon: Agree, ListColumnReader can just populate the offsets and leave the pos calculation done by the client. Will remove this field.

c++/include/orc/Common.hh
```diff
@@ -122,6 +122,11 @@ namespace orc {
     StreamKind_BLOOM_FILTER_UTF8 = 8
   };
 
+  enum ReadIntent {
```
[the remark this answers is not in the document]
rizaon: Following the discussion in [Vector.hh](https://github.com/apache/orc/pull/990/files/155cf1a171bcb498731bc95c4a939a445fa23114#r777800353), I am thinking to rewrite this enum as follow

```
  enum ArrayReadIntent {
    ArrayReadIntent_ALL = 0,
    ArrayReadIntent_OFFSETS = 1
  };
```

So the choice is either to read all (data+offsets) or offsets only (we can not read data without reading offsets).

Other than ListColumnReader, it looks like MapColumnReader and UnionColumnReader can benefit from this selective read intent, such as when dealing with "select count(*)" kind of query where no materialization is required. This is out of the scope of ORC-450 though.

c++/src/Options.hh
```diff
@@ -174,19 +175,39 @@ namespace orc {
...
+    privateBits->readIntents.clear();
     return *this;
   }
 
   RowReaderOptions& RowReaderOptions::includeTypes(const std::list<uint64_t>& types) {
     privateBits->selection = ColumnSelection_TYPE_IDS;
     privateBits->includedColumnIndexes.assign(types.begin(), types.end());
+    privateBits->readIntents.clear();
+    privateBits->includedColumnNames.clear();
+    privateBits->readIntents.clear();
```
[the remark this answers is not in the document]
rizaon: Done

c++/include/orc/Reader.hh
```diff
@@ -149,6 +149,28 @@ namespace orc {
...
+     */
+    typedef std::map<uint64_t, std::set<ReadIntent>> TypeReadIntents;
+
+    /**
+     * Selects which type ids to read and specific ReadIntents for each
+     * type id. The root type is always 0 and the rest of the types are
+     * labeled in a preorder traversal of the tree. The parent types are
+     * automatically selected, but the children are not. If a type id is
+     * selected with an empty ReadIntent set, the empty set will be replaced
+     * by {ReadIntent_DATA}.
```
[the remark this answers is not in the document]
rizaon: Now we default to ArrayReadIntent_ALL

c++/include/orc/Reader.hh
```diff
@@ -149,6 +149,28 @@ namespace orc {
...
+     *
+     * This option clears any previous setting of the selected columns or
+     * types.
+     * @param typesAndIntents a map of <typeId, set<ReadIntent>>.
+     * @return this
+     */
+    RowReaderOptions&
+    includeTypesAndIntents(const TypeReadIntents& typesAndIntents);
```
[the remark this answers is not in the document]
rizaon: Done

c++/include/orc/Reader.hh
```diff
@@ -149,6 +149,28 @@ namespace orc {
      */
     RowReaderOptions& includeTypes(const std::list<uint64_t>& types);
 
+    /**
+     * A map of <typeId, set<ReadIntent>> used as a parameter in
+     * RowReaderOptions::includeTypesAndIntents.
+     */
+    typedef std::map<uint64_t, std::set<ReadIntent>> TypeReadIntents;
```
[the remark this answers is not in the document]
rizaon: Done

c++/src/Vector.cc
```diff
@@ -242,7 +242,6 @@ namespace orc {
     }
     return false;
   }
-
```
dongjoon-hyun, line 245: Could you revert this irrelevant change, @rizaon ?

c++/include/orc/Common.hh
```diff
@@ -122,6 +122,11 @@ namespace orc {
     StreamKind_BLOOM_FILTER_UTF8 = 8
   };
 
+  enum ArrayReadIntent {
```
dongjoon-hyun, line 125: Could you add some description for this new enum?

c++/include/orc/Reader.hh
```diff
@@ -149,6 +149,26 @@ namespace orc {
      */
     RowReaderOptions& includeTypes(const std::list<uint64_t>& types);
 
+    /**
+     * A map of <typeId, ArrayReadIntent> used as a parameter in
+     * RowReaderOptions::includeTypesWithIntents.
```
dongjoon-hyun, line 154: Actually, this PR also uses it at `getReadIntents` too. We had better describe the definition a little more instead of listing the references.

c++/include/orc/Reader.hh
```diff
@@ -149,6 +149,26 @@ namespace orc {
...
+
+    /**
+     * Selects which type ids to read and specific ArrayReadIntents for each
+     * type id. The root type is always 0 and the rest of the types are
+     * labeled in a preorder traversal of the tree. The parent types are
+     * automatically selected, but the children are not.
+     *
+     * This option clears any previous setting of the selected columns or
+     * types.
+     * @param typesWithIntents a map of <typeId, ArrayReadIntent>.
```
dongjoon-hyun, line 166: Please use `TypeReadIntents` instead of `<typeId, ArrayReadIntent>`.

c++/include/orc/Reader.hh
```diff
@@ -266,6 +286,12 @@ namespace orc {
      * Get desired timezone to return data of timestamp type
      */
     const std::string& getTimezoneName() const;
+
+    /**
+     * Get the <typeId, ArrayReadIntent> map that was supplied through
```
dongjoon-hyun, line 291: Please use `TypeReadIntents` instead of `<typeId, ArrayReadIntent>`.

c++/src/ColumnReader.cc
```diff
@@ -993,6 +993,9 @@ namespace orc {
     void nextInternal(ColumnVectorBatch& rowBatch,
                       uint64_t numValues,
                       char *notNull);
+
+    void readArrayIndices(const int64_t *arrayOffsets, uint64_t arrayCount,
```
dongjoon-hyun, line 997: If you don't mind, please put one argument per line.
```cpp
void readArrayIndices(const int64_t *arrayOffsets,
                      uint64_t arrayCount,
```

c++/src/ColumnReader.cc
```diff
@@ -1006,9 +1009,12 @@ namespace orc {
...
     rle = createRleDecoder(std::move(stream), false, vers, memoryPool);
-    const Type& childType = *type.getSubtype(0);
-    if (selectedColumns[static_cast<uint64_t>(childType.getColumnId())]) {
-      child = buildReader(childType, stripe);
+    ArrayReadIntent intent = stripe.getReadIntent(columnId);
+    if (intent == ArrayReadIntent_ALL) {
+      const Type &childType = *type.getSubtype(0);
+      if (selectedColumns[static_cast<uint64_t>(childType.getColumnId())]) {
+        child = buildReader(childType, stripe);
+      }
```
dongjoon-hyun, line 1017: Do we have a test coverage for `else` case? Could you point me the test case for `ArrayReadIntent_OFFSETS`?

c++/src/Reader.cc
```diff
@@ -290,6 +291,15 @@ namespace orc {
     return selectedColumns;
   }
 
+  ArrayReadIntent RowReaderImpl::getReadIntent(uint64_t typeId) const {
+    auto elem = readIntents.find(typeId);
+    if (elem == readIntents.end()) {
+      return ArrayReadIntent_ALL;
+    } else {
+      return elem->second;
```
dongjoon-hyun, line 299: Do we have a test coverage for this code path?

dongjoon-hyun (commented):
I added a few comments and questions.

c++/include/orc/Common.hh
rizaon: Yes, we can extend this enum to support MAP and UNION type. However, I'd like to focus on LIST first in this PR, if it is OK. I'd be happy to extend it in the future JIRA.

Renaming this enum back to 'ReadIntent' sounds better if we anticipate extending it for MAP and UNION.

c++/src/ColumnReader.cc
rizaon: It is tested at [TestColumnReaderEncoded.testListOffsetsOnly](https://github.com/apache/orc/pull/990/files#diff-e0eca596eb78b0688bdb1d04e6a23d5b7c101ff686e99050dbe15147d275ec47R1474), which is modified from TestColumnReaderEncoded.testList.

The assertion is
```
ASSERT_EQ(0, longs->numElements);
```

c++/src/Reader.cc
rizaon: The TestColumnReaderEncoded.testListOffsetsOnly that I add only mock up to StripeStreamsImpl::getReadIntent(), and I believe it didn't touch this part.

I will try to add more comprehensive tests. Do you know any existing unit tests I can refer to?

c++/include/orc/Reader.hh
```diff
@@ -149,6 +149,26 @@ namespace orc {
...
+     * labeled in a preorder traversal of the tree. The parent types are
+     * automatically selected, but the children are not.
+     *
+     * This option clears any previous setting of the selected columns or
+     * types.
+     * @param typesWithIntents a map of <typeId, ArrayReadIntent>.
+     * @return this
+     */
+    RowReaderOptions&
+    includeTypesWithIntents(const TypeReadIntents& typesWithIntents);
```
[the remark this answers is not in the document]
rizaon: Suppose a customer table as described at https://impala.apache.org/docs/build/html/topics/impala_complex_types.html#complex_sample_schema

And a query
```
SELECT
  c.c_name,
  o.o_orderkey,
  l.pos
FROM customer c
  INNER JOIN c.c_orders o
  INNER JOIN o.o_lineitems l
LIMIT 5
```
We need to materialize `c.c_orders` (read all) to read `o.o_orderkey` but can omit materialization of `o.o_lineitems` elements (read offsets only) since we only need `o.o_lineitems.pos`.

c++/include/orc/Reader.hh
```diff
@@ -565,6 +591,15 @@ namespace orc {
...
     virtual void seekToRow(uint64_t rowNumber) = 0;
 
+    /**
+     * Get an ArrayReadIntent for a given typeId.
+     * @param typeId the type id to look up.
+     * @return an ArrayReadIntent that was specified for given typeId through
+     *    RowReaderOptions::includeTypesWithIntents. If no ArrayReadIntent was
+     *    specified for typeId, return ArrayReadIntent_ALL.
+     */
+    virtual ArrayReadIntent getReadIntent(uint64_t typeId) const = 0;
```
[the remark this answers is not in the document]
rizaon: My plan is to return `ArrayReadIntent_ALL` if typeId is not selected or selected without any specific read intent (in case `RowReaderOptions::includeTypes()` is being used). I understand the naming is a bit confusing. Will think more about this.

c++/src/Options.hh
```diff
@@ -174,20 +175,36 @@ namespace orc {
...
+    return *this;
+  }
+
+  RowReaderOptions&
+  RowReaderOptions::includeTypesWithIntents(const TypeReadIntents& typesAndIntents) {
+    privateBits->selection = ColumnSelection_TYPE_IDS;
+    privateBits->includedColumnIndexes.clear();
+    privateBits->readIntents.clear();
+    for (const auto& typeIntentPair : typesAndIntents ) {
+      privateBits->readIntents[typeIntentPair.first] = typeIntentPair.second;
```
rizaon, line 204: I think I missed a case here.
Suppose we select two nested list columns, both with `ArrayReadIntent_OFFSETS`. We need to override the parent list intent to `ArrayReadIntent_ALL` in `ColumnSelector::selectParents()`. Otherwise, the children list will not be read.

rizaon:
Thank you for all your feedback. I will try to address them in my next commit.

c++/src/Options.hh
wgtmac: OK, that makes sense.

c++/include/orc/Reader.hh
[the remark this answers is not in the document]
wgtmac: OK, that makes sense.

rizaon:
I came to a realization that in order to avoid List materialization, it is sufficient to unselect its children Type in ColumnSelector. ListColumnReader will skip creating childReader and won't read the elements altogether.

The latest patch changes the focus into resolving the selectedColumns for a given a readIntents map.

c++/include/orc/Common.hh
rizaon: Renamed back to `ReadIntent`, gave description comments, and TODO for MAP and UNION type.

c++/include/orc/Reader.hh
rizaon: Deleted in latest commit.

c++/include/orc/Reader.hh
rizaon: Deleted in latest commit.

c++/include/orc/Reader.hh
rizaon: Deleted in latest commit.

c++/include/orc/Reader.hh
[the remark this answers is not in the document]
rizaon: Deleted in latest commit.

c++/src/ColumnReader.cc
rizaon: Deleted in latest commit.

c++/src/Reader.cc
rizaon: Created 4 new test:

- TestReadIntent.testListAll
- TestReadIntent.testListOffsets
- TestReadIntent.testListAllAndOffsets
- TestReadIntent.testListConflictingIntent

c++/src/ColumnReader.cc
rizaon: I replace this with 4 new tests at `c++/test/TestReader.cc`.
However, they stop at column selection verification and do not read a row batch.

c++/src/Options.hh
rizaon: It is now resolved by relying on `ColumnSelector::selectParents()` to override the decision.

c++/src/Vector.cc
rizaon: Reverted.

c++/include/orc/Common.hh
```diff
@@ -122,6 +122,19 @@ namespace orc {
     StreamKind_BLOOM_FILTER_UTF8 = 8
   };
 
+  /**
+   * Specific read intention when selecting a certain TypeId.
+   * This enum currently only being utilized by LIST type selection.
+   *
+   * TODO: extend this to support MAP and UNION type.
```
dongjoon-hyun, line 129: Apache ORC community recommends to use an IDed TODO with JIRA ID if needed.
```
- * TODO: extend this to support MAP and UNION type.
+ * TODO(ORC-XXX): extend this to support MAP and UNION type.
```

For this code, I'd like to recommend to remove this TODO and file an ORC JIRA instead.

c++/include/orc/Reader.hh
```diff
@@ -149,6 +149,25 @@ namespace orc {
...
     RowReaderOptions& includeTypes(const std::list<uint64_t>& types);
 
+    /**
+     * A map type of <typeId, ReadIntent>.
+     */
+    typedef std::map<uint64_t, ReadIntent> TypeReadIntents;
+
+    /**
+     * Selects which type ids to read and specific ReadIntents for each
+     * type id. The root type is always 0 and the rest of the types are
```
dongjoon-hyun, line 159: `root type is always 0` might sound weird. Technically, `root type ID is always 0`.
BTW, I'd like to recommend to remove this sentence, `The root type is always 0 and the rest of the types are labeled in a preorder traversal of the tree` because this is redundant.

c++/include/orc/Reader.hh
```diff
@@ -149,6 +149,25 @@ namespace orc {
...
+     * labeled in a preorder traversal of the tree. The parent types are
```
dongjoon-hyun, line 160: `parent` -> `ancestor`? `parent` means only the direct nodes technically.

c++/include/orc/Reader.hh
```diff
@@ -149,6 +149,25 @@ namespace orc {
      */
     RowReaderOptions& includeTypes(const std::list<uint64_t>& types);
 
+    /**
+     * A map type of <typeId, ReadIntent>.
+     */
+    typedef std::map<uint64_t, ReadIntent> TypeReadIntents;
```
dongjoon-hyun, line 155: Maybe, `TypeReadIntents` -> `TypeReadIntentMap` sounds better?

c++/src/Reader.hh
```diff
@@ -88,10 +88,12 @@ namespace orc {
     // Select a field by id
     void updateSelectedByFieldId(std::vector<bool>& selectedColumns, uint64_t fieldId);
     // Select a type by id
-    void updateSelectedByTypeId(std::vector<bool>& selectedColumns, uint64_t typeId);
+    void updateSelectedByTypeId(std::vector<bool>& selectedColumns, uint64_t typeId,
+                                const RowReaderOptions::TypeReadIntents* readIntents = nullptr);
```
dongjoon-hyun, line 92: Shall we use reference instead of pointer?

c++/src/Reader.hh
```diff
@@ -88,10 +88,12 @@ namespace orc {
...
 
     // Select all of the recursive children of the given type.
-    void selectChildren(std::vector<bool>& selectedColumns, const Type& type);
+    void selectChildren(std::vector<bool>& selectedColumns, const Type& type,
+                        const RowReaderOptions::TypeReadIntents* readIntents = nullptr);
```
dongjoon-hyun, line 96: ditto. Shall we use a reference?

c++/test/TestReader.cc
```diff
@@ -137,4 +139,89 @@ namespace orc {
...
+   *     >
+   *   >
+   *
+   * @param readIntents TypeReadIntents describing the section.
+   * @param expectedSelection expected TypeIds that will be selected from given
+   * readIntents.
+   */
+  void verifySelection(const RowReaderOptions::TypeReadIntents &readIntents,
+                       const std::vector<uint32_t> &expectedSelection) {
+    std::string fileName = "complextypes_iceberg.orc";
```
dongjoon-hyun, line 176: Shall we remove `_iceberg` part of this test case file?

c++/include/orc/Common.hh
rizaon: Removed the TODO. I will file a follow-up JIRA after this PR merged.

c++/include/orc/Reader.hh
rizaon: Done.

c++/include/orc/Reader.hh
rizaon: Removed the sentece.

c++/include/orc/Reader.hh
rizaon: Done.

c++/src/Reader.hh
rizaon: Done.

c++/src/Reader.hh
rizaon: Done.

dongjoon-hyun:
Thank you for updates, @rizaon .

c++/src/Reader.hh
```diff
@@ -88,10 +88,12 @@ namespace orc {
...
     void updateSelectedByFieldId(std::vector<bool>& selectedColumns, uint64_t fieldId);
     // Select a type by id
-    void updateSelectedByTypeId(std::vector<bool>& selectedColumns, uint64_t typeId);
+    void updateSelectedByTypeId(std::vector<bool>& selectedColumns, uint64_t typeId,
+                                const RowReaderOptions::IdReadIntentMap& idReadIntentMap);
 
     // Select all of the recursive children of the given type.
-    void selectChildren(std::vector<bool>& selectedColumns, const Type& type);
+    void selectChildren(std::vector<bool>& selectedColumns, const Type& type,
+                        const RowReaderOptions::IdReadIntentMap& idReadIntentMap);
```
dongjoon-hyun, line 96: Actually, this doesn't match with the function description.
> // Select all of the recursive children of the given type.

c++/src/Reader.cc
```diff
@@ -41,6 +41,8 @@ namespace orc {
     "1.6.0", "1.6.1", "1.6.2", "1.6.3", "1.6.4", "1.6.5", "1.6.6", "1.6.7", "1.6.8",
     "1.6.9", "1.6.10", "1.6.11", "1.7.0"};
 
+  static const RowReaderOptions::IdReadIntentMap EMPTY_ID_READ_INTENT_MAP;
```
dongjoon-hyun, line 44: Actually, when I test this PR on my Mac. `Clang 13` complains with `declaration requires an exit-time destructor` and this is a regression although this is a warning. Could you try `Clang 13` and remove the warning?
```
$ make package
[  5%] Built target snappy_ep
[ 11%] Built target zlib_ep
[ 17%] Built target zstd_ep
[ 23%] Built target lz4_ep
[ 29%] Built target googletest_ep
[ 35%] Built target protobuf_ep
Consolidate compiler generated dependencies of target orc
[ 35%] Building CXX object c++/src/CMakeFiles/orc.dir/Reader.cc.o
/Users/dongjoon/APACHE/orc-merge/c++/src/Reader.cc:44:50: error: declaration requires an exit-time destructor [-Werror,-Wexit-time-destructors]
  static const RowReaderOptions::IdReadIntentMap EMPTY_ID_READ_INTENT_MAP;
                                                 ^
/Users/dongjoon/APACHE/orc-merge/c++/src/Reader.cc:44:50: error: declaration requires a global destructor [-Werror,-Wglobal-constructors]
2 errors generated.
make[2]: *** [c++/src/CMakeFiles/orc.dir/Reader.cc.o] Error 1
make: *** [all] Error 2
```

dongjoon-hyun (commented):
Could you keep the existing methods and add yours like the following? Although there exists some duplication, but it would be safer.
```c++
     // Select a type by id
     void updateSelectedByTypeId(std::vector<bool>& selectedColumns, uint64_t typeId);
+    // Select a type by id and readIntent
+    void updateSelectedByTypeId(std::vector<bool>& selectedColumns, uint64_t typeId,
+                                const RowReaderOptions::IdReadIntentMap& idReadIntentMap);

     // Select all of the recursive children of the given type.
     void selectChildren(std::vector<bool>& selectedColumns, const Type& type);
+    // Select ... 
+    void selectChildren(std::vector<bool>& selectedColumns, const Type& type,
+                        const RowReaderOptions::IdReadIntentMap& idReadIntentMap);
```

c++/src/Reader.cc
rizaon: I replaced it with function `emptyIdReadIntentMap()`

c++/src/Reader.hh
rizaon: Clarified the comment as follow
```
    // Select a type id of the given type.
    // This function may also select all of the recursive children of the given type
    // depending on the read intent of that type in idReadIntentMap.
    void selectChildren(std::vector<bool>& selectedColumns, const Type& type,
                        const RowReaderOptions::IdReadIntentMap& idReadIntentMap);
```

dongjoon-hyun (commented):
Thank you for updates!

c++/test/TestReader.cc
```diff
@@ -137,4 +139,89 @@ namespace orc {
...
+   *       >,
+   *       g:map<string,struct<
+   *         h:struct<
+   *           i:array<double>
+   *         >
+   *       >>
+   *     >
+   *   >
+   *
+   * @param readIntents TypeReadIntents describing the section.
```
stiga-huang: typo? "section" => "selection" ?

c++/test/TestReader.cc
```diff
@@ -137,4 +139,89 @@ namespace orc {
...
+  }
+
+  TEST(TestReadIntent, testListAll) {
+    // select all of int_array_array.
+    verifySelection({{4, ReadIntent_ALL}}, {0, 4, 5, 6});
+  }
+
+  TEST(TestReadIntent, testListOffsets) {
+    // select only the offsets of int_array_array.
+    verifySelection({{4, ReadIntent_OFFSETS}}, {0, 4});
```
stiga-huang, line 207: Can we also verify that the output batches only contains offsets (i.e. don't have elements)? Or add another simple test for verifying the batches?

c++/test/TestReader.cc
```diff
@@ -137,4 +139,89 @@ namespace orc {
...
+
+    // select only the offsets of int_array_array.item.
+    verifySelection({{4, ReadIntent_OFFSETS}, {5, ReadIntent_OFFSETS}},
+                    {0, 4, 5});
```
stiga-huang, line 211: Can we add a test for a column of array<array<array<int>>>, and selecting the most outer and most inner offsets together?

c++/test/TestReader.cc
rizaon: I believe this has been fixed at commit [11d66a2](https://github.com/apache/orc/pull/990/commits/11d66a2035f1d91346e1e44986e7ffd4e20e5b0f)

c++/test/TestReader.cc
rizaon: OK. Will add such test in my next commit.

c++/test/TestReader.cc
rizaon: complextypes_iceberg.orc is the most complex ORC file in the example dir that I can find. I can't find any that has 3 nested arrays.
Would you like me to create a new ORC file? Is there any tool that I can use?

c++/test/TestReader.cc
rizaon: Added in TestReadIntent.testRowBatchContent

c++/test/TestReader.cc
rizaon: Added as the 3rd test in TestReadIntent.testListOffsets

rizaon:
[6b16e86](https://github.com/apache/orc/pull/990/commits/6b16e867a63460a0137798978df9cb18dad20523) add `examples/TestOrcFile.nestedList.orc` and use it for all TestReadIntent.*

dongjoon-hyun (approved):
I changed the following directly, @rizaon .
- Rename `emptyIdReadIntentMap` to `EMPTY_IDREADINTENTMAP`.
- Fix indentation

+1, LGTM from my side.

Could you review this once more, @wgtmac , @williamhyun , and @stiga-huang ?

stiga-huang (approved):
LGTM. Thanks for adding more tests!

dongjoon-hyun:
Thank you so much, @rizaon , @wgtmac , @stiga-huang .

rizaon:
Thank you @dongjoon-hyun @wgtmac and @stiga-huang for shepherding this PR!

PR #992, opened 2022-01-06
ORC-1075: Support reading ORC files with no column statistics

guiyanakuang:

This PR aims to fix evaluation predicate proto. When the row index statistics only implements ColumnStatistics and does not provide information other than min or max, the predicate cannot be evaluated correctly

To make all files compatible with the orc specification readable by the official library

Added unit test. 
`testWithoutStatistics` for tests that do not provide statistics
`testMissMinOrMaxInStatistics` is used to test the case where min max statistics are not provided


dongjoon-hyun (commented):

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -434,18 +434,33 @@ public long getNext() {
...
       this.lower = getBaseObjectForComparison(type, lower);
       this.upper = getBaseObjectForComparison(type, upper);
       this.hasNulls = hasNulls;
       this.onlyLowerBound = onlyLowerBound;
       this.onlyUpperBound = onlyUpperBound;
+      this.hasValue = hasValue;
+      this.comparable = comparable;
+    }
+
+    ValueRange(PredicateLeaf predicate,
```
dongjoon-hyun, line 457: Thank you for adding this.

dongjoon-hyun:
BTW, why the new test file is `unofficial` ORC?
- `java/core/src/test/resources/unofficial.orc`

java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
```diff
@@ -2449,4 +2449,41 @@ public void testSargApplier() throws Exception {
     f2.setAccessible(true);
     assertFalse((boolean)f2.get(applier1));
   }
+
+  @Test
+  public void testCompatibleSpecificationUnofficialFile() throws IOException {
+    // unofficial.orc is a file compatible with the ORC specification,
```
dongjoon-hyun, line 2455: What do you mean by `compatible`? Is it because ORC spec doesn't enforce `min` and `max`?

guiyanakuang: > What do you mean by `compatible`? Is it because ORC spec doesn't enforce `min` and `max`?

Yes, the column statistics interface only needs to provide numberOfValues / hasNull / bytesOnDisk

guiyanakuang:
> BTW, why the new test file is `unofficial` ORC? Which writer is used?
> 
> * `java/core/src/test/resources/unofficial.orc`

The orc file provided in ORC-1075 was created via CUDF, and I looked at the binary and wrote a similar file

dongjoon-hyun:
Could you revise the PR title more specific for your code? Actually, the following title can be used any PRs about predicates.
```
The correct evaluation of the predicate proto for ORC files written in other ways that conform to the specification
```

dongjoon-hyun:
cc @omalley , @wgtmac , @pgaref , @williamhyun and @stiga-huang , too.

guiyanakuang:
@dongjoon-hyun How about the current title. : )

wbo4958:
@guiyanakuang, I just tested your Patch on none-1.orc, looks like it doesn't work. Could you help to double-check it?

guiyanakuang:
> @guiyanakuang, I just tested your Patch on none-1.orc, looks like it doesn't work. Could you help to double-check it?

@wbo4958 In the file nones-1.orc, index.getNumberOfValues() returns 0 instead of the actual 3 rows of data. So it doesn't work, we need to make sure the statistics are consistent with the actual.

It is important to note that none-1.orc contains statistics.
You can write an orc file without the rowIndex statistics for comparison by doing the following
```java
OrcFile.createWriter(new Path(".orc"),
                                         OrcFile.writerOptions(conf)
                                          .rowIndexStride(0)  // Set rowIndexStride to 0 to avoid writing rowIndex
                                          .setSchema(schema));
```

java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
```diff
@@ -2449,4 +2449,70 @@ public void testSargApplier() throws Exception {
...
+        assertEquals(xRow, x.vector[xRow]);
+      } else {
+        assertTrue(x.isNull[xRow]);
+      }
+    }
+    rowIterator.close();
+  }
+
+  @Test
+  public void testCDUFUnofficialFile() throws IOException {
```
wbo4958, line 2491: Could we change the name to testRowIndexWithoutStatistics ?

guiyanakuang: Looks better, I'll change later

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -1134,7 +1183,10 @@ public void setRowIndexCols(boolean[] rowIndexCols) {
             if (entry == null) {
               throw new AssertionError("RG is not populated for " + columnIx + " rg " + rowGroup);
             }
-            OrcProto.ColumnStatistics stats = entry.getStatistics();
+            OrcProto.ColumnStatistics stats = EMPTY_COLUMN_STATISTICS;
+            if (entry.hasStatistics()) {
+              stats = entry.getStatistics();
```
guiyanakuang, line 1188: Here is the key to fix.

When no column statistics are provided, entry.getStatistics() does not return null, but a default object that we must first determine by entry.hasStatistics() before getting it.

JsonFileDump has the same misuse as FileDump, I'll fix them later

wbo4958:
The title seems to make the PR fix the orc file written by cudf-specific. But it's a generic issue.

Could we change the title to  "Fix the issue of failing to read orc file without statistics in RowIndex when filter is pushed down"

wgtmac:
> ry and wrote a similar file

Does cuDF implement an unofficial ORC writer? If that is true, we may work with them to assign a unique writer id to cuDF.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -462,8 +482,31 @@ public long getNext() {
...
+      return hasValue;
+    }
+
+    /**
+     * This method is used separately when the hasValue is true
+     * Whether min or max is provided for comparison
+     * @return is it comparable
+     */
+    boolean comparable() {
+      return comparable;
```
wgtmac, line 499: According to the comment above, should we check hasValue just in case any misuse.

guiyanakuang:
> Does cuDF implement an unofficial ORC writer? If that is true, we may work with them to assign a unique writer id to cuDF.

Thanks for the reminder, I think this can be done in a new pr

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -706,13 +749,19 @@ static TruthValue evaluatePredicateRange(PredicateLeaf predicate,
                                            ValueRange range,
                                            BloomFilter bloomFilter,
                                            boolean useUTCTimestamp) {
+    if (!range.valid()) {
+      return TruthValue.YES_NO_NULL;
```
wgtmac, line 753: I have checked that the C++ implementation has already handled this case.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
guiyanakuang: How about this, return the expected value at all times
```java
    /**
     * Whether min or max is provided for comparison
     * @return is it comparable
     */
    boolean comparable() {
      return hasValue && comparable;
    }
```

guiyanakuang:
> The title seems to make the PR fix the orc file written by cudf-specific. But it's a generic issue.
> 
> Could we change the title to "Fix the issue of failing to read orc file without statistics in RowIndex when filter is pushed down"

@wbo4958 
In fact, this pr fixes two close cases
1. without statistics
2. missing min/max info in statistics

How about this title
Fix the issue of failing to read orc file **without statistics or missing min/max info in statistics** in RowIndex when filter is pushed down

guiyanakuang:
I found that I can test with static method to avoid uploading specific files, I will update the test cases later

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -462,8 +483,30 @@ public long getNext() {
...
     boolean hasValues() {
-      return lower != null;
+      return hasValue;
+    }
+
+    /**
+     * Whether min or max is provided for comparison
+     * @return is it comparable
+     */
+    boolean comparable() {
```
dongjoon-hyun, line 498: Maybe, `isComparable`?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -462,8 +483,30 @@ public long getNext() {
...
+      return hasValue && comparable;
+    }
+
+    /**
+     * value range is invalid if the column statistics are non-existent
+     * @see ColumnStatisticsImpl#isStatsExists()
+     * this method is similar to isStatsExists
+     * @return value range is valid or not
+     */
+    boolean valid() {
```
dongjoon-hyun, line 508: `valid` -> `isValid`?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -575,7 +618,7 @@ static ValueRange getValueRange(ColumnStatistics index,
       Boolean max = stats.getTrueCount() != 0;
       return new ValueRange<>(predicate, min, max, stats.hasNull());
     } else {
-      return new ValueRange(predicate, null, null, true);
+      return new ValueRange(predicate, index.hasNull(), true, false);
```
dongjoon-hyun, line 621: This change looks a little misleading. Could you use the full constructor?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -462,8 +483,30 @@ public long getNext() {
       this(predicate, null, null, hasNulls, false, false);
     }
 
+    ValueRange(PredicateLeaf predicate, boolean hasNulls, boolean hasValue, boolean comparable) {
```
dongjoon-hyun, line 486: Shall we remove this addition? I believe we can use the other constructor instead of adding this short cut.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
guiyanakuang: Looks better

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
guiyanakuang: This is a great suggestion, already using the full constructor

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
guiyanakuang: You're right. The previous place already uses the full constructor, so this addition is redundant and I have removed it

java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
pgaref: lets add the above info to the comment -- ideally with a link to the interface

pgaref:
> BTW, why the new test file is `unofficial` ORC? Which writer is used?
> 
> * `java/core/src/test/resources/unofficial.orc`

Had the exact same question. Ideally we should be able to write the actual file using a custom unofficial writer to repro the case.

guiyanakuang:
> Had the exact same question. Ideally we should be able to write the actual file using a custom unofficial writer to repro the case.

Initially I used the user-supplied file in jira for testing, but later I removed it and tested with the exposed static method.

It seems that CUDF requires the GPU to generate the ORC file, so it may not be practical for our test cases to depend on cudf to generate the file and then test it

dongjoon-hyun:
Usually, NVidia frameworks like RAPIDS support both CPU and GUI mode. Are you sure GUDF doesn't work on CPU only system. , @guiyanakuang ?

dongjoon-hyun (commented):
Why do you say we need to use CUDF framework here? According to ORC-1082, you mentioned that Apache ORC writer doesn't generate this issue, right, @guiyanakuang ? In principle, if Apache Spark writers doesn't have an issue, we should mark ORC-1075 as an improvement instead of a blocker of 1.7/1.6. Could you confirm the current status?
> It seems that CUDF requires the GPU to generate the ORC file, so it may not be practical for our test cases to depend on cudf to generate the file and then test it

guiyanakuang:
Yes, I confirm that.

dongjoon-hyun:
Thank you for the confirmation, @guiyanakuang .

dongjoon-hyun:
Hi, All.
I updated this PR's milestone and ORC-1075's issue type and priority.

dongjoon-hyun (approved):
+1, LGTM. Thank you, @guiyanakuang and @wgtmac .
Merged to main.

PR #996, opened 2022-01-06
ORC-1078: Row group end offset doesn't accommodate all the blocks

hsnusonic:
To calculate the row group end offset based on the buffer size. Current implementation assumes stretching the slop by a constant factor of 2 to accommodate the compression block that is not always the case.

When compression size is smaller than 2048 and two blocks are not enough for all the values left in last row group, we might end up seeking outside of range.

Unit test added TestRecordReaderImpl#testSmallCompressionSizeOrc


dongjoon-hyun (commented):
Thank you for making a PR, @hsnusonic .

java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
```diff
@@ -1538,7 +1538,7 @@ public void testPartialPlanCompressed() throws Exception {
         new InStream.StreamOptions()
             .withCodec(OrcCodecPool.getCodec(CompressionKind.ZLIB))
             .withBufferSize(1024);
-    final int SLOP = 2 * (OutStream.HEADER_SIZE + options.getBufferSize());
+    final int SLOP = 5 * (OutStream.HEADER_SIZE + options.getBufferSize());
```
dongjoon-hyun, line 1541: In this case, `stretchFactor` is `5`?

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -182,10 +182,13 @@ public static long estimateRgEndOffset(boolean isCompressed,
...
+    // by a factor to safely accommodate the next compression block.
+    // 512 is the MAX_SCOPE defined in RunLengthIntegerWriterV2.
+    // 8 is the maximum size of bytes for each value (see RunLengthIntegerWriterV2.zzBits100p).
+    // We need to calculate the maximum number of blocks by bufferSize accordingly.
+    int stretchFactor = bufferSize > 0 ? 2 + (512 * 8 - 1) / bufferSize : 2;
     long slop = isCompressed
-                    ? 2 * (OutStream.HEADER_SIZE + bufferSize)
+                    ? stretchFactor * (OutStream.HEADER_SIZE + bufferSize)
                     : WORST_UNCOMPRESSED_SLOP;
     return isLast ? streamLength : Math.min(streamLength, nextGroupOffset + slop);
```
dongjoon-hyun, line 193: According to the change on the old test case, is there a chance of performance regression by causing more data traffics when we read ORC files on S3?
```scala
- final int SLOP = 2 * (OutStream.HEADER_SIZE + options.getBufferSize());
+ final int SLOP = 5 * (OutStream.HEADER_SIZE + options.getBufferSize());
...
assertEquals(START + 1000, result.get(1).getOffset());
assertEquals(20000 + SLOP, result.get(1).getLength());
assertEquals(START + 41000, result.get(2).getOffset());
assertEquals(10000 + SLOP, result.get(2).getLength());
```

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -182,10 +182,13 @@ public static long estimateRgEndOffset(boolean isCompressed,
                                          long streamLength) {
     // figure out the worst case last location
     // if adjacent groups have the same compressed block offset then stretch the slop
-    // by factor of 2 to safely accommodate the next compression block.
-    // One for the current compression block and another for the next compression block.
+    // by a factor to safely accommodate the next compression block.
+    // 512 is the MAX_SCOPE defined in RunLengthIntegerWriterV2.
+    // 8 is the maximum size of bytes for each value (see RunLengthIntegerWriterV2.zzBits100p).
+    // We need to calculate the maximum number of blocks by bufferSize accordingly.
+    int stretchFactor = bufferSize > 0 ? 2 + (512 * 8 - 1) / bufferSize : 2;
```
dongjoon-hyun, line 189: It seems that we truncate values here. I'm wondering if we need a round-up operation for `(512 * 8 - 1) / bufferSize`? Or, is it safe to truncate?

java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
```diff
@@ -2449,4 +2449,23 @@ public void testSargApplier() throws Exception {
     f2.setAccessible(true);
     assertFalse((boolean)f2.get(applier1));
   }
+
+  @Test
+  public void testRgEndOffset() throws IOException {
+    Configuration conf = new Configuration();
+    Path filePath = new Path(getClass().getClassLoader().
+        getSystemResource("orc-small-compression-size.orc").getPath());
```
dongjoon-hyun, line 2457: Do you think we can create this file in this test case, @hsnusonic ?

java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
```diff
@@ -2449,4 +2449,23 @@ public void testSargApplier() throws Exception {
...
+    Reader reader = OrcFile.createReader(filePath, OrcFile.readerOptions(conf));
+
+    SearchArgument sarg = SearchArgumentFactory.newBuilder()
+        .startNot()
+        .lessThan("x", PredicateLeaf.Type.LONG, 120000L)
+        .end().build();
+    VectorizedRowBatch batch = reader.getSchema().createRowBatch();
+    try (RecordReader rows = reader.rows(reader.options().searchArgument(sarg, null))) {
+      while (rows.nextBatch(batch)) {
+        // do nothing but check all the data can be read
```
dongjoon-hyun, line 2467: Could you check the actual values to ensure they are read correctly?

pgaref: +1 -- it would be much easier if we generate the data as well making sure there are no inconsistencies

java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
pgaref: I would use the same formula as above for calculating stretchFactor for consistency

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
pgaref: Not sure bufferSize can be zero/negative -- can you explain a bit more the logic here?

pgaref: Also an example with compression blocks would like help

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -182,10 +182,13 @@ public static long estimateRgEndOffset(boolean isCompressed,
                                          long streamLength) {
     // figure out the worst case last location
     // if adjacent groups have the same compressed block offset then stretch the slop
-    // by factor of 2 to safely accommodate the next compression block.
-    // One for the current compression block and another for the next compression block.
+    // by a factor to safely accommodate the next compression block.
+    // 512 is the MAX_SCOPE defined in RunLengthIntegerWriterV2.
```
pgaref, line 186: Btw, we should probably expose and use MAX_SCOPE and MAX_SIZE directly from RunLengthIntegerWriterV2 instead of relying on plain values

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
hsnusonic: It should not cause performance issues here, since files should be read in stripe level (not row goup level) and we are fixing the boundary of each row group. @pgaref Please correct me if I misunderstood anything.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
hsnusonic: We are actually doing a round-up operation for `(512 * 8) / bufferSize`. We can interpret the formula as `1 (for uncompressed bytes in buffer) + Math.ceil(512 * 8 / (float) bufferSize)`. The second part can be expanded to `1 + (512 * 8 - 1) / bufferSize`. I just thought that we can avoid floating point calculation here.

java/core/src/test/org/apache/orc/impl/TestRecordReaderImpl.java
hsnusonic: Good point. I can try to generate files for the test.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
hsnusonic: @pgaref `bufferSize == 0` if `isCompressed == false`. In other words, `bufferSize` doesn't matter in uncompressed stream.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -180,12 +180,18 @@ public static long estimateRgEndOffset(boolean isCompressed,
...
+      throw new IllegalArgumentException("BufferSize must be > 0 but was " + bufferSize);
+    }
     // figure out the worst case last location
     // if adjacent groups have the same compressed block offset then stretch the slop
-    // by factor of 2 to safely accommodate the next compression block.
-    // One for the current compression block and another for the next compression block.
+    // by a factor to safely accommodate the next compression block.
+    // We need to calculate the maximum number of blocks by bufferSize accordingly.
+    final int stretchFactor = isCompressed
+        ? 2 + (MAX_VALUES_LENGTH * MAX_BIT_WIDTH / 8 - 1) / bufferSize
```
pgaref, line 191: I am not sure I understand the formula here -- first of all we divide by 8 is the conversion from bits to Bytes?

I thought this is a round up operation but for different buffer sizes like 512, 1024 this is evaluates to: ``2 + 0``

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
pgaref: Thats correct @hsnusonic  -- we start by reading each stripe footer and RowIndexes and check if we have rowgroups that match our filters. If there is no match we proceed to the next stripe. In that sense the change in the SLOP above wont make any difference as the data should be already in memory.

More details  https://github.com/apache/orc/blob/main/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java#L1264

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
hsnusonic: Thanks for the suggestion! That's better.

Yes, divided by 8 is the conversion from bits to bytes. Let me use number to explain.
```
2 + (MAX_VALUES_LENGTH * MAX_BIT_WIDTH / 8 - 1) / bufferSize
= 1 + 1 + (512 * 64 / 8 - 1) / bufferSize
= 1 + 1 + (4096 - 1) / bufferSize
// the first 1 is for the bytes in uncompressed buffer since these bytes will fit into 1 block
// the remain part should be viewed together as round up operation of 4096 / bufferSize
// if bufferSize = 4096 => stretchFactor = 1 + 1 + 0 
// if bufferSize = 2048 => stretchFactor = 1 + 1 + 1
// if bufferSize = 2047 => stretchFactor = 1 + 1 + 2
```
Please let me know if you have any suggestions to make the formula easier to read.

dongjoon-hyun:
I added a milestone label, `1.6.13`.

dongjoon-hyun:
Could you resolve the conflicts, @hsnusonic ?

dongjoon-hyun:
To @pgaref , the due date is next Monday (2022-01-17).

hsnusonic:
> Could you resolve the conflicts, @hsnusonic ?

Done. @dongjoon-hyun Could you please approve the test run?

dongjoon-hyun:
It seems to be running now, @hsnusonic . Maybe, one one did approve already.

dongjoon-hyun:
Gentle ping, @pgaref .
As [we discussed on dev@orc mailing list](https://lists.apache.org/thread/rxo1yzp8l4nr8covjx221k3jc084p5np), we can have this at `1.6.14` if this is not merged until 17th January.

pgaref (approved):
Latest changes LGTM -- thanks for the updated PR @hsnusonic !

dongjoon-hyun:
Thank you, @hsnusonic and @pgaref !

dongjoon-hyun:
FYI, `branch-1.6` seems to be broken on MacOS 10.5 after this commit, but it looks irrelevant to this PR.
- https://github.com/apache/orc/commits/branch-1.6

PR #998, opened 2022-01-08
ORC-1081: Fix heap-use-after-free in SearchArgumentBuilderImpl::end()

stiga-huang:

This PR fixes the heap-use-after-free issue in `SearchArgumentBuilderImpl::end()`. `std::deque::pop_front` will delete the element. We should call it after using the element.

This crashes the test when building ORC with AddressSanitizer. The ASAN report is attached in the JIRA.

After this fix, I'm able to run all existing tests with AddressSanitizer enabled.


dongjoon-hyun (commented):
Thank you, @stiga-huang .

dongjoon-hyun (approved):
+1, LGTM because `pop_front()` destroys the object.
Merged to main/1.7.

PR #1003, opened 2022-01-09
ORC-1082: Improve `FileDump` and `JsonFileDump` to be robust on missing column statistics

guiyanakuang:

This pr is aims to fixing code in the FileDump and JsonFileDump commands that determines whether a column statistic exists or not.

Before we can get the ColumnStatistics from RowIndex we need to determine whether it exists or not.

`entry.getStatistics()` does not return null at any time, and will return a default object when it is not set.

The current impl is wrong
```java
      OrcProto.ColumnStatistics colStats = entry.getStatistics();
      if (colStats == null) {
        buf.append("no stats at ");
      }
```

This can be tested with the following code

```java
  @Test
  public void test() {
    OrcProto.RowIndexEntry entry = OrcProto.RowIndexEntry.newBuilder().addPositions(0).build();
    assertFalse(entry.hasStatistics());
    assertNotNull(entry.getStatistics());
  }
```
But since this is just testing Protobuf generated code, I'm not sure if I should add it to the unit tests.

On the other hand, our proto is defined as follows, statistics is optional

https://github.com/apache/orc/blob/7d2838b4b38111bc1bfa00592d84a8b36286b9f2/proto/orc_proto.proto#L102-L105

Our orc writer implementation, however, will definitely write statistics, and there is no optional configuration, so it cannot be verified by the files written by orc

https://github.com/apache/orc/blob/7d2838b4b38111bc1bfa00592d84a8b36286b9f2/java/core/src/java/org/apache/orc/impl/writer/TreeWriterBase.java#L322-L330

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/FileDump.java
+++ b/java/tools/src/java/org/apache/orc/tools/FileDump.java
@@ -747,10 +747,10 @@ public final class FileDump {
         buf.append("unknown\n");
         continue;
       }
-      OrcProto.ColumnStatistics colStats = entry.getStatistics();
-      if (colStats == null) {
+      if (!entry.hasStatistics()) {
         buf.append("no stats at ");
       } else {
+        OrcProto.ColumnStatistics colStats = entry.getStatistics();
         ColumnStatistics cs =
             ColumnStatisticsImpl.deserialize(colSchema, colStats,
                 reader.writerUsedProlepticGregorian(),
--- a/java/tools/src/java/org/apache/orc/tools/JsonFileDump.java
+++ b/java/tools/src/java/org/apache/orc/tools/JsonFileDump.java
@@ -444,7 +444,7 @@ public class JsonFileDump {
       writer.beginObject();
       writer.name("entryId").value(entryIx);
       OrcProto.RowIndexEntry entry = index.getEntry(entryIx);
-      if (entry == null) {
+      if (entry == null || !entry.hasStatistics()) {
         continue;
       }
       OrcProto.ColumnStatistics colStats = entry.getStatistics();


dongjoon-hyun (commented):
Thanks. Could you elaborate how to verify your PR in the PR description, @guiyanakuang ?

guiyanakuang:
> Thanks. Could you elaborate how to verify your PR in the PR description, @guiyanakuang ?

@dongjoon-hyun This can be tested with the following code

```java
  @Test
  public void test() {
    OrcProto.RowIndexEntry entry = OrcProto.RowIndexEntry.newBuilder().addPositions(0).build();
    assertFalse(entry.hasStatistics());
    assertNotNull(entry.getStatistics());
  }
```
But since this is just testing Protobuf generated code, I'm not sure if I should add it to the unit tests.

On the other hand, our proto is defined as follows, statistics is optional

https://github.com/apache/orc/blob/7d2838b4b38111bc1bfa00592d84a8b36286b9f2/proto/orc_proto.proto#L102-L105

Our orc writer implementation, however, will definitely write statistics, and there is no optional configuration, so it cannot be verified by the files written by orc

https://github.com/apache/orc/blob/7d2838b4b38111bc1bfa00592d84a8b36286b9f2/java/core/src/java/org/apache/orc/impl/writer/TreeWriterBase.java#L322-L330

dongjoon-hyun (changes requested):
@guiyanakuang . I asked you to add it to the PR description for the other reviewers and to make it commit logs. Comments are not helpful for that purpose. As a committer,  you should ask the same thing to the contributor always. The commit log is the only communication rule we use among developers. :)

guiyanakuang:
You are right. Accept criticism :grimacing:

dongjoon-hyun (commented):
Thank you for updating the PR description.

dongjoon-hyun (approved):
+1, LGTM.

PR #1016, opened 2022-01-16
ORC-1095: Deprecate `UnknownFormatException`

dongjoon-hyun:

This PR aims to deprecate `UnknownFormatException` at Apache ORC 1.7.3.

This extends `IOException` to expose more data and used once here. However, the downstream projects are not using this information because `checkOrcVersion` returns `IOException`.  The error message is enough.

https://github.com/apache/orc/blob/80b98126cf8a15d9d44c67a9aabfabfdcc7f913e/java/core/src/java/org/apache/orc/impl/ReaderImpl.java#L517-L525

This PR replaces it with `IOException` with the same message generated by `UnknownFormatException` previously.

https://github.com/apache/orc/blob/80b98126cf8a15d9d44c67a9aabfabfdcc7f913e/java/core/src/java/org/apache/orc/UnknownFormatException.java#L33-L35

Pass the CIs.

### DIFF
--- a/java/core/src/java/org/apache/orc/UnknownFormatException.java
+++ b/java/core/src/java/org/apache/orc/UnknownFormatException.java
@@ -23,6 +23,10 @@ import org.apache.hadoop.fs.Path;
 
 import java.io.IOException;
 
+/**
+ * @deprecated This will be removed in the future releases.
+ */
+@Deprecated
 public class UnknownFormatException extends IOException {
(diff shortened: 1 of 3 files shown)


williamhyun (approved):
+1 LGTM

dongjoon-hyun:
Thank you, @williamhyun !
Merged to main/1.7.

PR #1020, opened 2022-01-19
ORC-1098: [C++] Support specifying type ids or column names in cpp tools

stiga-huang:

This is a follow-up task of #921. Currently we have options for the tools to work on specified top-level column fields. However, ACID ORC files usually have nested structure. We need the type ids to specify nested columns. As an extension, adding support for column names will also be helpful. So we don't need to manually convert column names to type ids. Also reports the valid values when an invalid column name is given.

This PR extracts the option parsing codes into ToolsHelper. So similiar cpp tools can share the same option set.

It makes the tools more useful in practice.

Added unit tests for the new options.

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -216,7 +216,15 @@ namespace orc {
     if (ite != nameIdMap.end()) {
       updateSelectedByTypeId(selectedColumns, ite->second);
     } else {
-      throw ParseError("Invalid column selected " + fieldName);
+      bool first = true;
+      std::ostringstream ss;
+      ss << "Invalid column selected " << fieldName << ". Valid names are ";
+      for (auto it = nameIdMap.begin(); it != nameIdMap.end(); ++it) {
+        if (!first) ss << ", ";
+        ss << it->first;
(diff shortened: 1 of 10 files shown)


dongjoon-hyun (commented):
Thank you so much for the PR, @stiga-huang .

Could you review this please, @wgtmac ?

And, if possible, could you participate the on-going Apache ORC 1.6.13 vote, @stiga-huang and @wgtmac ?

stiga-huang:
Thank @dongjoon-hyun! Just voted.

dongjoon-hyun:
Thank you so much for your participation, @stiga-huang .

dongjoon-hyun (approved):
+1, LGTM. Thank you, @stiga-huang .

dongjoon-hyun:
@stiga-huang . After reviewing this PR once more and testing on `branch-1.7`, I decided to re-target this to v1.7.3 instead of v1.8.0 because this is a safe `tools` only PR and the change on `Reader.cc` is also about `ParseError` message instead of a `core` code change. Thank you again.

PR #1024, opened 2022-01-27
ORC-1099: [C++] Extend ReadIntent to support MAP and UNION type

rizaon:
This patch is a follow-up of ORC-450. Similar to LIST, we can also avoid
the materialization of MAP and UNION column elements by specifying
ReadIntent_OFFSETS in selection.

This patch also adds a requirement that subtypes of UNION should be all
selected or not at all. UnionColumnReader will crash if only part of the
subtypes were selected. This patch modifies ColumnSelector to silently
select all the UNION subtypes if they were partially selected.

This patch helps avoid materialization of MAP and UNION elements,
such as in the case of "select count(*)" query.

- Added several unit tests under TestReadIntent.*


dongjoon-hyun (commented):
Thank you, @rizaon .

c++/test/TestReader.cc
```diff
@@ -250,4 +252,206 @@ namespace orc {
...
+   * Read TestOrcFile.nestedUnion.orc and verify the resolved selections.
+   *
+   * The ORC fie has the following schema:
+   * struct<
+   *   id:int,
+   *   single_union:uniontype<int,string>,
+   *   nested_union:uniontype<uniontype<int,uniontype<int,string>>,int>
+   * >
+   */
+  void verifyUnionSelection(const RowReaderOptions::IdReadIntentMap& idReadIntentMap,
```
stiga-huang, line 361: The same as `verifyMapSelection`, can we merge this function with `verifySelection`?

c++/test/TestReader.cc
```diff
@@ -250,4 +252,206 @@ namespace orc {
...
+    rowReaderOpts.includeTypesWithIntents(idReadIntentMap);
+    std::unique_ptr<RowReader> rowReader = reader->createRowReader(rowReaderOpts);
+
+    // Read a row batch.
+    std::unique_ptr<ColumnVectorBatch> batch = rowReader->createRowBatch(1024);
+    EXPECT_TRUE(rowReader->next(*batch));
+    EXPECT_EQ(1, batch->numElements);
+    auto& structBatch = dynamic_cast<StructVectorBatch&>(*batch);
+
+    // verify content of int_array selection.
```
stiga-huang, line 326: `int_array` should be `single_map`

c++/test/TestReader.cc
```diff
@@ -250,4 +252,206 @@ namespace orc {
...
+    rowReaderOpts.includeTypesWithIntents(idReadIntentMap);
+    std::unique_ptr<RowReader> rowReader = reader->createRowReader(rowReaderOpts);
+
+    // Read a row batch.
+    std::unique_ptr<ColumnVectorBatch> batch = rowReader->createRowBatch(1024);
+    EXPECT_TRUE(rowReader->next(*batch));
+    EXPECT_EQ(4, batch->numElements);
+    auto& structBatch = dynamic_cast<StructVectorBatch&>(*batch);
+
+    // verify content of int_array selection.
```
stiga-huang, line 424: `int_array` should be `single_union`

c++/test/TestReader.cc
```diff
@@ -250,4 +252,206 @@ namespace orc {
...
+    EXPECT_EQ(1, stringBatch.numElements);
+    // verify content of the first row.
+    EXPECT_EQ(0, unionBatch.tags.data()[0]);
+    EXPECT_EQ(0, unionBatch.offsets.data()[0]);
+    EXPECT_EQ(0, longBatch.data.data()[0]);
+    // verify content of the second row.
+    EXPECT_EQ(1, unionBatch.tags.data()[1]);
+    EXPECT_EQ(0, unionBatch.offsets.data()[1]);
+    EXPECT_EQ(2, stringBatch.length.data()[0]);
+    EXPECT_EQ(0, strncmp("s1", stringBatch.data.data()[0], 2));
```
stiga-huang, line 440: I feel like it's more readable if we create the test ORC files in flight. E.g. https://github.com/apache/orc/blob/0622002bd140d7bd88af0ae71d73fd1b3a101c17/c%2B%2B/test/TestPredicatePushdown.cc#L29

Otherwise, each developer should manually dump the file content to understand these. So I think we should try creating the ORC files in tests unless we can't (e.g. for corrupt files). Creating the ORC files inflight also reduce the repo size. But I'm not strongly opposing the current approach.
@dongjoon-hyun Could you share your opinion?

dongjoon-hyun: +1 for @stiga-huang 's comment about creating the test file in the test cases.

Thank you for your review, @stiga-huang .

c++/src/Reader.cc
```diff
@@ -120,10 +120,24 @@ namespace orc {
...
+        numSubtypeSelected++;
+      }
     }
     selectedColumns[id] = result;
+
+    if (type.getKind() == TypeKind::UNION && selectedColumns[id]) {
+      if (0 < numSubtypeSelected && numSubtypeSelected < type.getSubtypeCount()) {
+        // Subtypes of UNION should be fully selected or not selected at all.
+        // Override partial subtype selections with full selections.
+        for (uint64_t c = 0; c < type.getSubtypeCount(); ++c) {
```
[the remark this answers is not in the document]
rizaon: selectChildren will first check if type is selected or not. If type is already selected, then selectChildren will stop checking type's children.
Hence, we iterate the children here and select them directly.

c++/test/TestReader.cc
```diff
@@ -250,4 +252,206 @@ namespace orc {
     EXPECT_EQ(3, intArrayArrayArrayBatch.offsets.data()[1]);
     EXPECT_EQ(nullptr, intArrayArrayArrayBatch.elements.get());
   }
+
+  /**
+   * Read TestOrcFile.nestedMap.orc and verify the resolved selections.
+   *
+   * The ORC fie has the following schema:
```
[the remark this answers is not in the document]
rizaon: Removed this method in https://github.com/apache/orc/pull/1024/commits/9d4565d9ec7907ad20dcaab3f105bab265e164a0

c++/test/TestReader.cc
```diff
@@ -250,4 +252,206 @@ namespace orc {
...
+   * struct<
+   *   id:int,
+   *   single_map:map<string,string>,
+   *   nested_map:map<string,map<string,map<string,string>>>
+   * >
+   */
+  void verifyMapSelection(const RowReaderOptions::IdReadIntentMap& idReadIntentMap,
```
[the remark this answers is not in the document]
rizaon: Done

c++/test/TestReader.cc
rizaon: Done

c++/test/TestReader.cc
rizaon: Done

c++/test/TestReader.cc
rizaon: Done

c++/test/TestReader.cc
rizaon: Done

rizaon:
9d4565d change TestReadIntent.* to create the ORC test file inflight. It also remove examples/TestOrcFile.nestedList.orc and replaces it with a simpler test case in `createNestedListMemReader`.

I'm not sure why it failed tests orc-test in Windows build. Is there any way I can see the content of `D:/a/orc/orc/build/Testing/Temporary/LastTest.log` ?

dongjoon-hyun:
I re-triggered the tests to see `Windows` result again.

c++/src/Reader.cc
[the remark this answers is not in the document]
stiga-huang: Ah, right. Thanks for the explanation!

rizaon:
799db28 is a rebase and squash of the previous commits.

wgtmac (approved):
LGTM +1. Thanks for adding these test cases @rizaon !

dongjoon-hyun:
Thank you, @wgtmac . I re-triggered the CIs.

dongjoon-hyun (commented):
@rizaon . This patch seems to break Window build. Could you take a look as a last step?

rizaon:
The Windows build failed the test with the following error
```
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
```
This is due to the MemoryOutputStream being initialized in the createReader method and destructed before we start reading rows. This is fixed in [4745805](https://github.com/apache/orc/pull/1024/commits/474580534889b9a47fc7d71352ee9f01780f2192) by moving the MemoryOutputStream initialization to the Test body.

A ctest argument `--output-on-failure` is also added in Windows workflow for better observability.

@dongjoon-hyun @wgtmac this passes all tests now.

dongjoon-hyun:
Thank you!

dongjoon-hyun:
Merged to main for Apache ORC 1.8.0.
Thank you, @rizaon , @wgtmac , @stiga-huang .

PR #1025, opened 2022-01-27
ORC-236: Support `UNION` type in Java Convert tool

rizaon:
This patch add support to convert json to UNION type in orc file. For
example, for schema `struct<foo:uniontype<int,string>>`, the following
json lines can be parsed into UNION type.

```
{"foo": {"tag": 0, "value": 1}}
{"foo": {"tag": 1, "value": "testing"}}
{"foo": {"tag": 0, "value": 3}}
```

This add a missing support for UNION type in java convert tool.

Manually test against handcrafted json file.

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/convert/JsonReader.java
+++ b/java/tools/src/java/org/apache/orc/tools/convert/JsonReader.java
@@ -63,6 +63,8 @@ public class JsonReader implements RecordReader {
   private final FSDataInputStream input;
   private long rowNumber = 0;
   private final DateTimeFormatter dateTimeFormatter;
+  private final String unionTag;
+  private final String unionValue;
 
   interface JsonConverter {
     void convert(JsonElement value, ColumnVector vect, int row);
@@ -313,9 +315,9 @@ public class JsonReader implements RecordReader {
       } else {
         UnionColumnVector vector = (UnionColumnVector) vect;
         JsonObject obj = value.getAsJsonObject();
-        int tag = obj.get("tag").getAsInt();
+        int tag = obj.get(unionTag).getAsInt();
         vector.tags[row] = tag;
-        childConverter[tag].convert(obj.get("value"), vector.fields[tag], row);
+        childConverter[tag].convert(obj.get(unionValue), vector.fields[tag], row);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


dongjoon-hyun (commented):
Thank you, @rizaon .

dongjoon-hyun:
cc @guiyanakuang and @williamhyun , too

dongjoon-hyun:
Is this applicable to branch-1.7, @rizaon ?

rizaon:
@dongjoon-hyun If backport is clean, it should be applicable.

dongjoon-hyun:
Thank you, @rizaon . I set the milestone, `v1.7.3`.

guiyanakuang:
It might be nice to make Convert more generic, some existing json data may not have tag and value fields, adding an option for the user to customize the method of getting tag and value would be great. : )

rizaon:
Hello @guiyanakuang, thanks for your feedback!
Does it help if I add parameters to specify the tag and value json-key? Maybe, `--union_tag` and `--union_key` like this?
```
java -jar ./java/tools/target/orc-tools-1.8.0-SNAPSHOT-uber.jar convert -o sample.orc \
  -s "struct<foo:uniontype<int,string>>" --union_tag="tag" --union_value="value" sample.json
```

guiyanakuang:
@rizaon. I think it's good that this can handle simple situations. If there is complex json data we can keep upgrading

dongjoon-hyun:
I switched the milestone to 1.8.0 since 1.7.3 will be released in two weeks and this is a new improvement which is not a blocker for https://github.com/apache/orc/milestone/4?closed=1 .

java/tools/src/java/org/apache/orc/tools/convert/JsonReader.java
```diff
@@ -333,15 +364,20 @@ public JsonReader(Reader reader,
...
   }
 
   public JsonReader(Iterator<JsonElement> parser,
                     FSDataInputStream underlying,
                     long size,
                     TypeDescription schema,
-                    String timestampFormat) throws IOException {
+                    String timestampFormat,
+                    String unionTag,
+                    String unionValue) throws IOException {
```
guiyanakuang, line 380: @rizaon. Should the original constructor be retained so that the code that already exists can be seamlessly compatible looks better. I suggest adding a new constructor to handle the case of specifying unionTag/unionValue

rizaon: Done

dongjoon-hyun (approved):
+1, LGTM. Thank you, @rizaon and @guiyanakuang .
Also, cc @williamhyun because he is the release manager of Apache ORC 1.7.4.

Merged to main/1.7.

dongjoon-hyun:
@rizaon . I added you to the Apache ORC contributor group and assigned ORC-236 to you.
Welcome to the Apache ORC community!

rizaon:
Thank you, @dongjoon-hyun @guiyanakuang !

PR #1034, opened 2022-01-31
ORC-1108: Use `RawLocalFileSystem` to skip checksum files during benchmark data generation

williamhyun:
This PR aims to use `RawLocalFileSystem` to skip checksum files during benchmark data generation. 

There is no need to check for checksums. 

Manually.


dongjoon-hyun (approved):
+1, LGTM. Ya, this looks reasonable because we don't benchmark the data generation time and we don't use the checksum.

dongjoon-hyun:
BTW, I backported this to branch-1.7 too.

PR #1044, opened 2022-02-12
ORC-1116: [C++] Fix csv-import tool when exporting long bytes

KyleGrains:
Let each string type columns use its own databuffer before writing data into orc file, so that when one column is calling buffer.resize(), the invalidated buffer.data() is not affecting other columns which points to it.

When using the csv-import tool to export a orc file with schema like "struct<a:string,b:binary>", if the data in column "b" has very long bytes (over 4MB), the process could segmentation fault or the exported data in column "a" could become empty string.

Following the code in CSVFileImport.cc, when writing a orc file, all string type columns is using one databuffer inside function fillStringValues(). If a data length is larger than the buffer, the buffer will be resized. The resize() operation will cause all previous result of buffer.data() become invalid. 

In this case, when field "a" finished writing data into buffer, field "b" begin writing will resize the buffer, invalidate previous buffer.data(), so field "a"'s stringBatch is not refering to a valid buffer.data() any more.

A workaround could use different databuffers for each string type column, however requires allocating 4MB memory each. Alternatively using one same databuffer, but let all previous stringBatch re-points to databuffer's new address, after a resize() operation happens.

Prepare a csv file with two columns, while the second column has data larger than 4MB, like:
_str1,longbinaryabcd.........._
Run following command which finished successfully:
_$ csv-import "struct<a:string,b:binary>" Testbinary.csv /tmp/test.orc_
Run the following command shows data correctly:
_$ orc-contents /tmp/test.orc --columns=0_
_{"a": "str1"}_

### DIFF
--- a/tools/src/CSVFileImport.cc
+++ b/tools/src/CSVFileImport.cc
@@ -23,6 +23,7 @@
 #include <algorithm>
 #include <fstream>
 #include <iostream>
+#include <list>
 #include <memory>
 #include <getopt.h>
 #include <string>
@@ -77,8 +78,8 @@ void fillStringValues(const std::vector<std::string>& data,
                       orc::ColumnVectorBatch* batch,
                       uint64_t numValues,
                       uint64_t colIndex,
-                      orc::DataBuffer<char>& buffer,
-                      uint64_t& offset) {
+                      orc::DataBuffer<char>& buffer) {
+  uint64_t offset = 0;
   orc::StringVectorBatch* stringBatch =
     dynamic_cast<orc::StringVectorBatch*>(batch);
   bool hasNull = false;
@@ -365,7 +366,8 @@ int main(int argc, char* argv[]) {
   double totalElapsedTime = 0.0;
   clock_t totalCPUTime = 0;
 
-  orc::DataBuffer<char> buffer(*orc::getDefaultPool(), 4 * 1024 * 1024);
+  typedef std::list<orc::DataBuffer<char>> DataBufferList;
+  DataBufferList bufferList;
 
   orc::WriterOptions options;
   options.setStripeSize(stripeSize);
@@ -385,7 +387,6 @@ int main(int argc, char* argv[]) {
   std::ifstream finput(input.c_str());
   while (!eof) {
     uint64_t numValues = 0;      // num of lines read in a batch
-    uint64_t bufferOffset = 0;   // current offset in the string buffer
 
     data.clear();
     memset(rowBatch->notNull.data(), 1, batchSize);
@@ -420,13 +421,13 @@ int main(int argc, char* argv[]) {
           case orc::STRING:
           case orc::CHAR:
           case orc::VARCHAR:
-          case orc::BINARY:
+          case orc::BINARY: 
+            bufferList.emplace_back(*orc::getDefaultPool(), 4 * 1024 * 1024);
             fillStringValues(data,
                              structBatch->fields[i],
                              numValues,
                              i,
-                             buffer,
-                             bufferOffset);
+                             bufferList.back());
             break;
           case orc::FLOAT:
           case orc::DOUBLE:
--- a/tools/src/CSVFileImport.cc
+++ b/tools/src/CSVFileImport.cc
@@ -422,7 +422,7 @@ int main(int argc, char* argv[]) {
           case orc::CHAR:
           case orc::VARCHAR:
           case orc::BINARY: 
-            bufferList.emplace_back(*orc::getDefaultPool(), 4 * 1024 * 1024);
+            bufferList.emplace_back(*orc::getDefaultPool(), 1 * 1024 * 1024);
             fillStringValues(data,
                              structBatch->fields[i],
                              numValues,
--- a/tools/test/TestCSVFileImport.cc
+++ b/tools/test/TestCSVFileImport.cc
@@ -24,6 +24,8 @@
 #include "wrap/gmock.h"
 #include "wrap/gtest-wrapper.h"
 
+#include <fstream>
+
 TEST (TestCSVFileImport, test10rows) {
   // create an ORC file from importing the CSV file
   const std::string pgm1 = findProgram("tools/src/csv-import");
@@ -89,4 +91,37 @@ TEST (TestCSVFileImport, testTimezoneOption) {
     EXPECT_EQ(expected, output);
     EXPECT_EQ("", error);
   }
+}
+
+TEST (TestCSVFileImport, testLongString) {
+  // create an ORC file from importing the CSV file
+  const std::string pgm1 = findProgram("tools/src/csv-import");
+  const std::string csvFile = "/tmp/test_csv_import_test_long_string.csv";
+  const std::string orcFile = "/tmp/test_csv_import_test_long_string.orc";
+  const std::string schema = "'struct<_a:string,b_:binary,_c:varchar(10)>'";
+  std::string output;
+  std::string error;
+
+  std::ofstream csvFileStream(csvFile, std::ios::binary | std::ios::out | std::ios::trunc);
+  if(csvFileStream.is_open())
+  {
+    std::string longStr;
+    longStr.resize(4 * 1024 * 1024 + 1, 'x');
+    csvFileStream << "str1," << longStr << ",var1\n";
+    csvFileStream << "str2," << longStr << ",var2\n";
+    csvFileStream.close();
+  }
+
+  EXPECT_EQ(0, runProgram({pgm1, schema, csvFile, orcFile}, output, error));
+  EXPECT_EQ("", error);
+
+  // verify the ORC file content
+  const std::string pgm2 = findProgram("tools/src/orc-contents");
+  std::string option = "--columns=0,2";
+  const std::string expected =
+    "{\"_a\": \"str1\", \"_c\": \"var1\"}\n"
+    "{\"_a\": \"str2\", \"_c\": \"var2\"}\n";
+  EXPECT_EQ(0, runProgram({pgm2, option, orcFile}, output, error));
+  EXPECT_EQ(expected, output);
+  EXPECT_EQ("", error);
 }


dongjoon-hyun (commented):
Thank you for making a PR, @KyleGrains .

dongjoon-hyun:
cc @wgtmac , @guiyanakuang , @stiga-huang too

dongjoon-hyun (commented):
Could you add a test coverage at `TestCSVFileImport.cc`?

tools/src/CSVFileImport.cc
```diff
@@ -420,13 +421,13 @@ int main(int argc, char* argv[]) {
           case orc::STRING:
           case orc::CHAR:
           case orc::VARCHAR:
-          case orc::BINARY:
+          case orc::BINARY: 
+            bufferList.emplace_back(*orc::getDefaultPool(), 4 * 1024 * 1024);
```
stiga-huang, line 425: Maybe we can reduce the buffer size to 1 * 1024 * 1024 now since it's used per column.
We use 1MB buffer size in column writers:
https://github.com/apache/orc/blob/6f4481565761400587c7a266ca4c0a4244c0aa34/c%2B%2B/src/ColumnWriter.cc#L49-L62

stiga-huang (commented):
Thanks for fixing this! The solution LGTM. Just have a minor comment.

tools/src/CSVFileImport.cc
KyleGrains: Yes, that's reasonable, I had been considering changing the memory allocaton too.

KyleGrains:
> PR

Yes, however it may take some time, as I am new to github..

KyleGrains:
> Could you add a test coverage at `TestCSVFileImport.cc`, @KyleGrains?
Yes, however it may take some time, as I am new to github..

tools/test/TestCSVFileImport.cc
```diff
@@ -90,3 +90,27 @@ TEST (TestCSVFileImport, testTimezoneOption) {
     EXPECT_EQ("", error);
   }
 }
+
+TEST (TestCSVFileImport, testLongBinary) {
+  // create an ORC file from importing the CSV file
+  const std::string pgm1 = findProgram("tools/src/csv-import");
+  const std::string csvFile = findExample("TestCSVFileImport.testLongBinary.csv");
```
stiga-huang, line 97: I think this test does not reveal the bug. We need a test that would fail without the fix. So I think we need a csv file with strings larger than 4MB. It'd be better to create the csv file in the test instead of importing it staticly in this project.

KyleGrains: Yes, the csv file is not covering this case. Sorry I am not familiar with gtest enough, is there any recommended way doing this, like generating data on the fly?

KyleGrains: I added the the csv file creation in TestCSVFileImport.cc to create temp file in the test

tools/test/TestCSVFileImport.cc
```diff
@@ -90,3 +92,36 @@ TEST (TestCSVFileImport, testTimezoneOption) {
     EXPECT_EQ("", error);
   }
 }
+
+TEST (TestCSVFileImport, testLongString) {
```
dongjoon-hyun, line 96: Thank you for adding this.

tools/test/TestCSVFileImport.cc
```diff
@@ -90,3 +92,36 @@ TEST (TestCSVFileImport, testTimezoneOption) {
...
+
+  // verify the ORC file content
+  const std::string pgm2 = findProgram("tools/src/orc-contents");
+  std::string option = "--columns=0,2";
+  const std::string expected =
+    "{\"_a\": \"str1\", \"_c\": \"var1\"}\n"
+    "{\"_a\": \"str2\", \"_c\": \"var2\"}\n";
+  EXPECT_EQ(0, runProgram({pgm2, option, orcFile}, output, error));
+  EXPECT_EQ(expected, output);
+  EXPECT_EQ("", error);
```
dongjoon-hyun, line 126: Hmm. The test case seems to pass without your patch still.
```
$ make package test-out
...
Test project /Users/dongjoon/APACHE/orc-merge/build
    Start 1: orc-test
1/2 Test #1: orc-test .........................   Passed    2.91 sec
    Start 2: tool-test
2/2 Test #2: tool-test ........................   Passed   12.29 sec

100% tests passed, 0 tests failed out of 2

Total Test time (real) =  15.21 sec
Built target test-out

$ git diff main --stat
 tools/test/TestCSVFileImport.cc | 35 +++++++++++++++++++++++++++++++++++
 1 file changed, 35 insertions(+)
```

The valid test case should fail on main branch without your patch. Could you take a look at this once more and make it sure?

KyleGrains: I am not sure if this is my code program or some other issues, the output like:
$make package test-out
...
Test project /home/kai/github/orc/build
    Start 1: orc-test
1/2 Test #1: orc-test .........................   Passed    7.05 sec
    Start 2: tool-test
2/2 Test #2: tool-test ........................***Failed   17.12 sec
ORC version: 1.8.0-SNAPSHOT
example dir = /home/kai/github/orc/examples
build dir = /home/kai/github/orc/build
[==========] Running 85 tests from 10 test cases.
[----------] Global test environment set-up.
[----------] 3 tests from TestCSVFileImport
[ RUN      ] TestCSVFileImport.test10rows
[       OK ] TestCSVFileImport.test10rows (15 ms)
[ RUN      ] TestCSVFileImport.testTimezoneOption
[       OK ] TestCSVFileImport.testTimezoneOption (29 ms)
[ RUN      ] TestCSVFileImport.testLongString
/home/kai/github/orc/tools/test/TestCSVFileImport.cc:124: Failure
      Expected: 0
To be equal to: runProgram({pgm2, option, orcFile}, output, error)
      Which is: 1
/home/kai/github/orc/tools/test/TestCSVFileImport.cc:125: Failure
      Expected: expected
      Which is: "{\"_a\": \"str1\", \"_c\": \"var1\"}\n{\"_a\": \"str2\", \"_c\": \"var2\"}\n"
To be equal to: output
      Which is: ""
With diff:
@@ -1,2 +1,1 @@
-{\"_a\": \"str1\", \"_c\": \"var1\"}
-{\"_a\": \"str2\", \"_c\": \"var2\"}\n
+""

/home/kai/github/orc/tools/test/TestCSVFileImport.cc:126: Failure
      Expected: ""
To be equal to: error
      Which is: "Caught exception in /tmp/test_csv_import_test_long_string.orc: File size too small\n"
[  FAILED  ] TestCSVFileImport.testLongString (58 ms)

$git diff main --stat
 tools/test/TestCSVFileImport.cc | 35 +++++++++++++++++++++++++++++++++++
 1 file changed, 35 insertions(+)

$md5sum tools/test/TestCSVFileImport.cc
73b578e945183da5c3fdd8d16afda5c6  tools/test/TestCSVFileImport.cc

$cat /etc/redhat-release
CentOS Linux release 7.9.2009 (Core)

stiga-huang: Without the fix, the test failed in my env too
```
$ tools/test/tool-test --gtest_filter='*testLongString*'
ORC version: 1.8.0-SNAPSHOT
example dir = --gtest_filter=*testLongString*
build dir = .
Note: Google Test filter = *testLongString*
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from TestCSVFileImport
[ RUN      ] TestCSVFileImport.testLongString
/home/quanlong/workspace/orc/tools/test/TestCSVFileImport.cc:113: Failure
      Expected: 0
To be equal to: runProgram({pgm1, schema, csvFile, orcFile}, output, error)
      Which is: 139
/home/quanlong/workspace/orc/tools/test/TestCSVFileImport.cc:114: Failure
      Expected: ""
To be equal to: error
      Which is: "Segmentation fault (core dumped)\n"
/home/quanlong/workspace/orc/tools/test/TestCSVFileImport.cc:122: Failure
      Expected: 0
To be equal to: runProgram({pgm2, option, orcFile}, output, error)
      Which is: 1
/home/quanlong/workspace/orc/tools/test/TestCSVFileImport.cc:123: Failure
      Expected: expected
      Which is: "{\"_a\": \"str1\", \"_c\": \"var1\"}\n{\"_a\": \"str2\", \"_c\": \"var2\"}\n"
To be equal to: output
      Which is: ""
With diff:
@@ -1,2 +1,1 @@
-{\"_a\": \"str1\", \"_c\": \"var1\"}
-{\"_a\": \"str2\", \"_c\": \"var2\"}\n
+""

/home/quanlong/workspace/orc/tools/test/TestCSVFileImport.cc:124: Failure
      Expected: ""
To be equal to: error
      Which is: "Caught exception in /tmp/test_csv_import_test_long_string.orc: File size too small\n"
[  FAILED  ] TestCSVFileImport.testLongString (687 ms)
[----------] 1 test from TestCSVFileImport (687 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (687 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] TestCSVFileImport.testLongString

 1 FAILED TEST
```

I'm on Ubuntu 16.04.6. The invalid pointers might have undefined behaviors. I think it's ok if this catch the bug in some env but not all.

KyleGrains: I tried it on macOS 10.13.6, the test passed with and without the fix.

stiga-huang: I can't reproduce the problem on MacOS either.

KyleGrains: Maybe the clang++ compiled code is not reusing the invalidated address, so the referenced data is still there.

dongjoon-hyun: Got it. Thank you for confirming, @stiga-huang and @KyleGrains .

dongjoon-hyun (approved):
+1, LGTM.
- Thank you for your first contribution, @KyleGrains .
- Thank you for your time during your review/collaboration/approval here, @stiga-huang .

Merged to master/1.7.

cc @williamhyun because he is the release manager of Apache ORC 1.7.3.

dongjoon-hyun:
@KyleGrains  I added you to the Apache ORC contributor group and assigned ORC-1116 to you.
Welcome to the Apache ORC community again.

tools/test/TestCSVFileImport.cc
KyleGrains: Thank both of you too.

KyleGrains:
> @KyleGrains I added you to the Apache ORC contributor group and assigned [ORC-1116](https://issues.apache.org/jira/browse/ORC-1116) to you. Welcome to the Apache ORC community again.

Thanks, do I need to click any button on the issue page?

dongjoon-hyun:
No need from your side, @KyleGrains . :)

PR #1057, opened 2022-03-04
ORC-1123: Add `estimationMemory` method for writer

guiyanakuang:

Add `estimationMemory` method for writer. It exposes the internal treeWriter to estimate the memory used for the buffer.

This method help the write engine to better control the refresh policy of the ORC.

The source of this issue comes from Iceberg's discussion.
https://lists.apache.org/thread/g6yo7m46mr86ov1vkm9wnmshgw7hcl6b

Pass the CIs

### DIFF
--- a/java/core/src/java/org/apache/orc/Writer.java
+++ b/java/core/src/java/org/apache/orc/Writer.java
@@ -165,4 +165,11 @@ public interface Writer extends Closeable {
    * @since 1.6.8
    */
   List<StripeInformation> getStripes() throws IOException;
+
+  /**
+   * Estimate the memory currently used by the writer to buffer the stripe.
+   * `This method help write engine to control the refresh policy of the ORC.`
+   * @return the number of bytes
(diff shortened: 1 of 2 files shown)


dongjoon-hyun (approved):
I also read the background and get pings.
Merged to main/1.7 for Apache ORC 1.7.4.

dongjoon-hyun:
Also, cc @williamhyun since he is the release manager for Apache ORC 1.7.4.

pgaref:
Seems like we might be overestimating memory usage when taking into account unused buffers -- would probably make sense documenting/testing these estimates.

dongjoon-hyun:
+1 for doing that separately in a new ORC JIRA.

hililiwei:
> Seems like we might be overestimating memory usage when taking into account unused buffers -- would probably make sense documenting/testing these estimates.

In my tests, it did overestimate a bit.

PR #1062, opened 2022-03-10
ORC-1125: [C++] Support reading decimal64 in ORCv2

stiga-huang:

This PR adds a reader for decimal64 in ORCv2 (added in ORC-49). Also extends `FileVersion` to recognize the version of "UNSTABLE-PRE-2.0". Decimal columns written in such version will be read using the new reader.

The original reader codes come from #147 

Before this PR, the cpp reader is unable to read decimal columns in ORCv2.

Adds unit tests in TestColumnReader.cc
Adds an end to end test in TestFileContents.cc. 
Unfortunately, I can't find a clean way to generate such ORC files. So this PR introduces a new test file (decimal_v2.orc) generated by Hive.

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -502,6 +502,10 @@ namespace orc {
     return throwOnHive11DecimalOverflow;
   }
 
+  bool RowReaderImpl::getIsDecimalAsLong() const {
+    return contents->isDecimalAsLong;
+  }
+
   int32_t RowReaderImpl::getForcedScaleOnHive11Decimal() const {
     return forcedScaleOnHive11Decimal;
   }
@@ -616,9 +620,7 @@ namespace orc {
     if (contents->postscript->version_size() != 2) {
       return FileVersion::v_0_11();
     }
-    return FileVersion(
-                contents->postscript->version(0),
-                contents->postscript->version(1));
+    return {contents->postscript->version(0), contents->postscript->version(1)};
   }
 
   uint64_t ReaderImpl::getNumberOfRows() const {
@@ -1388,6 +1390,13 @@ namespace orc {
 
       contents->footer = REDUNDANT_MOVE(readFooter(stream.get(), buffer.get(),
         footerOffset, *contents->postscript,  *contents->pool));
+    }
+    contents->isDecimalAsLong = false;
+    if (contents->postscript->version_size() == 2) {
+      FileVersion v(contents->postscript->version(0), contents->postscript->version(1));
+      if (v == FileVersion::UNSTABLE_PRE_2_0()) {
+        contents->isDecimalAsLong = true;
+      }
     }
     contents->stream = std::move(stream);
     return std::unique_ptr<Reader>(new ReaderImpl(std::move(contents),
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -1656,6 +1656,60 @@ namespace orc {
     }
   }
 
+  class Decimal64ColumnReaderV2: public ColumnReader {
+  protected:
+    std::unique_ptr<RleDecoder> valueDecoder;
+    int32_t precision;
+    int32_t scale;
+
+  public:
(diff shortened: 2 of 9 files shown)


dongjoon-hyun (commented):
Thank you, @stiga-huang . Could you fix the build error?

c++/include/orc/Common.hh
```diff
@@ -34,6 +34,7 @@ namespace orc {
   public:
     static const FileVersion& v_0_11();
     static const FileVersion& v_0_12();
+    static const FileVersion& UNSTABLE_PRE_2_0();
```
dongjoon-hyun, line 37: Could you spin off this `UNSTABLE_PRE_2_0`-related part, @stiga-huang ?

c++/src/Common.cc
```diff
@@ -131,8 +131,11 @@ namespace orc {
   }
 
   std::string FileVersion::toString() const {
+    if (majorVersion == 1 && minorVersion == 9999) {
+      return "UNSTABLE-PRE-2.0";
```
wgtmac, line 135: Can we also skip writing & reading scale stream for Decimal128 in UNSTABLE-PRE-2.0?

wgtmac (commented):
Thanks @stiga-huang . Should we also implement the write path? I can do that later.

c++/src/Reader.cc
```diff
@@ -1389,6 +1389,13 @@ namespace orc {
       contents->footer = REDUNDANT_MOVE(readFooter(stream.get(), buffer.get(),
         footerOffset, *contents->postscript,  *contents->pool));
     }
+    if (contents->postscript->version_size() != 2) {
+      contents->fileVersion = std::unique_ptr<FileVersion>(new FileVersion(0, 11));
```
coderex2522, line 1393: Can this statement 'new FileVersion(0,11)' be changed to 'new FileVersion(FileVersion::v_0_11())'.

c++/src/Common.cc
stiga-huang: I think yes, but it'd be better to use RLEv3 since the values might have more tailing zeros than what we observed in https://github.com/apache/orc/pull/257#issuecomment-385100087

c++/src/Reader.cc
stiga-huang: Yes. BTW, I refactored the codes to just keep an `isDecimalAsLong` flag in `FileContents` so this can be simpler.

c++/include/orc/Common.hh
stiga-huang: Sorry that I don't understand this comment. Are you suggesting adding a check for this version instead of introducing an unstable version?

BTW, I use the same name we used in Java: https://github.com/apache/orc/blob/77b4d5acb58e03d50503d1b687421cac627a6e0b/java/core/src/java/org/apache/orc/OrcFile.java#L79

stiga-huang:
> Thanks @stiga-huang . Should we also implement the write path? I can do that later.

@wgtmac Sure. I recently won't have time on implementing the writer. Feel free to take it.

c++/include/orc/Common.hh
dongjoon-hyun: I'm suggesting to making another PR for `UNSTABLE_PRE_2.0` handling because it's orthogonal to `Decimal64` read support. If you spin off them, we can backport that PR to branch-1.7, @stiga-huang .

stiga-huang: Oh, that's a good point! Sure, let me do it.

stiga-huang: Created the PR: #1064

dongjoon-hyun:
Thank you, @stiga-huang . https://github.com/apache/orc/pull/1064 is merged to `main`/`branch-1.7`.
Could you rebase this PR to `main` branch once more?

stiga-huang:
> Thank you, @stiga-huang . #1064 is merged to `main`/`branch-1.7`. Could you rebase this PR to `main` branch once more?

Done

wgtmac (approved):
LGTM +1

PR #1066, opened 2022-03-16
ORC-1129: [C++] Make tool-test depends on cpp tools

stiga-huang:

Make `tool-test` depends on cpp tools so when any of the tool codes change, recompile tool-test will also recompile the tool.

This avoids mistakes when `tool-test` is using the old tool binaries that are inconsistent with the current codes.

Manually tested `make -j8 tool-test` and `make -j8 DESTDIR=./tmp install`

### DIFF
--- a/tools/src/CMakeLists.txt
+++ b/tools/src/CMakeLists.txt
@@ -94,12 +94,18 @@ target_link_libraries (csv-import
   ${CMAKE_THREAD_LIBS_INIT}
   )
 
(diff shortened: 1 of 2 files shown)


wgtmac (approved):
LGTM +1

dongjoon-hyun (approved):

dongjoon-hyun:
You can merge this, @stiga-huang . :)

dongjoon-hyun:
Thanks! :)

PR #1071, opened 2022-03-28
ORC-1137: [C++] Unroll loops and copy data directly in DoubleColumnReader::next()

stiga-huang:

This PR unrolls the loops inside DoubleColumnReader::readDouble() and readFloat(). Also refactor DoubleColumnReader to be a template class to save some runtime checks.
When running on a little-endian machine, the layout of the DATA stream matches the memory layout of the output double array. We use std::memcpy to copy data as more as we can in reading double columns.

This PR improves the performance in DoubleColumnReader::next(). The screenshot in the JIRA description shows it as a hot path.

Ran a performance test on reading 4 double columns on a file of tpch lineitem table using orc-scan. The file is 264MB with 7,296,488 rows.
I can see #instructions reduce 20% and time elapsed reduce 13%. More details here (collected by perf-stat):

|                   | original   | optimized   | reduce rate |
|-------------------|------------|-------------|-------------|
| task-clock (msec) | 786.314603 | 686.153474  | 12.74%      |
| context-switches  | 0          | 0           |             |
| cpu-migrations    | 0          | 0           |             |
| page-faults       | 6116       | 6113        | 0.05%       |
| cycles            | 3269319861 | 2865623800  | 12.35%      |
| instructions      | 8419802727 | 6739792930  | 19.95%      |
| branches          | 1185171422 | 925421384   | 21.92%      |
| branch-misses     | 989081     | 941046      | 4.86%       |
| time elapsed (s)  | 0.78855796 | 0.687502573 | 12.82%      |

Ran existing tests.

### DIFF
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -412,7 +412,7 @@ namespace orc {
     nanoRle->seek(positions.at(columnId));
   }
 
-  template<TypeKind COLUMN_KIND, bool IS_LITTLE_ENDIAN>
+  template<TypeKind columnKind, bool isLittleEndian>
   class DoubleColumnReader: public ColumnReader {
   public:
     DoubleColumnReader(const Type& type, StripeStreams& stripe);
@@ -429,7 +429,7 @@ namespace orc {
 
   private:
     std::unique_ptr<SeekableInputStream> inputStream;
-    const uint64_t bytesPerValue = (COLUMN_KIND == FLOAT) ? 4 : 8;
+    const uint64_t bytesPerValue = (columnKind == FLOAT) ? 4 : 8;
     const char *bufferPointer;
     const char *bufferEnd;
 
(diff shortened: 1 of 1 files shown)


wgtmac (approved):
LGTM +1

dongjoon-hyun (commented):
Thank you, @stiga-huang and @wgtmac .
Could you add the performance change between **before** and **after** into the PR description?

c++/src/ColumnReader.cc
```diff
@@ -412,10 +412,11 @@ namespace orc {
     nanoRle->seek(positions.at(columnId));
   }
 
+  template<TypeKind COLUMN_KIND, bool IS_LITTLE_ENDIAN>
```
dongjoon-hyun, line 415: Although I know that this pattern is used before, this still looks misleading. For example, this PR is introducing the following which is very confusing because both one looks like CONSTANT.
```
if (COLUMN_KIND == FLOAT) {
```

dongjoon-hyun: Is this the best practice for us to use as Apache ORC C++ coding standard, @stiga-huang ? It's up to you.

stiga-huang: Thanks for raising this question! I think I should rename `COLUMN_KIND` to `columnKind` and `IS_LITTLE_ENDIAN` to `isLittleEndian` since they are not type names.

stiga-huang:
> Could you add the performance change between before and after into the PR description?

Done

dongjoon-hyun (approved):
Thank you for updates.

wgtmac:
I have merged this. Thanks @stiga-huang @dongjoon-hyun !

PR #570, opened 2020-11-21
ORC-685: Add `ReaderImpl.extractFileTail` back

pgaref:
### What changes were proposed in this pull request?
This PR aims to add ReaderImpl.extractFileTail back to mitigate backward compatibility (partially as StripeStatistics are removed from ORCTail).

### Why are the changes needed?
Apache ORC 1.6.0 ~ 1.6.5 removes extractFileTail  from the ReaderImpl. This PR adds API backward compatibility.

### How was this patch tested?
TestReaderImp. testOrcTailStripeStats()

PR #557, opened 2020-10-28
ORC-677: Add a deprecated legacy constructor SargApplier back

dongjoon-hyun:
### What changes were proposed in this pull request?

This PR aims to add a deprecated legacy constructor `SargApplier` back.

### Why are the changes needed?

ORC-27 removes this at Apache ORC 1.5.9 and 1.6.3 and this causes a compilation error at Apache Hive 3.1 branch (https://github.com/apache/hive/pull/1615).
- ORC-27: https://github.com/apache/orc/commit/949c7444ef98680026d9f2b93d7299b76cdfb030#diff-8678415090c906d11e0aa6241ba24643192cd66617c68d8deb0f7efe74dd2ba3L876-R883

### How was this patch tested?

Pass the CI with the newly added test case.

PR #555, opened 2020-10-27
ORC-676. Add getRawDataSizeFromColIndices back to ReaderImpl

dongjoon-hyun:
### What changes were proposed in this pull request?

This PR aims to add `getRawDataSizeFromColIndices` back to `ReaderImpl`.

### Why are the changes needed?

This helps the migration path from Apache Orc 1.5 to 1.6+

### How was this patch tested?

Pass the newly added test case.

PR #1072, opened 2022-03-28
ORC-1138

pavibhai:
Optimizes the read of streams in ORC by combining multiple nearby reads a single read, optionally allowing the retention or drop of the extra bytes.

* minSeekSize: If separation between multiple reads is within minSeekSize then these are combined into a single read
* minSeekSizeTolerance: Helps in the decisioning of whether to retain the extra bytes (extra memory) or take extra CPU to drop the unwanted bytes

We are introducing two new configuration parameters that control how read of streams takes place in ORC
* minSeekSize: If separation between multiple reads is within minSeekSize then these are combined into a single read
* minSeekSizeTolerance: Helps in the decisioning of whether to retain the extra bytes (extra memory) or take extra CPU to drop the unwanted bytes

This leads to significant time savings (and cost also) when dealing with AWS S3. Reads with gaps e.g. reading alternate columns shows a significant penalty 5.8s vs 1.4s with the patch.

* New Unit Tests were added
* None of the existing tests were changed

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -194,6 +194,18 @@ public enum OrcConf {
   ORC_MAX_DISK_RANGE_CHUNK_LIMIT("orc.max.disk.range.chunk.limit",
       "hive.exec.orc.max.disk.range.chunk.limit",
     Integer.MAX_VALUE - 1024, "When reading stripes >2GB, specify max limit for the chunk size."),
+  ORC_MIN_DISK_SEEK_SIZE("orc.min.disk.seek.size",
+                                 "hive.exec.orc.min.disk.seek.size",
+                                 0,
+                         "When determining contiguous reads, gaps within this size are "
+                         + "read contiguously and not seeked. Default value of zero disables this "
+                         + "optimization"),
+  ORC_MIN_DISK_SEEK_SIZE_TOLERANCE("orc.min.disk.seek.size.tolerance",
+                          "hive.exec.orc.min.disk.seek.size.tolerance", 0.00,
+                          "Define the tolerance for for extra bytes read as a result of "
+                          + "orc.min.disk.seek.size. If the (bytesRead - bytesNeeded) / bytesNeeded"
+                          + "is greater than this threshold then extra work is performed to drop "
+                          + "the extra bytes from memory after the read."),
   ENCRYPTION("orc.encrypt", "orc.encrypt", null, "The list of keys and columns to encrypt with"),
   DATA_MASK("orc.mask", "orc.mask", null, "The masks to apply to the encrypted columns"),
   KEY_PROVIDER("orc.key.provider", "orc.key.provider", "hadoop",
--- a/java/core/src/java/org/apache/orc/Reader.java
+++ b/java/core/src/java/org/apache/orc/Reader.java
@@ -234,6 +234,9 @@ public interface Reader extends Closeable {
     private boolean allowSARGToFilter = false;
     private boolean useSelected = false;
     private boolean allowPluginFilters = false;
+    private int minSeekSize = (int) OrcConf.ORC_MIN_DISK_SEEK_SIZE.getDefaultValue();
+    private double minSeekSizeTolerance = (double) OrcConf.ORC_MIN_DISK_SEEK_SIZE_TOLERANCE
+      .getDefaultValue();
 
     /**
      * @since 1.1.0
@@ -256,6 +259,8 @@ public interface Reader extends Closeable {
       allowSARGToFilter = OrcConf.ALLOW_SARG_TO_FILTER.getBoolean(conf);
       useSelected = OrcConf.READER_USE_SELECTED.getBoolean(conf);
       allowPluginFilters = OrcConf.ALLOW_PLUGIN_FILTER.getBoolean(conf);
+      minSeekSize = OrcConf.ORC_MIN_DISK_SEEK_SIZE.getInt(conf);
+      minSeekSizeTolerance = OrcConf.ORC_MIN_DISK_SEEK_SIZE_TOLERANCE.getDouble(conf);
     }
 
     /**
@@ -646,6 +651,24 @@ public interface Reader extends Closeable {
 
     public Options allowPluginFilters(boolean allowPluginFilters) {
       this.allowPluginFilters = allowPluginFilters;
+      return this;
+    }
+
+    public int minSeekSize() {
+      return minSeekSize;
+    }
+
+    public Options minSeekSize(int minSeekSize) {
+      this.minSeekSize = minSeekSize;
+      return this;
+    }
+
+    public double minSeekSizeTolerance() {
+      return minSeekSizeTolerance;
+    }
+
+    public Options minSeekSizeTolerance(double value) {
+      this.minSeekSizeTolerance = value;
       return this;
     }
   }
--- a/java/core/src/java/org/apache/orc/impl/BufferChunk.java
+++ b/java/core/src/java/org/apache/orc/impl/BufferChunk.java
@@ -89,7 +89,10 @@ public class BufferChunk extends DiskRangeList {
       return false;
     }
     BufferChunk ob = (BufferChunk) other;
-    return chunk.equals(ob.chunk);
+    return (chunk != null && ob.chunk != null && chunk.equals(ob.chunk))
+           || (chunk == null && ob.chunk == null
+               && offset == ob.offset
+               && getLength() == ob.getLength());
   }
 
   @Override
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
@@ -53,6 +53,8 @@ public class RecordReaderUtils {
     private final Supplier<FileSystem> fileSystemSupplier;
     private final Path path;
     private final boolean useZeroCopy;
+    private final int minSeekSize;
+    private final double minSeekSizeTolerance;
     private InStream.StreamOptions options;
     private boolean isOpen = false;
 
@@ -62,6 +64,8 @@ public class RecordReaderUtils {
       this.file = properties.getFile();
       this.useZeroCopy = properties.getZeroCopy();
       this.options = properties.getCompression();
+      this.minSeekSize = properties.getMinSeekSize();
+      this.minSeekSizeTolerance = properties.getMinSeekSizeTolerance();
     }
 
     @Override
@@ -99,7 +103,8 @@ public class RecordReaderUtils {
     public BufferChunkList readFileData(BufferChunkList range,
                                         boolean doForceDirect
                                         ) throws IOException {
-      RecordReaderUtils.readDiskRanges(file, zcr, range, doForceDirect);
+      RecordReaderUtils.readDiskRanges(file, zcr, range, doForceDirect, minSeekSize,
+                                       minSeekSizeTolerance);
       return range;
     }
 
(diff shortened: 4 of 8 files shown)


dongjoon-hyun (commented):
Thank you, @pavibhai .

cc @williamhyun

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -194,6 +194,18 @@
   ORC_MAX_DISK_RANGE_CHUNK_LIMIT("orc.max.disk.range.chunk.limit",
       "hive.exec.orc.max.disk.range.chunk.limit",
     Integer.MAX_VALUE - 1024, "When reading stripes >2GB, specify max limit for the chunk size."),
+  ORC_MIN_DISK_SEEK_SIZE("orc.min.disk.seek.size",
+                                 "hive.exec.orc.min.disk.seek.size",
+                                 0,
```
dongjoon-hyun, line 199: indentation?

java/core/src/java/org/apache/orc/impl/BufferChunk.java
```diff
@@ -89,7 +89,10 @@ public boolean equals(Object other) {
       return false;
     }
     BufferChunk ob = (BufferChunk) other;
-    return chunk.equals(ob.chunk);
+    return (chunk != null && ob.chunk != null && chunk.equals(ob.chunk))
+           || (chunk == null && ob.chunk == null
+               && offset == ob.offset
+               && getLength() == ob.getLength());
```
dongjoon-hyun, line 95: This looks like breaking the contact of `equals`.

dongjoon-hyun (dismissed):
I'm not sure about the change on `BufferChunk.equals` method. Are you sure, @pavibhai ?

dongjoon-hyun:
cc @pgaref and @guiyanakuang ,  too

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -194,6 +194,18 @@
...
+                         "When determining contiguous reads, gaps within this size are "
+                         + "read contiguously and not seeked. Default value of zero disables this "
+                         + "optimization"),
+  ORC_MIN_DISK_SEEK_SIZE_TOLERANCE("orc.min.disk.seek.size.tolerance",
```
rbalamohan, line 203: Thanks for sharing the patch. AWS S3 connectors by default has readahead (mostly set to 64 or 128KB). So in a way, the data is read in addition to what is requested for. 
1. Would this patch be different from that?
2. There can be cases when this could be reading more than necessary and throwing off the read bytes later. Would that cause perf penalties?

java/core/src/java/org/apache/orc/impl/BufferChunk.java
guiyanakuang: It looks like the offset/length comparison is missing when chunck != null, how about this
```java
return Objects.equals(chunk, ob.chunk) && offset == ob.offset && getLength() == ob.getLength();
```

java/core/src/java/org/apache/orc/OrcConf.java
pavibhai: Fixed

java/core/src/java/org/apache/orc/impl/BufferChunk.java
dongjoon-hyun: Thank you for reverting, @pavibhai .

java/core/src/test/org/apache/orc/impl/TestRecordReaderUtils.java
```diff
@@ -0,0 +1,168 @@
...
+  private void validateChunks(RecordReaderUtils.ChunkReader chunkReader) {
+    BufferChunk current = chunkReader.getFrom();
+    while (current != chunkReader.getTo().next) {
+      assertTrue(current.hasData());
+      assertEquals(current.getOffset() % Byte.MAX_VALUE, current.getData().get(),
+                   String.format("Failed for %s", current));
+      current = (BufferChunk) current.next;
+    }
+  }
+}
```
guiyanakuang, line 168: Missing a newline at the end

pavibhai: Done

java/core/src/java/org/apache/orc/impl/BufferChunk.java
pavibhai: @dongjoon-hyun I have reverted this and included this comparison logic in the unit test.

IMO it is more thorough to check for offset and length in the equals. We can take that up independent of this patch.

@omalley or @pgaref any thoughts on this?

java/core/src/java/org/apache/orc/Reader.java
```diff
@@ -648,6 +653,24 @@ public Options allowPluginFilters(boolean allowPluginFilters) {
       this.allowPluginFilters = allowPluginFilters;
       return this;
     }
+
+    public int minSeekSize() {
```
dongjoon-hyun, line 657: Could you add `@since` to newly added methods?

pavibhai: Sorry missed that, added to the new methods in Options.

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -194,6 +194,18 @@
...
       "hive.exec.orc.max.disk.range.chunk.limit",
     Integer.MAX_VALUE - 1024, "When reading stripes >2GB, specify max limit for the chunk size."),
+  ORC_MIN_DISK_SEEK_SIZE("orc.min.disk.seek.size",
+                         "hive.exec.orc.min.disk.seek.size",
+                         0,
+                         "When determining contiguous reads, gaps within this size are "
+                         + "read contiguously and not seeked. Default value of zero disables this "
+                         + "optimization"),
+  ORC_MIN_DISK_SEEK_SIZE_TOLERANCE("orc.min.disk.seek.size.tolerance",
+                          "hive.exec.orc.min.disk.seek.size.tolerance", 0.00,
```
dongjoon-hyun, line 204: Can we remove this Hive conf? Actually, Apache Hive has no plan to use this, doesn't it?

pavibhai: @dongjoon-hyun Agreed, I have reverted both the properties to use the ORC configuration

dongjoon-hyun:
Thank you for updates, @pavibhai .

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -194,6 +194,18 @@ public enum OrcConf {
...
+  ORC_MIN_DISK_SEEK_SIZE("orc.min.disk.seek.size",
+                         "orc.min.disk.seek.size",
+                         0,
+                         "When determining contiguous reads, gaps within this size are "
+                         + "read contiguously and not seeked. Default value of zero disables this "
+                         + "optimization"),
+  ORC_MIN_DISK_SEEK_SIZE_TOLERANCE("orc.min.disk.seek.size.tolerance",
+                          "orc.min.disk.seek.size.tolerance", 0.00,
+                          "Define the tolerance for for extra bytes read as a result of "
+                          + "orc.min.disk.seek.size. If the (bytesRead - bytesNeeded) / bytesNeeded"
```
dongjoon-hyun, line 206: `bytesNeeded"` -> `bytesNeeded "`

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -405,72 +410,31 @@ static void zeroCopyReadRanges(FSDataInputStream file,
...
   }
 
-  /**
-   * Read the data from the file based on a list of ranges in a single read.
-   * @param file the file to read from
-   * @param first the first range to read
-   * @param last the last range to read
-   * @param allocateDirect should we use direct buffers
-   */
-  static void readRanges(FSDataInputStream file,
```
dongjoon-hyun, line 415: Do you think we can keep this in some way, @pavibhai ? Although it's `impl` package, we know that some downstream like Hive and Flink is using our internals. If possible, we should not remove anything for safety.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -405,72 +410,31 @@ static void zeroCopyReadRanges(FSDataInputStream file,
...
-               last.next.getOffset() <= currentEnd &&
+               last.next.getOffset() <= (currentEnd + minSeekSize) &&
                last.next.getEnd() - first.getOffset() < Integer.MAX_VALUE) {
       last = (BufferChunk) last.next;
       currentEnd = Math.max(currentEnd, last.getEnd());
     }
     return last;
   }
 
+  static BufferChunk findSingleRead(BufferChunk first) {
```
dongjoon-hyun, line 434: Please move function to the original line after the original comment.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -481,26 +445,40 @@ static BufferChunk findSingleRead(BufferChunk first) {
...
-        readRanges(file, current, last, doForceDirect);
+        ChunkReader chunkReader = ChunkReader.create(current, minSeekSize);
+        chunkReader.readRanges(file, doForceDirect, minSeekSizeTolerance);
+        current = (BufferChunk) chunkReader.to.next;
       }
-      current = (BufferChunk) last.next;
     }
   }
 
+  static void readDiskRanges(FSDataInputStream file,
```
dongjoon-hyun, line 475: ditto. We should put this public method after the method description.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -481,26 +445,40 @@ static BufferChunk findSingleRead(BufferChunk first) {
    * @param zcr the zero copy shim
    * @param list the disk ranges within the file to read
    * @param doForceDirect allocate direct buffers
+   * @param minSeekSize the minimum gap to prefer seek vs read
+   * @param minSeekSizeTolerance allowed tolerance for extra bytes in memory as a result of
+   *                             minSeekSize
    */
   static void readDiskRanges(FSDataInputStream file,
```
dongjoon-hyun, line 452: Do we need to add this method as public? Shall we add it as `private`?

java/core/src/java/org/apache/orc/OrcConf.java
pavibhai: done

dongjoon-hyun:
Could you revise the PR title, @pavibhai ?

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
pavibhai: I checked out hive@master, flink@master and I don't see any references to this method.

I think it should be fine to remove this method. As you called out given that it is in impl I think we should be ok cleaning up a method that is package local.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
pavibhai: it is package local moved

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
pavibhai: moved the package local method as suggested

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
pavibhai: sure, made it private

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
dongjoon-hyun: Thank you for checking, @pavibhai . However, it wasn't okay at all. Please see the history.

ORC-685: Add `ReaderImpl.extractFileTail` back (#570)
ORC-677: Add a deprecated legacy constructor SargApplier back (#557)
ORC-676. Add getRawDataSizeFromColIndices back to ReaderImpl (#555)
ORC-671. Add OrcTail.getStripeStatistics back for backward compatiblility

Is there any difficulty to keep this method, @pavibhai ?

dongjoon-hyun: Hive and Flink were just examples here because they didn't use the last Apache ORC yet due to those breaking change from our side. Please check our Adopter page. I don't think it's safe for new `Improvement` PR to remove something in the feature release. 

https://orc.apache.org/docs/adopters.html

Even Apache ORC 2.0, I'd avoid this kind approach.

dongjoon-hyun: Hive is using `<orc.version>1.6.9</orc.version>` and Flink is using `<orc.version>1.5.6</orc.version>`.
Apache ORC community had better conservative instead of moving away from our customers.

pavibhai: So you are recommending that any member that is not private should be assumed to be public? Is that correct?

pavibhai: Thanks for the quick call to clarify this.

I have added the deleted method back. There is some code duplication as a result of this in the method, if you are ok with rewriting this method with the new code while retaining the same interface I can do that as well.

dongjoon-hyun (approved):
+1, LGTM. Thank you, @pavibhai .

PR #1078, opened 2022-04-05
ORC-1146: Float category missing check if the statistic sum is a finite value

dongjoon-hyun:

This pr is aimed at checking whether the float category statistic sum has a finite value.

When the orc float category is written with NaN, pushing down is not supported.

Added unit test.


dongjoon-hyun:
cc @guiyanakuang . This will be merged with your authorship after passing CIs.

PR #1080, opened 2022-04-05
ORC-1147: Use `isNaN` instead of `isFinite` to determine the contain NaN values

guiyanakuang:

This pr is aimed at using `isNaN` instead of `isFinite` to determine the contain NaN values.
I want to exclude Double.POSITIVE_INFINITY / Double.NEGATIVE_INFINITY both cases, and only match NaN.

In the case of a sum overflow we can also predicate down to skip the corresponding strip.

Added unit test.

### DIFF
--- a/java/core/src/test/org/apache/orc/TestVectorOrcFile.java
+++ b/java/core/src/test/org/apache/orc/TestVectorOrcFile.java
@@ -4384,6 +4384,91 @@ public class TestVectorOrcFile {
     assertEquals(0, batch.size);
   }
 
+  @ParameterizedTest
+  @MethodSource("data")
+  public void testPredicatePushdownWithSumOverflow(Version fileFormat) throws Exception {
+    TypeDescription schema = TypeDescription.createStruct()
+        .addField("double1", TypeDescription.createDouble())
+        .addField("float1", TypeDescription.createFloat());
+
+    Writer writer = OrcFile.createWriter(testFilePath,
+        OrcFile.writerOptions(conf)
+            .setSchema(schema)
+            .stripeSize(400000L)
+            .compress(CompressionKind.NONE)
+            .bufferSize(500)
+            .rowIndexStride(1000)
+            .version(fileFormat));
+    VectorizedRowBatch batch = schema.createRowBatch();
+    batch.ensureSize(3500);
+    batch.size = 3500;
+    batch.cols[0].noNulls = true;
+    batch.cols[1].noNulls = true;
+
+    DoubleColumnVector dbcol = ((DoubleColumnVector) batch.cols[0]);
+    DoubleColumnVector fcol = ((DoubleColumnVector) batch.cols[1]);
+
+    // First two rows of data cause sum overflow, sum is not a finite value,
+    // but this does not prevent pushing down (range comparisons work fine)
+    // The same applies to the middle stripe
+    fcol.vector[0] = dbcol.vector[0] = Double.MAX_VALUE / 2 + Double.MAX_VALUE / 4;
+    fcol.vector[1] = dbcol.vector[1] = Double.MAX_VALUE / 2 + Double.MAX_VALUE / 4;
+    for (int i=2; i < 3500; ++i) {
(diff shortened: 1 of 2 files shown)


dongjoon-hyun (changes requested):
Oh, @guiyanakuang . This is not correct, @guiyanakuang .

There exists multiple NaN values. Double.NaN is just one of them.

dongjoon-hyun:
Please note that IEEE Standard defines NaN as range, not a single value.

guiyanakuang:
> Oh, @guiyanakuang . This is not correct, @guiyanakuang .
> 
> There exists multiple NaN values. Double.NaN is just one of them.

Maybe I should use the JDK's own method `new Double(dstas.getSum()).isNaN()`, where I want to exclude Double.POSITIVE_INFINITY / Double.NEGATIVE_INFINITY both cases, and only match NaN

dongjoon-hyun:
Thank sounds better.

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4384,6 +4384,91 @@ public void testPredicatePushdownWithNan(Version fileFormat) throws Exception {
...
+    batch.cols[0].noNulls = true;
+    batch.cols[1].noNulls = true;
+
+    DoubleColumnVector dbcol = ((DoubleColumnVector) batch.cols[0]);
+    DoubleColumnVector fcol = ((DoubleColumnVector) batch.cols[1]);
+
+    // First two rows of data cause sum overflow, sum is not a finite value,
+    // but this does not prevent pushing down (range comparisons work fine)
+    // The same applies to the middle stripe
+    fcol.vector[0] = dbcol.vector[0] = Double.MAX_VALUE / 2 + Double.MAX_VALUE / 4;
```
dongjoon-hyun, line 4414: Please define a constant variable once for `Double.MAX_VALUE / 2 + Double.MAX_VALUE / 4` and use it.

dongjoon-hyun:
cc @williamhyun

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4384,6 +4384,91 @@ public void testPredicatePushdownWithNan(Version fileFormat) throws Exception {
...
+        .end()
+        .build();
+
+    rows = reader.rows(reader.options()
+        .range(0L, Long.MAX_VALUE)
+        .searchArgument(sarg, new String[]{"float1"}));
+    batch = reader.getSchema().createRowBatch(3500);
+
+    rows.nextBatch(batch);
+    // First stride should be read
```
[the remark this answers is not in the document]
guiyanakuang: Sorry for the confusion between Stripe and Stride, this test only generates a stripe and the index is split into 4 strides.
The doc notes have been amended.

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
guiyanakuang: Resolved

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4384,6 +4384,96 @@ public void testPredicatePushdownWithNan(Version fileFormat) throws Exception {
...
+    batch.cols[1].noNulls = true;
+
+    DoubleColumnVector dbcol = ((DoubleColumnVector) batch.cols[0]);
+    DoubleColumnVector fcol = ((DoubleColumnVector) batch.cols[1]);
+
+    double largeNumber = Double.MAX_VALUE / 2 + Double.MAX_VALUE / 4;
+
+    // Here we are writing 3500 rows of data, with stripeSize set to 400000
+    // and rowIndexStride set to 1000, so 1 stripe will be written,
+    // indexed in 4 strides.
```
dongjoon-hyun, line 4415: Thank you for the details.

dongjoon-hyun (approved):

PR #1082, opened 2022-04-06
ORC-1147: Use isNaN instead of isFinite to determine the contain NaN values

dongjoon-hyun:

This pr is aimed at using `isNaN` instead of `isFinite` to determine the contain NaN values.
I want to exclude Double.POSITIVE_INFINITY / Double.NEGATIVE_INFINITY both cases, and only match NaN.

In the case of a sum overflow we can also predicate down to skip the corresponding strip.

Added unit test.

### DIFF
--- a/java/core/src/test/org/apache/orc/TestVectorOrcFile.java
+++ b/java/core/src/test/org/apache/orc/TestVectorOrcFile.java
@@ -4153,9 +4153,7 @@ public class TestVectorOrcFile {
     assertEquals(0, batch.size);
   }
 
-  @ParameterizedTest
-  @MethodSource("data")
-  public void testPredicatePushdownWithSumOverflow(Version fileFormat) throws Exception {
+  public void testPredicatePushdownWithSumOverflow() throws Exception {
     TypeDescription schema = TypeDescription.createStruct()
         .addField("double1", TypeDescription.createDouble())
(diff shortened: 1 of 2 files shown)


dongjoon-hyun:
After passing tests, this will be merged as @guiyanakuang 's authorship.

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4153,6 +4153,96 @@ public void testPredicatePushdownWithNan() throws Exception {
     assertEquals(0, batch.size);
   }
 
+  @ParameterizedTest
+  @MethodSource("data")
+  public void testPredicatePushdownWithSumOverflow(Version fileFormat) throws Exception {
```
dongjoon-hyun, line 4158: Ah, this is incompatible with branch-1.6. Let me fix it, @guiyanakuang .
- https://github.com/apache/orc/runs/5844990614?check_suite_focus=true

guiyanakuang: Thanks for fixing this incompatibility. 😃

dongjoon-hyun:
This is merged via https://github.com/apache/orc/commit/d15022fe7ce007533ea66918db877024eed3403b

PR #1087, opened 2022-04-16
ORC-1150: [C++] Optimize RowReaderImpl::computeBatchSize() by pre-computation

stiga-huang:

In case of PPD, SargsApplier mantains a bool vector representing whether a RowGroup is picked after evaluating the SearchArguments. The bool vector is then used in RowReaderImpl::computeBatchSize() to compute the next skipped row id for  the current row range. The computation is done for each batch.

In the worst case, if all RowGroups are picked, computeBatchSize() has to iterate the whole vector starts from the current RowGroup. In practise, a stripe can have thousands of RowGroups. The complexity of iterating the bool vector is comparable to materializing the batch (usually have size as 1024). The perf result in the JIRA description shows that it can take 1/4 of the scan time.

Instead of maintaining the bool vector, we can pre-compute the next skipped row id of the row range that each RowGroup locates. RowReaderImpl::computeBatchSize() can use the results directly. This PR replaces the bool vector with the vector of row ids explained above. If the current RowGroup is skipped, the value is 0. This matches the case when the original bool value is false.

The pre-computation is still done in SargsApplier::pickRowGroups(). RowReaderImpl::computeBatchSize() and RowReaderImpl::advanceToNextRowGroup() are optimized to use the pre-computed results.

This PR improves the performance in case of PPD. I pick a 252MB file of the tpcds inventory table, use a modified orc-scan with PPD that picks all RowGroups to scan it. Here are the results measure by perf-stat:

|                | original    | optimized   | speedup      |
|----------------|-------------|-------------|--------------|
| Time taken (s) | 0.722492053 | 0.527221077 | 1.3703777875 |
| instructions   | 8725349519  | 5989083282  | 1.4568756366 |
| branches       | 1143322885  | 752821194   | 1.5187177169 |

Passed all existing tests.

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -1110,7 +1110,7 @@ namespace orc {
         currentRowInStripe = advanceToNextRowGroup(currentRowInStripe,
                                                    rowsInCurrentStripe,
                                                    footer->rowindexstride(),
-                                                   sargsApplier->getRowGroups());
+                                                   sargsApplier->getNextSkippedRows());
         previousRow = firstRowOfStripe[currentStripe] + currentRowInStripe - 1;
         if (currentRowInStripe > 0) {
           seekToRowGroup(static_cast<uint32_t>(currentRowInStripe / footer->rowindexstride()));
@@ -1141,7 +1141,7 @@ namespace orc {
                                     currentRowInStripe,
                                     rowsInCurrentStripe,
                                     footer->rowindexstride(),
-                                    sargsApplier->getRowGroups());
+                                    sargsApplier->getNextSkippedRows());
     }
     data.numElements = rowsToRead;
     if (rowsToRead == 0) {
@@ -1165,7 +1165,7 @@ namespace orc {
       uint64_t nextRowToRead = advanceToNextRowGroup(currentRowInStripe,
                                                      rowsInCurrentStripe,
                                                      footer->rowindexstride(),
-                                                     sargsApplier->getRowGroups());
+                                                     sargsApplier->getNextSkippedRows());
       if (currentRowInStripe != nextRowToRead) {
         // it is guaranteed to be at start of a row group
         currentRowInStripe = nextRowToRead;
@@ -1186,21 +1186,18 @@ namespace orc {
                                            uint64_t currentRowInStripe,
                                            uint64_t rowsInCurrentStripe,
                                            uint64_t rowIndexStride,
-                                           const std::vector<bool>& includedRowGroups) {
+                                           const std::vector<uint64_t>& nextSkippedRows) {
     // In case of PPD, batch size should be aware of row group boundaries. If only a subset of row
     // groups are selected then marker position is set to the end of range (subset of row groups
     // within stripe).
     uint64_t endRowInStripe = rowsInCurrentStripe;
-    if (!includedRowGroups.empty()) {
-      endRowInStripe = currentRowInStripe;
-      uint32_t rg = static_cast<uint32_t>(currentRowInStripe / rowIndexStride);
-      for (; rg < includedRowGroups.size(); ++rg) {
-        if (!includedRowGroups[rg]) {
-          break;
-        } else {
-          endRowInStripe = std::min(rowsInCurrentStripe, (rg + 1) * rowIndexStride);
-        }
-      }
+    uint64_t groupsInStripe = nextSkippedRows.size();
+    if (groupsInStripe > 0) {
+      auto rg = static_cast<uint32_t>(currentRowInStripe / rowIndexStride);
+      if (rg >= groupsInStripe) return 0;
+      uint64_t nextSkippedRow = nextSkippedRows[rg];
+      if (nextSkippedRow == 0) return 0;
+      endRowInStripe = nextSkippedRow;
     }
     return std::min(requestedSize, endRowInStripe - currentRowInStripe);
   }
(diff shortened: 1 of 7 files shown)


c++/src/Reader.cc
```diff
@@ -1186,41 +1186,46 @@ namespace orc {
...
+    }
+    auto rg = static_cast<uint32_t>(currentRowInStripe / rowIndexStride);
+    if (rg >= groupsInStripe) {
+      // Points to the end of the stripe
+      return rowsInCurrentStripe;
+    }
+    if (nextSkippedRows[rg] != 0) {
+      // Current row group is selected
+      return currentRowInStripe;
+    }
```
coderex2522, line 1222: Does Line 1219-1221 have redundancy with the following logic(Line 1224-1227)? It seems that deleting Line 1219-1221 does not affect the results.

stiga-huang: I think we need this. This is the only case that returns `currentRowInStripe`. If we delete this, we will return `rg * rowIndexStride` which is a different result. Or did I miss something?

wgtmac:
Thanks @stiga-huang for the investigation and a quick fix. I will take a look next week.

c++/src/Reader.cc
coderex2522: You're right. When the nextSkippedRows[rg] is not zero, it should return currentRowInStripe.

c++/src/Reader.cc
```diff
@@ -1186,41 +1186,46 @@ namespace orc {
...
-                                           const std::vector<bool>& includedRowGroups) {
+                                           const std::vector<uint64_t>& nextSkippedRows) {
     // In case of PPD, batch size should be aware of row group boundaries. If only a subset of row
     // groups are selected then marker position is set to the end of range (subset of row groups
     // within stripe).
     uint64_t endRowInStripe = rowsInCurrentStripe;
-    if (!includedRowGroups.empty()) {
-      endRowInStripe = currentRowInStripe;
-      uint32_t rg = static_cast<uint32_t>(currentRowInStripe / rowIndexStride);
-      for (; rg < includedRowGroups.size(); ++rg) {
```
wgtmac, line 1197: Just curious: how much it will ease the pain if we simply break the loop when it reaches the requestedSize of rows?

wgtmac (commented):
The fix LGTM. Left an inline question.

c++/src/Reader.cc
stiga-huang: Ah, this is a much simpler fix! Maybe the current PR is just better than this in some corner cases, e.g. small RowGroup stride + large requestedSize.

I redo the experiments and they have similar improvements:

|| Original | Current PR | New approach |
-- | -- | -- | --
Time taken (s) | 0.720241394 | 0.514005243 | 0.517543660
task-clock (msec) | 719.469567 | 513.202998 | 516.743815
context-switches | 0 | 0 | 0
cpu-migrations | 0 | 0 | 0
page-faults | 3,209 | 3,216 | 3,210
cycles | 3,184,577,093 | 2,309,953,292 | 2,257,880,102
instructions | 8,725,559,709 | 5,989,406,509 | 5,993,877,602
branches | 1,143,366,754 | 752,891,327 | 753,289,704
branch-misses | 12,242,784 | 11,990,373 | 12,041,485

I'm ok if we choose the simpler fix.

wgtmac: Thanks for the further investigation! I am OK with either approach. Will it do better if we break the loop in the computeBatchSize() even in your fix?

stiga-huang: `computeBatchSize()` doesn't have loops in this PR. Do you mean `advanceToNextRowGroup()`?

wgtmac: My apology. I was looking at the left side of the change.

wgtmac (approved):
+1 LGTM.

c++/src/sargs/SargsApplier.hh
```diff
@@ -60,10 +60,11 @@ namespace orc {
                       const std::map<uint32_t, BloomFilterIndex>& bloomFilters);
 
     /**
-     * Return a vector of bool for each row group for their selection
-     * in the last evaluation
+     * Return a vector of the next skipped row for each RowGroup. Each value is the row id
+     * in stripe. 0 means the current RowGroup is entirely skipped.
+     * Only valid after invoking pickRowGroups().
      */
-    const std::vector<bool>& getRowGroups() const { return mRowGroups; }
```
dongjoon-hyun, line 66: There is no way to keep the original method? This looks like a breaking change if we remove this public method.

dongjoon-hyun (commented):
The patch itself looks good. I'm just curious about the side-effect on our C++ library downstream projects. If there is no worry, I'm fine too.

c++/src/sargs/SargsApplier.hh
wgtmac: IIUC, _**SargsApplier.hh**_ is not a public header file.

stiga-huang: I think all the public header files that our users can use are https://github.com/apache/orc/tree/main/c%2B%2B/include/orc
So this is not a breaking change.

BTW, the current `vector<uint64_t>` result is a superset of the original `vector<bool>` in the way that 0 values mean not selected and non-zero values mean selected. If your downstream projects port the ORC codes directly and depend on this method, you can simply change the usages in this way.

If this is still not good enough, we can choose the simpler fix proposed by @wgtmac.

dongjoon-hyun: I'm fine with the AS-IS PR if you folks are sure. :)

dongjoon-hyun (approved):

PR #1088, opened 2022-04-17
ORC-1151: [C++] Fix `ColumnWriter` for non-UTC Timestamp columns

noirello:
Fix converting non UTC timestamps for statistics.

Currently, the statistics for timestamp columns are incorrect, when the writer's time zone is not UTC.

Ran the existing test cases.

### DIFF
--- a/c++/src/ColumnWriter.cc
+++ b/c++/src/ColumnWriter.cc
@@ -1837,7 +1837,7 @@ namespace orc {
         // TimestampVectorBatch already stores data in UTC
         int64_t millsUTC = secs[i] * 1000 + nanos[i] / 1000000;
         if (!isUTC) {
-          millsUTC = timezone.convertToUTC(millsUTC);
+          millsUTC = timezone.convertToUTC(secs[i]) * 1000 + nanos[i] / 1000000;
         }
         ++count;
         if (enableBloomFilter) {
--- a/c++/test/TestTimestampStatistics.cc
+++ b/c++/test/TestTimestampStatistics.cc
@@ -21,10 +21,15 @@
 
 #include "Adaptor.hh"
 
+#include "MemoryInputStream.hh"
+#include "MemoryOutputStream.hh"
+
 #include "wrap/gmock.h"
 #include "wrap/gtest-wrapper.h"
 
 namespace orc {
+
+  static const int DEFAULT_MEM_STREAM_SIZE = 1024 * 1024; // 1M
 
   TEST(TestTimestampStatistics, testOldFile) {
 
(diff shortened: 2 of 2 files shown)


c++/src/ColumnWriter.cc
```diff
@@ -1837,7 +1837,7 @@ namespace orc {
         // TimestampVectorBatch already stores data in UTC
         int64_t millsUTC = secs[i] * 1000 + nanos[i] / 1000000;
         if (!isUTC) {
-          millsUTC = timezone.convertToUTC(millsUTC);
+          millsUTC = timezone.convertToUTC(secs[i]) * 1000 + nanos[i] / 1000000;
```
dongjoon-hyun, line 1840: Thank you, @noirello . Could you make a test case for this fix?

noirello: Sure, added a UTC and a non UTC test case.

noirello: Please, double check the timestamps in the test cases. Time zones always can be confusing.

dongjoon-hyun: It's a little counter-intuitive because we only convert `secs[i]` only. Could you add some comment here about why we don't need to convert `nanos[i]`?

c++/test/TestTimestampStatistics.cc
```diff
@@ -57,4 +62,97 @@ namespace orc {
...
+    EXPECT_TRUE(stripeColStats->hasUpperBound());
+    EXPECT_TRUE(stripeColStats->hasMinimum());
+    EXPECT_TRUE(stripeColStats->hasMaximum());
+    EXPECT_EQ(stripeColStats->getMinimum(), expectedMinMillis);
+    EXPECT_EQ(stripeColStats->getMaximum(), expectedMaxMillis);
+    EXPECT_EQ(stripeColStats->getLowerBound(), expectedMinMillis);
+    EXPECT_EQ(stripeColStats->getUpperBound(), expectedMaxMillis + 1);
+  }
+
+  TEST(TestTimestampStatistics, testTimezoneNonUTC) {
```
dongjoon-hyun, line 110: Thank you for adding this.

c++/test/TestTimestampStatistics.cc
```diff
@@ -57,4 +62,97 @@ namespace orc {
...
+      memStream.getData(), memStream.getLength()));
+    ReaderOptions rOptions;
+    rOptions.setMemoryPool(*pool);
+    std::unique_ptr<Reader> reader = createReader(std::move(inStream), rOptions);
+
+    std::unique_ptr<StripeStatistics> stripeStats = reader->getStripeStatistics(0);
+    const TimestampColumnStatistics* stripeColStats =
+      reinterpret_cast<const TimestampColumnStatistics*>(stripeStats->getColumnStatistics(1));
+
+
```
dongjoon-hyun, line 99: Shall we remove an empty line?

c++/src/ColumnWriter.cc
dongjoon-hyun: cc @wgtmac , @stiga-huang too because this is a correctness issue.

wgtmac: If we have set nanosecond in the test case _**WriterTest.writeTimestampWithTimezone**_, this issue may be fixed earlier.

wgtmac: > Yes, it's a little counter-intuitive because we only convert `secs[i]` only. So, `Timezone.convertToUTC` doesn't care `nanos` part properly and we should not put it here?
> 
> https://github.com/apache/orc/blob/f4c7cc14b31876e3dd4934b4bfc40c05d4b8ebca/c%2B%2B/src/Timezone.cc#L604-L606
> 
> If then, can we fix it in `Timezone.convertToUTC` instead? Is there a side-effect?

I think the current fix is good enough. As time_t uses second internally, we'd better keep the contract of Timezone.convertToUTC. We may add some comment above the fix to help understanding.

dongjoon-hyun: Got it. +1 for @wgtmac 's final decision.

c++/test/TestTimestampStatistics.cc
noirello: It's done.

dongjoon-hyun (approved):
+1, LGTM from my side. It would be great if we can get @wgtmac or @stiga-huang 's final sign-off.

wgtmac (approved):
LGTM. Thanks @noirello @dongjoon-hyun

stiga-huang (approved):
LGTM. Thanks for fixing this, @noirello !

williamhyun (commented):
+1, LGTM. 
Do we need this at ORC 1.7.5?

noirello:
> +1, LGTM. Do we need this at ORC 1.7.5?
Yes, I think this should be included in the 1.7 branch.

dongjoon-hyun:
+1 for backporting.

dongjoon-hyun:
Hi All. I backported this to branch-1.7.

PR #1096, opened 2022-04-25
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -8,6 +8,9 @@ on:
     branches:
     - main
 
+permissions:
+  contents: read
+
 jobs:
   build:
     name: "Java ${{ matrix.java }} and ${{ matrix.cxx }} on ${{ matrix.os }}"
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -20,8 +20,14 @@
 name: "On pull requests"
 on: pull_request_target
 
+permissions:
+  contents: read
+
 jobs:
   label:
+    permissions:
+      contents: read  # for actions/labeler to determine modified files
+      pull-requests: write  # for actions/labeler to add labels to PRs
     name: Label pull requests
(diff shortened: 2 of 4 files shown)


dongjoon-hyun (commented):
Could you elaborate a little more about which case the attackers succeed?
> even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

dongjoon-hyun:
Gentle ping, @naveensrinivasan ~

naveensrinivasan:
> Gentle ping, @naveensrinivasan ~

With the restricted permissions, they won't be able to compromise. HTH

dongjoon-hyun:
What I asked is `without this patch`, @naveensrinivasan .
IIRC, I've never heard about that kind of vulnerability on Apache repos from ASF Infra team.

naveensrinivasan:
> What I asked is `without this patch`, @naveensrinivasan .
> IIRC, I've never heard about that kind of vulnerability on Apache repos from ASF Infra team.

I will let you decide!

dongjoon-hyun:
What do you mean by the following? If there is no possibility of vulnerability, we don't need this PR by the definition. It's not up to anyone's decision.
> I will let you decide!

dongjoon-hyun:
Thank you for closing, @naveensrinivasan .

PR #1073, opened 2022-03-29
ORC-969: [C++] Evaluate SearchArguments using file and stripe level stats

coderex2522:
### What changes were proposed in this pull request?
PPD use file stats and stripe stats to filter file content.

### Why are the changes needed?
To make better use of file and stripe stats.

### How was this patch tested?
use the UT testStripeAndFileStats test.

PR #1099, opened 2022-04-27
ORC-1159: [C++] Fix crash when the last stripe is skipped

stiga-huang:

This PR fixes a bug introduced by a recent improvement (#1073). When the last stripe is skipped by StartNextStripe(), we should return in RowReaderImpl::next().

This fixes a crash of the C++ client when sargs exist and the last stripe is skipped.

We are lack of test coverage for the above case. Add more tests in TestPredicatePushdown.

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -1137,7 +1137,7 @@ namespace orc {
     if (currentRowInStripe == 0) {
       startNextStripe();
     }
-    uint64_t rowsToRead =
+    uint64_t rowsToRead = currentStripe >= lastStripe ? 0 :
       std::min(static_cast<uint64_t>(data.capacity),
                rowsInCurrentStripe - currentRowInStripe);
     if (sargsApplier && rowsToRead > 0) {
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -1035,6 +1035,20 @@ namespace orc {
     return memory + decompressorMemory ;
   }
 
+  // Update fields to indicate we've reached the end of file
+  void RowReaderImpl::markEndOfFile() {
+    currentStripe = lastStripe;
+    currentRowInStripe = 0;
+    rowsInCurrentStripe = 0;
+    if (lastStripe == 0) {
+      // Empty file
+      previousRow = 0;
+    } else {
+      previousRow = firstRowOfStripe[lastStripe - 1] +
+          footer->stripes(static_cast<int>(lastStripe - 1)).numberofrows();
+    }
+  }
+
   void RowReaderImpl::startNextStripe() {
     reader.reset(); // ColumnReaders use lots of memory; free old memory first
     rowIndexes.clear();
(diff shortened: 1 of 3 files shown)


dongjoon-hyun (commented):
Thank you, @stiga-huang .

c++/test/TestPredicatePushdown.cc
```diff
@@ -281,6 +282,45 @@ namespace orc {
     for (uint64_t i = 0; i < 2000; ++i) {
       EXPECT_EQ(i + 3500 , batch1.data[i]);
     }
+    // 2nd batch of 1500 rows
```
dongjoon-hyun, line 285: Please make a new test case, @stiga-huang .

dongjoon-hyun:
cc @coderex2522 and @wgtmac

dongjoon-hyun:
Also, cc @williamhyun if you have some time.

c++/test/TestPredicatePushdown.cc
stiga-huang: This line belongs to the current test. I'll move the following test of "Sargs: col1 < 3500" to a separated test.

dongjoon-hyun: +1 for that. Thanks!

c++/test/TestPredicatePushdown.cc
```diff
@@ -281,6 +282,50 @@ namespace orc {
...
+    EXPECT_EQ(5500, rowReader->getRowNumber());
+    EXPECT_EQ(1500, readBatch->numElements);
+    for (uint64_t i = 0; i < 1500; ++i) {
+      EXPECT_EQ(i + 5500 , batch1.data[i]) << i;
+    }
+    // no more batches
+    EXPECT_EQ(false, rowReader->next(*readBatch));
+  }
+
+  void TestSelectedWithStripeStats2(Reader* reader) {
```
dongjoon-hyun, line 297: Can we have a more meaningful name? Or, could you add a comment as a function description?

c++/src/Reader.cc
```diff
@@ -1133,7 +1133,7 @@ namespace orc {
     if (currentRowInStripe == 0) {
       startNextStripe();
     }
-    uint64_t rowsToRead =
+    uint64_t rowsToRead = currentStripe >= lastStripe ? 0 :
```
wgtmac, line 1136: Is it better to add a else branch to line 1094 like below:

```
if (currentStripe < lastStripe) {  // line 1094
  ...
} else {
  currentRowInStripe = rowsInCurrentStripe = 0;
  previousRow = firstRowOfStripe[lastStripe - 1] +
    footer->stripes(static_cast<int>(lastStripe - 1)).numberofrows();
}
```

stiga-huang: Sure. I think these are all better since they fix the issue inside `startNextStripe()` which is also used in `RowReaderImpl::seekToRow()`.

While adding test coverage for `seekToRow()`, I encountered another issue (ORC-1160) and I realized it also exists in 1.7.x releases. I think we need to resolve it first: #1100 . Then I'll rebase the current PR on it and continue to add tests.

stiga-huang: I continue the work after #1100 is merged. The current PR is rebased onto it.

I think all the above solutions fix the issue and don't have significant performance difference. So I tend to focus on readability. Added a new method `markEndOfFile()` to clean up some duplicated codes. Also added the test coverage on `seekToRow()`.

dongjoon-hyun (approved):

williamhyun (approved):

wgtmac:
I have merged it. Thanks @stiga-huang @dongjoon-hyun @williamhyun @coderex2522

PR #1108, opened 2022-05-07
ORC-1167: Support `orc.row.batch.size` configuration

cxzl25:
Support `orc.row.batch.size` configuration in `Reader.Options`.

Now create `OrcMapreduceRecordReader`, the default value of batch size is 1024.

If we read 1024 relatively large strings, we might get `NegativeArraySizeException`, but no configuration to reduce batch size.

```
java.lang.NegativeArraySizeException
	at org.apache.orc.impl.TreeReaderFactory$BytesColumnVectorUtil.commonReadByteArrays(TreeReaderFactory.java:1544)
	at org.apache.orc.impl.TreeReaderFactory$BytesColumnVectorUtil.readOrcByteArrays(TreeReaderFactory.java:1566)
	at org.apache.orc.impl.TreeReaderFactory$StringDirectTreeReader.nextVector(TreeReaderFactory.java:1662)
	at org.apache.orc.impl.TreeReaderFactory$StringTreeReader.nextVector(TreeReaderFactory.java:1508)
	at org.apache.orc.impl.TreeReaderFactory$StructTreeReader.nextBatch(TreeReaderFactory.java:2047)
	at org.apache.orc.impl.RecordReaderImpl.nextBatch(RecordReaderImpl.java:1219)
	at org.apache.orc.mapreduce.OrcMapreduceRecordReader.ensureBatch(OrcMapreduceRecordReader.java:84)
	at org.apache.orc.mapreduce.OrcMapreduceRecordReader.nextKeyValue(OrcMapreduceRecordReader.java:102)
	at org.apache.spark.sql.execution.datasources.RecordReaderIterator.hasNext(RecordReaderIterator.scala:39)
```

- Add UT
- Production environment verification

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -221,7 +221,7 @@ public enum OrcConf {
       "calendar. Only Hive 3.1 and the C++ library wrote using the proleptic, so hybrid\n" +
       "is the default."),
   ROW_BATCH_SIZE("orc.row.batch.size", "orc.row.batch.size", 1024,
-          "The number of rows to include in a orc vectorized reader batch. " +
+      "The number of rows to include in a orc vectorized reader batch. " +
       "The value should be carefully chosen to minimize overhead and avoid OOMs in reading data.")
   ;
 
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -222,8 +222,7 @@ public enum OrcConf {
       "is the default."),
   ROW_BATCH_SIZE("orc.row.batch.size", "orc.row.batch.size", 1024,
           "The number of rows to include in a orc vectorized reader batch. " +
-                  "The value should be carefully chosen to minimize overhead " +
-                  "and avoid OOMs in reading data.")
+      "The value should be carefully chosen to minimize overhead and avoid OOMs in reading data.")
   ;
 
   private final String attribute;
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -222,7 +222,7 @@ public enum OrcConf {
       "is the default."),
   ROW_BATCH_SIZE("orc.row.batch.size", "orc.row.batch.size", 1024,
           "The number of rows to include in a orc vectorized reader batch. " +
-                  "The number should be carefully chosen to minimize overhead " +
+                  "The value should be carefully chosen to minimize overhead " +
                   "and avoid OOMs in reading data.")
   ;
 
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -219,7 +219,11 @@ public enum OrcConf {
       "orc.proleptic.gregorian.default", false,
       "This value controls whether pre-ORC 27 files are using the hybrid or proleptic\n" +
       "calendar. Only Hive 3.1 and the C++ library wrote using the proleptic, so hybrid\n" +
-      "is the default.")
+      "is the default."),
+  ROW_BATCH_SIZE("orc.row.batch.size", "orc.row.batch.size", 1024,
+          "The number of rows to include in a orc vectorized reader batch. " +
+                  "The number should be carefully chosen to minimize overhead " +
+                  "and avoid OOMs in reading data.")
   ;
 
   private final String attribute;
--- a/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
@@ -1994,7 +1994,16 @@ public class TreeReaderFactory {
           totalLength = (int) (batchSize * scratchlcv.vector[0]);
         }
       }
-
+      if (totalLength < 0) {
+        StringBuilder sb = new StringBuilder("totalLength:" + totalLength
+                + " is a negative number.");
+        if (batchSize > 1) {
+          sb.append(" The current batch size is ")
+                  .append(batchSize)
+                  .append(", you can reduce the value by 'orc.row.batch.size'.");
+        }
+        throw new IOException(sb.toString());
+      }
       // Read all the strings for this batch
       byte[] allBytes = new byte[totalLength];
       int offset = 0;
--- a/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
@@ -1999,11 +1999,11 @@ public class TreeReaderFactory {
         StringBuilder sb = new StringBuilder("totalLength:" + totalLength
                 + " is a negative number.");
         if (batchSize > 1) {
-          sb.append(" The current batch size is ")
-                  .append(batchSize)
-                  .append(", you can reduce the value by '")
-                  .append(OrcConf.ROW_BATCH_SIZE.getAttribute())
-                  .append("'.");
+          sb.append(" The current batch size is ");
+          sb.append(batchSize);
+          sb.append(", you can reduce the value by '");
+          sb.append(OrcConf.ROW_BATCH_SIZE.getAttribute());
+          sb.append("'.");
         }
         throw new IOException(sb.toString());
       }
--- a/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
@@ -34,6 +34,7 @@ import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
 import org.apache.hadoop.hive.ql.exec.vector.expressions.StringExpr;
 import org.apache.hadoop.hive.ql.io.filter.FilterContext;
 import org.apache.hadoop.hive.serde2.io.HiveDecimalWritable;
+import org.apache.orc.OrcConf;
 import org.apache.orc.OrcFile;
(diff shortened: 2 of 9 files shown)


java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -219,7 +219,11 @@ public enum OrcConf {
       "orc.proleptic.gregorian.default", false,
       "This value controls whether pre-ORC 27 files are using the hybrid or proleptic\n" +
       "calendar. Only Hive 3.1 and the C++ library wrote using the proleptic, so hybrid\n" +
-      "is the default.")
+      "is the default."),
+  ROW_BATCH_SIZE("orc.row.batch.size", "orc.row.batch.size", 1024,
+          "The number of rows to include in a orc vectorized reader batch. " +
+                  "The number should be carefully chosen to minimize overhead " +
+                  "and avoid OOMs in reading data.")
```
dongjoon-hyun, line 226: The following will be enough. I also change `number` to `value` in the second sentence.
```java
  ROW_BATCH_SIZE("orc.row.batch.size", "orc.row.batch.size", 1024,
      "The number of rows to include in a orc vectorized reader batch. " +
      "The value should be carefully chosen to minimize overhead and avoid OOMs in reading data.")
```

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -1994,7 +1994,16 @@ private static byte[] commonReadByteArrays(InStream stream, IntegerReader length
...
-
+      if (totalLength < 0) {
+        StringBuilder sb = new StringBuilder("totalLength:" + totalLength
+                + " is a negative number.");
+        if (batchSize > 1) {
+          sb.append(" The current batch size is ")
+                  .append(batchSize)
+                  .append(", you can reduce the value by 'orc.row.batch.size'.");
+        }
+        throw new IOException(sb.toString());
```
dongjoon-hyun, line 2005: We need a test coverage for this new code path, @cxzl25 .

dongjoon-hyun (changes requested):
Thank you for filing a JIRA and making a PR, @cxzl25 . I left a few comments.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
cxzl25: Thanks for such a quick review, I added some UT. @dongjoon-hyun

java/core/src/java/org/apache/orc/OrcConf.java
dongjoon-hyun: @cxzl25 . Here I changed your indentation too. I understand the indentation is heterogenous in this file, but please follow my example.

cxzl25: Sorry I didn't notice the indentation issue here, I was breaking lines because it was over 100 characters. After changing the indentation, it is now 98 characters, thanks.

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -219,7 +219,10 @@ public enum OrcConf {
       "orc.proleptic.gregorian.default", false,
       "This value controls whether pre-ORC 27 files are using the hybrid or proleptic\n" +
       "calendar. Only Hive 3.1 and the C++ library wrote using the proleptic, so hybrid\n" +
-      "is the default.")
+      "is the default."),
+  ROW_BATCH_SIZE("orc.row.batch.size", "orc.row.batch.size", 1024,
+          "The number of rows to include in a orc vectorized reader batch. " +
```
dongjoon-hyun, line 224: My example included this line too.

cxzl25: IntelliJ IDEA sometimes hinted that I didn't notice it was a few characters indent, I fixed it now, thanks again for your patience.

image

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -1994,7 +1995,18 @@ private static byte[] commonReadByteArrays(InStream stream, IntegerReader length
...
-
+      if (totalLength < 0) {
+        StringBuilder sb = new StringBuilder("totalLength:" + totalLength
+                + " is a negative number.");
+        if (batchSize > 1) {
+          sb.append(" The current batch size is ")
+                  .append(batchSize)
+                  .append(", you can reduce the value by '")
+                  .append(OrcConf.ROW_BATCH_SIZE.getAttribute())
+                  .append("'.");
```
dongjoon-hyun, line 2006: Instead of this chaining, the following will be better in the readability.
```java
sb.append(" The current batch size is ")
sb.append(batchSize)
sb.append(", you can reduce the value by '")
sb.append(OrcConf.ROW_BATCH_SIZE.getAttribute())
sb.append("'.");
```

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -1994,7 +1995,18 @@ private static byte[] commonReadByteArrays(InStream stream, IntegerReader length
           totalLength = (int) (batchSize * scratchlcv.vector[0]);
         }
       }
-
+      if (totalLength < 0) {
+        StringBuilder sb = new StringBuilder("totalLength:" + totalLength
+                + " is a negative number.");
+        if (batchSize > 1) {
```
dongjoon-hyun, line 2001: This PR seems to be mixing three things. 

First, when `totalLength < 0` happens, do we need to throw a new exception from ORC side?

Second, in the above case, when `batchSize == 1` additionally, this PR is adding a new exception here. Logically, the following.
```java
if (totalLength < 0 && batchSize == 1) {
    throw new IOException("totalLength:" + totalLength + " is a negative number."
}
```
- If this is a meaningful error, can we proceed that in a separate PR because we may want to backport that.

dongjoon-hyun (commented):
I left a few comments including one for `assumeTrue(rt.maxMemory() > 4_000_000_000L);`. Please check them.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
cxzl25: It makes sense to add this exception.
Because when the user encounters the `NegativeArraySizeException`, user does not know what to do. The user can only read the source code of the orc to know that the batch size can be adjusted.

I can remove it in this PR first and then raise another jira to add this exception.

I encountered this problem when reading orc datasource in Spark 3.2, because the orc structure is complex, so the default is to use `OrcMapreduceRecordReader` to read.
Then I use config
```
spark.sql.orc.enableNestedColumnVectorizedReader=true
spark.sql.orc.columnarReaderBatchSize=1
```
Temporarily solved the problem.

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -219,7 +219,10 @@ public enum OrcConf {
       "orc.proleptic.gregorian.default", false,
       "This value controls whether pre-ORC 27 files are using the hybrid or proleptic\n" +
       "calendar. Only Hive 3.1 and the C++ library wrote using the proleptic, so hybrid\n" +
-      "is the default.")
+      "is the default."),
+  ROW_BATCH_SIZE("orc.row.batch.size", "orc.row.batch.size", 1024,
```
cxzl25, line 223: Do we need to split the two configurations? This way the reader and writer use different batch sizes.
I see the discussion result of [ORC-777](https://issues.apache.org/jira/browse/ORC-777) is not introducing configuration items to confuse users.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
dongjoon-hyun: Yes, I agree with you that the directional error message will be helpful.

java/core/src/test/org/apache/orc/impl/TestOrcLargeStripe.java
```diff
@@ -214,4 +215,77 @@ public void testStringDirectGreaterThan2GB() throws IOException {
       fs.delete(testFilePath, false);
     }
   }
+
+  @Test
+  public void testRowBatchSizeWhenStringDirectGreaterThan2GB() throws IOException {
+    final Runtime rt = Runtime.getRuntime();
+    assumeTrue(rt.maxMemory() > 4_000_000_000L);
```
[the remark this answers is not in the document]
dongjoon-hyun: No, increasing the memory requirements is harmful because we will miss the memory leakage issues in the future development. That's the point why we requires `2g` in our tests. Can you revise the test case instead?

java/core/src/java/org/apache/orc/OrcConf.java
dongjoon-hyun: cc @pgaref

java/core/src/test/org/apache/orc/impl/TestOrcLargeStripe.java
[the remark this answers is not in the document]
cxzl25: Make sense.
Now the trigger condition is that a relatively large string object is required. I tried the size as much as possible in this test case, but 2g is still not enough.
I now use the mock method to test this exception.

java/core/src/java/org/apache/orc/OrcConf.java
cxzl25: How about the orc.read.row.batch.size` configuration item in this PR?
Then I raise another PR about `NegativeArraySizeException`.
If the writer needs to support configuration in the future, I will raise another PR and configure it with `orc.write.row.batch.size`.

dongjoon-hyun (approved):
Merged to master for Apache ORC 1.9.

PR #1112, opened 2022-05-09
ORC-1139: Benchmark for ORC-1136 that combines multiple individual close reads into a single read

pavibhai:
Adds new benchmark to estimate the impact of combining multiple close reads into a single read.

Bench addition to ensure that we have not added any penalties as a result of this change

Running and verifying the bench results


dongjoon-hyun (commented):
Thank you for making a PR, @pavibhai !

dongjoon-hyun (approved):
+1, LGTM.

PR #1114, opened 2022-05-11
ORC-1170:[C++] Optimize the RowReader::seekToRow function in the case of seeking within the same stripe

coderex2522:
The patch will optimize the seekToRow function. The RowReader::seekToRow() function always call startNextStripe() to seek to the beginning of target stripe. However, when PPD is enabled many calls are still staying in the same stripe which results in many unnecessary calls. 

This PR improves the seekToRow function call. I pick a 259MB file of the TPCH lineitem table which contains 7200030
records. I use a modified orc-scan tool that calls the RowReader::seekToRow function once before each call the RowReader:: next() function to test this patch.  The results are shown below:

|| original | optimized | speedup |
|:-----|:----:|:----:|:----:|
|Elapsed Time(s)|43.69|27.13|1.61|

The original UTs(TestPredicatePushdown, DictionaryEncoding::multipleStripesWithIndex, DictionaryEncoding::multipleStripesWithoutIndex) will cover this patch.

### DIFF
--- a/c++/include/orc/MemoryPool.hh
+++ b/c++/include/orc/MemoryPool.hh
@@ -72,6 +72,10 @@ namespace orc {
       return currentCapacity;
     }
 
+    const T& operator[](uint64_t i) const {
+      return buf[i];
+    }
+
     T& operator[](uint64_t i) {
       return buf[i];
     }
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -424,8 +424,7 @@ namespace orc {
       if (rowIndexes.empty()) {
         loadStripeIndex();
       }
-      auto rowGroupId = static_cast<uint32_t>(rowsToSkip / rowIndexStride);
-      seekToRowGroup(rowGroupId);
+      seekToRowGroup(static_cast<uint32_t>(rowsToSkip / rowIndexStride));
       // skip leading rows in the target row group
       rowsToSkip %= rowIndexStride;
     }
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -282,7 +282,7 @@ namespace orc {
       }
     }
     firstStripe = currentStripe;
-    currentInitedStripe = lastStripe;
+    processingStripe = lastStripe;
 
     if (currentStripe == 0) {
       previousRow = (std::numeric_limits<uint64_t>::max)();
@@ -424,6 +424,7 @@ namespace orc {
       if (rowIndexes.empty()) {
         loadStripeIndex();
       }
+      // TODO: process the call of loadStripeIndex() failures
       seekToRowGroup(static_cast<uint32_t>(rowsToSkip / rowIndexStride));
       // skip leading rows in the target row group
       rowsToSkip %= rowIndexStride;
@@ -1088,7 +1089,7 @@ namespace orc {
       }
       currentStripeFooter = getStripeFooter(currentStripeInfo, *contents.get());
       rowsInCurrentStripe = currentStripeInfo.numberofrows();
-      currentInitedStripe = currentStripe;
+      processingStripe = currentStripe;
 
       if (sargsApplier) {
         bool isStripeNeeded = true;
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -282,6 +282,7 @@ namespace orc {
       }
     }
     firstStripe = currentStripe;
+    currentInitedStripe = lastStripe;
 
     if (currentStripe == 0) {
       previousRow = (std::numeric_limits<uint64_t>::max)();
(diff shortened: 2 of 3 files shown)


dongjoon-hyun (commented):
Thank you for making a PR, @coderex2522 .

c++/include/orc/MemoryPool.hh
```diff
@@ -72,6 +72,10 @@ namespace orc {
       return currentCapacity;
     }
 
+    const T& operator[](uint64_t i) const {
+      return buf[i];
+    }
+
```
dongjoon-hyun, line 78: Does this part have a big impact?

c++/src/Reader.hh
```diff
@@ -187,6 +188,16 @@ namespace orc {
     friend class TestRowReader_advanceToNextRowGroup_Test;
     friend class TestRowReader_computeBatchSize_Test;
 
+    inline uint64_t lastRowOfCurrentStripe() const {
```
dongjoon-hyun, line 191: Please add a function description comment.

c++/src/Reader.hh
```diff
@@ -187,6 +188,16 @@ namespace orc {
...
+      return currentStripe < lastStripe ?
+        firstRowOfStripe[currentStripe] + rowsInCurrentStripe :
+        footer->numberofrows();
+    }
+
+    inline bool isCurrentStripeInited() const {
```
dongjoon-hyun, line 197: Please add a function description comment.

c++/src/Reader.hh
coderex2522: Done

c++/src/Reader.hh
coderex2522: Done.

c++/include/orc/MemoryPool.hh
coderex2522: The DataBuffer class already has a non-const operator() function, so adding a const operator() function should have no effect.

stiga-huang (commented):
LGTM. It'd be nice if we can have a more specifit commit tittle. We are optimzing the case seeking within the same stripe. Seeking within the same RowGroup is not optimized in this PR. So having a title like "ORC-1170:[C++] Optimize seeking within the same stripe" seems better to me.

coderex2522:
> LGTM. It'd be nice if we can have a more specifit commit tittle. We are optimzing the case seeking within the same stripe. Seeking within the same RowGroup is not optimized in this PR. So having a title like "[ORC-1170](https://issues.apache.org/jira/browse/ORC-1170):[C++] Optimize seeking within the same stripe" seems better to me.

I think seeking within the same stripe  is inclusive of  seeking withing the same RowGroup. Seeking withing the same RowGroup will also call function startNextStripe under the original implementation.

c++/include/orc/MemoryPool.hh
wgtmac: suggest to use the trick to remove duplicates between const and non-const overloads: https://stackoverflow.com/questions/123758/how-do-i-remove-code-duplication-between-similar-const-and-non-const-member-func

c++/src/Reader.cc
```diff
@@ -391,27 +392,39 @@ namespace orc {
...
-          loadStripeIndex();
-        }
-        auto rowGroupId = static_cast<uint32_t>(rowsToSkip / rowIndexStride);
-        if (rowGroupId != 0) {
-          seekToRowGroup(rowGroupId);
-        }
+      if (rowIndexes.empty()) {
+        loadStripeIndex();
       }
+      seekToRowGroup(static_cast<uint32_t>(rowsToSkip / rowIndexStride));
```
wgtmac, line 427: If somehow it fails to parse the row index or finds some missing columns in the call of loadStripeIndex(), the seekToRowGroup here cannot proceed. Maybe add a TODO here to mark a future improvement.

c++/src/Reader.hh
```diff
@@ -146,6 +146,7 @@ namespace orc {
     uint64_t firstStripe;
     uint64_t currentStripe;
     uint64_t lastStripe; // the stripe AFTER the last one
+    uint64_t currentInitedStripe;
```
wgtmac, line 149: The naming is a little bit casual, maybe processingStripe?

c++/include/orc/MemoryPool.hh
coderex2522: I write a simple demo for testing the implement overload that use const function to  implement non-const function. 
The complex implement(return const_cast<T&>(static_cast<const DataBuffer<T>&>(*this)[i])) will cost twice the overhead of the simple implement(return buf[i]).
simple_demo

c++/src/Reader.hh
coderex2522: The function has remove.

c++/src/Reader.cc
coderex2522: Done

c++/src/Reader.hh
coderex2522: Done.

c++/include/orc/MemoryPool.hh
dongjoon-hyun: Thanks. Could you share your environment information and the numbers with the above, @coderex2522 ?

c++/src/Reader.cc
dongjoon-hyun: nit. 
For `TODO` comment, an IDed TODO style is recommended in order not to forget about it.
It means to file an official ORC JIRA and use `TODO(ORC-XXX)` style.

dongjoon-hyun:
Although I set the milestone for Apache ORC 1.9.0, please feel free to backport this if you want in Apache ORC 1.8.0, @wgtmac and @stiga-huang .

c++/include/orc/MemoryPool.hh
coderex2522: Test Environment: x86_64, gcc version 4.8.5
The test results were 7.03 seconds and 3.82 seconds respectively. @dongjoon-hyun

c++/src/Reader.cc
coderex2522: Improved the todo comment, thanks!

c++/include/orc/MemoryPool.hh
dongjoon-hyun: Which OS are you on? The build chain looks a little old to me. I expected `9.4.0` or `11.2.0`.

coderex2522: I retested it with gcc 9.2.1 and the results were similar. And both tests were on Linux OS x86_64.

wgtmac (approved):
LGTM, +1

dongjoon-hyun (approved):
+1, LGTM, too. Thank you, @coderex2522 , @stiga-huang , @wgtmac .
Merged to master.

dongjoon-hyun:
This is merged via https://github.com/apache/orc/commit/11a2544f4e54b29d5b0d3ca6aeb8dde5291b2163

PR #1137, opened 2022-05-23
ORC-1186: Limit `family` in `aarch64` profile

dongjoon-hyun:

This PR aims to limit OS family to `mac` in the AS-IS `aarch64` profile

ORC-814 added `applesilicon` profile and ORC-816 renamed it to `aarch64`.
However, the current Apache ORC doesn't support Linux `aarch64` yet.
We had better be clear what is working in that profile.

This profile should be tested on Apple Silicon manually. I tested it.

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -462,6 +462,7 @@
       </properties>
       <activation>
         <os>
+          <family>mac</family>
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
cc @williamhyun

PR #1143, opened 2022-05-26
ORC-1190: Fix `ORCWriterBenchMark` `dumpDir` initialization

mwlon:
Fix writer benchmark initialization

This does not affect main ORC or tests, only the benchmark suite.

```$ java -jar hive/target/orc-benchmarks-hive-*-uber.jar write data
$ java -jar hive/target/orc-benchmarks-hive-*-uber.jar write data
# JMH version: 1.20
# VM version: JDK 17.0.3, VM 17.0.3+7-LTS
# VM invoker: /Users/dongjoon/A/jdk-release/applejdk-17.0.3.7.3.jdk/Contents/Home/bin/java
# VM options: -server -Xms256m -Xmx2g -Dbench.root.dir=/Users/dongjoon/APACHE/orc-merge/java/bench/data
# Warmup: 2 iterations, 10 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: org.apache.orc.bench.hive.ORCWriterBenchMark.dictBench
# Parameters: (dictImpl = RBTREE, distinctCount = 10000)

# Run progress: 0.00% complete, ETA 00:10:30
# Fork: 1 of 1
# Warmup Iteration   1: [WARN ] Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
...
```

### DIFF
--- a/java/CMakeLists.txt
+++ b/java/CMakeLists.txt
@@ -88,7 +88,7 @@ add_test(
 
 add_test(
   NAME java-bench-hive-test
-  COMMAND java -Dbench.root.dir=. ${ADD_OPENS} ${JAVA_NIO} -jar bench/hive/orc-benchmarks-hive-${ORC_VERSION}-uber.jar write -i 1 -I 0 -t 1 data
+  COMMAND java ${ADD_OPENS} ${JAVA_NIO} -jar bench/hive/orc-benchmarks-hive-${ORC_VERSION}-uber.jar write -i 1 -I 0 -t 1 data
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
 install(
--- a/java/bench/hive/src/java/org/apache/orc/bench/hive/ORCWriterBenchMark.java
+++ b/java/bench/hive/src/java/org/apache/orc/bench/hive/ORCWriterBenchMark.java
@@ -59,8 +59,11 @@ import java.util.concurrent.TimeUnit;
 @AutoService(OrcBenchmark.class)
 public class ORCWriterBenchMark implements OrcBenchmark {
   private static final Path root = Utilities.getBenchmarkRoot();
-  private Path dumpDir = new Path(root, "dumpDir");
   private List<VectorizedRowBatch> batches = new ArrayList<>();
+
+  private Path dumpDir() {
+    return new Path(root, "dumpDir");
+  }
 
   @Override
   public String getName() {
@@ -112,7 +115,7 @@ public class ORCWriterBenchMark implements OrcBenchmark {
     Configuration conf = new Configuration();
     TrackingLocalFileSystem fs = new TrackingLocalFileSystem();
     fs.initialize(new URI("file:///"), conf);
-    fs.delete(dumpDir, true);
+    fs.delete(dumpDir(), true);
   }
 
   @Override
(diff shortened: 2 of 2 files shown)


dongjoon-hyun (commented):
Sorry but, for this specific one, I'm not sure this proposal is better than AS-IS behavior, @mwlon . For me, I prefer the early exception and termination because this is a benchmark driver.
> Allow running benchmarks even if irrelevant benchmarks are misconfigured

mwlon:
@dongjoon-hyun I agree we should have an early exception if the _intended_ benchmark is misconfigured, but shouldn't we be able to run a benchmark individually as long as it is correctly configured?

Right now I can't run any of the benchmarks - you can verify this behavior. For instance, trying to run the `"read-all"` benchmark fails simply because we can't initialize the irrelevant writer one to check if its name is equal to `"read-all"`.

dongjoon-hyun:
I agree with your user perspective. However, I prefer to push the community fix it immediately because we are the developers. :)

dongjoon-hyun:
BTW, FYI, we are trying to add a test coverage for benchmark. Here is the AS-IS progress. We can add your bug report here.

https://github.com/apache/orc/blob/dc2d86487e6e4479e8d674e64a851dd1a8da74aa/java/CMakeLists.txt#L79-L92

mwlon:
@dongjoon-hyun Do you prefer this solution?

dongjoon-hyun (commented):
Yes, this looks like fixing the real issue. Thanks for switching the direction, @mwlon .
Could you revise the PR title and JIRA title accordingly?

mwlon:
@dongjoon-hyun done.

dongjoon-hyun:
Thank you for updating. Let me verify manually in this evening. 
Since we don't have a test coverage in our CI, this should be tested manually.

dongjoon-hyun:
BTW, it would be better if you narrow down more into `dumpDir`. The current title is too broad. :)
```
Fix Writer benchmark initialization
```

dongjoon-hyun:
BTW, cc @autumnust and @pgaref , too.

dongjoon-hyun (changes requested):
Mostly looks good, but the following PR description `read-all` looks wrong to me.

> I tested by packaging and running java -jar hive/target/orc-benchmarks-hive-1.9.0-SNAPSHOT-uber.jar read-all data, observing that it no longer failed.

Could you replace the above test command with the following?
```
$ java -jar hive/target/orc-benchmarks-hive-*-uber.jar write data
# JMH version: 1.20
# VM version: JDK 17.0.3, VM 17.0.3+7-LTS
# VM invoker: /Users/dongjoon/A/jdk-release/applejdk-17.0.3.7.3.jdk/Contents/Home/bin/java
# VM options: -server -Xms256m -Xmx2g -Dbench.root.dir=/Users/dongjoon/APACHE/orc-merge/java/bench/data
# Warmup: 2 iterations, 10 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: org.apache.orc.bench.hive.ORCWriterBenchMark.dictBench
# Parameters: (dictImpl = RBTREE, distinctCount = 10000)

# Run progress: 0.00% complete, ETA 00:10:30
# Fork: 1 of 1
# Warmup Iteration   1: [WARN ] Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
...
```

In addition, as I mentioned before, we used with `-Dbench.root.dir=.` before. So, it has not been broken as you mentioned.

https://github.com/apache/orc/blob/dc2d86487e6e4479e8d674e64a851dd1a8da74aa/java/CMakeLists.txt#L90-L92

dongjoon-hyun (approved):
+1, LGTM (Pending CIs). Thank you, @mwlon .

`$ java -jar hive/target/orc-benchmarks-hive-*-uber.jar write data` command seems to be missed when you copied my comment. I added it to the PR description.

dongjoon-hyun:
Merged to main/1.8/1.7.

PR #1156, opened 2022-06-11
ORC-1200: Extracting encryption setup logic from `WriterImpl`

liujiawinds:
Extracting the encryption setup logic as a tool class.

Currently, the encryption setup at `WriterImpl` is so tightly coupled that the encryption variant cannot be obtained externally. This affects the integration of ORC and Flink.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/EncryptionConfiguration.java
+++ b/java/core/src/java/org/apache/orc/impl/EncryptionConfiguration.java
@@ -1,0 +1,212 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.orc.impl;
+
+
+import org.apache.orc.DataMask;
+import org.apache.orc.OrcFile;
+import org.apache.orc.TypeDescription;
+import org.apache.orc.impl.writer.WriterEncryptionKey;
+import org.apache.orc.impl.writer.WriterEncryptionVariant;
+
+import java.io.IOException;
+import java.security.SecureRandom;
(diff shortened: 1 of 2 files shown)


dongjoon-hyun:
Thank you for making a PR, @liujiawinds .
I set the milestone 1.9.0 which is the version of main branch.
For now, there is no plan of backporting this.

dongjoon-hyun (changes requested):
It's unclear to me why this is requested by Apache Flink community. For example,
1. Do you have some documentation or reference in Apache Flink community?
2. Is this Apache ORC design issue?
3. Is this request aligned with other encryption handling in Apache Flink community (like Apache Parquet Modular Encryption handling)?

In general, `branch-1.8` is already in feature freeze mode. I don't think this is required at v1.8.0 (currently).

Also, cc @omalley because he is the main author of Encryption Feature.

liujiawinds:
@dongjoon-hyun 

> 1. Do you have some documentation or reference in Apache Flink community?

Yes, first we must create a stream based `PhysicalFsWriter` before `WriterImpl` initial. [WriterImpl](https://github.com/apache/orc/blob/7facf81846a5e4395f2dbab335faef93f04c6aee/java/core/src/java/org/apache/orc/impl/WriterImpl.java#L223)
Then, You can look into flink orc writer from [OrcBulkWriterFactory](https://github.com/apache/flink/blob/99c2a415e9eeefafacf70762b6f54070f7911ceb/flink-formats/flink-orc/src/main/java/org/apache/flink/orc/writer/OrcBulkWriterFactory.java#L99). 
Given the above, now  I want to passing encryption settings to `PhysicalFsWriter` before `WriterImpl` initail, but the encryption setup at `WriterImpl ` is so tightly coupled that the encryption variant cannot be obtained externally. 

> 2. Is this Apache ORC design issue?

Yes, I think it is.

> 3. Is this request aligned with other encryption handling in Apache Flink community (like Apache Parquet Modular Encryption handling)?

Parquet has a similar class named `FileEncryptionProperties`.  [ParquetWriter](https://github.com/apache/parquet-mr/blob/59e9f78b8b3a30073db202eb6432071ff71df0ec/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetWriter.java#L283)

>In general, branch-1.8 is already in feature freeze mode. I don't think this is required at v1.8.0 (currently).

It's ok to release in later versions. I will extract this part of code in flink.

dongjoon-hyun:
Thank you for the details.

liujiawinds:
What should I do next for this pr? I don't quite understand the meaning of it being marked as `Changes requested` state.
:)

dongjoon-hyun:
Instead of commenting, you are supposed to revise your PR description with your previous comment's content because only the PR title and description becomes a commit message.
> What should I do next for this pr? I don't quite understand the meaning of it being marked as `Changes requested` state.
> :)

dongjoon-hyun:
After rethinking about this PR, I removed the milestone.

liujiawinds:
@dongjoon-hyun The PR description has been revised.

PR #1158, opened 2022-06-12
ORC-961: [C++] expose related metrics of the reader

coderex2522:
This patch keeps track of the time spent and the number of calls to each module. The current metrics mainly include the decompression time and number of calls, decoding time and number of calls and total elapsed time.

It exposes the relevant metrics of the reader, so that the user can visually see the metrics of each module which includes decompression, decoding.

The orc-scan tool with the -m parameter can output relevant metrics.


dongjoon-hyun (commented):
Could you fix C++ compilation error, @coderex2522?
```
/Users/runner/work/orc/orc/c++/src/Utils.hh:51:57: error: implicit use of sequentially-consistent atomic may incur stronger memory barriers than necessary [-Werror,-Watomic-implicit-seq-cst]
[1169](https://github.com/apache/orc/runs/6850860422?check_suite_focus=true#step:5:1170)
             __sync_fetch_and_add(latencyUs, elapsedTime);
[1170](https://github.com/apache/orc/runs/6850860422?check_suite_focus=true#step:5:1171)
             ~~~~~~~~~~~~~~~~~~~~                       ^
```

c++/include/orc/Reader.hh
```diff
@@ -39,6 +39,29 @@ namespace orc {
...
+    volatile uint64_t ReaderInclusiveLatencyUs;
+    volatile uint64_t DecompCount;
+    volatile uint64_t DecompLatencyUs;
+    volatile uint64_t DecodingCount;
+    volatile uint64_t DecodingLatencyUs;
+    volatile uint64_t ByteDecodingCount;
+    volatile uint64_t ByteDecodingLatencyUs;
+
+    ReaderMetrics() {
+      ReaderCount = 0;
```
wgtmac, line 53: nit: provide the default value in the declaration so we can avoid explicit constructor here.

c++/include/orc/Reader.hh
```diff
@@ -455,6 +478,12 @@ namespace orc {
      */
     virtual bool hasCorrectStatistics() const = 0;
 
+    /**
+     * Get metrics of the reader
+     * @return the accumulated reader metrics to current state.
+     */
+    virtual ReaderMetrics getReaderMetrics() const = 0;
```
wgtmac, line 485: const ReaderMetrics& ?

c++/include/orc/Reader.hh
```diff
@@ -39,6 +39,29 @@ namespace orc {
   struct ReaderOptionsPrivate;
   struct RowReaderOptionsPrivate;
 
+  struct ReaderMetrics {
```
wgtmac, line 42: Please add comment to help understanding.

c++/include/orc/Reader.hh
```diff
@@ -39,6 +39,29 @@ namespace orc {
...
+
+    ReaderMetrics() {
+      ReaderCount = 0;
+      ReaderInclusiveLatencyUs = 0;
+      DecompCount = 0;
+      DecompLatencyUs = 0;
+      DecodingCount = 0;
+      DecodingLatencyUs = 0;
+      ByteDecodingCount = 0;
+      ByteDecodingLatencyUs = 0;
```
wgtmac, line 60: Can we support blocking io latency as well?

c++/src/Utils.hh
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef ORC_UTILS_HH
+#define ORC_UTILS_HH
+
+#include <stdint.h>
+#include <time.h>
+#include <sys/time.h>
```
wgtmac, line 24: Windows may fail with this inclusion. You may want to use std chrono library. See : https://en.cppreference.com/w/cpp/chrono/high_resolution_clock/now

c++/src/Utils.hh
```diff
@@ -0,0 +1,61 @@
...
+                  volatile uint64_t* cnt) {
+        start = getCurrentTimeUs();
+        latencyUs = latency;
+        count = cnt;
+    }
+
+    ~AutoStopwatch() {
+        if (latencyUs) {
+            uint64_t elapsedTime = getCurrentTimeUs() - start;
+             __sync_fetch_and_add(latencyUs, elapsedTime);
```
wgtmac, line 51: Windows does not provide this API, please use std::atomic instead.

c++/src/Reader.cc
```diff
@@ -1154,6 +1165,8 @@ namespace orc {
   }
 
   bool RowReaderImpl::next(ColumnVectorBatch& data) {
+    AutoStopwatch measure(&contents->readerMetrics->ReaderInclusiveLatencyUs,
+                  &contents->readerMetrics->ReaderCount);
```
wgtmac, line 1169: Fix alignment

tools/src/ToolsHelper.cc
```diff
@@ -80,10 +83,25 @@ bool parseOptions(int* argc, char** argv[], uint64_t* batchSize, orc::RowReaderO
...
     }
   } while (opt != -1);
   *argc -= optind;
   *argv += optind;
   return true;
 }
+
+void printReaderMetrics(std::ostream& out, const orc::ReaderMetrics& metrics) {
+  out << "ReaderCount: " << metrics.ReaderCount << std::endl;
+  out << "ElapsedTimeUs: " << metrics.ReaderInclusiveLatencyUs << std::endl;
```
wgtmac, line 100: nit: user may expect to read time in seconds which is much easier to interpret.

wgtmac:
I have requested @stiga-huang for review because he has opened the JIRA.

c++/include/orc/Reader.hh
```diff
@@ -39,6 +39,29 @@ namespace orc {
   struct ReaderOptionsPrivate;
   struct RowReaderOptionsPrivate;
 
+  struct ReaderMetrics {
+    volatile uint64_t ReaderCount;
```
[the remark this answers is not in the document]
coderex2522: Done

c++/include/orc/Reader.hh
coderex2522: has removed the explicit constructor.

c++/include/orc/Reader.hh
coderex2522: Done

c++/include/orc/Reader.hh
coderex2522: I add the IO latency and count in the OrcFile.cc and OrcHdfsFile.cc

c++/src/Utils.hh
coderex2522: Done

dongjoon-hyun (commented):
FYI, this PR still fails on all MacOSes, @coderex2522 .

c++/src/Compression.cc
```diff
@@ -633,9 +642,10 @@ DIAGNOSTIC_PUSH
   ZlibDecompressionStream::ZlibDecompressionStream
                    (std::unique_ptr<SeekableInputStream> inStream,
                     size_t bufferSize,
-                    MemoryPool& _pool
+                    MemoryPool& _pool,
+                    ReaderMetrics& metrics
```
wgtmac, line 646: Rename the argument to _metrics, this is why the MasOS build fails. Same for constructors below.

> [ 48%] Building CXX object c++/src/CMakeFiles/orc.dir/Compression.cc.o
/Users/runner/work/orc/orc/c++/src/Compression.cc:646:36: error: parameter 'metrics' shadows member inherited from type 'DecompressionStream' [-Werror,-Wshadow-field]
                    ReaderMetrics& metrics
                                   ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:396:20: note: declared here
    ReaderMetrics& metrics;
                   ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:767:36: error: parameter 'metrics' shadows member inherited from type 'DecompressionStream' [-Werror,-Wshadow-field]
                    ReaderMetrics& metrics
                                   ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:396:20: note: declared here
    ReaderMetrics& metrics;
                   ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:816:46: error: parameter 'metrics' shadows member inherited from type 'DecompressionStream' [-Werror,-Wshadow-field]
                              ReaderMetrics& metrics
                                             ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:396:20: note: declared here
    ReaderMetrics& metrics;
                   ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:861:43: error: parameter 'metrics' shadows member inherited from type 'DecompressionStream' [-Werror,-Wshadow-field]
                           ReaderMetrics& metrics
                                          ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:396:20: note: declared here
    ReaderMetrics& metrics;
                   ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:895:43: error: parameter 'metrics' shadows member inherited from type 'DecompressionStream' [-Werror,-Wshadow-field]
                           ReaderMetrics& metrics
                                          ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:396:20: note: declared here
    ReaderMetrics& metrics;
                   ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:[1201](https://github.com/apache/orc/runs/6884140918?check_suite_focus=true#step:5:1202):44: error: parameter 'metrics' shadows member inherited from type 'DecompressionStream' [-Werror,-Wshadow-field]
                            ReaderMetrics& metrics)
                                           ^
/Users/runner/work/orc/orc/c++/src/Compression.cc:396:20: note: declared here
    ReaderMetrics& metrics;
                   ^
6 errors generated.
make[2]: *** [c++/src/CMakeFiles/orc.dir/Compression.cc.o] Error 1
make: *** [all] Error 2

coderex2522:
> FYI, this PR still fails on all MacOSes, @coderex2522 .

fixed.

c++/src/Utils.hh
```diff
@@ -0,0 +1,58 @@
...
+#ifndef ORC_UTILS_HH
+#define ORC_UTILS_HH
+
+#include <atomic>
+#include <chrono>
+
+namespace orc {
+
+class AutoStopwatch {
+    std::chrono::high_resolution_clock::time_point start;
```
stiga-huang, line 28: This file uses 4 spaces indention. I think our code style uses 2 spaces.

stiga-huang:
Thanks for adding these metrics! Can we also expose the stats of `SargsApplier`?
https://github.com/apache/orc/blob/afc671d214020116cb5d6cde8c3b2dfc945295c5/c%2B%2B/src/sargs/SargsApplier.hh#L122-L123

c++/src/Utils.hh
coderex2522: Yes, I fixed the Utils.hh file format.

coderex2522:
> Thanks for adding these metrics! Can we also expose the stats of `SargsApplier`?
> 
> https://github.com/apache/orc/blob/afc671d214020116cb5d6cde8c3b2dfc945295c5/c%2B%2B/src/sargs/SargsApplier.hh#L122-L123

Sure, I add the SelectedRowGroupCount and EvaluatedRowGroupCount to expose the PPD metrics.

tools/src/ToolsHelper.cc
coderex2522: The output of time metrics is uniformly changed to be in seconds.

c++/src/ByteRLE.cc
```diff
@@ -433,6 +441,13 @@ namespace orc {
 
   void ByteRleDecoderImpl::next(char* data, uint64_t numValues,
                                 char* notNull) {
+    AutoStopwatch meassure(&metrics.ByteDecodingLatencyUs,
```
stiga-huang, line 444: This is not the pure decoding time since it contains reading buffers from the input stream. So it's actually ioTime + decompressionTime + decodingTime. I think we should either fix it or document it.

c++/include/orc/Reader.hh
```diff
@@ -32,13 +32,34 @@
...
 
 namespace orc {
 
   // classes that hold data members so we can maintain binary compatibility
   struct ReaderOptionsPrivate;
   struct RowReaderOptionsPrivate;
 
+  /**
+   * Expose the reader metrics including the latency and
+   * number of calls of the decompression/decoding/IO modules.
```
stiga-huang, line 45: Please document the relation of the latency counters. IIUC, ReaderInclusiveLatencyUs contains the time of decoding latency. Decoding latency contains the time of decompression latency which finally contains the time of IO blocking.

c++/include/orc/OrcFile.hh
```diff
@@ -105,20 +105,26 @@ namespace orc {
   /**
    * Create a stream to a local file or HDFS file if path begins with "hdfs://"
    * @param path the name of the file in the local file system or HDFS
+   * @param metrics the metrics of the reader
    */
-  ORC_UNIQUE_PTR<InputStream> readFile(const std::string& path);
+  ORC_UNIQUE_PTR<InputStream> readFile(const std::string& path,
+                                       ReaderMetrics& metrics);
```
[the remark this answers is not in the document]
coderex2522: Parameter references for the metrics do cause compatibility issues. I replaced the all parameter reference with metrics pointer. And use macro DEFINE_AUTO_STOPWATCH  to do the metrics pointer null checks.

c++/include/orc/Reader.hh
coderex2522: Done. I added some comments to the metrics latency fields.

c++/src/Utils.hh
```diff
@@ -0,0 +1,63 @@
...
+      latencyUs->fetch_add(static_cast<uint64_t>(elapsedTime.count()));
+    }
+
+    if (count) {
+      count->fetch_add(1);
+    }
+  }
+};
+
+#define DEFINE_AUTO_STOPWATCH(METRICS_PTR, LATENCY_VAR, COUNT_VAR)    \
```
wgtmac, line 56: I'd suggest to rename DEFINE_AUTO_STOPWATCH to SCOPED_STOPWATCH because usually DEFINE_ prefix is used to define a class or something via marco expansion.

c++/src/sargs/SargsApplier.cc
```diff
@@ -126,10 +127,14 @@ namespace orc {
...
     // update stats
-    mStats.first = std::accumulate(
-      mNextSkippedRows.cbegin(), mNextSkippedRows.cend(), mStats.first,
-      [](bool rg, uint64_t s) { return rg ? 1 : 0 + s; });
-    mStats.second += groupsInStripe;
+    uint64_t selectedRGs = std::accumulate(
+      mNextSkippedRows.cbegin(), mNextSkippedRows.cend(), 0UL,
+      [](uint64_t initVal, uint64_t rg) {
+        return rg > 0 ? initVal + 1 : initVal; });
+    if (mMetrics != nullptr) {
```
wgtmac, line 134: What if the file or stripe has been skipped and no row group evaluation happens here? We may have missed them.

c++/src/Compression.cc
```diff
@@ -463,6 +469,7 @@ DIAGNOSTIC_PUSH
   }
 
   bool DecompressionStream::Next(const void** data, int*size) {
+    DEFINE_AUTO_STOPWATCH(metrics, DecompLatencyUs, DecompCount);
```
wgtmac, line 472: The stop watch should not be put here since it includes the io latency. If you move this line to line 498 right before the definition of **size_t availableSize**, what we got is the pure decompression time.

c++/src/RleDecoderV2.cc
```diff
@@ -432,6 +435,7 @@ void RleDecoderV2::skip(uint64_t numValues) {
 void RleDecoderV2::next(int64_t* const data,
                         const uint64_t numValues,
                         const char* const notNull) {
+  DEFINE_AUTO_STOPWATCH(metrics, DecodingLatencyUs, DecodingCount);
```
wgtmac, line 438: Where is the stop watch call in the RleDecoderV1 ?

wgtmac: In addition, all IOs happen in the call of readByte(). We probably should make metrics a member varialbe of the RleDecoder and counts its io time in readByte(). Finally, we can exlucde the io time from decoding latancy here to get the pure RLE decoding time.

c++/include/orc/Reader.hh
```diff
@@ -32,13 +32,40 @@
...
+  struct ReaderMetrics {
+    std::atomic<uint64_t> ReaderCount{0};
+    // ReaderInclusiveLatencyUs contains the latency of
+    // the decompression/decoding/IO modules.
+    std::atomic<uint64_t> ReaderInclusiveLatencyUs{0};
+    std::atomic<uint64_t> DecompCount{0};
+    // DecompLatencyUs contains the IO blocking latency.
+    std::atomic<uint64_t> DecompLatencyUs{0};
+    std::atomic<uint64_t> DecodingCount{0};
+    // DecodingLatencyUs contains the latency of the decompression/IO modules.
```
wgtmac, line 56: I think we can do better here to get the pure decoding time. Please see the inline comments below.

c++/include/orc/Reader.hh
```diff
@@ -32,13 +32,40 @@
...
+  /**
+   * Expose the reader metrics including the latency and
+   * number of calls of the decompression/decoding/IO modules.
+   */
+  struct ReaderMetrics {
+    std::atomic<uint64_t> ReaderCount{0};
+    // ReaderInclusiveLatencyUs contains the latency of
+    // the decompression/decoding/IO modules.
+    std::atomic<uint64_t> ReaderInclusiveLatencyUs{0};
+    std::atomic<uint64_t> DecompCount{0};
```
wgtmac, line 52: I'd prefer to use the full name here. And maybe replace **Count** with **Call** ?

wgtmac (changes requested):

c++/include/orc/Reader.hh
coderex2522: The DecompCount and DecompLatencyUs have been replaced with DecompressionCall and DecompressionLatencyUs.

c++/src/Utils.hh
coderex2522: Ok, the macro DEFINE_AUTO_STOPWATCH was replaced with SCOPED_STOPWATCH.

c++/src/Compression.cc
coderex2522: The NextDecompress function will call the readBuffer function which includes the io operation. Currently the pure decompression time is got by subtracting the time taken by the readBuffer function.

c++/src/RleDecoderV2.cc
coderex2522: Currently the pure decoding time is got by subtracting the time taken by the readByte() function.

c++/src/sargs/SargsApplier.cc
```diff
@@ -173,16 +174,23 @@ namespace orc {
...
-  bool SargsApplier::evaluateFileStatistics(const proto::Footer& footer) {
+  bool SargsApplier::evaluateFileStatistics(const proto::Footer& footer,
+                                            uint64_t fileRowGroupCount) {
     if (!mHasEvaluatedFileStats) {
       if (footer.statistics_size() == 0) {
         mFileStatsEvalResult = true;
       } else {
         mFileStatsEvalResult = evaluateColumnStatistics(footer.statistics());
+        if (!mFileStatsEvalResult && mMetrics != nullptr) {
+          mMetrics->EvaluatedRowGroupCount.fetch_add(fileRowGroupCount);
```
wgtmac, line 192: The unit of input split is stripe. If the current input split does not contain all stripes in the file, it is incorrect to count row groups of the whole file.

c++/src/Reader.cc
```diff
@@ -286,6 +287,10 @@ namespace orc {
           lastStripe = i + 1;
         }
       }
+      if (footer->rowindexstride() > 0) {
+        fileRowGroupCount +=
```
wgtmac, line 291: I'd suggest to rename fileRowGroupCount to something like numSplitRowGroups,  numRequestedRowGroups or numRowGroupsInRange. Then refine it to count stripe in range only.

wgtmac (commented):
Thanks for the update. Please see my new comments.

c++/src/Reader.cc
coderex2522: Done. The fileRowGroupCount is replaced with numRowGroupsInStripeRange.

wgtmac (approved):
+1 LGTM. Thanks @coderex2522 !

c++/src/Compression.hh
```diff
@@ -30,12 +30,14 @@ namespace orc {
...
    * @param pool the memory pool
+   * @param metrics the reader metrics
    */
   std::unique_ptr<SeekableInputStream>
      createDecompressor(CompressionKind kind,
                         std::unique_ptr<SeekableInputStream> input,
                         uint64_t bufferSize,
-                        MemoryPool& pool);
+                        MemoryPool& pool,
+                        ReaderMetrics* metrics);
```
dongjoon-hyun, line 40: If you don't mind, shall we have the default value, `ReaderMetrics* metrics = nullptr`, to be more backward-compatible?

dongjoon-hyun (commented):
Thank you, @coderex2522 and @wgtmac .

I have three questions as the final pieces:
- Backward-compatibility: Apache ORC C++ API is used in the downstream. So, although this PR looks good, it would be great to make it sure that we don't break the backward compatibility once more. I believe the default value of new parameters could reduce the pain.
- The ability to disable this metrics. According to the code, we can disable this feature by hand over NULL pointer? Could you double-confirm it? Could you add a test coverage to confirm that this case is possible? It will be helpful to prevent accidentally segfault in the future.
- Last question is about the target version. For this PR, if you want, you can backport this to branch-1.8 for Apache ORC 1.8.0. Which one do you prefer v1.9.0 (Next year) or v1.8.0 (This year).

dongjoon-hyun:
Gentle ping, @coderex2522 .

wgtmac:
> Thank you, @coderex2522 and @wgtmac .
> 
> I have three questions as the final pieces:
> 
> * Backward-compatibility: Apache ORC C++ API is used in the downstream. So, although this PR looks good, it would be great to make it sure that we don't break the backward compatibility once more. I believe the default value of new parameters could reduce the pain.
> * The ability to disable this metrics. According to the code, we can disable this feature by hand over NULL pointer? Could you double-confirm it? Could you add a test coverage to confirm that this case is possible? It will be helpful to prevent accidentally segfault in the future.
> * Last question is about the target version. For this PR, if you want, you can backport this to branch-1.8 for Apache ORC 1.8.0. Which one do you prefer v1.9.0 (Next year) or v1.8.0 (This year).

Good suggestion @dongjoon-hyun.

We should be able to set nullptr to ReaderMetrics to disable it at run time. BTW, we can totally disable it at compile time if we have something like this: https://github.com/apache/impala/blob/master/be/src/util/runtime-profile-counters.h#L53

dongjoon-hyun:
Thank you,  @wgtmac

coderex2522:
> Thank you, @coderex2522 and @wgtmac .
> 
> I have three questions as the final pieces:
> 
> * Backward-compatibility: Apache ORC C++ API is used in the downstream. So, although this PR looks good, it would be great to make it sure that we don't break the backward compatibility once more. I believe the default value of new parameters could reduce the pain.
> * The ability to disable this metrics. According to the code, we can disable this feature by hand over NULL pointer? Could you double-confirm it? Could you add a test coverage to confirm that this case is possible? It will be helpful to prevent accidentally segfault in the future.
> * Last question is about the target version. For this PR, if you want, you can backport this to branch-1.8 for Apache ORC 1.8.0. Which one do you prefer v1.9.0 (Next year) or v1.8.0 (This year).

* External projects will depend on the api interface in the include directory and need to maintain compatibility, while the header files in the src directory will not be dependent on external projects, so I prefer not to add the default value.
* I will add the new test to coverage the disable metrics case.
* The target version v1.9.0 is fine, I have no requirement to support in branch-1.8.

wgtmac (approved):
LGTM

dongjoon-hyun (approved):
+1, LGTM. Thank you, @coderex2522 , @wgtmac , @stiga-huang .
Merged to main for Apache Orc 1.9.0.

XinyuZeng:
It seems like this feature is enabled by default and will bring extra overhead to the decoding performance, as I tested via orc-scan. Could this feature be disabled by default?

dongjoon-hyun:
Thank you for suggestion, @XinyuZeng . Yes, it was also discussed before https://github.com/apache/orc/pull/1158#pullrequestreview-1031811661 .

Could you open a new official JIRA for that? We can add that configuration for that at Apache ORC 1.9.0 too.

dongjoon-hyun:
BTW, please include the reproducible step for the following observation, @XinyuZeng .
> reading time increased by more than 2x with orc-scan on my machine.

XinyuZeng:
@dongjoon-hyun I created jira-issue 1232 for this problem. Also, even the metrics is disabled by setting nullptr, there is still some cost of initing the AutoStopWatch object and destructing it. The following perf flamegraph shows the cost.
image

dongjoon-hyun:
Thank you for the detail, @coderex2522 . Could you make a PR to provide an opt-out way to achieve your goal?

coderex2522:
> @dongjoon-hyun I created jira-issue 1232 for this problem. Also, even the metrics is disabled by setting nullptr, there is still some cost of initing the AutoStopWatch object and destructing it. The following perf flamegraph shows the cost. Although the extra time may be trivial, it may be better to be disabled at compile time.


I am confused about why does the chrono time function call still appear in the flamegraph after disabling the metrics.

XinyuZeng:
> > @dongjoon-hyun I created jira-issue 1232 for this problem. Also, even the metrics is disabled by setting nullptr, there is still some cost of initing the AutoStopWatch object and destructing it. The following perf flamegraph shows the cost. Although the extra time may be trivial, it may be better to be disabled at compile time.
> 
> I am confused about why does the chrono time function call still appear in the flamegraph after disabling the metrics.

The SCOPED_STOPWATCH and SCOPED_MINUS_STOPWATCH macros still create the AutoStopwatch object which has chrono members, even after we pass in null METRICS_PTR.

dongjoon-hyun:
Since Apache ORC 1.9.0 is planned on September 2023, I believe we have enough time to remove the regression before release. Please feel free to open a PR for the improvement. Everyone can make a PR and in the worst case (if there is no volunteer to fix the regression), we can revert ORC-961 from `branch-1.9` before creating 1.9.0 RC1 vote.
- https://github.com/apache/orc/milestone/10

BTW, please let us know your opinions about the above approach, @wgtmac , @williamhyun , @guiyanakuang , @stiga-huang .

coderex2522:
> 

I submitted [ORC-1232 PR](https://github.com/apache/orc/pull/1206) in order to diable the metrics collector at compile phase. If there is no requirement for metrics collection, users do not have to turn on BUILD_CPP_ENABLE_METRICS.

PR #1161, opened 2022-06-16
ORC-1204. ORC MapReduce writer to flush when long arrays

omalley:

This adds a new configuration for the row by row writer that sets a limit of how many items will be buffered in the VectorizedRowBatch before it is sent to the ORC Writer. This does not change the behavior of the core writer, where the application already has control over the size of the batch.

We are getting OOM when writing rows with long arrays with the row-by-row writer. Heap dumps show that it is the ColumnVector inside the array that is taking large amounts of memory.

Updated a couple unit tests and added two more.

### DIFF
--- a/java/mapreduce/src/java/org/apache/orc/mapred/OrcOutputFormat.java
+++ b/java/mapreduce/src/java/org/apache/orc/mapred/OrcOutputFormat.java
@@ -75,6 +75,8 @@ public class OrcOutputFormat<V extends Writable>
     Path path = getTaskOutputPath(conf, name);
     Writer writer = OrcFile.createWriter(path,
         buildOptions(conf).fileSystem(fileSystem));
-    return new OrcMapredRecordWriter<>(writer);
+    return new OrcMapredRecordWriter<>(writer,
+        OrcConf.ROW_BATCH_SIZE.getInt(conf),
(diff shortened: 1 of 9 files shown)


omalley:
32k was based on the 1k rows and wanting to let lists have ~30 elements/row before we cause an early flush.

dongjoon-hyun:
Thank you! Merged to main.

java/mapreduce/src/java/org/apache/orc/mapred/OrcOutputFormat.java
```diff
@@ -75,6 +75,8 @@ public RecordWriter<NullWritable, V> getRecordWriter(FileSystem fileSystem,
     Path path = getTaskOutputPath(conf, name);
     Writer writer = OrcFile.createWriter(path,
         buildOptions(conf).fileSystem(fileSystem));
-    return new OrcMapredRecordWriter<>(writer);
+    return new OrcMapredRecordWriter<>(writer,
+        OrcConf.ROW_BATCH_SIZE.getInt(conf),
```
cxzl25, line 79: https://github.com/apache/orc/pull/1108#discussion_r869544673

Should we split into two configurations? In scenarios with read and write orc, the same configuration should not be mixed.

PR #1164, opened 2022-06-22
ORC-1205: Size of batches in some ConvertTreeReaders should be ensured before using

abstractdog:
This fix introduces a new API for TreeReaders: ensureSize, which is supposed to be called by TreeReader clients before making a call for reading a new batch.


dongjoon-hyun (changes requested):
Thank you for making a PR but could you open a PR to `main` first?
To prevent a regression in the future release, Apache projects have a backporting rule.
We land at `main` branch first and backport to `branch-1.8`, then `branch-1.7`, and so on.

abstractdog:
makes sense @dongjoon-hyun , I'm porting the change to main soon

PR #1166, opened 2022-06-23
ORC-1205: `nextVector` should invoke `ensureSize` when reusing vectors

abstractdog:
The fix simply ensures the size of some batches in ConvertTreeReader instances without API change.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
@@ -464,6 +464,8 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         // Allocate column vector for file; cast column vector for reader.
         doubleColVector = new DoubleColumnVector(batchSize);
         longColVector = (LongColumnVector) previousVector;
+      } else {
+        doubleColVector.ensureSize(batchSize, false);
       }
       // Read present/isNull stream
       fromReader.nextVector(doubleColVector, isNull, batchSize, filterContext, readPhase);
@@ -550,6 +552,8 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         // Allocate column vector for file; cast column vector for reader.
         decimalColVector = new DecimalColumnVector(batchSize, precision, scale);
         longColVector = (LongColumnVector) previousVector;
+      } else {
+        decimalColVector.ensureSize(batchSize, false);
       }
       // Read present/isNull stream
       fromReader.nextVector(decimalColVector, isNull, batchSize, filterContext, readPhase);
@@ -591,6 +595,8 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         // Allocate column vector for file; cast column vector for reader.
         bytesColVector = new BytesColumnVector(batchSize);
         longColVector = (LongColumnVector) previousVector;
+      } else {
+        bytesColVector.ensureSize(batchSize, false);
       }
       // Read present/isNull stream
       fromReader.nextVector(bytesColVector, isNull, batchSize, filterContext, readPhase);
@@ -628,6 +634,8 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         // Allocate column vector for file; cast column vector for reader.
         timestampColVector = new TimestampColumnVector(batchSize);
         longColVector = (LongColumnVector) previousVector;
+      } else {
+        timestampColVector.ensureSize(batchSize, false);
       }
       // Read present/isNull stream
       fromReader.nextVector(timestampColVector, isNull, batchSize, filterContext, readPhase);
@@ -668,6 +676,8 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         // Allocate column vector for file; cast column vector for reader.
         longColVector = new LongColumnVector(batchSize);
         doubleColVector = (DoubleColumnVector) previousVector;
+      } else {
+        longColVector.ensureSize(batchSize, false);
       }
       // Read present/isNull stream
       fromReader.nextVector(longColVector, isNull, batchSize, filterContext, readPhase);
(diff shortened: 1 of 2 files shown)


dongjoon-hyun:
BTW, @abstractdog . Which ORC version are you using now?
- The latest one is https://orc.apache.org/news/2022/06/16/ORC-1.7.5/ 
- 1.6.x will be EOL soon.

guiyanakuang:
Thanks to @abstractdog  for the extensive test coverage. I verified that the patch works.

The current reader inheritance relationship looks like this

```
TypeReader +--> TreeReader +---> ConvertTreeReader +-> Various ConvertTreeReaderImpl
                    +
                    |
                    +--------------------------------> Various TreeReaderImpl
```
The new ensureSize interface is called recursively to ensure the size of the vector, which is a good solution to the out-of-bounds problem in Various ConvertTreeReaderImpl. However, in the Various TreeReaderImpl code path we had the ensureSize method being called repeatedly. I think this is a good refactoring design and I wonder if we should simply fix the out-of-bounds issue in the ConvertTreeReader first. Then refactor the ensureSize entry in another pr to be consistent.

Please let me know if I am not correct.

abstractdog:
thanks for the comments so far!

@dongjoon-hyun: hive is on ORC 1.6.9, so I'm motivated to backport the patch to branch-1.6 too if it can make its way to at least a single release, but if it becomes EOL, I don't mind, we're fixing this on 1.7.x too, hive should upgrade eventually anyway :)

@guiyanakuang: can you please elaborate on this:
```
in the Various TreeReaderImpl code path we had the ensureSize method being called repeatedly
```
it's not clear where exactly this happens, and why it is a problem, but otherwise, I'm fine with fixing this on ConvertTreeReaders only: so did you mean that the ensureSize method can stay exposed in TypeReader interface (as it is in the current patch), and let it be a no-op in TreeReaderFactory (so ConvertTreeReader will implement it instead of calling super.ensureSize)?

guiyanakuang:
What I mean is that this pr is more like a refactoring than a problem fix.

The ensureSize interface is defined in the TypeReader, and if this pr is treated as a refactoring, our modification looks only half done. There are still many temporary reuse vector calls to ensureSize in Various TreeReaderImpl, rather than using the new interface to control size (e.g. line 2131).

https://github.com/apache/orc/blob/3eac56761f48f3e315eea56dc63de9f2cbadcdbb/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java#L2126-L2134

So I think it would be better to use a separate pr to abstract the code and use the public ensureSize interface (both ConvertTreeReaderImpl and TreeReaderImpl use this interface to control the size of the vectors).

guiyanakuang:
> I see, makes sense @guiyanakuang as I'm now interested in fixing the issue ASAP for ConvertTreeReaders, the best way is what you proposed in Jira, right?
> 
> ```
>     if (doubleColVector == null) {
>         // Allocate column vector for file; cast column vector for reader.
>         doubleColVector = new DoubleColumnVector(batchSize);
>         longColVector = (LongColumnVector) previousVector;
>     }
> +   else {
> +       doubleColVector.ensureSize(batchSize, false);
> +   }
> ```
> 
> without touching the interface
> 
> if so, let me refresh the PR soon with the fix

Yes I do think so.

abstractdog:
thanks @guiyanakuang, just force pushed the fix

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -1764,6 +1812,9 @@ public void nextVector(ColumnVector previousVector,
                                                    " proleptic Gregorian dates.");
           }
         }
+      } else {
+        bytesColVector.ensureSize(batchSize, false);
+        dateColumnVector.ensureSize(batchSize, false);
```
guiyanakuang, line 1817: `dateColumnVector.ensureSize(batchSize, false);` may throw a null pointer exception and we may need to check additionally.

guiyanakuang: `dateColumnVector` can be ignored. This is because it is the vector obtained by forcing `previousVector` .

https://github.com/apache/orc/blob/3eac56761f48f3e315eea56dc63de9f2cbadcdbb/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java#L1423

`computeBatchSize` ensures that batchSize <= the size of the input vector.

abstractdog: handled it, thanks!

guiyanakuang: > `dateColumnVector` can be ignored. This is because it is the vector obtained by forcing `previousVector` .
> 
> https://github.com/apache/orc/blob/3eac56761f48f3e315eea56dc63de9f2cbadcdbb/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java#L1423
> 
> `computeBatchSize` ensures that batchSize <= the size of the input vector.

@abstractdog, in addition to this, LGTM

abstractdog: okay, removed
thanks a lot for the quick reviews @guiyanakuang !

abstractdog:
thanks for your comments and approving this PR @guiyanakuang 
if patch looks good, I'm about to open some PR with the very same contents, I'll update jira

dongjoon-hyun:
In addition, 
- Please revise the PR description. The current one is invalid.
> This fix introduces a new API for TreeReaders: ensureSize

- After revising the PR description, please copy and paste it into your new PRs. PR title and description become a commit title and message. So, the following is an invalid PR description in your backporting PRs in Apache ORC community.
> same patch as main

abstractdog:
finally, this doesn't introduce any APIs, updating descriptions

dongjoon-hyun (approved):
+1, LGTM. Thank you so much, @abstractdog and @guiyanakuang !

PR #1167, opened 2022-06-24
ORC-1205: `nextVector` should invoke `ensureSize` when reusing vectors

abstractdog:
The fix simply ensures the size of some batches in ConvertTreeReader instances without API change.


dongjoon-hyun (changes requested):
Please see
- https://github.com/apache/orc/pull/1166#issuecomment-1165879657

dongjoon-hyun (approved):
+1, LGTM. Thank you, @abstractdog .
Merged to branch-1.8 for Apache ORC 1.8.0.

PR #1169, opened 2022-06-24
ORC-1205: Size of batches in some ConvertTreeReaders should be ensured before using

abstractdog:


dongjoon-hyun (changes requested):
Please see
- https://github.com/apache/orc/pull/1166#issuecomment-1165872257

dongjoon-hyun:
For now, I'll close this PR.

PR #1177, opened 2022-07-17
ORC-1215: Remove a wrong `NotNull` annotation on `value` of `setAttribute`

williamhyun:
This PR aims to remove a wrong `NotNull` annotation on `value`. 

According to ORC-529, attribute value can be null to clear the existing value.

https://github.com/apache/orc/blob/8eef0c495ffe334825655ec22c1516b24ecd9f48/java/core/src/java/org/apache/orc/TypeDescription.java#L245-L249

Pass the CIs with the newly added test case.


dongjoon-hyun (approved):
+1, LGTM. Thank you, @williamhyun .

dongjoon-hyun:
Merged to main/1.8/1.7.

PR #1182, opened 2022-07-19
ORC-1217: Downgrade `org.jetbrains.annotations` to 17.0.0

williamhyun:
This PR aims to downgrade `org.jetbrains.annotations` to 17.0.0.

ORC-916, ORC-956, and ORC-1056 bumped `annotations`, but we only use `@NotNull` so we do not need to update the `annotations` version. 

Pass the CIs.

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -823,7 +823,7 @@
       <dependency>
         <groupId>org.jetbrains</groupId>
         <artifactId>annotations</artifactId>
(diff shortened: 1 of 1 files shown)


williamhyun:
cc: @dongjoon-hyun

dongjoon-hyun (approved):
+1, I agree with you. Thank you, @williamhyun . LGTM.
Merged to main/1.8.

PR #1196, opened 2022-07-26
ORC-1228: Fix `setAttribute` to handle null value

williamhyun:
This PR aims to fix `setAttribute` to handle null value.

In case of clearing the value with null, we should remove the corresponding key also to save memory. 

Pass the CIs with the new test case.


williamhyun:
cc: @dongjoon-hyun

dongjoon-hyun (approved):
+1, LGTM. I agree with you that the new behavior is much better.

dongjoon-hyun:
Feel free to merge and backport, @williamhyun . You are the release manager of Apache ORC 1.7.6.

PR #1223, opened 2022-08-17
ORC-1252:[C++] Expose IO metrics for write operation

coderex2522:
The pull request will add IO metrics for the write operation. The csv-import tool can display relevant IO metrics.

This patch can expose the metrics information related to the process of writing ORC files.

Test IO metrics in BufferedOutputStream's UTs. And the csv-import tool with option -m  can be used for testing.

### DIFF
--- a/c++/include/orc/Writer.hh
+++ b/c++/include/orc/Writer.hh
@@ -24,11 +24,11 @@
 #include "orc/Type.hh"
 #include "orc/Vector.hh"
 
+#include <atomic>
 #include <memory>
 #include <set>
 #include <string>
 #include <vector>
-#include <atomic>
 
 namespace orc {
 
@@ -51,7 +51,9 @@ namespace orc {
    * Expose the IO metrics for write operation.
    */
   struct WriterMetrics {
+    // Record the number of IO requests written to the output file
     std::atomic<uint64_t> IOCount{0};
+    // Record the lantency of IO blocking
     std::atomic<uint64_t> IOBlockingLatencyUs{0};
   };
   /**
--- a/c++/include/orc/Writer.hh
+++ b/c++/include/orc/Writer.hh
@@ -28,6 +28,7 @@
 #include <set>
 #include <string>
 #include <vector>
+#include <atomic>
 
 namespace orc {
 
--- a/c++/include/orc/Writer.hh
+++ b/c++/include/orc/Writer.hh
@@ -47,6 +47,13 @@ namespace orc {
   class Timezone;
 
   /**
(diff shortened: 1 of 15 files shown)


dongjoon-hyun (commented):
Thank you, @coderex2522 . Do we need to consider the perf regression in this PR?

cc @wgtmac , @stiga-huang , too.

coderex2522:
> * Do we need to consider the perf regression in this PR?

IO metrics for write operation reuse the macro definition SCOPED_STOPWATCH. By default the build option BUILD_CPP_ENABLE_METRICS is off, so there is no perf regression here.

c++/include/orc/Writer.hh
```diff
@@ -28,6 +28,7 @@
 #include <set>
 #include <string>
 #include <vector>
+#include <atomic>
```
wgtmac, line 31: please sort alphabetically.

c++/src/OrcFile.cc
```diff
@@ -161,6 +163,7 @@ DIAGNOSTIC_POP
     }
 
     void write(const void* buf, size_t length) override {
+      SCOPED_STOPWATCH(metrics, IOBlockingLatencyUs, IOCount);
```
wgtmac, line 166: This may not be the right place. We should measure where it is called. Otherwise we cannot support HDFS and other file systems.

c++/include/orc/Writer.hh
```diff
@@ -46,6 +47,13 @@ namespace orc {
 
   class Timezone;
 
+  /**
+   * Expose the IO metrics for write operation.
+   */
+  struct WriterMetrics {
+    std::atomic<uint64_t> IOCount{0};
```
wgtmac, line 54: Add some comment to explain them

c++/include/orc/Writer.hh
coderex2522: done

c++/src/OrcFile.cc
coderex2522: To support all file systems, the macro SCOPED_STOPWATCH for IO metrics is place in class BufferedOutputStream(OutputStream.cc).

c++/include/orc/Writer.hh
coderex2522: done

wgtmac (approved):
LGTM +1. Thanks @coderex2522

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Thank you, @coderex2522 and @wgtmac !

PR #1244, opened 2022-09-09
ORC-1266: DecimalColumnVector resets the isRepeating flag in the nextVector method

cxzl25:
`DecimalColumnVector` resets the `isRepeating` flag in the `nextVector` method.

`DecimalColumnVector` does not set `isRepeating` when reading decimal data, which leads to the use of type promotion to read decimal, and the obtained data may be wrong.

add UT

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
@@ -554,6 +554,7 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         longColVector = (LongColumnVector) previousVector;
       } else {
         decimalColVector.ensureSize(batchSize, false);
+        decimalColVector.reset();
       }
       // Read present/isNull stream
       fromReader.nextVector(decimalColVector, isNull, batchSize, filterContext, readPhase);
@@ -718,6 +719,7 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         doubleColVector = (DoubleColumnVector) previousVector;
       } else {
         decimalColVector.ensureSize(batchSize, false);
+        decimalColVector.reset();
       }
       // Read present/isNull stream
       fromReader.nextVector(decimalColVector, isNull, batchSize, filterContext, readPhase);
@@ -1061,6 +1063,7 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         decimalColVector = previousVector;
       } else {
         fileDecimalColVector.ensureSize(batchSize, false);
+        fileDecimalColVector.reset();
       }
       // Read present/isNull stream
       fromReader.nextVector(fileDecimalColVector, isNull, batchSize, filterContext, readPhase);
@@ -1215,6 +1218,7 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         bytesColVector = (BytesColumnVector) previousVector;
       } else {
         decimalColVector.ensureSize(batchSize, false);
+        decimalColVector.reset();
       }
       // Read present/isNull stream
       fromReader.nextVector(decimalColVector, isNull, batchSize, filterContext, readPhase);
@@ -1652,6 +1656,7 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         timestampColVector = (TimestampColumnVector) previousVector;
       } else {
         decimalColVector.ensureSize(batchSize, false);
+        decimalColVector.reset();
       }
       timestampColVector.changeCalendar(fileUsedProlepticGregorian, false);
       // Read present/isNull stream
--- a/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
@@ -554,7 +554,6 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         longColVector = (LongColumnVector) previousVector;
       } else {
         decimalColVector.ensureSize(batchSize, false);
-        decimalColVector.reset();
       }
       // Read present/isNull stream
       fromReader.nextVector(decimalColVector, isNull, batchSize, filterContext, readPhase);
@@ -719,7 +718,6 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         doubleColVector = (DoubleColumnVector) previousVector;
       } else {
         decimalColVector.ensureSize(batchSize, false);
-        decimalColVector.reset();
       }
       // Read present/isNull stream
       fromReader.nextVector(decimalColVector, isNull, batchSize, filterContext, readPhase);
@@ -1063,7 +1061,6 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         decimalColVector = previousVector;
       } else {
         fileDecimalColVector.ensureSize(batchSize, false);
-        fileDecimalColVector.reset();
       }
       // Read present/isNull stream
       fromReader.nextVector(fileDecimalColVector, isNull, batchSize, filterContext, readPhase);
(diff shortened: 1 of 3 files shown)


java/core/src/test/org/apache/orc/impl/TestConvertTreeReaderFactory.java
```diff
@@ -639,4 +641,94 @@ private void testConvertToDateIncreasingSize() throws Exception {
...
+    StringBuilder sb = new StringBuilder();
+    batch.cols[0].stringifyValue(sb, 1023);
+    assertEquals(sb.toString(), "null");
+
+    rows.nextBatch(batch);
+    assertEquals(1024, batch.size);
+    assertEquals(expected, options.toString());
+    assertEquals(batch.cols.length, 1);
+    assertEquals(batch.cols[0].getClass(), expectedColumnType);
+    assertFalse(batch.cols[0].isRepeating);
```
cxzl25, line 709: isRepeating is true, which causes `ConvertTreeReader#convertVector` to only retain the first data of the entire batch, and the rest of the data are ignored, resulting in incorrect results.
In this case, 1024 entries are all 1.

cxzl25:
Reading data may be wrong, please help to review when you have time.
Many thanks in advance.
@dongjoon-hyun  @omalley

guiyanakuang:
Thank you for making a PR, @cxzl25 

I've done some investigation into this issue. `decimalColVector.reset()` seems to fix only part of the bug. I think the fundamental issue is that ORC is missing code to check and set isRepeating when reading all decimal vectors that are not null. 

Make a few simple changes in your test code, the first 1024 values we set to 1.
```java
      VectorizedRowBatch b = schema.createRowBatch();
      DecimalColumnVector f1 = (DecimalColumnVector) b.cols[0];
      f1.isRepeating = true;
      f1.set(0, HiveDecimal.create(1));
      b.size = 1024;
      w.addRowBatch(b);

     // skip some code
     assertTrue(batch.cols[0].isRepeating);  // We actually get false. 
```
There is no code to check and set isRepeating to a filled vector.

cxzl25:
>  I think the fundamental issue is that ORC is missing code to check and set isRepeating when reading all decimal vectors that are not null.

Yes, I agree.
I checked all `ColumnVectors`, only `DecimalColumnVector` isRepeating is not being reset.

`StructBatchReader#readBatchColumn` always resets the vector, so on reads without type promotion, there is no problem.

guiyanakuang:
https://github.com/apache/orc/blob/d085b647eec6b4e0ffc61ee0481049507b741746/java/core/src/java/org/apache/orc/impl/RunLengthIntegerReaderV2.java#L389-L394

So I think it would be better to port the above check code to the following. What do you think? @cxzl25   : ) 

https://github.com/apache/orc/blob/d085b647eec6b4e0ffc61ee0481049507b741746/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java#L1553-L1575

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -554,6 +554,7 @@ public void nextVector(ColumnVector previousVector,
         longColVector = (LongColumnVector) previousVector;
       } else {
         decimalColVector.ensureSize(batchSize, false);
+        decimalColVector.reset();
```
guiyanakuang, line 557: Should you revert the changes to `ConvertTreeReaderFactory.java` ?
There is now code to check and set `isRepeating`.
After reverting, decimal will be handled in the same way as the other types.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -1551,17 +1551,25 @@ private void nextVector(DecimalColumnVector result,
...
+        result.isRepeating = true;
+        for (int r = 0; r < batchSize; ++r) {
           decWritable = vector[r];
           if (!decWritable.serializationUtilsRead(
               valueStream, scratchScaleVector[r],
               scratchBytes)) {
             result.isNull[r] = true;
             result.noNulls = false;
           }
+          if (result.isRepeating
```
guiyanakuang, line 1563: Would it be better to extract it as a common method.

java/core/src/test/org/apache/orc/impl/TestConvertTreeReaderFactory.java
```diff
@@ -639,4 +641,94 @@ private void testConvertToDateIncreasingSize() throws Exception {
...
+      f1.isRepeating = true;
+      f1.set(0, (HiveDecimal) null);
+      b.size = 1024;
+      w.addRowBatch(b);
+      b.reset();
+      for (int i = 0; i < 1024; i++) {
+        f1.set(i, HiveDecimal.create(i + 1));
+      }
+      b.size = 1024;
+      w.addRowBatch(b);
```
guiyanakuang, line 667: Can we add a third batch of repeated data so we can test for repeated to non-repeated and non-repeated to repeated.

guiyanakuang (approved):

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -1708,6 +1724,32 @@ private void nextVector(Decimal64ColumnVector result,
       result.scale = (short) scale;
     }
 
+    private void setIsRepeatingIfNeeded(Decimal64ColumnVector result, int index) {
+      if (result.isRepeating
+          && index > 0
+          && (result.vector[0] != result.vector[index] ||
+          result.isNull[0] != result.isNull[index])) {
```
dongjoon-hyun, line 1731: Indentation on line 1731 looks wrong to me. Did you want the following?
```java
&& (result.vector[0] != result.vector[index] || result.isNull[0] != result.isNull[index])) {
```

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -1708,6 +1724,32 @@ private void nextVector(Decimal64ColumnVector result,
...
+        result.isRepeating = false;
+      }
+    }
+
+    private void setIsRepeatingIfNeeded(DecimalColumnVector result, int index) {
+      if (result.isRepeating
+          && index > 0
+          && (!eq(result.vector[0], result.vector[index]) ||
+          result.isNull[0] != result.isNull[index])) {
```
dongjoon-hyun, line 1740: ditto. Indentation.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -1864,6 +1914,23 @@ private void nextVector(Decimal64ColumnVector result,
       result.scale = (short) scale;
     }
 
+    private void setIsRepeatingIfNeeded(DecimalColumnVector result, int index) {
+      if (result.isRepeating
+          && index > 0
+          && (!eq(result.vector[0], result.vector[index]) ||
+          result.isNull[0] != result.isNull[index])) {
```
dongjoon-hyun, line 1921: ditto.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -1708,6 +1724,32 @@ private void nextVector(Decimal64ColumnVector result,
...
+        result.isRepeating = false;
+      }
+    }
+
+    private boolean eq(HiveDecimalWritable first, HiveDecimalWritable other) {
```
dongjoon-hyun, line 1745: Is there a reason why we cannot use `HiveDecimalWritable.equals` here?
- https://github.com/apache/hive/blob/293a448296933b7498a91e7eeb91edc88dfaa07e/storage-api/src/java/org/apache/hadoop/hive/serde2/io/HiveDecimalWritable.java#L1003-L1019
```java
  public boolean equals(Object other) {
    if (!isSet) {
      return false;
    }
    if (this == other) {
      return true;
    }
    if (other == null || getClass() != other.getClass()) {
      return false;
    }
    HiveDecimalWritable otherHiveDecWritable = (HiveDecimalWritable) other;
    if (!otherHiveDecWritable.isSet()) {
      return false;
    }
    return fastEquals((FastHiveDecimal) otherHiveDecWritable);

  }
```

dongjoon-hyun (approved):
+1, LGTM. Merged to main.
Thank you, @cxzl25 and @guiyanakuang .

Issue #1240, opened 2022-08-30
Huge memory taken for each field when exporting


LouisClt:
Hello,
Using arrow adapter, I became aware that the memory (RAM)  footprint of the export (exporting an orc file) was very huge for each field. For instance, exporting a table with 10000 fields can take up to 30Go, even if there is only 10 records.
Even for 100 fields, that could take 100Mo+.
The "issue" seems to be coming from here :
https://github.com/apache/orc/blob/432a7aade9ea8d3cd705d315da21c2c859bce9ef/c%2B%2B/src/ColumnWriter.cc#L59

When we create a writer with the "createWriter" (https://github.com/apache/orc/blob/432a7aade9ea8d3cd705d315da21c2c859bce9ef/c%2B%2B/src/Writer.cc#L681-L684 ), a stream (compressor) is created for each field. As we allocate a Buffer of 1 * 1024 *1024 we get as a minimum 1Mo additionnal size taken in memory for each field. 

Is there a reason the BufferedOutputStream initial capacity is that high ? I circumvented my problem by lowering it to 1Ko (it didn't change much the performance according to my testing, but it may depend on usecases). Could it be envisaged to put a global variable (or static one) to parametrize this to allow changing this hard coded parameter ?
Thanks

dongjoon-hyun:
cc @wgtmac , @stiga-huang , @coderex2522

coderex2522:
@LouisClt  To support the zero-copy mechanism, class BufferedOutputStream will have an internal data buffer. And the default capacity of  the internal data buffer is 1MB. This default capacity size should be able to be modified, but here's a hint that if the buffer capacity is set too small, it may cause the buffer to expand and trigger memcpy function frequently.

wgtmac:
We may replace the DataBuffer by a new Buffer implementation with a much smarter memory management to automatically grow and shrink its size according to actual usage. This management can happen on the column basis.

LouisClt:
Thanks everyone for your answers. I understand the possible performances issues linked with lowering too much the size of the buffer (on my testing it was OK in my case though).
I think the solution given by @wgtmac would be fine for me, and better than passing by global variables, if it is feasible.

wgtmac:
I have created a JIRA to track the progress: https://issues.apache.org/jira/browse/ORC-1280

coderex2522:
@dongjoon-hyun @wgtmac @LouisClt I will follow up on this issues(ORC-1290) and implement a much smarter memory management.

[...]

PR #1271, opened 2022-10-06
ORC-1280:[C++] Implement block-based buffer(Part I)

coderex2522:
This patch implements a block-based buffer allocator. Provides smarter memory management for  the buffer of BufferedOutputStream class.

This patch implements smarter memory management, which can effectively solve the [issue-1240](https://github.com/apache/orc/issues/1240). And replacing DataBuffer with BlockBuffer in BufferedOutputStream class will be implemented in another patch.

The UTs in TestBlockBuffer.cc can test this patch.

### DIFF
--- a/c++/src/BlockBuffer.cc
+++ b/c++/src/BlockBuffer.cc
@@ -62,8 +62,17 @@ namespace orc {
     }
   }
 
-  void BlockBuffer::reserve(uint64_t capacity) {
-    while (currentCapacity < capacity) {
+  void BlockBuffer::resize(uint64_t size) {
+    reserve(size);
+    if (currentCapacity >= size) {
+      currentSize = size;
+    } else {
+      throw std::logic_error("Block buffer resize error");
+    }
+  }
+
+  void BlockBuffer::reserve(uint64_t newCapacity) {
+    while (currentCapacity < newCapacity) {
       char* newBlockPtr = memoryPool.malloc(blockSize);
       if (newBlockPtr != nullptr) {
         blocks.push_back(newBlockPtr);
@@ -71,15 +80,6 @@ namespace orc {
       } else {
         break;
       }
-    }
-  }
-
-  void BlockBuffer::resize(uint64_t size) {
-    reserve(size);
-    if (currentCapacity >= size) {
-      currentSize = size;
-    } else {
-      throw std::logic_error("Block buffer resize error");
     }
   }
 } // namespace orc
--- a/c++/src/BlockBuffer.cc
+++ b/c++/src/BlockBuffer.cc
@@ -41,7 +41,7 @@ namespace orc {
     currentSize = currentCapacity = 0;
   }
 
-  Block BlockBuffer::getBlock(uint64_t blockIndex) const {
+  BlockBuffer::Block BlockBuffer::getBlock(uint64_t blockIndex) const {
     if (blockIndex >= getBlockNumber()) {
       throw std::out_of_range("Block index out of range");
     }
@@ -49,7 +49,7 @@ namespace orc {
                  std::min(currentSize - blockIndex * blockSize, blockSize));
   }
 
-  Block BlockBuffer::getEmptyBlock() {
+  BlockBuffer::Block BlockBuffer::getNextBlock() {
     if (currentSize < currentCapacity) {
       Block emptyBlock(
           blocks[currentSize / blockSize] + currentSize % blockSize,
--- a/c++/src/BlockBuffer.cc
+++ b/c++/src/BlockBuffer.cc
@@ -1,0 +1,88 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "BlockBuffer.hh"
+
+#include <algorithm>
+
+namespace orc {
+
+  BlockBuffer::BlockBuffer(MemoryPool& pool, uint64_t _blockSize)
+      : memoryPool(pool),
+        currentSize(0),
+        currentCapacity(0),
+        blockSize(_blockSize) {
+    if (blockSize == 0) {
+      throw std::logic_error("Block size cannot be zero");
+    }
+    reserve(blockSize);
+  }
+
+  BlockBuffer::~BlockBuffer() {
+    for (size_t i = 0; i < blocks.size(); ++i) {
+      memoryPool.free(blocks[i]);
+    }
+    blocks.clear();
+    currentSize = currentCapacity = 0;
+  }
+
+  uint64_t BlockBuffer::getBlockNumber() const {
+    return (currentSize + blockSize - 1) / blockSize;
+  }
+
+  Block BlockBuffer::getBlock(uint64_t blockIndex) {
+    if (blockIndex >= getBlockNumber()) {
+      throw std::out_of_range("Block index out of range");
+    }
+    return Block(blocks[blockIndex],
+                 std::min(currentSize - blockIndex * blockSize, blockSize));
+  }
+
+  Block BlockBuffer::getEmptyBlock() {
+    if (currentSize < currentCapacity) {
+      Block emptyBlock(blocks[currentSize / blockSize] + currentSize % blockSize,
+          blockSize - currentSize % blockSize);
+      currentSize = (currentSize / blockSize + 1) * blockSize;
+      return emptyBlock;
+    } else {
+      resize(currentSize + blockSize);
+      return Block(blocks.back(), blockSize);
+    }
+  }
+
+  void BlockBuffer::reserve(uint64_t capacity) {
+    while (currentCapacity < capacity) {
+      char* newBlockPtr = memoryPool.malloc(blockSize);
+      if (newBlockPtr != nullptr) {
+        blocks.push_back(newBlockPtr);
+        currentCapacity += blockSize;
+      } else {
+        break;
+      }
+    }
+  }
+
+  void BlockBuffer::resize(uint64_t size) {
+    reserve(size);
+    if (currentCapacity >= size) {
+      currentSize = size;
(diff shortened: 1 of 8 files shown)


c++/src/MemoryPool.hh
```diff
@@ -0,0 +1,92 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef ORC_MEMORYPOOL_IMPL_HH
+#define ORC_MEMORYPOOL_IMPL_HH
+
+#include "orc/MemoryPool.hh"
```
dongjoon-hyun, line 23: It's a little confusing making another `MemoryPool.hh` like this. Could you create a new file instead of `MemoryPool.hh` file?

dongjoon-hyun:
Thank you for making a PR. I had one comment about file structure, https://github.com/apache/orc/pull/1271#discussion_r989278403.

cc @wgtmac , @stiga-huang

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,90 @@
+
```
dongjoon-hyun, line 1: Redundant space.

c++/test/TestBlockBuffer.cc
```diff
@@ -0,0 +1,82 @@
+
```
dongjoon-hyun, line 1: Redundant space.

dongjoon-hyun:
It seems that the labeler is broken. Let me fix it.

c++/src/MemoryPool.hh
coderex2522: Sure, I rename src/MemoryPool.hh into src/BlockBuffer.hh.

c++/src/BlockBuffer.hh
coderex2522: Done.

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,89 @@
...
+namespace orc {
+
+  struct Block {
+    char* data;
+    uint64_t size;
+
+    Block() : data(nullptr), size(0) {}
+    Block(char* _data, uint64_t _size) : data(_data), size(_size) {}
+    Block(const Block& block) = default;
+    ~Block() {}
```
wgtmac, line 35: ~Block() = default;

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    uint64_t capacity() const {
+      return currentCapacity;
+    }
+
+    void resize(uint64_t size);
+    void reserve(uint64_t capacity);
+  };
+} // namespace
```
wgtmac, line 87: // namespace orc

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,89 @@
...
+
+  public:
+    BlockBuffer(MemoryPool& pool, uint64_t blockSize);
+
+    ~BlockBuffer();
+
+    /**
+     * Get the Block object
+     */
+    Block getBlock(uint64_t blockIndex);
```
wgtmac, line 64: We need some detail explanation with regard to the mechanism and behavior of the BlockBuffer. Especially concepts like block, block index and block number.

wgtmac, line 64: Should we provide mutable and immutable overloads?

wgtmac (commented):
Thanks @coderex2522 for the improvement! Would be great to equip the new class with more comment in detail.

c++/src/BlockBuffer.hh
coderex2522: done.

c++/src/BlockBuffer.hh
coderex2522: done.

c++/src/BlockBuffer.hh
coderex2522: Added some description for block, blockbuffer, block index and block number.

coderex2522: Because the getBlock function does not modify class member variables, the Constant attribute is added to the function.

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,108 @@
...
+    uint64_t size() const {
+      return currentSize;
+    }
+
+    uint64_t capacity() const {
+      return currentCapacity;
+    }
+
+    void resize(uint64_t size);
+    void reserve(uint64_t capacity);
```
wgtmac, line 104: Insert a new line before reserve function. Better to explain what will happen in the reserve function

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,108 @@
...
+
+    uint64_t size() const {
+      return currentSize;
+    }
+
+    uint64_t capacity() const {
+      return currentCapacity;
+    }
+
+    void resize(uint64_t size);
```
wgtmac, line 103: Should we support shrinkToFit parameter?

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,108 @@
...
+
+#include <vector>
+
+namespace orc {
+
+  /**
+   * Block points to a section of memory allocated by BlockBuffer,
+   * containing the corresponding physical memory address and size.
+   */
+  struct Block {
```
wgtmac, line 32: Is it better to make it nested class of BlockBuffer?

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,108 @@
...
+     * and the rest of the blocks are all of size blockSize.
+     * @param blockIndex the index of blocks
+     * @return the allocated block object
+     */
+    Block getBlock(uint64_t blockIndex) const;
+
+    /**
+     * Get a empty block or a new block if the buffer is exhausted.
+     * If the last allocated block size is less than blockSize,
+     * the empty block size is blockSize - lastBlockSize.
```
wgtmac, line 82: Where is the lastBlockSize?
It would be more user-friendly to get the available size of the returned block via output parameter.

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,108 @@
...
+     * Otherwise, the empty block size is blockSize.
+     * @return a empty block object
+     */
+    Block getEmptyBlock();
```
wgtmac, line 86: rename to getNextBlock or getBlockToWrite?

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,108 @@
...
+    ~BlockBuffer();
+
+    /**
+     * Get the allocated block object.
+     * The last allocated block size may be less than blockSize,
+     * and the rest of the blocks are all of size blockSize.
+     * @param blockIndex the index of blocks
+     * @return the allocated block object
+     */
+    Block getBlock(uint64_t blockIndex) const;
```
wgtmac, line 77: User can still ruin the memory of the returned Block. Make the return type const Block?

c++/src/BlockBuffer.hh
```diff
@@ -0,0 +1,108 @@
...
+     * Get a empty block or a new block if the buffer is exhausted.
+     * If the last allocated block size is less than blockSize,
+     * the empty block size is blockSize - lastBlockSize.
+     * Otherwise, the empty block size is blockSize.
+     * @return a empty block object
+     */
+    Block getEmptyBlock();
+
+    /**
+     * Get the number of allocated blocks
```
wgtmac, line 89: Get the number of blocks that are fully or partially occupied

c++/src/BlockBuffer.hh
coderex2522: Added some comments to explain the reserve function.

c++/src/BlockBuffer.hh
coderex2522: As it stands struct Block only serves class BlockBuffer, so it makes reasonable to treat it as a nested class of  class BlockBuffer.

c++/src/BlockBuffer.hh
coderex2522: lastBlockSize means the size of the last allocated block. The comment has been modified here.
The returned block is used to describe the section of memory that can be read or written. And I add comment to the size in struct Block. So I don't recommend adding extra output parameter.

c++/src/BlockBuffer.hh
coderex2522: Done. Rename to getNextBlock.

c++/src/BlockBuffer.hh
coderex2522: Done.

c++/src/BlockBuffer.hh
coderex2522: Since the struct Block contains bare data pointer, user can always write the memory even if it is const Block.

c++/src/BlockBuffer.cc
```diff
@@ -0,0 +1,85 @@
...
+    return Block(blocks[blockIndex],
+                 std::min(currentSize - blockIndex * blockSize, blockSize));
+  }
+
+  BlockBuffer::Block BlockBuffer::getNextBlock() {
+    if (currentSize < currentCapacity) {
+      Block emptyBlock(
+          blocks[currentSize / blockSize] + currentSize % blockSize,
+          blockSize - currentSize % blockSize);
+      currentSize = (currentSize / blockSize + 1) * blockSize;
```
wgtmac, line 57: Should we provide a function to update currentSize to reflect the actual size written? Maybe something setSize() or backup(). This looks weird when the returned block is written partially meaning that currentSize is larger than used

To provide better usability, we may provide a function like **void append(const char* data, size_t size)** to append data to the buffer and manage the blocks internally. This can be a separate patch.

wgtmac (commented):
I have left an inline comment. Overall looks good.

c++/src/BlockBuffer.hh
coderex2522: Class DataBuffer only recycle memory during destructor, so class BlockBuffer does not support shrink function, which can be further optimized later.

c++/src/BlockBuffer.cc
coderex2522: If user only uses part of the block, the size of BlockBuffer can currently be set via the resize() function. BlockBuffer class is temporarily used to replace DataBuffer in BufferedOutputStream, and the existing functions conform to BufferedOutputStream's usage behavior.

The append function can be added additionally later if there are usage scenarios.

coderex2522: Since the flush() function of class BufferedOutputStream needs to access all allocated memory blocks, it needs class BlockBuffer to provide an interface in order for BufferedOutputStream to access all allocated memory blocks.

wgtmac (approved):
+1 LGTM

dongjoon-hyun (approved):
+1, LGTM, too.

wgtmac:
I've submitted and closed it. Thanks @coderex2522 and @dongjoon-hyun

PR #1272, opened 2022-10-07
ORC-1284: Add permissions to GitHub Action labeler

dongjoon-hyun:

This PR aims to recover GitHub Action `labeler` job.

As we can see in this PR, the labeler is broken. We need permissions.

Unfortunately, this should be tested after merging.


dongjoon-hyun:
Let me merge this to recover the PR builders.

PR #1273, opened 2022-10-08
ORC-1285: [Java] Refactor selected vector support

deshanxiao:
This pr is aimed to refactor the support of selected vector.

reduce code duplication

UT(TestSelectedVector)

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/WriterImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/WriterImpl.java
@@ -1006,6 +1006,12 @@ public class WriterImpl implements WriterInternal, MemoryManager.Callback {
     return this.treeWriter.estimateMemory();
   }
 
+  /**
+   * Get next chunk size when row selected enable.
+   * We will find the longest size of chunk that will be less than maxSize.
+   * @param posn the position where we are
+   * @param maxSize the largest chunk size in this selection
+   */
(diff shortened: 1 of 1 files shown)


java/core/src/java/org/apache/orc/impl/WriterImpl.java
```diff
@@ -1018,4 +1005,20 @@ private WriterEncryptionVariant[] setupEncryption(
...
     return this.treeWriter.estimateMemory();
   }
+
+  /**
+   * Get next chunk size if row selected enable.
+   * We will find the longest size of chunk that will be less than maxSize.
+   * @param posn the position where we are
+   * @param maxSize the largest chunk size in this selection
+   */
+  private int nextSelectedRowSize(int posn, int maxSize, int[] selected) {
```
wgtmac, line 1015: Good refactoring. The only concern is that before the change we know posn and (posn + len) will never exceed selected.length. The new method has potential misuse if someone else has passed wrong maxSize when calling it.

deshanxiao: Good catching! Will add an assertion here.

deshanxiao:
Also CC @dongjoon-hyun @guiyanakuang Thanks~

dongjoon-hyun:
cc @pavibhai too

PR #1275, opened 2022-10-12
ORC-1286: [C++] replace DataBuffer with BlockBuffer in class BufferedOutputStream

coderex2522:
This PR can solve the huge memory taken by BufferedOutputStream and refactor the write data logic in class CompressionBase.

This patch use BlockBuffer to replace DataBuffer  of class BufferedOutputStream in order to solve the [issue](https://github.com/apache/orc/issues/1240).

The UTs in TestBufferedOutputStream.cc and TestCompression.cc can cover this patch. Add the TestBlockBuffer.write_to UT.

### DIFF
--- a/c++/src/BlockBuffer.cc
+++ b/c++/src/BlockBuffer.cc
@@ -17,6 +17,7 @@
  */
 
 #include "BlockBuffer.hh"
+#include "orc/Writer.hh"
 #include "orc/OrcFile.hh"
 
 #include <algorithm>
@@ -84,41 +85,54 @@ namespace orc {
     }
   }
 
-  uint64_t BlockBuffer::writeTo(OutputStream* output) {
+  void BlockBuffer::writeTo(OutputStream* output,
+                            WriterMetrics* metrics) {
+    if (currentSize == 0) {
+      return;
+    }
     static uint64_t MAX_CHUNK_SIZE = 1024 * 1024 * 1024;
     uint64_t chunkSize = std::min(output->getNaturalWriteSize(), MAX_CHUNK_SIZE);
     if (chunkSize == 0) {
       throw std::logic_error("Natural write size cannot be zero");
     }
-    char* chunk = memoryPool.malloc(chunkSize);
-    uint64_t chunkOffset = 0;
-
-    uint64_t IOCount = 0;
+    uint64_t ioCount = 0;
     uint64_t blockNumber = getBlockNumber();
-    for (uint64_t i = 0; i < blockNumber; ++i) {
-      Block block = getBlock(i);
-      uint64_t blockOffset = 0;
-      while (blockOffset < block.size) {
-        // copy current block into chunk
-        uint64_t copySize =
-          std::min(chunkSize - chunkOffset, block.size - blockOffset);
-        memcpy(chunk + chunkOffset, block.data + blockOffset, copySize);
-        chunkOffset += copySize;
-        blockOffset += copySize;
+    // if only exists one block, currentSize is equal to first block size
+    if (blockNumber == 1 && currentSize <= chunkSize) {
+      Block block = getBlock(0);
+      output->write(block.data, block.size);
+      ++ioCount;
+    } else {
+      char* chunk = memoryPool.malloc(chunkSize);
+      uint64_t chunkOffset = 0;
+      for (uint64_t i = 0; i < blockNumber; ++i) {
+        Block block = getBlock(i);
+        uint64_t blockOffset = 0;
+        while (blockOffset < block.size) {
+          // copy current block into chunk
+          uint64_t copySize =
+            std::min(chunkSize - chunkOffset, block.size - blockOffset);
+          memcpy(chunk + chunkOffset, block.data + blockOffset, copySize);
+          chunkOffset += copySize;
+          blockOffset += copySize;
 
-        // chunk is full
-        if (chunkOffset >= chunkSize) {
-          output->write(chunk, chunkSize);
-          chunkOffset = 0;
-          ++IOCount;
+          // chunk is full
+          if (chunkOffset >= chunkSize) {
+            output->write(chunk, chunkSize);
+            chunkOffset = 0;
+            ++ioCount;
+          }
         }
       }
+      if (chunkOffset != 0) {
+        output->write(chunk, chunkOffset);
+        ++ioCount;
+      }
+      memoryPool.free(chunk);
     }
-    if (chunkOffset != 0) {
-      output->write(chunk, chunkOffset);
-      ++IOCount;
+
+    if (metrics != nullptr) {
+      metrics->IOCount.fetch_add(ioCount);
     }
-    memoryPool.free(chunk);
-    return IOCount;
   }
 } // namespace orc
--- a/c++/src/BlockBuffer.cc
+++ b/c++/src/BlockBuffer.cc
@@ -17,6 +17,7 @@
  */
 
 #include "BlockBuffer.hh"
+#include "orc/OrcFile.hh"
 
 #include <algorithm>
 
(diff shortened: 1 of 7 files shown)


dongjoon-hyun (commented):
@coderex2522 . Could you use a new JIRA instead of using `PART 1` and `PART II`?
Apache community uses ORC JIRA ID for trace-ability.

c++/src/io/OutputStream.cc
```diff
@@ -95,9 +91,11 @@ namespace orc {
 
   uint64_t BufferedOutputStream::flush() {
     uint64_t dataSize = dataBuffer->size();
+    for (uint64_t i = 0; i < dataBuffer->getBlockNumber(); ++i)
     {
       SCOPED_STOPWATCH(metrics, IOBlockingLatencyUs, IOCount);
-      outputStream->write(dataBuffer->data(), dataSize);
+      auto block = dataBuffer->getBlock(i);
+      outputStream->write(block.data, block.size);
```
wgtmac, line 98: Here may introduce a performance regression by issuing more write I/Os. If writing to cloud object store such as S3, concatenating buffer and making single write is probably a better choice. A possible solution is to add void BlockBuffer::writeTo(OutputStream*) and handle the logic internally.

c++/src/io/OutputStream.cc
```diff
@@ -95,9 +91,11 @@ namespace orc {
 
   uint64_t BufferedOutputStream::flush() {
     uint64_t dataSize = dataBuffer->size();
+    for (uint64_t i = 0; i < dataBuffer->getBlockNumber(); ++i)
     {
       SCOPED_STOPWATCH(metrics, IOBlockingLatencyUs, IOCount);
```
wgtmac, line 96: Measure it outside the loop.

c++/src/io/OutputStream.cc
```diff
@@ -37,7 +37,7 @@ namespace orc {
                                     : outputStream(outStream),
                                       blockSize(blockSize_),
                                       metrics(metrics_) {
-    dataBuffer.reset(new DataBuffer<char>(pool));
+    dataBuffer.reset(new BlockBuffer(pool, blockSize));
```
wgtmac, line 40: Does the **blockSize_** here have the same meaning of BlockBuffer?

wgtmac (changes requested):

c++/src/Compression.cc
```diff
@@ -112,7 +118,10 @@ namespace orc {
                                                 bufferSize(0),
                                                 outputPosition(0),
                                                 outputSize(0) {
-    // PASS
+    // init header pointer array
+    for (int i = 0; i < HEADER_SIZE; ++i) {
```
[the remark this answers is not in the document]
coderex2522: use array.fill() function to initial the header array here.

c++/src/Compression.cc
```diff
@@ -93,6 +95,10 @@ namespace orc {
 
     // Compress output buffer size
     int outputSize;
+
+    // Compression block header pointer array
+    static const int HEADER_SIZE = 3;
+    char* header[HEADER_SIZE];
```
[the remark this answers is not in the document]
coderex2522: done.

c++/src/Compression.cc
```diff
@@ -138,19 +147,44 @@ namespace orc {
            static_cast<uint64_t>(outputSize - outputPosition);
   }
 
+  void CompressionStreamBase::writeData(const unsigned char* data, int size) {
```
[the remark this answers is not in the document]
coderex2522: done.

c++/src/Compression.cc
```diff
@@ -138,19 +147,44 @@ namespace orc {
...
+      if (outputPosition == outputSize) {
+        if (!BufferedOutputStream::Next(
+          reinterpret_cast<void **>(&outputBuffer),
+          &outputSize)) {
+            throw std::runtime_error(
+                "Failed to get next output buffer from output stream.");
+        }
+        outputPosition = 0;
+      } else  if (outputPosition > outputSize) {
+        // this will unlikely happen, but we have seen a few on zstd v1.1.0
```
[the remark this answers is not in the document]
coderex2522: done.

c++/src/io/OutputStream.cc
coderex2522: I try to merge the adjacent write requests here. This part of the logic is like IO merge, so I thought it would be more appropriate to put it here.

c++/src/io/OutputStream.cc
coderex2522: Before this patch, Data buffer will increase memory in blockSize_ during the next() function of class BufferedOutputStream. So the blockSize_ is reasonable as a construction parameter for class BlockBuffer.

wgtmac: OK, that makes sense.

c++/src/io/OutputStream.cc
```diff
@@ -95,9 +91,30 @@ namespace orc {
...
     {
-      SCOPED_STOPWATCH(metrics, IOBlockingLatencyUs, IOCount);
-      outputStream->write(dataBuffer->data(), dataSize);
+      uint64_t ioCount = 0;
+      SCOPED_STOPWATCH(metrics, IOBlockingLatencyUs, nullptr);
+      // try to merge adjacent IO requests
+      BlockBuffer::Block mergeBlock = dataBuffer->getBlock(0);
+      for (uint64_t i = 1; i < dataBuffer->getBlockNumber(); ++i) {
+        auto curBlock = dataBuffer->getBlock(i);
+        if (mergeBlock.data + mergeBlock.size == curBlock.data) {
```
wgtmac, line 103: This leaves the work to the memory allocator and it may never happen.

c++/src/io/OutputStream.cc
```diff
@@ -95,9 +91,30 @@ namespace orc {
...
+          mergeBlock.size += curBlock.size;
+        } else {
+          outputStream->write(mergeBlock.data, mergeBlock.size);
```
wgtmac, line 106: We may also limit each write not to exceed outputStream->getNaturalWriteSize()

c++/src/io/OutputStream.cc
coderex2522: High probability of not happening. So remove this logic.

c++/src/io/OutputStream.cc
coderex2522: The writeTo() of BlockBuffer will limit each write not to exceed NaturalWriteSize of outputStream.

c++/src/BlockBuffer.cc
```diff
@@ -82,4 +83,42 @@ namespace orc {
...
     }
   }
+
+  uint64_t BlockBuffer::writeTo(OutputStream* output) {
+    static uint64_t MAX_CHUNK_SIZE = 1024 * 1024 * 1024;
+    uint64_t chunkSize = std::min(output->getNaturalWriteSize(), MAX_CHUNK_SIZE);
+    if (chunkSize == 0) {
+      throw std::logic_error("Natural write size cannot be zero");
+    }
+    char* chunk = memoryPool.malloc(chunkSize);
```
wgtmac, line 93: If blockNumber == 1, we should avoid this unnecessary allocation

coderex2522: if natural write size less than block size, here need to malloc chunk.

wgtmac: Yes, however you can still get rid of the allocation when blockNumber == 1 && naturalWriteSize >= block.size. BTW, we can keep the current implementation for now

c++/src/io/OutputStream.cc
```diff
@@ -95,9 +91,14 @@ namespace orc {
 
   uint64_t BufferedOutputStream::flush() {
     uint64_t dataSize = dataBuffer->size();
+    // flush data buffer into outputStream
+    if (dataSize > 0)
     {
-      SCOPED_STOPWATCH(metrics, IOBlockingLatencyUs, IOCount);
-      outputStream->write(dataBuffer->data(), dataSize);
+      SCOPED_STOPWATCH(metrics, IOBlockingLatencyUs, nullptr);
+      uint64_t IOCount = dataBuffer->writeTo(outputStream);
```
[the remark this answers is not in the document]
coderex2522: done

wgtmac (approved):
LGTM +1

c++/src/BlockBuffer.cc
coderex2522: Allocation can be avoided here when blockNumber == 1 && currentSize <= NaturalWriteSize.

dongjoon-hyun (approved):
+1, LGTM for Apache ORC 1.9.

PR #1278, opened 2022-10-17
ORC-1289: Refine DynamicArray

deshanxiao:
Refine DynamicIntArray and DynamicByteArray

This PR introduces the following changes:
1. Removed unused methods DynamicByteArray#readAll
2. Use Arrays.fill to replace the original implementation
3. In DynamicIntArray#getSizeInBytes, the method should return long to avoid out of bounds.
4. Update some comments.

UT

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
+++ b/java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
@@ -20,7 +20,6 @@ package org.apache.orc.impl;
 import org.apache.hadoop.io.Text;
 
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
@@ -123,29 +122,6 @@ public final class DynamicByteArray {
   }
 
   /**
-   * Read the entire stream into this array.
-   * @param in the stream to read from
-   * @throws IOException
-   */
-  public void readAll(InputStream in) throws IOException {
-    int currentChunk = length / chunkSize;
-    int currentOffset = length % chunkSize;
-    grow(currentChunk);
-    int currentLength = in.read(data[currentChunk], currentOffset,
-      chunkSize - currentOffset);
-    while (currentLength > 0) {
-      length += currentLength;
-      currentOffset = length % chunkSize;
-      if (currentOffset == 0) {
-        currentChunk = length / chunkSize;
-        grow(currentChunk);
-      }
-      currentLength = in.read(data[currentChunk], currentOffset,
-        chunkSize - currentOffset);
-    }
-  }
-
-  /**
    * Byte compare a set of bytes against the bytes in this dynamic array.
    * @param other source of the other bytes
    * @param otherOffset start offset in the other array
@@ -190,9 +166,7 @@ public final class DynamicByteArray {
    */
   public void clear() {
     length = 0;
-    for(int i=0; i < data.length; ++i) {
-      data[i] = null;
-    }
+    Arrays.fill(data, null);
     initializedChunks = 0;
   }
 
@@ -221,7 +195,7 @@ public final class DynamicByteArray {
    * @param out the stream to write to
    * @param offset the first offset to write
    * @param length the number of bytes to write
-   * @throws IOException
+   * @throws IOException if an I/O error occurs.
    */
   public void write(OutputStream out, int offset,
                     int length) throws IOException {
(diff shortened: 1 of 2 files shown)


java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
```diff
@@ -190,9 +166,7 @@ public int size() {
    */
   public void clear() {
     length = 0;
-    for(int i=0; i < data.length; ++i) {
-      data[i] = null;
-    }
+    Arrays.fill(data, null);
```
dongjoon-hyun, line 169: This is orthogonal to redefine something, @deshanxiao .

java/core/src/java/org/apache/orc/impl/DynamicIntArray.java
```diff
@@ -113,9 +113,7 @@ public int size() {
 
   public void clear() {
     length = 0;
-    for(int i=0; i < data.length; ++i) {
-      data[i] = null;
-    }
+    Arrays.fill(data, null);
```
dongjoon-hyun, line 116: This is orthogonal to redefine something, @deshanxiao .

java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
```diff
@@ -122,29 +121,6 @@ public int add(byte[] value, int valueOffset, int valueLength) {
...
-      currentOffset = length % chunkSize;
-      if (currentOffset == 0) {
-        currentChunk = length / chunkSize;
-        grow(currentChunk);
-      }
-      currentLength = in.read(data[currentChunk], currentOffset,
-        chunkSize - currentOffset);
-    }
-  }
-
```
dongjoon-hyun, line 147: Sorry but I'm -1 for removing like this at 1.9.0.

java/core/src/java/org/apache/orc/impl/DynamicIntArray.java
```diff
@@ -137,8 +135,8 @@ public String toString() {
     return sb.append('}').toString();
   }
 
-  public int getSizeInBytes() {
-    return 4 * initializedChunks * chunkSize;
+  public long getSizeInBytes() {
```
dongjoon-hyun, line 138: From this PR, I believe this is the valid one.

deshanxiao:
Thank you @dongjoon-hyun

java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
deshanxiao: Sorry, will remove it.

java/core/src/java/org/apache/orc/impl/DynamicIntArray.java
```diff
@@ -22,14 +22,14 @@
...
+ * <p>
  * The motivation for this class is memory optimization, i.e. space efficient
  * storage of potentially huge arrays without good a-priori size guesses.
- *
+ * <p>
  * The API of this class is between a primitive array and a AbstractList. It's
  * not a Collection implementation because it handles primitive types, but the
  * API could be extended to support iterators and the like.
- *
+ * <p>
```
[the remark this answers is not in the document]
deshanxiao: Yes, will remove it and create a new one for it.

java/core/src/java/org/apache/orc/impl/DynamicIntArray.java
deshanxiao: Sure

java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
deshanxiao: Yes, it is equivalent to the previous one.

deshanxiao: I will remove it.

dongjoon-hyun (commented):
Could you rebase this PR to the `main` branch, @deshanxiao ?

deshanxiao:
> Could you rebase this PR to the `main` branch, @deshanxiao ?

Hi @dongjoon-hyun 
I tried adding this unit test, but it requires a lot of memory(2GB * 4 =  8GB).
I'm worried that this will cause problems with the whole test system.
But it's easy to reproduce in DynamicByteArray because it requires less memory.

do you have any good suggestions?

```
  @Test
  public void testIntArray2() throws Exception {
    DynamicIntArray array = new DynamicIntArray(10);
    array.set(Integer.MAX_VALUE, 10);

    assertEquals(1L + Integer.MAX_VALUE, array.size()) ;
  }
```

deshanxiao:
Turn it off before thinking about how to do it.

Issue #1277, opened 2022-10-13
[JAVA] mvn package fails if test compiling was skipped


zjx20:
I would like to run `mvn -Dmaven.test.skip=true clean package`, but `maven-dependency-plugin` complains `Unused declared dependencies` for some libraries used by the test code, which breaks the compilation. Please check the attached logs.

I'm not a java expert, but I'm guessing the cause of the problem is the misuse of `analyze-only` in the package phase. According to the [documentation](https://maven.apache.org/plugins/maven-dependency-plugin/analyze-only-mojo.html), the `analyze-only` goal is meant to be used during the `test-compile` phase. In our case, the test class was not compiled, so the dependency analyzer treated some libraries as unused. I don't know the right way to fix it though.

The setting of maven-dependency-plugin:
https://github.com/apache/orc/blob/8cf1047f9ace3799df12f24d2a5096b17a9a6ed0/java/pom.xml#L373-L388

Logs:
```
$ cd orc/java

$ mvn -Dmaven.test.skip=true clean package
[INFO] Scanning for projects...
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO]
[INFO] Apache ORC                                                         [pom]
[INFO] ORC Shims                                                          [jar]
[INFO] ORC Core                                                           [jar]
[INFO] ORC MapReduce                                                      [jar]
[INFO] ORC Tools                                                          [jar]
[INFO] ORC Examples                                                       [jar]
[INFO]
[INFO] -------------------------< org.apache.orc:orc >-------------------------
[INFO] Building Apache ORC 1.9.0-SNAPSHOT                                 [1/6]
[INFO] --------------------------------[ pom ]---------------------------------
[INFO]
[INFO] --- maven-clean-plugin:3.2.0:clean (default-clean) @ orc ---
[INFO] Deleting /Users/x/Documents/playground/orc/java/target
[INFO]
[INFO] --- maven-enforcer-plugin:3.1.0:enforce (enforce-maven-version) @ orc ---
[INFO]
[INFO] --- maven-enforcer-plugin:3.1.0:enforce (enforce-java-version) @ orc ---
[INFO]
[INFO] --- maven-enforcer-plugin:3.1.0:enforce (enforce-maven) @ orc ---
[INFO]
[INFO] --- maven-remote-resources-plugin:1.7.0:process (process-resource-bundles) @ orc ---
[INFO] Preparing remote bundle org.apache:apache-jar-resource-bundle:1.4
[...]

PR #1286, opened 2022-10-21
ORC-1294: Fix build error when skip tests build

deshanxiao:
This PR aims to fix build error when skip tests build. It has two modifications:
- Add `ignoreNonCompile` in maven-dependency-plugin to skip all Runtime/Provided/Test/System scope dependency checking.
- change slf4j-api scope to test.

The root cause of building failed is that -Dmaven.test.skip=true does not compile the test classes.
Apache Maven Dependency Plugin will throw an error if a dependency is not used. We can use ignoreNonCompile defined in https://maven.apache.org/plugins/maven-dependency-plugin/analyze-only-mojo.html#ignoreNonCompile to avoid the problem.

After adding this tag, I found that we have some classes such as slf4j-api in MapReduce module that only appear in test but scope is defined as compile. I adjusted its scope to test.

For a related discussion, see https://github.com/apache/orc/issues/1277

UT

### DIFF
--- a/java/mapreduce/pom.xml
+++ b/java/mapreduce/pom.xml
@@ -63,13 +63,13 @@
       <groupId>org.apache.hive</groupId>
       <artifactId>hive-storage-api</artifactId>
     </dependency>
+
+    <!-- test inter-project -->
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
       <scope>test</scope>
     </dependency>
-
-    <!-- test inter-project -->
     <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-mapreduce-client-jobclient</artifactId>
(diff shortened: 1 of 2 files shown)


deshanxiao:
CC @zjx20 @dongjoon-hyun @pavibhai

java/mapreduce/pom.xml
```diff
@@ -66,6 +66,7 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
+      <scope>test</scope>
     </dependency>
```
dongjoon-hyun, line 70: Could you move this to the next section, `<!-- test inter-project -->`, please?

deshanxiao: Sure, thank you @dongjoon-hyun

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Thank you, @deshanxiao .

PR #1299, opened 2022-10-28
ORC-1304: [C++] Fix bug of seeking over empty PRESENT stream

coderex2522:
This pr doesn't call readHeader function when the seek location is all zero.

When the length of present stream is zero, the seek function pre-call readHeader() function will throw ParseError. 

Add the ByteRle.seekOverEmptyPresentStream and  RLEv1.seekOverEmptyPresentStream UT.


wgtmac (commented):
We need to add e2e test cases to cover at least empty struct/map/list mentioned in the issue.

stiga-huang:
+1. The fix LGTM. I suggest a more specific commit title, e.g. "ORC-1304: [C++] Fix bug of seeking in empty PRESENT stream".

For the missing e2e tests, we can add them in TestPredicatePushdown.cc.

dongjoon-hyun (commented):
+1 for @wgtmac 's and @stiga-huang 's comment.

coderex2522:
> We need to add e2e test cases to cover at least empty struct/map/list mentioned in the issue.

I add the new UT WriterTest.testSeekWithNestedType to cover the child column of the struct/map/list type has empty present stream.

c++/test/TestPredicatePushdown.cc
```diff
@@ -557,4 +557,103 @@ namespace orc {
       TestFirstStripeSelectedWithStripeStats(reader.get(), pos);
     }
   }
+
+  TEST(TestPredicatePushdown, testSeekOverEmptyPresentStream) {
```
wgtmac, line 561: nit: the bug has nothing to do with predicate pushdown, we'd better move it out of here.

c++/test/TestWriter.cc
```diff
@@ -2044,103 +2044,5 @@ namespace orc {
     testSuppressPresentStream(CompressionKind_SNAPPY);
   }
 
-  TEST(WriterTest, testSeekWithNestedType) {
```
wgtmac: Why is it removed?

c++/test/TestPredicatePushdown.cc
coderex2522: I move the ut into the TestReader.cc file because the ParseError exception was not handled during the read operation.

c++/test/TestWriter.cc
coderex2522: It just move into TestReadIntent.testSeekOverEmptyPresentStream in the TestReader.cc file.

wgtmac (approved):
+1 LGTM.

dongjoon-hyun:
cc @williamhyun , too

PR #1305, opened 2022-11-04
ORC-1306: Fixed indented code style for Java modules

guiyanakuang:
This pr is aimed at fixing all indentation style issues with java modules. And add indentation checking rules to checkstyle.xml.

```xml
    <module name="Indentation">
      <property name="severity" value="error"/>
      <property name="basicOffset" value="2"/>
      <property name="braceAdjustment" value="0"/>
      <property name="caseIndent" value="2"/>
      <property name="throwsIndent" value="2"/>
      <property name="lineWrappingIndentation" value="4"/>
      <property name="arrayInitIndent" value="2"/>
    </module>
```

Automated checks save more time and nothing is missed.

Checking with maven-check-plugin.

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcUtils.java
+++ b/java/core/src/java/org/apache/orc/OrcUtils.java
@@ -122,83 +122,83 @@ public class OrcUtils {
               .build());
     }
     switch (typeDescr.getCategory()) {
-    case BOOLEAN:
-      type.setKind(OrcProto.Type.Kind.BOOLEAN);
-      break;
-    case BYTE:
-      type.setKind(OrcProto.Type.Kind.BYTE);
-      break;
-    case SHORT:
-      type.setKind(OrcProto.Type.Kind.SHORT);
-      break;
-    case INT:
-      type.setKind(OrcProto.Type.Kind.INT);
-      break;
-    case LONG:
-      type.setKind(OrcProto.Type.Kind.LONG);
-      break;
-    case FLOAT:
-      type.setKind(OrcProto.Type.Kind.FLOAT);
-      break;
-    case DOUBLE:
-      type.setKind(OrcProto.Type.Kind.DOUBLE);
-      break;
-    case STRING:
-      type.setKind(OrcProto.Type.Kind.STRING);
-      break;
-    case CHAR:
-      type.setKind(OrcProto.Type.Kind.CHAR);
-      type.setMaximumLength(typeDescr.getMaxLength());
-      break;
-    case VARCHAR:
-      type.setKind(OrcProto.Type.Kind.VARCHAR);
-      type.setMaximumLength(typeDescr.getMaxLength());
-      break;
-    case BINARY:
-      type.setKind(OrcProto.Type.Kind.BINARY);
-      break;
-    case TIMESTAMP:
-      type.setKind(OrcProto.Type.Kind.TIMESTAMP);
-      break;
-    case TIMESTAMP_INSTANT:
+      case BOOLEAN:
+        type.setKind(OrcProto.Type.Kind.BOOLEAN);
+        break;
+      case BYTE:
+        type.setKind(OrcProto.Type.Kind.BYTE);
+        break;
+      case SHORT:
+        type.setKind(OrcProto.Type.Kind.SHORT);
+        break;
+      case INT:
+        type.setKind(OrcProto.Type.Kind.INT);
+        break;
+      case LONG:
+        type.setKind(OrcProto.Type.Kind.LONG);
+        break;
+      case FLOAT:
+        type.setKind(OrcProto.Type.Kind.FLOAT);
+        break;
+      case DOUBLE:
+        type.setKind(OrcProto.Type.Kind.DOUBLE);
+        break;
+      case STRING:
+        type.setKind(OrcProto.Type.Kind.STRING);
+        break;
+      case CHAR:
+        type.setKind(OrcProto.Type.Kind.CHAR);
+        type.setMaximumLength(typeDescr.getMaxLength());
+        break;
+      case VARCHAR:
+        type.setKind(OrcProto.Type.Kind.VARCHAR);
+        type.setMaximumLength(typeDescr.getMaxLength());
+        break;
+      case BINARY:
+        type.setKind(OrcProto.Type.Kind.BINARY);
+        break;
+      case TIMESTAMP:
+        type.setKind(OrcProto.Type.Kind.TIMESTAMP);
+        break;
+      case TIMESTAMP_INSTANT:
         type.setKind(OrcProto.Type.Kind.TIMESTAMP_INSTANT);
         break;
-    case DATE:
-      type.setKind(OrcProto.Type.Kind.DATE);
-      break;
-    case DECIMAL:
-      type.setKind(OrcProto.Type.Kind.DECIMAL);
-      type.setPrecision(typeDescr.getPrecision());
-      type.setScale(typeDescr.getScale());
-      break;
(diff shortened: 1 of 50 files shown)


wgtmac (approved):

guiyanakuang:
cc @dongjoon-hyun @williamhyun.
Btw, there are very few conflicts with the port patch. Therefore it does not cause additional work for further ports.

java/core/src/java/org/apache/orc/OrcUtils.java
```diff
@@ -318,15 +317,15 @@ TypeDescription convertTypeFromProtobuf(List<OrcProto.Type> types,
...
+        result = TypeDescription.createDecimal();
+        if (type.hasScale()) {
+          result.withScale(type.getScale());
         }
-        break;
+        if (type.hasPrecision()) {
+          result.withPrecision(type.getPrecision());
+        }
+      }
+      break;
```
dongjoon-hyun, line 328: It seems that I missed this in `main` branch but `break` indentation looks weird to me. Shall we remove the extra `{}` instead?
```
- case DECIMAL: {
+ case DECIMAL:
```
- Do you think newly added indentation rule looks insufficient to catch this?

java/core/src/java/org/apache/orc/OrcUtils.java
```diff
@@ -345,27 +344,27 @@ TypeDescription convertTypeFromProtobuf(List<OrcProto.Type> types,
...
+        result = TypeDescription.createStruct();
+        for(int f=0; f < type.getSubtypesCount(); ++f) {
+          String name = type.getFieldNames(f);
+          name = name.startsWith("`") ? name : "`" + name + "`";
+          String fieldName = ParserUtils.parseName(new ParserUtils.StringPosition(name));
+          result.addField(fieldName, convertTypeFromProtobuf(types, type.getSubtypes(f)));
         }
-        break;
+      }
+      break;
```
dongjoon-hyun, line 355: ditto.
```
- case STRUCT: {
+ case STRUCT:
```

java/core/src/java/org/apache/orc/OrcUtils.java
```diff
@@ -345,27 +344,27 @@ TypeDescription convertTypeFromProtobuf(List<OrcProto.Type> types,
...
       case UNION: {
```
dongjoon-hyun, line 356: ditto. 
```
- case UNION: {
+ case UNION:
```

dongjoon-hyun (changes requested):

dongjoon-hyun:
Thank you for making a backport, @guiyanakuang . There exist a few case where `break` is indented in the same level with the next `case` statement in this PR. It looks weird to me. I guess we can fix them by eliminating the redundant `{}`.
```
break;
case ...
```

guiyanakuang:
This is really weird, let me fix it. Also the master branch has been merged, so I'll fix it after this pr review is over.

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -118,62 +121,62 @@ protected void assignStringGroupVectorEntry(BytesColumnVector bytesColVector,
...
+        case STRING:
+          bytesColVector.setVal(elementNum, bytes, start, length);
+          break;
+        case CHAR: {
           int adjustedDownLen =
               StringExpr.rightTrimAndTruncate(bytes, start, length, readerType.getMaxLength());
           bytesColVector.setVal(elementNum, bytes, start, adjustedDownLen);
+          break;
         }
-        break;
```
guiyanakuang, line 132: @dongjoon-hyun. Here I chose to move the break inside the brackets. Because if remove the {} . The context will show the problem of duplicate variables with the same name (`adjustedDownLen `) being defined.

java/core/src/java/org/apache/orc/OrcUtils.java
guiyanakuang: When {} is present, the new rule expects the break indent to be the same as the case, 6 spaces.
I have now removed the {}.

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
dongjoon-hyun: If you don't mind, please rename `adjustedDownLen` and remove `{}`.

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -118,62 +121,62 @@ protected void assignStringGroupVectorEntry(BytesColumnVector bytesColVector,
...
-      case VARCHAR:
-        {
+        case VARCHAR: {
           int adjustedDownLen =
               StringExpr.truncate(bytes, start, length, readerType.getMaxLength());
           bytesColVector.setVal(elementNum, bytes, start, adjustedDownLen);
+          break;
```
dongjoon-hyun, line 143: ditto.

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -118,62 +121,62 @@ protected void assignStringGroupVectorEntry(BytesColumnVector bytesColVector,
...
-            .rightTrimAndTruncate(bytesColVector.vector[elementNum],
-                bytesColVector.start[elementNum], length,
-                readerType.getMaxLength());
+              .rightTrimAndTruncate(bytesColVector.vector[elementNum],
+                  bytesColVector.start[elementNum], length,
+                  readerType.getMaxLength());
           if (adjustedDownLen < length) {
             bytesColVector.length[elementNum] = adjustedDownLen;
           }
+          break;
```
dongjoon-hyun, line 165: ditto.

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -118,62 +121,62 @@ protected void assignStringGroupVectorEntry(BytesColumnVector bytesColVector,
...
-            .truncate(bytesColVector.vector[elementNum],
-                bytesColVector.start[elementNum], length,
-                readerType.getMaxLength());
+              .truncate(bytesColVector.vector[elementNum],
+                  bytesColVector.start[elementNum], length,
+                  readerType.getMaxLength());
           if (adjustedDownLen < length) {
             bytesColVector.length[elementNum] = adjustedDownLen;
           }
+          break;
```
dongjoon-hyun, line 176: ditto

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
guiyanakuang: Here I chose to move the break inside the brackets. Because if remove the {} . The context will show the problem of duplicate variables with the same name (`adjustedDownLen`) being defined.

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
guiyanakuang: OK, following your advice.

guiyanakuang:
I have eliminated the redundant `{}`. Thank you for your careful review. : ) @dongjoon-hyun

dongjoon-hyun (approved):
+1, LGTM. I double-checked the renaming is correct. Thank you so much!

dongjoon-hyun:
Merged to branch-1.8.

PR #1307, opened 2022-11-04
ORC-1298:[C++] Support dedicated ColumnVectorBatch of numeric types

luffy-zh:

This patch implements int8/int16/int32/float buffer and refactors the RLE encoder/decoder, IntegerColumnWriter/Reader, and DoubleColumnWriter/Reader.

For the context of [ORC-1298](https://issues.apache.org/jira/projects/ORC/issues/ORC-1298?filter=allopenissues), this change makes user no need to focus on overflow.

Add the WriterTest.testWriterFixedWidthNumericVectorBatch UT.

### DIFF
--- a/c++/include/orc/Type.hh
+++ b/c++/include/orc/Type.hh
@@ -72,6 +72,10 @@ namespace orc {
     virtual ORC_UNIQUE_PTR<ColumnVectorBatch> createRowBatch(uint64_t size, MemoryPool& pool,
                                                              bool encoded = false) const = 0;
 
+    virtual ORC_UNIQUE_PTR<ColumnVectorBatch> createRowBatch(
+        uint64_t size, MemoryPool& pool, bool encoded = false,
+        bool useTightNumericVector = false) const = 0;
+
     /**
      * Add a new field to a struct type.
      * @param fieldName the name of the new field
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -25,9 +25,9 @@
 
 #include <cstdlib>
 #include <cstring>
-#include <iostream>
 #include <list>
 #include <memory>
+#include <sstream>
 #include <stdexcept>
 #include <vector>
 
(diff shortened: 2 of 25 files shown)


wgtmac:
Thanks @luffy-zh for the contribution!

cc @stiga-huang @coderex2522

c++/include/orc/Type.hh
```diff
@@ -70,10 +70,11 @@ namespace orc {
...
      */
-    virtual ORC_UNIQUE_PTR<ColumnVectorBatch> createRowBatch(uint64_t size,
-                                                             MemoryPool& pool,
-                                                             bool encoded = false
-                                                             ) const = 0;
+    virtual ORC_UNIQUE_PTR<ColumnVectorBatch> createRowBatch(
+      uint64_t size,
+      MemoryPool& pool,
+      bool encoded = false,
+      bool enableFixedWidthNumericVectorBatch = false) const = 0;
```
dongjoon-hyun, line 77: @luffy-zh When you add one parameter, please avoid any change to the existing parameters. For example, indentation changes on line 73 to 76.

dongjoon-hyun: In addition, can we keep the original three-parameter method and add new four-parameter method with new parameter additionally?
```
bool enableFixedWidthNumericVectorBatch = false) const = 0;
```

c++/src/ColumnWriter.cc
```diff
@@ -456,126 +456,107 @@ namespace orc {
...
   class IntegerColumnWriter : public ColumnWriter {
   public:
     IntegerColumnWriter(
                         const Type& type,
                         const StreamsFactory& factory,
-                        const WriterOptions& options);
+                        const WriterOptions& options) :
+                          ColumnWriter(type, factory, options),
+                          rleVersion(options.getRleVersion()) {
+      std::unique_ptr<BufferedOutputStream> dataStream =
```
wgtmac, line 468: nit: you can still separate template class declaration with definition like before to keep the class declaration compact. See here for reference: https://stackoverflow.com/questions/13351474/c-separate-declaration-definition-for-template-function-in-template-class

c++/src/ColumnWriter.cc
```diff
@@ -456,126 +456,107 @@ namespace orc {
     }
   }
 
+  template <typename BatchType, typename IntType>
```
wgtmac, line 459: IntType can be deduced from BatchType, isn't it?

c++/src/ColumnWriter.cc
```diff
@@ -828,16 +810,17 @@ namespace orc {
...
-                          bool isFloatType) :
-                              ColumnWriter(type, factory, options),
-                              isFloat(isFloatType),
-                              buffer(*options.getMemoryPool()) {
+  template<typename ValueType, typename BatchType>
+  DoubleColumnWriter<ValueType, BatchType>::DoubleColumnWriter(
+    const Type& type,
+    const StreamsFactory& factory,
+    const WriterOptions& options,
+    bool isFloatType) :
```
wgtmac, line 818: We should throw if isFloatType == false and ValueType is float

c++/src/ColumnWriter.cc
```diff
@@ -3061,15 +3050,36 @@ namespace orc {
...
+                                                             options));
+        }
       case LONG:
-      case SHORT:
         return std::unique_ptr<ColumnWriter>(
-          new IntegerColumnWriter(
-                                  type,
-                                  factory,
-                                  options));
+          new IntegerColumnWriter<LongVectorBatch, int64_t>(
```
wgtmac, line 3071: It seems redundant to define the 2nd template parameter

c++/src/ColumnWriter.cc
```diff
@@ -3083,18 +3093,26 @@ namespace orc {
                                   options));
       case DOUBLE:
         return std::unique_ptr<ColumnWriter>(
-          new DoubleColumnWriter(
-                                 type,
-                                 factory,
-                                 options,
-                                 false));
+          new DoubleColumnWriter<double, DoubleVectorBatch>(
```
wgtmac, line 3096: ditto

c++/src/RLE.hh
```diff
@@ -59,8 +59,16 @@ namespace orc {
...
-                      const char* notNull);
+                     const char* notNull);
 
+    virtual void add(const int32_t* data, uint64_t numValues,
+                     const char* notNull);
+
+    virtual void add(const int16_t* data, uint64_t numValues,
+                     const char* notNull);
+
+    virtual void add(const int8_t* data, uint64_t numValues,
```
wgtmac, line 70: RLE only supports int64_t, int32_t, and int16_t. Please remove the int8_t overload.

c++/src/RLE.hh
```diff
@@ -126,6 +134,15 @@ namespace orc {
     virtual void next(int64_t* data, uint64_t numValues,
                       const char* notNull) = 0;
 
+    virtual void next(int32_t* data, uint64_t numValues,
+                      const char* notNull) = 0;
+
+    virtual void next(int16_t* data, uint64_t numValues,
+                      const char* notNull) = 0;
+
+    virtual void next(int8_t* data, uint64_t numValues,
```
wgtmac, line 143: ditto

c++/src/RLE.cc
```diff
@@ -71,14 +71,41 @@ namespace orc {
...
-                         const char* notNull) {
+                       const char* notNull) {
     for (uint64_t i = 0; i < numValues; ++i) {
       if (!notNull || notNull[i]) {
         write(data[i]);
       }
     }
   }
 
+  void RleEncoder::add(const int32_t* data, uint64_t numValues,
```
wgtmac, line 82: Why not use a template function to repeat the code? static_cast does not have any runtime penalty.

c++/include/orc/Vector.hh
```diff
@@ -88,28 +89,88 @@ namespace orc {
...
-    void resize(uint64_t capacity);
-    void clear();
-    uint64_t getMemoryUsage();
+    virtual ~IntegerVectorBatch() {
+      // PASS
+    }
+
+    std::string toString() const {
+      std::ostringstream buffer;
+      buffer << "Long vector <" << numElements << " of " << capacity << ">";
```
wgtmac, line 106: It may not be "Long vector" any more

c++/include/orc/Vector.hh
```diff
@@ -88,28 +89,88 @@ namespace orc {
...
+      // PASS
+    }
+
+    virtual ~BaseDoubleVectorBatch() {
+      // PASS
+    }
+
+    std::string toString() const {
+      std::ostringstream buffer;
+      buffer << "Double vector <" << numElements << " of " << capacity << ">";
```
wgtmac, line 148: It may not be "Double vector" any more.

c++/test/TestWriter.cc
```diff
@@ -2044,5 +2048,91 @@ namespace orc {
     testSuppressPresentStream(CompressionKind_SNAPPY);
   }
 
+  TEST_P(WriterTest, testWriteFixedWidthNumericVectorBatch) {
+    MemoryOutputStream memStream(DEFAULT_MEM_STREAM_SIZE);
+    MemoryPool * pool = getDefaultPool();
+    std::unique_ptr<Type> type(Type::buildTypeFromString(
+      "struct<col1:double,col2:float,col3:int,col4:smallint,col5:tinyint>"));
```
coderex2522, line 2055: It's better to test the bigint type as well.

.gitignore
```diff
@@ -1,14 +0,0 @@
-build
```
coderex2522, line 1: This file is not associated with this pull request. Whether it is a misoperation?

c++/src/ColumnReader.cc
```diff
@@ -474,31 +465,32 @@ namespace orc {
...
+  void DoubleColumnReader<columnKind, isLittleEndian, ValueType, BatchType>::next(
+      ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) {
     ColumnReader::next(rowBatch, numValues, notNull);
     // update the notNull from the parent class
     notNull = rowBatch.hasNulls ? rowBatch.notNull.data() : nullptr;
-    double* outArray = dynamic_cast<DoubleVectorBatch&>(rowBatch).data.data();
+    ValueType* outArray =
+        reinterpret_cast<ValueType*>(dynamic_cast<BatchType&>(rowBatch).data.data());
 
     if (columnKind == FLOAT) {
```
wgtmac, line 477:
```suggestion
    if constexpr (columnKind == FLOAT) {
```

c++/src/Timezone.cc
```diff
@@ -24,6 +24,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#include <iostream>
```
wgtmac, line 27: Remove it

tools/src/FileMetadata.cc
```diff
@@ -26,7 +26,7 @@
 #include "orc/Exceptions.hh"
 #include "orc/OrcFile.hh"
 
-//#include "Adaptor.hh"
+// #include "Adaptor.hh"
```
wgtmac, line 29: Remove this line

wgtmac (changes requested):

c++/src/MemoryPool.cc
```diff
@@ -180,6 +198,60 @@ namespace orc {
...
+
+  template <>
+  DataBuffer<int16_t>::~DataBuffer() {
+    if (buf) {
+      memoryPool.free(reinterpret_cast<char*>(buf));
+    }
+  }
+
+  template <>
+  void DataBuffer<int16_t>::resize(uint64_t newSize) {
```
wgtmac, line 229: resize of different vectors should also be in the unit test

c++/src/ColumnWriter.cc
```diff
@@ -415,109 +415,97 @@ namespace orc {
...
+        recordPosition();
+      }
+    }
 
-    virtual void flush(std::vector<proto::Stream>& streams) override;
+    virtual void add(ColumnVectorBatch& rowBatch, uint64_t offset, uint64_t numValues,
+                     const char* incomingMask) override {
+      const BatchType* intBatch = dynamic_cast<const BatchType*>(&rowBatch);
+      if (intBatch == nullptr) {
+        throw InvalidArgument("Failed to cast to LongVectorBatch");
```
wgtmac, line 438:
```suggestion
        throw InvalidArgument("Failed to cast to integer vector batch");
```

wgtmac (approved):
LGTM +1

c++/src/ColumnReader.cc
```diff
@@ -1696,13 +1690,25 @@ namespace orc {
...
     switch (static_cast<int64_t>(type.getKind())) {
-      case DATE:
-      case INT:
+      case SHORT: {
+        if (useTightNumericVector) {
+          return std::unique_ptr<ColumnReader>(
+              new IntegerColumnReader<ShortVectorBatch>(type, stripe));
+        }
+      }
+      case INT: {
```
dongjoon-hyun, line 1702: This seems to cause `implicit-fallthrough` warning and compilation failure.

wgtmac: Should we fix it?

dongjoon-hyun: Yes, we need to fix or mitigate this for Apache ORC 1.9.0 because this causes Docker Test failures.

dongjoon-hyun: I filed a JIRA and PR for `branch-1.9` first. Will make another PR for main branch..
- https://issues.apache.org/jira/browse/ORC-1453
- https://github.com/apache/orc/pull/1548 (for branch-1.9)

PR #1314, opened 2022-11-14
ORC-1310: Whitelist Support for plugin filter

deshanxiao:
This PR is aimed to add whitelist support for plugin filter.

**ServiceLoader** will load all the interfaces in current classpath. When we have two implementations of PluginFilterService, Both of them will be loaded as plugin filters. This PR will provide a configuration "orc.filter.plugin.whitelist". ORC reader will only load the class which is in the whitelist when the configuration is not null.

UT

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -194,11 +194,11 @@ public enum OrcConf {
                       + "order of application."),
 
   PLUGIN_FILTER_WHITELIST("orc.filter.plugin.whitelist",
-                      "orc.filter.plugin.whitelist",
-                      "",
-                      "A comma-separated list of whitelist classes. It is used to specify the "
-                      + "className which will be loaded as PluginFilterService. "
-                      + "Classes that not in the list will be ignored."),
+                          "orc.filter.plugin.whitelist",
+                          "",
+                          "A comma-separated list of whitelist classes. It is used to specify the "
+                          + "className which will be loaded as PluginFilterService. "
+                          + "Classes that not in the list will be ignored."),
 
   WRITE_VARIABLE_LENGTH_BLOCKS("orc.write.variable.length.blocks", null, false,
       "A boolean flag as to whether the ORC writer should write variable length\n"
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -196,9 +196,9 @@ public enum OrcConf {
   PLUGIN_FILTER_WHITELIST("orc.filter.plugin.whitelist",
                       "orc.filter.plugin.whitelist",
                       "",
-                      "A comma-separated list of whitelist classes. It is used to specify the " +
-                          "className which will be loaded as PluginFilterService. " +
-                          "Classes that not in the list will be ignored."),
+                      "A comma-separated list of whitelist classes. It is used to specify the "
+                          + "className which will be loaded as PluginFilterService. "
+                          + "Classes that not in the list will be ignored."),
 
   WRITE_VARIABLE_LENGTH_BLOCKS("orc.write.variable.length.blocks", null, false,
       "A boolean flag as to whether the ORC writer should write variable length\n"
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -197,7 +197,7 @@ public enum OrcConf {
                       "orc.filter.plugin.whitelist",
                       "",
                       "A comma-separated list of whitelist classes. It is used to specify the " +
-                          "className which will be loaded as org.apache.orc.filter.PluginFilterService. " +
+                          "className which will be loaded as PluginFilterService. " +
                           "Classes that not in the list will be ignored."),
 
   WRITE_VARIABLE_LENGTH_BLOCKS("orc.write.variable.length.blocks", null, false,
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -20,6 +20,8 @@ package org.apache.orc;
 
 import org.apache.hadoop.conf.Configuration;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Properties;
 
 /**
@@ -190,6 +192,14 @@ public enum OrcConf {
                       + "determined, they are combined using AND. The order of application is "
                       + "non-deterministic and the filter functionality should not depend on the "
                       + "order of application."),
+
+  PLUGIN_FILTER_WHITELIST("orc.filter.plugin.whitelist",
+                      "orc.filter.plugin.whitelist",
+                      "",
+                      "A comma-separated list of whitelist classes. It is used to specify the " +
+                          "className which will be loaded as org.apache.orc.filter.PluginFilterService. " +
+                          "Classes that not in the list will be ignored."),
+
   WRITE_VARIABLE_LENGTH_BLOCKS("orc.write.variable.length.blocks", null, false,
       "A boolean flag as to whether the ORC writer should write variable length\n"
       + "HDFS blocks."),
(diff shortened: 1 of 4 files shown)


dongjoon-hyun (changes requested):
Thank you for making a PR, however, Apache ORC intentionally aims to avoid `non-inclusive` languages like `whitelist`. Could you revise your PR to use more conscious language?

FYI, we are checking like the following.
- https://clc.diversity.apache.org/analysis.html?project=orc.git

Screenshot 2022-11-14 at 9 15 53 AM

dongjoon-hyun:
cc @williamhyun and @wgtmac , too

williamhyun:
+1 for @dongjoon-hyun's comment

deshanxiao:
Sorry @dongjoon-hyun , will update it. Thanks~

deshanxiao:
@dongjoon-hyun I have updated the jira and the pull request.

dongjoon-hyun:
Thank you for update, @deshanxiao .

dongjoon-hyun (changes requested):
It looks like a breaking change to the existing production, @deshanxiao . Is the intention of this PR correct?

deshanxiao:
> It looks like a breaking change to the existing production, @deshanxiao . Is the intention of this PR correct?

Actually if orc.filter.plugin.allowlist is empty. Everything will the same as before. I think the it will be better  if change the default value to *

java/core/src/java/org/apache/orc/impl/filter/FilterFactory.java
```diff
@@ -184,4 +186,24 @@ static List<BatchFilter> findPluginFilters(String filePath, Configuration conf)
...
+
+  /**
+   * Delete filter(s) which is not in the allowlist.
+   *
+   * @param filters fully filter list we load from class path.
+   * @param allowList filter allowList. The plugin service will not be
+   *                  loaded if it is not in allowList.
+   */
+  static void deleteFilterNotInAllowList(List<BatchFilter> filters, List<String> allowList) {
+    if (allowList != null && allowList.size() > 0) {
```
dongjoon-hyun, line 198: Given the function name, `deleteFilterNotInAllowList` and description, this logic looks wrong to me.

java/core/src/test/org/apache/orc/impl/filter/TestPluginFilterService.java
```diff
@@ -56,4 +60,24 @@ public void testErrorCreatingFilter() {
   public void testMissingFilter() {
     assertTrue(FilterFactory.findPluginFilters("file://db/table11/file1", conf).isEmpty());
   }
+
+  @Test
+  public void testAllowListFilter() {
```
dongjoon-hyun, line 65: Please add additional test cases explicitly and revise the logic.
- `testAllowListFilterAllowAll`
- `testAllowListFilterDisallowAll`

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -190,6 +193,14 @@ public enum OrcConf {
                       + "determined, they are combined using AND. The order of application is "
                       + "non-deterministic and the filter functionality should not depend on the "
                       + "order of application."),
+
+  PLUGIN_FILTER_ALLOWLIST("orc.filter.plugin.allowlist",
+                          "orc.filter.plugin.allowlist",
+                          "*",
+                          "A comma-separated list of allowlist classes. It is used to specify the "
```
dongjoon-hyun, line 200: It looks like recursive definition. We cannot use `allowlist` in the definition sentence of `allowlist`. Maybe, the following is better?
```
- A comma-separated list of allowlist classes
+ A list of comma-separated class names
```

pavibhai:
Thanks @deshanxiao for raising this. Can you please help me understand the use a little better? Is it something like

```text
Is this more like I have started a spark session with a set of dependencies (that in this case might include plugin filter classes). However I might not want to engage these filters for some queries while I might use it for other queries from the same session.
```

deshanxiao:
@pavibhai Yes, you are right. In fact, a spark query does not want to use all plugin filter classes because some plugin filter classes are used by other query. Now, according to the mechanism of ServiceLoader , all plugin filter classes will be loaded into any quey.
With this PR we can choose the plugin filter classes we want to use.

java/core/src/java/org/apache/orc/impl/filter/FilterFactory.java
```diff
@@ -71,9 +71,11 @@ public static BatchFilter createBatchFilter(Reader.Options opts,
     // 2. Process PluginFilter
     if (opts.allowPluginFilters()) {
       List<BatchFilter> pluginFilters = findPluginFilters(filePath, conf);
-      if (!pluginFilters.isEmpty()) {
-        LOG.debug("Added plugin filters {} to the read", pluginFilters);
-        filters.addAll(pluginFilters);
+      List<BatchFilter> allowFilters =
```
pavibhai, line 74: Maybe we can move this logic into `findPluginFilters`?

deshanxiao:
@pavibhai Thanks for your suggestions!

deshanxiao:
Hi @pavibhai, I have updated all the comments, please help to see if you have time, thanks~

pavibhai:
> Hi @pavibhai, I have updated all the comments, please help to see if you have time, thanks~

Sure. I shall take a look first thing tomorrow morning

java/core/src/java/org/apache/orc/OrcConf.java
```diff
@@ -329,6 +341,14 @@ public String getString(Configuration conf) {
     return getString(null, conf);
   }
 
+  public List<String> getStringAsList(Configuration conf) {
+    String confStr = getString(null, conf);
+    if (confStr == null || confStr.isEmpty()) {
+      return new ArrayList<>();
+    }
+    return Arrays.asList(confStr.split(","));
+  }
```
dongjoon-hyun, line 350: If you don't mind, could you open a new JIRA and PR for this, @deshanxiao ?

And, it would be great to remove white space and empty strings after `split`. For example, `a,,b,c,`?

dongjoon-hyun: I saw your JIRA and have been waiting for your PR, @deshanxiao  :)

dongjoon-hyun: I merged https://github.com/apache/orc/pull/1337 .

dongjoon-hyun (commented):
https://github.com/apache/orc/pull/1338 is merged. Could you rebase to `main` branch and resolve the conflicts?

deshanxiao:
@dongjoon-hyun I have resolved the conflicts, thank you!

dongjoon-hyun:
Thank you for update, @deshanxiao .

java/core/src/java/org/apache/orc/impl/filter/FilterFactory.java
dongjoon-hyun: Please rename this to `pluginFilters` to minize the diff.

java/core/src/java/org/apache/orc/impl/filter/FilterFactory.java
```diff
@@ -184,4 +186,32 @@ static List<BatchFilter> findPluginFilters(String filePath, Configuration conf)
...
     return filters;
   }
+
+  /**
+   * Filter BatchFilter which is in the allowList.
+   *
+   * @param filters whole BatchFilter list we load from class path.
+   * @param allowList a Class-Name list that we want to load in.
+   */
+  static List<BatchFilter> getAllowedFilters(List<BatchFilter> filters, List<String> allowList) {
```
dongjoon-hyun, line 196: Is it open for `test`? I'd recommend to switch to `private static`?

dongjoon-hyun (changes requested):
I opened a PR to you, @deshanxiao .
- https://github.com/deshanxiao/orc/pull/1/files

deshanxiao:
Thank you, I merge it.

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Merged to main for Apache ORC 1.9.0. Thank you, @deshanxiao and @pavibhai .

Issue #1318, opened 2022-11-18
Byte to integer conversions fail on platforms with unsigned char type


johnwheffner:
When building the C++ library on a platform where `char` is by default unsigned, byte-to-integer expansion is incorrect in `orc::expandBytesToIntegers` as well as in a few unit tests.

This can be reproduced on any CPU architecture when building with gcc by compiling with `-funsigned-char`.

dongjoon-hyun:
Thank you, @johnwheffner and @wgtmac .

dongjoon-hyun:
FYI. Apache ORC 1.8.1 is scheduled on December 3rd.

cc @williamhyun

wgtmac:
> Thank you, @johnwheffner and @wgtmac . For now, I set the milestone 1.9.0, but we can backport it to [Milestone 1.8.1](https://github.com/apache/orc/milestone/13).

I can confirm that this issue only exists in the main branch, therefore we don't need backport the fix to branch-1.8.

dongjoon-hyun:
It's great. Thank you for the confirmation, @wgtmac .

wgtmac:
@johnwheffner FYI, this issue has been fixed in the main branch. Please let me know if there is any remaining issue. Thanks!

PR #1323, opened 2022-11-23
ORC-1315:[C++] Fix byte to integer conversions fail on platforms with unsigned char type

luffy-zh:

This patch adds static_cast for the char type cast in RLE/ColumnReader.

For the context of [ORC-1315](https://issues.apache.org/jira/browse/ORC-1315), this patch fixes signed char to unsigned char conversions fail where char is by default unsigned.

No new tests were added.

### DIFF
--- a/c++/src/ByteRLE.cc
+++ b/c++/src/ByteRLE.cc
@@ -354,7 +354,7 @@ namespace orc {
     if (bufferStart == bufferEnd) {
       nextBuffer();
     }
-    return *(bufferStart++);
+    return static_cast<signed char>(*(bufferStart++));
   }
 
   void ByteRleDecoderImpl::readHeader() {
@@ -365,7 +365,7 @@ namespace orc {
     } else {
       remainingValues = static_cast<size_t>(ch) + MINIMUM_REPEAT;
       repeating = true;
-      value = readByte();
+      value = static_cast<char>(readByte());
     }
   }
 
@@ -458,7 +458,7 @@ namespace orc {
         if (notNull) {
           for (uint64_t i = 0; i < count; ++i) {
             if (notNull[position + i]) {
-              data[position + i] = readByte();
+              data[position + i] = static_cast<char>(readByte());
               consumed += 1;
             }
           }
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -116,7 +116,7 @@ namespace orc {
   }
 
   /**
-   * Expand an array of bytes in place to the corresponding array of longs/.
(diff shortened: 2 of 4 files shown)


c++/src/ByteRLE.cc
```diff
@@ -354,7 +354,7 @@ namespace orc {
     if (bufferStart == bufferEnd) {
       nextBuffer();
     }
-    return *(bufferStart++);
+    return static_cast<signed char>(*(bufferStart++));
```
dongjoon-hyun, line 357: Unlike the other places, this is `static_cast<signed char>`?

dongjoon-hyun (commented):
Thank you for making a PR, @luffy-zh . BTW, according to the JIRA, do we need this at Apache ORC 1.8.1?

cc @wgtmac , @stiga-huang , @williamhyun .

c++/src/ByteRLE.cc
wgtmac: How about changing the return type from `signed char` to `int8_t`? But this may introduce a larger change.

wgtmac (approved):

wgtmac:
> 

I believe we need to port to 1.8.1.

dongjoon-hyun:
Got it. Thanks, @wgtmac .

c++/src/ColumnReader.cc
```diff
@@ -129,7 +129,7 @@ namespace orc {
       return;
     }
     for (size_t i = numValues - 1; i < numValues; --i) {
```
coderex2522, line 131: Very poor readability of code logic. The line should be optimized.

coderex2522, line 131: The line should be optimized. It looks a bit confusing.

coderex2522 (commented):
+1, LGTM.

wgtmac (approved):

dongjoon-hyun:
Hi, @wgtmac . Is there a reason why this is not backported to `branch-1.8`?
> I believe we need to port to 1.8.1. @dongjoon-hyun

And, do we need 2 patches to be complete for ORC-1315?
```
$ git log --oneline main | grep ORC-1315
91ded769 ORC-1315: [C++] Fix test failure when unsigned char is in effect
acb8d6ca ORC-1315: [C++] Fix byte to integer conversion failure on platforms with unsigned char type
```

dongjoon-hyun:
Sorry, I was confused. I found your comment, https://github.com/apache/orc/issues/1318#issuecomment-1328183678
> I can confirm that this issue only exists in the main branch, therefore we don't need backport the fix to branch-1.8.

I'll change the milestone of this PR.

PR #1338, opened 2022-12-10
ORC-1329: Add `OrcConf.getStringAsList` method

deshanxiao:
This PR is aimed to add a new OrcConf utility: getStringAsList.

We need to support a new OrcConf utility - getStringAsList, which can handle the scenario where the configuration item is a String, but the parameter received by the orc is a List.

UT


dongjoon-hyun (approved):
+1, LGTM. Thank you, @deshanxiao .
Merged to main/1.8.

PR #1355, opened 2023-01-07
ORC-1345: Use `makeBom` and skip snapshot check in GitHub Action `publish_snapshot` job

dongjoon-hyun:

This PR aims to improve GitHub Action `publish_snapshot` job by using `makeBom` and skipping checking snapshot versions. In addition, the download transfer progress is also disabled to reduce the log size.

`publish_snapshot` is always generating new set of snapshot files.

Manual review. This should be verified after merging.

https://github.com/apache/orc/blob/eee67e192017ee78fa89061ffc8a013c0daf14c7/.github/workflows/publish_snapshot.yml#L3-L6


dongjoon-hyun:
This is tested manually, but should be verified after merging. Let me merge this.

dongjoon-hyun:
After checking the result, it turns out that `Uploading` is also skipped mistakenly.

williamhyun (commented):
+1 Late LGTM

PR #1331, opened 2022-12-01
ORC-1324: Use Java 19 instead of 18 in GHA

williamhyun:
### What changes were proposed in this pull request?
This PR aims to use Java 19 instead of Java 18 in GHA testing.

### Why are the changes needed?
Java 19 is the latest version. 

### How was this patch tested?
Pass the new CIs.

PR #1358, opened 2023-01-08
ORC-1348: [C++] TimezoneImpl constructor should pass vector ref

zhjwpku:

timezone file might be hundreds or thousands of bytes, pass references of byte vector should be more efficient.
so change `TimezoneImpl(const std::string& name, const std::vector<unsigned char> bytes)` to
`TimezoneImpl(const std::string& name, const std::vector<unsigned char>& bytes)`.

this can reduce vector copy so improve performance.

### DIFF
--- a/c++/src/Timezone.cc
+++ b/c++/src/Timezone.cc
@@ -565,7 +565,7 @@ namespace orc {
 
   class TimezoneImpl : public Timezone {
    public:
-    TimezoneImpl(const std::string& name, const std::vector<unsigned char> bytes);
+    TimezoneImpl(const std::string& _filename, const std::vector<unsigned char>& buffer);
     virtual ~TimezoneImpl() override;
 
     /**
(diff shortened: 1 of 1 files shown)


dongjoon-hyun (commented):
Thank you for making a PR, @zhjwpku .

cc @wgtmac , @stiga-huang , @williamhyun

dongjoon-hyun (commented):
BTW, please preserve the Apache ORC PR description template like https://github.com/apache/orc/pull/1331 .

wgtmac (approved):
Good catch! LGTM +1 

Thanks @zhjwpku and @dongjoon-hyun

c++/src/Timezone.cc
```diff
@@ -565,7 +565,7 @@ namespace orc {
 
   class TimezoneImpl : public Timezone {
    public:
-    TimezoneImpl(const std::string& name, const std::vector<unsigned char> bytes);
+    TimezoneImpl(const std::string& name, const std::vector<unsigned char>& bytes);
```
wgtmac, line 568: nit: the function parameter names are inconsistent between declaration and definition. It would be nice to make them consistent.

dongjoon-hyun: +1 for the above comment.

zhjwpku: yeah, resolved.

dongjoon-hyun (approved):
+1, LGTM.

wgtmac:
I just committed it. Thanks @zhjwpku and @dongjoon-hyun!

PR #1370, opened 2023-01-10
ORC-1354: [C++] remove whitespace before the scope resolution operator

zhjwpku:

remove whitespace before the scope resolution operator

though spaces are allowed before/after the scope resolution operator, but this

is bad for the following case:

`CompressionStrategy ::CompressionStrategy_SPEED`

### DIFF
--- a/c++/src/Writer.cc
+++ b/c++/src/Writer.cc
@@ -179,7 +179,7 @@ namespace orc {
   }
 
   bool WriterOptions::getAlignedBitpacking() const {
-    return privateBits->compressionStrategy == CompressionStrategy ::CompressionStrategy_SPEED;
+    return privateBits->compressionStrategy == CompressionStrategy::CompressionStrategy_SPEED;
   }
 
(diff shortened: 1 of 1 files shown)


dongjoon-hyun (commented):
Can we enable some compiler-level check to prevent these kind of issues in the future, @zhjwpku ?

zhjwpku:
> Unfortunately, I have not found a way to enforce it by clang format.

Yeah, clang-format can remove the spaces after ::, but for the spaces before the ::, it always keeping one space.

dongjoon-hyun:
If we are unable to enforce this, is this worth for a JIRA and PR?
Sorry, but I'd recommend to ignore this kind of instances in C++ in the future, @wgtmac and @zhjwpku .

zhjwpku:
> If we are unable to enforce this programatically, is this worth for a JIRA and PR? Sorry, but I'd recommend to ignore this kind of instances in C++ in the future, @wgtmac and @zhjwpku .

Sure, agree. PR closed.

PR #1373, opened 2023-01-10
ORC-1343: Disable ENABLE_INDEXES

dongjoon-hyun:

This is a partial revert of ORC-1283.

ORC reader assumes built-in indexes always so far.

N/A

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/WriterImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/WriterImpl.java
@@ -188,7 +188,8 @@ public class WriterImpl implements WriterInternal, MemoryManager.Callback {
       this.rowIndexStride = 0;
     }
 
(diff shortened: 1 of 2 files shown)


dongjoon-hyun:
cc @deshanxiao and @PengleiShi

dongjoon-hyun:
cc @guiyanakuang and @williamhyun , too

guiyanakuang:
+1 LGTM.

BTW, with `orc.create.index` enabled in the future, I think we should, in any case, set `orc.row.index.stride` to 0 when `orc.create.index=false`.

Because this will ensure that readers of older versions can read files written by newer versions.

dongjoon-hyun:
Thank you, @deshanxiao . Merged to branch-1.7.

dongjoon-hyun:
Yes, we should proceed that very carefully, @guiyanakuang .

PR #1374, opened 2023-01-10
ORC-1355: [C++] Use const references in `addUserMetadata|printAttributes|stripPrefix`

zhjwpku:

change `void addUserMetadata(const std::string name, const std::string value) override;`
    to `void addUserMetadata(const std::string& name, const std::string& value) override;`
to reduce copy construct.

performance improvement.

This patch introduces no new features and all the exist test cases passed.

### DIFF
--- a/c++/include/orc/Writer.hh
+++ b/c++/include/orc/Writer.hh
@@ -289,7 +289,7 @@ namespace orc {
     /**
      * Add user metadata to the writer.
      */
-    virtual void addUserMetadata(const std::string name, const std::string value) = 0;
+    virtual void addUserMetadata(const std::string& name, const std::string& value) = 0;
   };
 }  // namespace orc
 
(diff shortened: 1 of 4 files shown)


dongjoon-hyun (changes requested):
It would be better to do this as a single PR instead of one-by-one, @zhjwpku .

zhjwpku:
> It would be better to do this as a single PR instead of one-by-one, @zhjwpku .

I am reading the codebase file by file, not sure how many this kind of drawbacks will be ;(

zhjwpku:
> It would be better to do this as a single PR instead of one-by-one, @zhjwpku .

hi @dongjoon-hyun , I have searched the codebase and found two other this kind
of bugs, seems these are no more others.

dongjoon-hyun:
cc @wgtmac , @stiga-huang , @williamhyun

wgtmac (approved):

williamhyun:
Thank you @zhjwpku I will merge this now.

PR #1387, opened 2023-01-30
ORC-1362: [Java] Enable ENABLE_INDEXES

deshanxiao:
Enable the functionality behind `orc.create.index`.

In https://github.com/apache/orc/pull/1373, we disabled the configuration "orc.create.index" because the orc reader cannot work well when orc.create.index=false.

During the write of a file with `orc.create.index=false` and `orc.row.index.stride > 0` no indexes are created in the ORC file. The `orc.row.index.stride` is persisted into the file metadata. During a read that requires the use of indexes e.g. in the presence of search arguments we get a read failure as the reader was expecting indexes based on the value of `orc.row.index.stride` but none exist.

Two changes are proposed here:
* Enable consistent writes i.e. assume orc.row.index.stride=0 if orc.create.index=false or if orc.row.index.stride >0 then assume orc.create.index=true
* Enable the reader to handle the absence of row index information. During the process of selecting row groups in a stripe if the index information is seen as missing then we assume that all row groups should be read as the index information is missing.

UT


java/core/src/java/org/apache/orc/impl/WriterImpl.java
```diff
@@ -189,8 +189,12 @@ public WriterImpl(FileSystem fs,
       this.rowIndexStride = 0;
     }
 
-    // ORC-1343: We ignore `opts.isBuildIndex` due to the lack of reader support
-    this.buildIndex = rowIndexStride > 0;
+    if (!opts.isBuildIndex() && rowIndexStride > 0) {
+      throw new IllegalArgumentException("Please set rowIndexStride to zero " +
```
deshanxiao, line 193: Or we could set the rowIndexStride=0 directly.

deshanxiao:
CC @dongjoon-hyun @guiyanakuang @pavibhai

java/core/src/java/org/apache/orc/impl/WriterImpl.java
guiyanakuang: So, based on this pr, it is not possible to write buildIndex = false and rowIndexStride > 0 , right?

java/core/src/test/org/apache/orc/impl/TestReaderImpl.java
```diff
@@ -527,4 +527,37 @@ public void testWithoutCompressionBlockSize() throws IOException {
       }
     }
   }
+
+  @Test
+  public void testSargSkipPickupGroupWithoutIndex() throws IOException {
+    Configuration conf = new Configuration();
+    Path[] paths = new Path[] {
+        new Path(workDir, "TestOrcFile.testSargSkipPickupGroupWithoutIndexCPlusPlus.orc"),
+        new Path(workDir, "TestOrcFile.testSargSkipPickupGroupWithoutIndexJava.orc"),
```
guiyanakuang, line 536: Please add some descriptions of the test file columns.

java/core/src/test/org/apache/orc/impl/TestReaderImpl.java
```diff
@@ -527,4 +527,37 @@ public void testWithoutCompressionBlockSize() throws IOException {
...
+    };
+    for (Path path: paths) {
+      FileSystem fs = path.getFileSystem(conf);
+      try (ReaderImpl reader = (ReaderImpl) OrcFile.createReader(path,
+          OrcFile.readerOptions(conf).filesystem(fs))) {
+
+        SearchArgument sarg = SearchArgumentFactory.newBuilder()
+            .startNot()
+            .lessThan("x", PredicateLeaf.Type.LONG, 100000L)
```
guiyanakuang, line 545: The test column data is [0-1023]. I think we should use a no-intersection condition to verify that the pickup is skipped.

dongjoon-hyun (changes requested):
Could you use a different title, @deshanxiao ? This PR only considers Java side, doesn't it?

java/core/src/java/org/apache/orc/impl/WriterImpl.java
deshanxiao: Yes, you are right.

deshanxiao:
> Could you use a different title, @deshanxiao ? This PR only considers Java side, doesn't it?

Done

java/core/src/test/org/apache/orc/impl/TestReaderImpl.java
deshanxiao: Thanks @guiyanakuang , done.

java/core/src/test/org/apache/orc/impl/TestReaderImpl.java
deshanxiao: Good catch! Done. Thanks @guiyanakuang

java/core/src/java/org/apache/orc/impl/WriterImpl.java
pavibhai: I was looking at the WriterOptions. Was thinking if we should integrate if there as well. There seem to be 2 places of interest right now.

Constructor. We currently have
```java
      buildIndex = OrcConf.ENABLE_INDEXES.getBoolean(tableProperties, conf);
      rowIndexStrideValue =
          (int) OrcConf.ROW_INDEX_STRIDE.getLong(tableProperties, conf);
```
We should set rowIndexStrideValue to 0 if `buildIndex` is false
 
Setter for rowIndexStride (I don't see a setter for buildIndex)
```java
    public WriterOptions rowIndexStride(int value) {
      rowIndexStrideValue = value;
      return this;
    }
```
Adapt this to say if rowIndexStride is 0 then set buildIndex false, else if rowIndexStride > 0 set buildIndex true.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -1290,7 +1290,12 @@ private void clearStreams() {
   private void readStripe() throws IOException {
     StripeInformation stripe = beginReadStripe();
     planner.parseStripe(stripe, fileIncluded);
-    includedRowGroups = pickRowGroups();
+    try {
+      includedRowGroups = pickRowGroups();
+    } catch (Throwable e) {
```
pavibhai, line 1295: this feels too broad to swallow any throwable. If the aim is to just deal with inconsistencies of not seeing an index stride and not seeing indexes.

I think a better change is inside the method `pickRowGroups` where after reading rowIndexes we test if row indexes are absent and treat that case as no row groups to skip.

pavibhai (changes requested):
Thanks for the patch @deshanxiao. The only item I have is on catch throwable which seems too broad that we are not propagating to callers.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
deshanxiao: Thank you @pavibhai 
My question is when `pickRowGroups `fails (not only the inconsistencies issue) should we continue to read the file ? Or throw an exception directly?

java/core/src/java/org/apache/orc/impl/WriterImpl.java
deshanxiao: @pavibhai Good suggesions, thank you, let me try.

deshanxiao: Now we have two options to support the configuration:
1. Throw an exception directly when the configuration mismatch.
2. Help users configure `orc.row.index.stride=0` when `orc.create.index=false` is configured, 

Using the first scheme will allow users to strongly perceive the orc.row.index.stride parameter, but it will increase the difficulty of use. Users need to manually set the orc.row.index.stride parameter.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
pavibhai: Let me give a split answer on this:

* When the inconsistent issue happens should we continue reading? Yes as this is well-defined on what the issue is i.e. row index stride is set but indexes have not been written.
* Should other issues in reading indexes be intercepted? The general answer here is no, as we don't know what the issue is that you are encountering so we don't know if it is correct to intercept it.

java/core/src/java/org/apache/orc/impl/WriterImpl.java
pavibhai: My preference would be for automatic determination so 2.

* if `orc.row.index.stride = 0` then set `orc.create.index=false`
* if `orc.row.index.stride > 0` then set `orc.create.index=true`

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
deshanxiao: OK, thank you @pavibhai.

java/core/src/java/org/apache/orc/impl/WriterImpl.java
deshanxiao: Thank you . Updated it.

deshanxiao:
Hi @pavibhai Sorry for my late update. I have modified all comments you mention. Could you please take a look?

pavibhai:
@deshanxiao thanks for the changes. I will get back to you soon.

java/core/src/test/org/apache/orc/impl/TestReaderImpl.java
```diff
@@ -527,4 +527,41 @@ public void testWithoutCompressionBlockSize() throws IOException {
...
     }
   }
+
+  @Test
+  public void testSargSkipPickupGroupWithoutIndex() throws IOException {
+    Configuration conf = new Configuration();
+    // We use ORC files in two languages to test, the previous Java version could not work
+    // well when stripeSize > 0 and orc.create.index=false, now it can skip these row groups.
+    Path[] paths = new Path[] {
+        // Writen by C++ API with schema struct<x:int,y:string> stripeSize=0
```
pavibhai, line 537: .nit do you mean row index size?

pavibhai (commented):
Almost there. I think one small typo

java/core/src/test/org/apache/orc/impl/TestReaderImpl.java
deshanxiao: Thank you @pavibhai  Fixed it.

pavibhai (approved):
+1 LGTM

pavibhai:
Thanks for the changes @deshanxiao.

While merging I felt that we should update the description of the PR. I have updated the description of the PR please check to see if this is acceptable.

cc @guiyanakuang @dongjoon-hyun

dongjoon-hyun (approved):
+1, LGTM for Apache ORC 1.9.0.

dongjoon-hyun:
Thank you so much, @deshanxiao , @guiyanakuang , @pavibhai !

Also, cc @wgtmac , @stiga-huang , @coderex2522 for C++ part compatibility. Please let us know if there is any concern for this.

dongjoon-hyun:
Additionally, cc @williamhyun

wgtmac:
I took a glance and found we may have a similar issue on the C++ side: https://github.com/apache/orc/blob/main/c%2B%2B/src/Writer.cc#L422

dongjoon-hyun:
Thank you for taking a look at it. Do you think we can have consistent behavior in both C++ and Java languages in Apache ORC 1.9?

wgtmac:
> Thank you for taking a look at it. Do you think we can have consistent behavior in both C++ and Java languages in Apache ORC 1.9?

Yes, I will fix it later this week.

wgtmac:
After a deep look, I can confirm that the C++ side does not have the issue.

The C++ writer only enables writing index with a positive rowIndexStride: https://github.com/apache/orc/blob/main/c%2B%2B/src/Writer.cc#L127

The C++ reader only creates SargsApplier with a positive rowIndexStride in the footer: https://github.com/apache/orc/blob/main/c%2B%2B/src/Reader.cc#L304

@dongjoon-hyun

dongjoon-hyun:
@wgtmac Thank you for the confirmation! Then, we are good.

deshanxiao:
Thank you very much! @pavibhai @dongjoon-hyun @guiyanakuang @wgtmac

PR #1402, opened 2023-02-08
ORC-1361: InvalidProtocolBufferException when reading large stripe statistics

zabetak:
Catch `InvalidProtocolBufferException` when parsing stripe statistics, log a warning message, and return an empty list.

In some cases ORC files may be created with very large Metadata section due to stripe statistics. The bigger the ORC file gets the easier is to endup with a large Metadata section. Nevertheless, it is still possible to hit the problem with smaller ORC files and `TestOrcWithLargeStripeStatistics` demonstrates some extremes.

Any attempt to read back the stripe statistics from the file will fail with `InvalidProtocolBufferException`. The exact exception may differ slighly and depending on: a) the size of stripe statistics; b) protobuf size limit.

Stripe statistics less than 2GB, and protobuf limit less than stats size:
```
com.google.protobuf.InvalidProtocolBufferException: Protocol message was too large.  May be malicious.  Use CodedInputStream.setSizeLimit() to increase the size limit.
```

Stripe statistics greater than 2GB, and protobuf limit 2GB (`Integer.MAX_VALUE`):
```
com.google.protobuf.InvalidProtocolBufferException: While parsing a protocol message, the input ended unexpectedly in the middle of a field.  This could mean either that the input has been truncated or that an embedded message misreported its own length.
```

The `Protocol message was too large` problem could be alleviated by increasing the limit as it was done in the past. However, increasing the limit would hide the underlying problem and probably lead to permanent metadata corruption in the near future. Moreover, the limit is already high enough (1GB) and bumping it further would lead to more memory being used potentially triggering `OutOfMemoryError`, OOM Killer, GC pauses, and other problems that are usually harder to debug and find the root cause.

When the stripe statistics exceeds the 2GB there is nothing to be done to parse back the statistics since protobuf cannot deserialize such messages (https://github.com/protocolbuffers/protobuf/issues/11729). The problem cannot be solved unless the metadata storage changes but this can only happen in newer versions.

On the other hand, stripe statistics are important but not vital for readers. In those situations where limits are breached it is acceptable to log a warning and return nothing instead of raising a fatal error to the caller.

Existing unit tests plus new tests added in `TestOrcWithLargeStripeStatistics`.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ReaderImpl.java
@@ -1043,7 +1043,7 @@ public class ReaderImpl implements Reader {
           InStream.createCodedInputStream(stream));
       return meta.getStripeStatsList();
     } catch (InvalidProtocolBufferException e) {
-      LOG.warn("Failed to parse stripe statistics; check ORC-1361 for more details.", e);
+      LOG.warn("Failed to parse stripe statistics", e);
       return Collections.emptyList();
     }
   }
--- a/java/core/src/java/org/apache/orc/impl/ReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ReaderImpl.java
@@ -19,6 +19,7 @@
 package org.apache.orc.impl;
 
 import com.google.protobuf.CodedInputStream;
+import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.TextFormat;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
@@ -55,6 +56,7 @@ import java.nio.ByteBuffer;
 import java.security.Key;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.function.Supplier;
@@ -1035,11 +1037,15 @@ public class ReaderImpl implements Reader {
       long offset,
       int length,
       InStream.StreamOptions options) throws IOException {
-    InStream stream = InStream.create("stripe stats", tailBuffer, offset,
-        length, options);
-    OrcProto.Metadata meta = OrcProto.Metadata.parseFrom(
-        InStream.createCodedInputStream(stream));
-    return meta.getStripeStatsList();
+    try (InStream stream = InStream.create("stripe stats", tailBuffer, offset,
+        length, options)) {
+      OrcProto.Metadata meta = OrcProto.Metadata.parseFrom(
+          InStream.createCodedInputStream(stream));
+      return meta.getStripeStatsList();
+    } catch (InvalidProtocolBufferException e) {
+      LOG.warn("Failed to parse stripe statistics; check ORC-1361 for more details.", e);
+      return Collections.emptyList();
+    }
   }
 
   private List<StripeStatistics> convertFromProto(List<OrcProto.StripeStatistics> list) {
--- a/java/core/src/test/org/apache/orc/TestOrcWithLargeStripeStatistics.java
+++ b/java/core/src/test/org/apache/orc/TestOrcWithLargeStripeStatistics.java
@@ -1,0 +1,115 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.orc;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.ql.exec.vector.BytesColumnVector;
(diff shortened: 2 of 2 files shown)


java/core/src/java/org/apache/orc/impl/ReaderImpl.java
```diff
@@ -1035,11 +1037,15 @@ private static List<OrcProto.StripeStatistics> deserializeStripeStats(
...
-    OrcProto.Metadata meta = OrcProto.Metadata.parseFrom(
-        InStream.createCodedInputStream(stream));
-    return meta.getStripeStatsList();
+    try (InStream stream = InStream.create("stripe stats", tailBuffer, offset,
+        length, options)) {
+      OrcProto.Metadata meta = OrcProto.Metadata.parseFrom(
+          InStream.createCodedInputStream(stream));
+      return meta.getStripeStatsList();
+    } catch (InvalidProtocolBufferException e) {
+      LOG.warn("Failed to parse stripe statistics; check ORC-1361 for more details.", e);
```
dongjoon-hyun, line 1046: This message could be wrong because `InvalidProtocolBufferException` can happen in another cases, right?

dongjoon-hyun (commented):
Thank you for making a PR, @zabetak . Do you think 2GB limitation exists in C++ code path?

cc @williamhyun , @wgtmac , @stiga-huang , @coderex2522 .

wgtmac:
> Thank you for making a PR, @zabetak . Do you think 2GB limitation exists in C++ code path?
> 
> cc @williamhyun , @wgtmac , @stiga-huang , @coderex2522 .

I think so. According to my experience, large message cannot be decoded with following warning:

> Reading dangerously large protocol message. If the message turns out to be larger than 67108864 bytes, parsing will be halted for security reasons. To increase the limit (or to disable these warnings), see CodedInputStream::SetTotalBytesLimit() in google/protobuf/io/coded_stream.h.

One has to workaround it manually by setting a larger limit.

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
deshanxiao: +1 
From [protobuf](https://github.com/protocolbuffers/protobuf/blob/520c601c99012101c816b6ccc89e8d6fc28fdbb8/java/core/src/main/java/com/google/protobuf/InvalidProtocolBufferException.java), there are many cases here.

zabetak: The first part of the message (`Failed to parse stripe statistics`) is accurate. The second part indeed can be misleading in some cases.

Instead of pointing to ORC-1361 what do you think of creating a page/section in the ORC website and document there whatever is needed? Then the message can contain the URL to this page.

zabetak:
@wgtmac Note that there is no workaround when the size reaches 2GB at least from the java perspective. The limit cannot be increased further and as I explained under ORC-1361 probably is not worth it.

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
wgtmac: Does the exception message provide detail information? It yes, we may not need to state `check ORC-1361 for more details` explicitly.

guiyanakuang: Maybe we just need to change the title to catch `InvalidProtocolBufferException` and let the program continue to run.
Reading large bars of statistical information is one of the causes of the exception. But throw any InvalidProtocolBufferException and this pr will allow the program to continue to run.

deshanxiao: We can keep running, but need to let the user know that we skipped these statistics. This can be achieved by catching exceptions, or requiring the user to configure a parameter such as skipCheckStatistics

zabetak: The details in the exception are probably enough for the user to understand that the stats are big/corrupted. What may be more difficult to understand is why the stats are big or why they were corrupted and what can they do about it. 

Providing some tips like check the strip size, number of columns, size of string stats, total size of the file, etc., should be useful and that's what I was thinking that could go in ORC-1361 or another resource.

I am fine with whatever you decide but I would strongly suggest to LOG at least the exception that was caught.

Let me know what you prefer and I will do the appropriate changes.

zabetak: It seems the content of the logging message is somewhat blocking this PR from being merged. I think this is a rather minor issue so how about keeping it dead simple (`Failed to parse stripe statistics`) and move the PR forward? WDYT?

deshanxiao:
I generally agree the PR although the PR does not directly solve the problem of large statistical information.
It can skip statistical information without affecting the normal reading of files.

zabetak:
@deshanxiao It doesn't solve the problem cause there is no a real fix :/ . The metadata structure has to change if we want to allow large metadata but this is relevant only for new versions of the ORC format. If there is enough interest we can try to brainstorm under https://github.com/apache/orc/issues/1404 or another place.

java/core/src/java/org/apache/orc/impl/ReaderImpl.java
wgtmac: > It seems the content of the logging message is somewhat blocking this PR from being merged. I think this is a rather minor issue so how about keeping it dead simple (`Failed to parse stripe statistics`) and move the PR forward? WDYT?

I'm +1 on this. Thanks @zabetak

zabetak: Simplified message in https://github.com/apache/orc/pull/1402/commits/55eebc6c6b779e6063528b3514670b33d46756cf

wgtmac (approved):

guiyanakuang (approved):

deshanxiao:
> @deshanxiao It doesn't solve the problem cause there is no a real fix :/ . The metadata structure has to change if we want to allow large metadata but this is relevant only for new versions of the ORC format. If there is enough interest we can try to brainstorm under #1404 or another place.

I see. +1 for the PR. Thank you @zabetak

deshanxiao (approved):

wgtmac:
I just submitted it. Thanks all!

PR #1403, opened 2023-02-08
ORC-1371: Remove unsupported SLF4J bindings from classpath

zabetak:
* Exclude `slf4j-log4j12` from Hadoop dependencies
* Exclude `logback-classic` from Zookeeper

Running `./mvnw clean install` shows a lot of warnings of the following form:
```
SLF4J: Class path contains SLF4J bindings targeting slf4j-api versions 1.7.x or earlier.
SLF4J: Ignoring binding found at [jar:file:/home/stamatis/.m2/repository/org/slf4j/slf4j-log4j12/1.7.10/slf4j-log4j12-1.7.10.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Ignoring binding found at [jar:file:/home/stamatis/.m2/repository/ch/qos/logback/logback-classic/1.2.10/logback-classic-1.2.10.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See https://www.slf4j.org/codes.html#ignoredBindings for an explanation.
```

The problem is that slf4j-log4j12-1.7.10.jar and logback-classic-1.2.10.jar are in the classpath but are incompatible with the current version of slf4j-api that ORC uses.

`./mvnw clean install 2>&1 | grep "Class path contains SLF4J bindings" | wc -l`

Before: 74
After: 0

### DIFF
--- a/java/mapreduce/pom.xml
+++ b/java/mapreduce/pom.xml
@@ -69,7 +69,6 @@
     <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-mapreduce-client-jobclient</artifactId>
-      <version>${min.hadoop.version}</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -227,6 +227,10 @@
             <groupId>tomcat</groupId>
             <artifactId>jasper-runtime</artifactId>
           </exclusion>
+          <exclusion>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
+          </exclusion>
         </exclusions>
       </dependency>
       <dependency>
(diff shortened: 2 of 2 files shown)


java/mapreduce/pom.xml
```diff
@@ -69,7 +69,6 @@
     <dependency>
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-mapreduce-client-jobclient</artifactId>
-      <version>${min.hadoop.version}</version>
```
dongjoon-hyun, line 72: Is this inevitably needed in this `SLF4J binding` PR? Otherwise, shall we move out from this PR?

dongjoon-hyun (commented):
Thank you for making a PR, @zabetak . I have only one comment about `java/mapreduce/pom.xml`.

dongjoon-hyun:
Also, cc @williamhyun

deshanxiao (commented):
Could you merge the change with origin source?

java/mapreduce/pom.xml
zabetak: We need to add an exclusion for `hadoop-mapreduce-client-jobclient`. This can be done here or in the `dependencyManagement` section in the parent `pom.xml`.

If we do it here then if someone in the future adds `hadoop-mapreduce-client-jobclient` in another module without copying the exclusion things will be broken again. The same can happen if `hadoop-mapreduce-client-jobclient` appears transitively from somewhere else. 

On the other hand, if we add `hadoop-mapreduce-client-jobclient` in the `dependencyManagement` we are sure that the problems mentioned previously will not happen.

In general it is a good practice to add all dependencies in the [dependency-management section|https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-management] so I am always in favor of doing that.

Opinions can differ though, so I am happy to make the changes if you think that it would be better to just add the exclusion here.

zabetak:
@deshanxiao See my comment here: https://github.com/apache/orc/pull/1403/files#r1101212637

dongjoon-hyun (approved):
+1, LGTM. Thank you, @zabetak and @deshanxiao .

dongjoon-hyun:
Merged to main branch for Apache ORC 1.9.0.

Welcome to the Apache ORC community. I added you to the Apache ORC contributor group and assigned ORC-1371 to you, @zabetak .

PR #1412, opened 2023-02-16
ORC-1373: Provide directional exceptions instead of NPE in case of `DynamicByteArray` length overflow

cxzl25:

When `DynamicByteArray` calculates `chunkIndex` overflow, it will throw NPE.
We can throw exception to remind users to avoid this problem by configuring what ORC parameters.

When the written string is very large, the grow calculation may overflow, causing the array data not to be expanded, and then NPE.

org.apache.orc.impl.DynamicByteArray#add(byte[], int, int)
```java
grow((length + valueLength) / chunkSize);
```

#### Log

```java
Caused by: java.lang.NullPointerException
	at java.lang.System.arraycopy(Native Method)
	at org.apache.orc.impl.DynamicByteArray.add(DynamicByteArray.java:115)
	at org.apache.orc.impl.StringRedBlackTree.addNewKey(StringRedBlackTree.java:48)
	at org.apache.orc.impl.StringRedBlackTree.add(StringRedBlackTree.java:60)
	at org.apache.orc.impl.writer.StringTreeWriter.writeBatch(StringTreeWriter.java:69)
	at org.apache.orc.impl.writer.StructTreeWriter.writeRootBatch(StructTreeWriter.java:56)
	at org.apache.orc.impl.WriterImpl.addRowBatch(WriterImpl.java:696)
```

#### Local Test
org.apache.orc.impl.TestDynamicArray#testBigByteArray
```java
 @Test
  public void testBigByteArray() {
    DynamicByteArray dba = new DynamicByteArray(128, 32 * 1024);

    byte[] val = new byte[1024];
    dba.add(val, 0, val.length);

    byte[] bigVal = new byte[Integer.MAX_VALUE - 16];
    dba.add(bigVal, 0, bigVal.length);
  }
```

#### Output

```java
java.lang.RuntimeException: chunkIndex overflow:-65535. You can set orc.column.encoding.direct=columnName, or orc.dictionary.key.threshold=0 to turn off dictionary encoding.
	at org.apache.orc.impl.DynamicByteArray.grow(DynamicByteArray.java:73)
	at org.apache.orc.impl.DynamicByteArray.add(DynamicByteArray.java:122)
	at org.apache.orc.impl.TestDynamicArray.testBigByteArray(TestDynamicArray.java:36)
```

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
+++ b/java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
@@ -53,8 +53,16 @@ public final class DynamicByteArray {
 
   /**
    * Ensure that the given index is valid.
+   * Throws an exception if chunkIndex is negative.
    */
   private void grow(int chunkIndex) {
+    if (chunkIndex < 0) {
+      throw new RuntimeException(String.format("chunkIndex overflow:%d. " +
+        "You can set %s=columnName, or %s=0 to turn off dictionary encoding.",
+        chunkIndex,
+        OrcConf.DIRECT_ENCODING_COLUMNS.getAttribute(),
+        OrcConf.DICTIONARY_KEY_SIZE_THRESHOLD.getAttribute()));
+    }
     if (chunkIndex >= initializedChunks) {
       if (chunkIndex >= data.length) {
         int newSize = Math.max(chunkIndex + 1, 2 * data.length);
@@ -64,12 +72,6 @@ public final class DynamicByteArray {
         data[i] = new byte[chunkSize];
       }
       initializedChunks = chunkIndex + 1;
-    } else if (chunkIndex < 0) {
-      throw new RuntimeException(String.format("chunkIndex overflow:%d. " +
-        "You can set %s=columnName, or %s=0 to turn off dictionary encoding.",
-        chunkIndex,
-        OrcConf.DIRECT_ENCODING_COLUMNS.getAttribute(),
-        OrcConf.DICTIONARY_KEY_SIZE_THRESHOLD.getAttribute()));
     }
   }
 
--- a/java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
+++ b/java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
@@ -18,6 +18,9 @@
 package org.apache.orc.impl;
 
 import org.apache.hadoop.io.Text;
+import org.apache.orc.OrcConf;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -30,6 +33,9 @@ import java.util.Arrays;
  * chunks that are allocated when needed.
  */
 public final class DynamicByteArray {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DynamicByteArray.class);
+
   static final int DEFAULT_CHUNKSIZE = 32 * 1024;
   static final int DEFAULT_NUM_CHUNKS = 128;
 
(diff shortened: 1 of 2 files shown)


java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
```diff
@@ -59,10 +65,16 @@ private void grow(int chunkIndex) {
         int newSize = Math.max(chunkIndex + 1, 2 * data.length);
         data = Arrays.copyOf(data, newSize);
       }
-      for(int i=initializedChunks; i <= chunkIndex; ++i) {
+      for (int i = initializedChunks; i <= chunkIndex; ++i) {
         data[i] = new byte[chunkSize];
       }
       initializedChunks = chunkIndex + 1;
+    } else if (chunkIndex < 0) {
+      LOG.error("chunkIndex overflow:{}. You can adjust the relevant configuration: {},{}.",
```
dongjoon-hyun, line 73: Thank you for adding this. Can we give some directional hints how to adjust those configurations, @cxzl25 ?

cxzl25: Usually this problem occurs in the production environment, I usually set `orc.dictionary.key.threshold=0`.  
Or find which field is a large string and skip it by `orc.column.encoding.direct=columnName`.   
Because sometimes it is difficult to find which field is a large string, at this time, we can configure `orc.column.encoding.direct=*`. This is equivalent to `orc.dictionary.key.threshold=0`.


How about this?
```bash
2023-02-15 23:37:26,658 [main] ERROR DynamicByteArray: chunkIndex overflow:-65535. You can set orc.column.encoding.direct=columnName, or orc.dictionary.key.threshold=0 to turn off dictionary encoding.
```

deshanxiao:
@cxzl25 
Could we throw a more accurate exception about it?

cxzl25:
> Could we throw a more accurate exception about it?

How about this?

```java
java.lang.RuntimeException: chunkIndex overflow:-65535. You can set orc.column.encoding.direct=columnName, or orc.dictionary.key.threshold=0 to turn off dictionary encoding.
	at org.apache.orc.impl.DynamicByteArray.grow(DynamicByteArray.java:73)
	at org.apache.orc.impl.DynamicByteArray.add(DynamicByteArray.java:122)
	at org.apache.orc.impl.TestDynamicArray.testBigByteArray(TestDynamicArray.java:36)
```

dongjoon-hyun:
+1 for @cxzl25 's new exception style because it was NPE already.

deshanxiao (approved):

java/core/src/java/org/apache/orc/impl/DynamicByteArray.java
```diff
@@ -59,10 +60,16 @@ private void grow(int chunkIndex) {
         int newSize = Math.max(chunkIndex + 1, 2 * data.length);
         data = Arrays.copyOf(data, newSize);
       }
-      for(int i=initializedChunks; i <= chunkIndex; ++i) {
+      for (int i = initializedChunks; i <= chunkIndex; ++i) {
         data[i] = new byte[chunkSize];
       }
       initializedChunks = chunkIndex + 1;
+    } else if (chunkIndex < 0) {
```
dongjoon-hyun, line 67: How many places do we have like the following?
```
grow((length + valueLength) / chunkSize);
```

dongjoon-hyun: If NPE happens always before, this looks like a precondition check for this method.
- Shall we place this at the beginning of this method as an independent `if` statement?
- Could you add your new logic to the function description too? The current one looks vague.
https://github.com/apache/orc/blob/d403e11b7eab126259fb0683c0eb5c6131fd347b/java/core/src/java/org/apache/orc/impl/DynamicByteArray.java#L54
```

BTW, I revise the PR title, @cxzl25 . You can change back if you want.

dongjoon-hyun (approved):
+1, LGTM. Thank you so much, @cxzl25 and @deshanxiao .

dongjoon-hyun:
Also, cc @williamhyun

PR #1413, opened 2023-02-17
Improve Specification documents

HuwCampbell:
The order of the data dictionary and length streams in dictionary columns is backwards compared to what all current writers emit; also, the readers gracefully handle absent streams (they're actually optional, as observed in the example files).

The documentation for nanosecond encoding was unclear and had broken sentences.

Improvement of documentation in specification documents.

### DIFF
--- a/site/specification/ORCv0.md
+++ b/site/specification/ORCv0.md
@@ -623,8 +623,8 @@ DIRECT        | PRESENT         | Yes      | Boolean RLE
               | LENGTH          | No       | Unsigned Integer RLE v1
 DICTIONARY    | PRESENT         | Yes      | Boolean RLE
               | DATA            | No       | Unsigned Integer RLE v1
-              | DICTIONARY_DATA | No       | String contents
-              | LENGTH          | No       | Unsigned Integer RLE v1
+              | LENGTH          | Yes      | Unsigned Integer RLE v1
+              | DICTIONARY_DATA | Yes      | String contents
 
 ## Boolean Columns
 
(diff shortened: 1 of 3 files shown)


dongjoon-hyun (commented):
Thank you for making a PR. It looks worthy having a JIRA. Could you file one, @HuwCampbell ?

cc @williamhyun , @wgtmac , @guiyanakuang , @deshanxiao

deshanxiao:
Hi @HuwCampbell 

In [  flushDictionary ](https://github.com/apache/orc/blob/d403e11b7eab126259fb0683c0eb5c6131fd347b/java/core/src/java/org/apache/orc/impl/writer/StringBaseTreeWriter.java#L155)

``` java
    context.writeBytes(stringOutput);
    lengthOutput.write(context.getLength());
    dumpOrder[context.getOriginalPosition()] = currentId++;
```

I saw that the `StringBaseTreeWriter` write `DICTIONARY_DATA` first and then write the `LENGTH`, right?

guiyanakuang:
https://github.com/apache/orc/blob/d403e11b7eab126259fb0683c0eb5c6131fd347b/java/core/src/java/org/apache/orc/impl/StreamName.java#L62-L78

https://github.com/apache/orc/blob/d403e11b7eab126259fb0683c0eb5c6131fd347b/proto/orc_proto.proto#L124-L129

> The order of the data dictionary and length streams in dictionary columns is backwards compared to what all current writers emit;

@deshanxiao ,  I think @HuwCampbell  is right in that the order in which the streams are flushed to the file is actually based on this `compareTo ` method. `LENGTH` is written before `DICTIONARY_DATA` if it is in the same column.

wgtmac:
On the C++ side, `DICTIONARY_DATA` is flushed before `LENGTH` stream. https://github.com/apache/orc/blob/main/c%2B%2B/src/ColumnWriter.cc#L1143

Actually the specs should not rule the order of the streams. In some implementations, streams of columns are even sorted to mitigate small I/Os.

HuwCampbell:
One can observe the streams with the orc tools.

I'll get around to a Jira in the coming days (it's Friday evening, so maybe Monday).

```
> orc-metadata --verbose examples/TestOrcFile.test1.orc
{
...
  "stripes": [
  ...
  "encodings": [
       ...
       { "column": 14, "encoding": "dictionary rle2", "count": 2 },
       ...
   ],
      "streams": [
        ...
        { "id": 37, "column": 14, "kind": "data", "offset": 683, "length": 6 },
        { "id": 38, "column": 14, "kind": "length", "offset": 689, "length": 6 },
        { "id": 39, "column": 14, "kind": "dictionary", "offset": 695, "length": 10 },
        ...
      ],
      ...
    }
  ] 
}

wgtmac:
BTW, do you have an interest to add https://github.com/HuwCampbell/orc-haskell to here: https://github.com/apache/orc/blob/main/proto/orc_proto.proto#L370 ? @HuwCampbell

guiyanakuang:
> Actually the specs should not rule the order of the streams. In some implementations, streams of columns are even sorted to mitigate small I/Os.

On the Java side, writers sort by Kind. But the reader doesn't care about the order. I agree with you.

dongjoon-hyun:
Given the current status, I'm also +1 for @wgtmac and @guiyanakuang 's comments. Maybe, shall we add an official note to the document?

deshanxiao:
Thank you @dongjoon-hyun  @guiyanakuang @wgtmac 

So, the order of these DICTIONARY doesn't really matter, right?
@HuwCampbell could you please update the document to add a note about it?

HuwCampbell:
That should be fine. I'll raise a Jira ticket first.

dongjoon-hyun:
Gentle ping, @HuwCampbell .

dongjoon-hyun:
The `specification` is spun off into a new `orc-format` repo. From now, we need to use `orc-format` as the ground-truth and maintains the files in `orc` repo by syncing before each releases.
- https://github.com/apache/orc-format/tree/main/specification

PR #1132, opened 2022-05-23
ORC-1182: Use `slf4j-simple` instead of deprecated `slf4j-log4j12`

dongjoon-hyun:
### What changes were proposed in this pull request?

This PR aims to use `slf4j-simple` instead of the deprecated `slf4j-log4j12`.

### Why are the changes needed?

https://www.slf4j.org/manual.html

> Binding/provider for [log4j version 1.2](http://logging.apache.org/log4j/1.2/index.html), a widely used logging framework. Given that log4j 1.x has been declared EOL in 2015 and again in 2022, as of SLF4J 1.7.35, the slf4j-log4j module automatically redirects to the slf4j-reload4j module at build time. Assuming you wish to continue to use the log4j 1.x framework, we strongly encourage you to use slf4j-reload4j instead. See below.

### How was this patch tested?

Pass the CI.

PR #1421, opened 2023-02-23
ORC-1378: [Java] Optimize logging dependency definition

deshanxiao:
This PR aims to optimize logging dependency definition.

Here are the changes:
1. Add slf4j-simple as the logging implementation for SLF4J in example module to remove warnings:
```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
```

2. Exclude log4j from hadoop dependencies to remove warnings in example module:
```
log4j:WARN No appenders could be found for logger (org.apache.hadoop.util.Shell).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```

3. Set slf4j-simple to runtime scope to avoid accidentally binding ORC
project to a single logging implementation.

4. Remove slf4j-simple declaration from bench module cause it is redundant.

5. Remove version from tools module since it is redundant.

When we run orc java examples, the following warning will show in output:
```
java -cp ./examples/target/orc-examples-1.9.0-SNAPSHOT-uber.jar org.apache.orc.examples.CoreWriter

SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
log4j:WARN No appenders could be found for logger (org.apache.hadoop.util.Shell).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info
```
This PR optimizes logging dependency definition.

After the PR:
```
java -cp ./examples/target/orc-examples-1.9.0-SNAPSHOT-uber.jar  org.apache.orc.examples.CoreWriter

org.apache.orc.examples.CoreWriter
Feb 23, 2023 7:03:48 PM org.apache.hadoop.util.NativeCodeLoader <clinit>
WARNING: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
[main] INFO org.apache.orc.impl.HadoopShimsCurrent - Can't get KeyProvider for ORC encryption from hadoop.security.key.provider.path.
[main] INFO org.apache.orc.impl.PhysicalFsWriter - ORC writer created for path: my-file.orc with stripeSize: 67108864 blockSize: 268435456 compression: Compress: ZLIB buffer: 262144
[main] INFO org.apache.orc.impl.WriterImpl - ORC writer created for path: my-file.orc with stripeSize: 67108864 options: Compress: ZLIB buffer: 262144

```

manual testing.

### DIFF
--- a/java/examples/pom.xml
+++ b/java/examples/pom.xml
@@ -63,6 +63,11 @@
       <groupId>org.apache.hive</groupId>
       <artifactId>hive-storage-api</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
--- a/java/bench/pom.xml
+++ b/java/bench/pom.xml
@@ -431,12 +431,6 @@
         <version>${slf4j.version}</version>
         <scope>runtime</scope>
       </dependency>
-      <dependency>
-        <groupId>org.slf4j</groupId>
-        <artifactId>slf4j-simple</artifactId>
-        <version>${slf4j.version}</version>
-        <scope>runtime</scope>
-      </dependency>
     </dependencies>
   </dependencyManagement>
 
--- a/java/core/pom.xml
+++ b/java/core/pom.xml
@@ -103,11 +103,6 @@
       <artifactId>auto-service</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-simple</artifactId>
-      <scope>provided</scope>
-    </dependency>
   </dependencies>
 
   <build>
--- a/java/core/pom.xml
+++ b/java/core/pom.xml
@@ -103,6 +103,11 @@
       <artifactId>auto-service</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/java/core/pom.xml
+++ b/java/core/pom.xml
@@ -103,6 +103,11 @@
       <artifactId>auto-service</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <scope>provided</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -231,6 +231,10 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
           </exclusion>
+          <exclusion>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
+          </exclusion>
         </exclusions>
       </dependency>
       <dependency>
@@ -286,6 +290,10 @@
           <exclusion>
             <groupId>xerces</groupId>
             <artifactId>xercesImpl</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
           </exclusion>
         </exclusions>
       </dependency>
@@ -382,6 +390,7 @@
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-simple</artifactId>
         <version>${slf4j.version}</version>
+        <scope>runtime</scope>
       </dependency>
       <dependency>
         <groupId>org.threeten</groupId>
--- a/java/tools/pom.xml
+++ b/java/tools/pom.xml
@@ -100,7 +100,6 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-simple</artifactId>
-      <version>${slf4j.version}</version>
     </dependency>
   </dependencies>
 


java/examples/pom.xml
```diff
@@ -53,16 +53,33 @@
       <groupId>org.apache.hadoop</groupId>
       <artifactId>hadoop-common</artifactId>
       <scope>compile</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>log4j</groupId>
+          <artifactId>log4j</artifactId>
+        </exclusion>
+      </exclusions>
```
zabetak, line 61: Isn't it better to put the exclusions in the main `pom.xml` file so that we don't have to define them for every module?

java/examples/pom.xml
```diff
@@ -53,16 +53,33 @@
...
     </dependency>
     <dependency>
       <groupId>org.apache.hive</groupId>
       <artifactId>hive-storage-api</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
```
zabetak, line 81: I noticed that slf4j-simple is declared at `compile` scope in the main `pom.xml`. Would't be sufficient to have it at `runtime` scope?

java/examples/pom.xml
deshanxiao: Great, let me try. Thank you @zabetak

deshanxiao:
Hi @zabetak, thank you for your suggestion, I update the PR. If you have time please take a look thanks~

deshanxiao:
cc @dongjoon-hyun @guiyanakuang

java/examples/pom.xml
zabetak: Actually, I meant to change the scope in `java/pom.xml` not here. Log bindings rarely need to be compile dependencies since we don't want projects to be tight into one specific log implementation.

guiyanakuang (changes requested):
> we don't want projects to be tight into one specific log implementation

+1

deshanxiao:
Sorry, I made a mistake. Fixed.

guiyanakuang:
> Sorry, I made a mistake. Fixed.

It's okay, other modules did it this way before. It's just that now I think @zabetak suggestion is better.

guiyanakuang:
https://github.com/guiyanakuang/orc/commit/66675a9bddecb76e3d19c4dfdf8934244ffe0e03

@deshanxiao , I made some attempts based on your pull request, what do you think about it.
cc @zabetak

deshanxiao:
> [guiyanakuang@66675a9](https://github.com/guiyanakuang/orc/commit/66675a9bddecb76e3d19c4dfdf8934244ffe0e03)
> 
> @deshanxiao , I made some attempts based on your pull request, what do you think about it. cc @zabetak

+1 for the change, Thank you. Please let me know if you think this deserves a new PR

deshanxiao:
If we remove the slf4j-simple from my PR, the slf4j may not be able to find the corresponding implementation when executing. This may have a warn log in the PR

deshanxiao:
I understand  @zabetak means to set it to runtime:
If there is a specific implementation class in the compiler phase, there will be a compilation error.

guiyanakuang:
@deshanxiao Please continue with the PR.

guiyanakuang:
The advantage of setting it to runtime is to avoid transitive dependencies, as projects that depend on Orc have their own logging implementation. We only set it in the parent pom.xml to avoid setting it in each submodule, and this way the dependencies won't be as tightly coupled.

deshanxiao:
> The advantage of setting it to runtime is to avoid transitive dependencies, as projects that depend on Orc have their own logging implementation. We only set it in the parent pom.xml to avoid setting it in each submodule, and this way the dependencies won't be as tightly coupled.

1. I found that there are two slf4j-simple dependencies in the pom file, one in dependency and one in dependencymanagement. I remove the dependencymanagement one.
2. If we add dependencies in parent pom. All modules (tools bench example) that depend on the pom.xml file will automatically inherit this dependency

zabetak:
I think it would be simpler if I would show you what I had in mind:
https://github.com/zabetak/orc/tree/ORC-1378

zabetak (changes requested):
The current PR also contains some files that shouldn't be there:
* `.my-file.orf.crc`
* `my-file.orc`
I somehow missed to spot them in the first pass.

deshanxiao:
> The current PR also contains some files that shouldn't be there:
> 
> * `.my-file.orf.crc`
> * `my-file.orc`
>   I somehow missed to spot them in the first pass.

yeah, It is good to fix git ignore to solve the problem.

zabetak:
> If we add dependencies in parent pom (not dependencymanagement). All modules (tools bench example) that depend on the pom.xml file will automatically inherit this dependency.

Note that adding the dependency in the global dependency section it will also lead to many log (INFO) messages appearing when running tests; not sure if this is desired.

deshanxiao:
> I think it would be simpler if I would show you what I had in mind: https://github.com/zabetak/orc/tree/ORC-1378

@guiyanakuang @zabetak The biggest difference between this PR and @zabetak 's PR ’is whether to add dependency (not dependency management) to pom.

image

deshanxiao:
image

guiyanakuang:
> > I think it would be simpler if I would show you what I had in mind: https://github.com/zabetak/orc/tree/ORC-1378
> 
> @guiyanakuang @zabetak The biggest difference between this PR and @zabetak 's PR ’is whether to add dependency (not dependency management) to pom.
> 
> image

Sorry, in the example I provided, I added the slf4j-simple with the runtime scope in the parent pom.xml dependencies instead of the dependency management section. I now realize that this could be harmful. Before this change, orc-core did not transitively pass any log implementation jar. After adding slf4j-simple to the parent pom.xml dependencies, it will transitively pass it, which may cause a log library conflict for projects that depend on orc-core. Therefore, I think @zabetak's approach is better.

Additionally, we can add slf4j-simple with the provided scope to orc-core, which can avoid transitive dependencies and also enable logging output when testing orc-core.

deshanxiao:
Thank you @zabetak. I have merged your commit and add provicde scope in orc-core.

deshanxiao:
Test has passed:
image

deshanxiao:
@guiyanakuang @zabetak I have changed the title and description quoted by @zabetak. Please take a look.

This PR should own to @zabetak.

Thank you very much @guiyanakuang @zabetak

guiyanakuang:
@deshanxiao , please modify the PR title to match the content. Because this PR not only solves the problem in the example module. Perhaps the title should be "Optimize logging dependency definition for ORC project."

guiyanakuang:
cc @dongjoon-hyun

guiyanakuang (approved):
+1 LGTM

zabetak (commented):
> After adding slf4j-simple to the parent pom.xml dependencies, it will transitively pass it,

The transitive inheritance can be avoided by marking the dependency as `optional`; probably not worth doing now but it is a useful tip to keep in mind for the future.

>Additionally, we can add slf4j-simple with the provided scope to orc-core, which can avoid transitive dependencies and also enable logging output when testing orc-core.

If we want to enable logging for tests then the best would be to declare the dependency in `test` scope; `provided` is a bit misleading. Furthermore, I think it is better to do it as part of another PR (to keep the scope here minimal) cause enable logging for tests could be something useful for all modules not only core I guess. 

>This PR should own to @zabetak.

@deshanxiao It is a team effort and you had the initiative. It's not necessary to worry about ownership ;)

dongjoon-hyun (commented):
I don't think the following is a precise claim because #1132 only changed `bench` modules which is not published to the Maven Central. It's just one of the `test` module which is irrelevant to the `example` module.
> When we run orc java examples, the following error will show in output:
> ...
> After https://github.com/apache/orc/pull/1132, we use slf4j-simple instead of log4j.

Could you revise the PR description a little more? Or, did I missed something?

deshanxiao:
> I don't think the following is a precise claim because #1132 only changed `bench` modules which is not published to the Maven Central. It's just one of the `test` module which is irrelevant to the `example` module.
> 
> > When we run orc java examples, the following error will show in output:
> > ...
> > After #1132, we use slf4j-simple instead of log4j.
> 
> Could you revise the PR description a little more? Or, did I miss something?

deshanxiao:
> I don't think the following is a precise claim because #1132 only changed `bench` modules which is not published to the Maven Central. It's just one of the `test` module which is irrelevant to the `example` module.
> 
> > When we run orc java examples, the following error will show in output:
> > ...
> > After #1132, we use slf4j-simple instead of log4j.
> 
> Could you revise the PR description a little more? Or, did I miss something?

Thank you @dongjoon-hyun.  You are right, let me fix it.

guiyanakuang:
@zabetak Thank you for the reminder. You're right. At that time, I found that only orc-core was unable to output log information during testing (because other modules already had runtime simple-slf4j dependencies). Using "test" instead of "provided" is better. I'm sorry for providing misleading information.

deshanxiao:
@zabetak @guiyanakuang 
The following is the latest slf4j implementation class of each module:

- orc-core : slf4j-simple:jar:2.0.6:test
- orc-tools : slf4j-simple:jar:2.0.6:runtime
- orc-examples : slf4j-simple:jar:2.0.6:runtime

dongjoon-hyun (approved):
+1, LGTM from my side. Thank you all!

As a last piece, I revise the PR description like the following because WARNING is not ERROR.
```
- When we run orc java examples, the following error will show in output:
+ When we run orc java examples, the following warning will show in output:
```

Also, I validated this PR with one of our downstream projects; the Apache Spark's dependency check script. There is no unintentional change from there as discussed here.

deshanxiao:
> 

Thank you @dongjoon-hyun

deshanxiao:
@dongjoon-hyun @guiyanakuang @zabetak Please let me know if you have any comments. Thank you!

zabetak (approved):
LGTM, don't have anything else to add!

guiyanakuang:
@deshanxiao thank you for your efforts, thank you @zabetak and @dongjoon-hyun for the review.  I just merged it into the main branch.

PR #1427, opened 2023-03-03
ORC-1376: [C++] Scaffolding of schema evolution

ffacs:
Add schema evolution and check PPD security

This pr adds a class `schema evolution` which checks validity and  PDD security of a type conversion. The convertions would be finished later.

To support schema evolution in c++

UT

### DIFF
--- a/c++/include/orc/Type.hh
+++ b/c++/include/orc/Type.hh
@@ -23,8 +23,6 @@
 #include "orc/Vector.hh"
 #include "orc/orc-config.hh"
 
-#include <optional>
-
 namespace orc {
 
   enum TypeKind {
@@ -67,7 +65,12 @@ namespace orc {
     virtual std::vector<std::string> getAttributeKeys() const = 0;
     virtual std::string getAttributeValue(const std::string& key) const = 0;
     virtual std::string toString() const = 0;
-    virtual std::optional<const Type*> getTypeByColumnId(uint64_t colId) const = 0;
+    /**
+     * Get the Type with the given column ID
+     * @param colId the column ID
+     * @return the type corresponding to the column Id, nullptr if not exists
+     */
+    virtual const Type* getTypeByColumnId(uint64_t colId) const = 0;
 
     /**
      * Create a row batch for this type.
--- a/c++/include/orc/Type.hh
+++ b/c++/include/orc/Type.hh
@@ -23,6 +23,8 @@
 #include "orc/Vector.hh"
 #include "orc/orc-config.hh"
 
+#include <optional>
+
 namespace orc {
 
   enum TypeKind {
@@ -65,6 +67,7 @@ namespace orc {
     virtual std::vector<std::string> getAttributeKeys() const = 0;
     virtual std::string getAttributeValue(const std::string& key) const = 0;
     virtual std::string toString() const = 0;
+    virtual std::optional<const Type*> getTypeByColumnId(uint64_t colId) const = 0;
 
     /**
      * Create a row batch for this type.
--- a/c++/src/SchemaEvolution.cc
+++ b/c++/src/SchemaEvolution.cc
@@ -1,0 +1,259 @@
+#include "SchemaEvolution.hh"
+#include "orc/Exceptions.hh"
+
+#include <unordered_map>
+#include <unordered_set>
+
+namespace orc {
+
+  SchemaEvolution::SchemaEvolution(const std::shared_ptr<Type>& _readType, const Type* fileType)
+      : readType(_readType) {
+    if (readType) {
+      buildConversion(readType.get(), fileType);
+    } else {
+      for (uint64_t i = 0; i <= fileType->getMaximumColumnId(); ++i) {
+        safePPDConversionMap.insert(i);
+      }
+    }
+  }
+
+  const Type* SchemaEvolution::getReadType(const Type& fileType) const {
+    auto ret = readTypeMap.find(fileType.getColumnId());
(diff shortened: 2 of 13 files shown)


wgtmac:
Thanks @ffacs for contribution! I will take a look next week.

cc @stiga-huang @coderex2522

c++/include/orc/Type.hh
```diff
@@ -65,6 +65,7 @@ namespace orc {
     virtual std::vector<std::string> getAttributeKeys() const = 0;
     virtual std::string getAttributeValue(const std::string& key) const = 0;
     virtual std::string toString() const = 0;
+    virtual const Type* getTypeByColumnId(uint64_t colId) const = 0;
```
wgtmac, line 68: We have supported C++17 now. It can be replaced by std::optional. Or we need to add a comment to say what will happen if the column does not exist.

c++/include/orc/Type.hh
```diff
@@ -97,6 +98,13 @@ namespace orc {
     static std::unique_ptr<Type> buildTypeFromString(const std::string& input);
   };
 
+  struct EnumClassHash {
```
wgtmac, line 101: Can we move it into `.cc` file?

c++/src/SchemaEvolution.cc
```diff
@@ -0,0 +1,233 @@
+#include "SchemaEvolution.hh"
+#include <unordered_map>
+#include <unordered_set>
+#include "orc/Exceptions.hh"
```
wgtmac, line 4:
```suggestion
#include "SchemaEvolution.hh"
#include "orc/Exceptions.hh"
#include <unordered_map>
#include <unordered_set>
```

c++/src/SchemaEvolution.cc
```diff
@@ -0,0 +1,233 @@
...
+        // support nothing now
+    };
+    auto iter = SUPPORTED_CONVERSIONS.find(fileType.getKind());
+    if (iter == SUPPORTED_CONVERSIONS.cend()) {
+      return false;
+    }
+    return iter->second.find(readType.getKind()) != iter->second.cend();
+  }
+
+  // return value: <valid convert, need convert>
```
wgtmac, line 44: Can we use a struct for better readability?

wgtmac (commented):
This is looking good. I have left some minor comments.

As this is the 1st patch of a large feature. I expect there are more patches coming. Could you please open a separate JIRA for each followup work item?

ffacs:
I`ve uploaded a new patch, please take a look. @wgtmac @coderex2522 @stiga-huang

c++/include/orc/Type.hh
```diff
@@ -65,6 +67,7 @@ namespace orc {
     virtual std::vector<std::string> getAttributeKeys() const = 0;
     virtual std::string getAttributeValue(const std::string& key) const = 0;
     virtual std::string toString() const = 0;
+    virtual std::optional<const Type*> getTypeByColumnId(uint64_t colId) const = 0;
```
wgtmac, line 70: I didn’t notice that this is a public header. As Apache impala is still on c++14 and ‘optional’ is a c++17 feature, it will break impala. Could you revert this and add a comment for the possible nullptr of return? Sorry about that.

wgtmac (changes requested):
It is better to add more commits instead of a forced push. This makes it harder to find what changes are made compared to previous commits.

Overall looks good to me.

c++/include/orc/Type.hh
ffacs: done

dongjoon-hyun (changes requested):
Thank you for making a PR, @ffacs .

This PR seems to introduce a breaking change to `SargsApplier::SargsApplier` constructor. Could you provide a backward compatibility?

wgtmac (approved):
LGTM. Thanks @ffacs!

c++/src/SchemaEvolution.cc
```diff
@@ -0,0 +1,259 @@
+#include "SchemaEvolution.hh"
+#include "orc/Exceptions.hh"
+
+#include <unordered_map>
```
coderex2522, line 4: These two header files are already included in the SchemaEvolution.hh. It is a redundant code.

c++/src/SchemaEvolution.hh
```diff
@@ -0,0 +1,45 @@
+#ifndef ORC_SCHEMA_EVOLUTION_HH
+#define ORC_SCHEMA_EVOLUTION_HH
+
+#include <unordered_map>
+#include <unordered_set>
+#include "orc/Type.hh"
```
coderex2522, line 6: It is recommended that this line ' #include "orc/Type.hh" ' be moved to precede line ' #include <unordered_map>'.

coderex2522 (commented):
Thanks @ffacs! And I've left some formatting comments.

c++/src/SchemaEvolution.hh
ffacs: done

c++/src/SchemaEvolution.cc
ffacs: donwe

coderex2522:
+1, LGTM.

dongjoon-hyun:
Thank you all!

dongjoon-hyun:
Is the JIRA ID correct?

wgtmac:
> Is the JIRA ID correct?

It seems that there isn't any JIRA for this subtask.

PR #1431, opened 2023-03-07
ORC-1384: Fix `ArrayIndexOutOfBoundsException` when reading dictionary stream bigger then dictionary

zratkai:
Avoid  ArrayIndexOutOfBoundsException when reading dictionary stream bigger then dictionary. Check the size of the dictionary and input and read only the min of those.

In Hive when reading with LLAP data is read in 4kB blocks which leads to ArrayIndexOutOfBoundsException when the dictionary is smaller.

It is tested with HIVE's qtest, since here we do not have the necessary subclasses.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
@@ -2292,10 +2292,15 @@ public class TreeReaderFactory {
           int dictionaryBufferSize = dictionaryOffsets[dictionaryOffsets.length - 1];
           dictionaryBuffer = new byte[dictionaryBufferSize];
           int pos = 0;
-          int chunkSize = in.available();
-          byte[] chunkBytes = new byte[chunkSize];
+          //check if dictionary size is smaller than available stream size
+          // to avoid ArrayIndexOutOfBoundsException
+          int readSize = Math.min(in.available(), dictionaryBufferSize);
+          byte[] chunkBytes = new byte[readSize];
           while (pos < dictionaryBufferSize) {
-            int currentLength = in.read(chunkBytes, 0, chunkSize);
+            int currentLength = in.read(chunkBytes, 0, readSize);
+            //check if dictionary size is smaller than available stream size
+            // to avoid ArrayIndexOutOfBoundsException
+            currentLength = Math.min(currentLength, dictionaryBufferSize - pos);
             System.arraycopy(chunkBytes, 0, dictionaryBuffer, pos, currentLength);
             pos += currentLength;
           }
--- a/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
@@ -2292,13 +2292,13 @@ public class TreeReaderFactory {
           int dictionaryBufferSize = dictionaryOffsets[dictionaryOffsets.length - 1];
           dictionaryBuffer = new byte[dictionaryBufferSize];
           int pos = 0;
-          //check if dictionary size is smaller than available stream size
+          // check if dictionary size is smaller than available stream size
           // to avoid ArrayIndexOutOfBoundsException
           int readSize = Math.min(in.available(), dictionaryBufferSize);
           byte[] chunkBytes = new byte[readSize];
           while (pos < dictionaryBufferSize) {
             int currentLength = in.read(chunkBytes, 0, readSize);
-            //check if dictionary size is smaller than available stream size
+            // check if dictionary size is smaller than available stream size
             // to avoid ArrayIndexOutOfBoundsException
             currentLength = Math.min(currentLength, dictionaryBufferSize - pos);
             System.arraycopy(chunkBytes, 0, dictionaryBuffer, pos, currentLength);


java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -2292,10 +2292,11 @@ private void readDictionaryStream(InStream in) throws IOException {
           int dictionaryBufferSize = dictionaryOffsets[dictionaryOffsets.length - 1];
           dictionaryBuffer = new byte[dictionaryBufferSize];
           int pos = 0;
-          int chunkSize = in.available();
-          byte[] chunkBytes = new byte[chunkSize];
+          //check if dictionary size is smaller than available stream size to avoid ArrayIndexOutOfBoundsException
+          int readSize = Math.min(in.available(), dictionaryBufferSize);
+          byte[] chunkBytes = new byte[readSize];
```
guiyanakuang, line 2297: The previous reading code was implemented on the premise that the total amount of data in the stream is the same as the expected amount to be read. I am curious as to why `in.available()` is greater than `dictionaryBufferSize` in your environment.

deshanxiao: I'm also curious, `available `currently returns the remaining value of the Bytebuffer, could you provide the specific scenario?
@zratkai

deshanxiao:
Thank you for making the PR.
Could you make the CI happy with `mvn checkstyle:check -Panalyze -Pbenchmark` ? @zratkai

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
zratkai: In Hive the ORC's Reader classes (e.g. StringTreeReader)  have sublclasses and ORC file is read with LLAP. LLAP uses a default 4kB blocks to read, and this causes that the in.available() can be bigger than dictionary size. (If dictionary is smaller than 4kB.) This fix will not cause any issue if ORC file is read without LLAP, only with ORC classes, since in System.arraycopy() the maximum amount which can be copied can not be bigger than destination or source size. The previous reading mode created a bigger destination array in Hive than necessary.

guiyanakuang: @zratkai Thank you for your detailed explanation. I understand your use case.

I think `in.available()` will be recalculated every time a new block is read, right? So this fix seems not perfect.
For example, at the beginning,
`in.available()` = 4k
`dictionaryBufferSize` = 5k
`readSize` = 4k
Entering the loop,
the first time` currentLength` is read as 4k, and then copied.
The second time `currentLength` may still be 4k, which will still cause an out of bounds error.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -2292,10 +2292,12 @@ private void readDictionaryStream(InStream in) throws IOException {
...
-          byte[] chunkBytes = new byte[chunkSize];
+          //check if dictionary size is smaller than available stream size
+          // to avoid ArrayIndexOutOfBoundsException
+          int readSize = Math.min(in.available(), dictionaryBufferSize);
+          byte[] chunkBytes = new byte[readSize];
           while (pos < dictionaryBufferSize) {
-            int currentLength = in.read(chunkBytes, 0, chunkSize);
+            int currentLength = in.read(chunkBytes, 0, readSize);
             System.arraycopy(chunkBytes, 0, dictionaryBuffer, pos, currentLength);
             pos += currentLength;
```
guiyanakuang, line 2302: Would this fix be better?
```suggestion
            currentLength = Math.min(currentLength, dictionaryBufferSize - pos);
            System.arraycopy(chunkBytes, 0, dictionaryBuffer, pos, currentLength);
            pos += currentLength;
```

deshanxiao:
Actually, the issue blocked the upgrade of ORC in Hive:
https://github.com/apache/hive/pull/3833

zratkai:
@deshanxiao thanks for adding this info. Yes, exactly, this was the original issue I started with. To finish it this fix is needed.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
guiyanakuang: > Actually this won't be a problem, because if the readSize is bigger than the available input, it just returns the available here.

The example I gave was specific to the LLAP environment. Could you test it out? I'm not worried about other cases, as available is always accurate.

zratkai: @guiyanakuang ok, I understand it now. I combined my changes with your suggestions. Combining this two reduces the memory usage also. With that this cases are covered:

Case 1: 
in.available() = 4096
dictionaryBuffer = 900
readSize = 900
It  creates a chunkBytes array only with 900 bytes (instead of 4096 bytes) and reads 900 bytes in one run.

Case2:
in.available() = 4k
dictionaryBufferSize = 5k
readSize = 4k
It  creates a chunkBytes with 4kB and reads 4kB at first run, then it reads 1kB and copies the remaining 1kB to the dictionaryBuffer.

dongjoon-hyun (commented):
Thank you for making a PR, @zratkai .
- Do you happen to know what caused this regression?
- Do you think you can add a small unit test which you described in the commets?

dongjoon-hyun:
Also cc @williamhyun

zratkai:
@dongjoon-hyun 

> Thank you for making a PR, @zratkai .
> 
> * Do you happen to know what caused this regression?
> * Do you think you can add a small unit test which you described in the commets?

1. It is caused by LLAP in Hive, since it reads in 4kB blocks as I described above.
2. I was thinking about that, but it is not easy for this modification, since it is in a private method, and we don't have any test even for the public method which calls this (org.apache.orc.impl.TreeReaderFactory.StringDictionaryTreeReader#nextVector) . So it would mean I need to create for that as well from scratch.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
zratkai: @guiyanakuang can we close this thread? Can you merge this PR?

dongjoon-hyun (approved):
+1, LGTM.
Merged to main/1.8 for Apache ORC 1.8.3

dongjoon-hyun:
Thank you, @zratkai and @guiyanakuang .

Also, cc @williamhyun

dongjoon-hyun:
BTW, welcome to the Apache ORC community, @zratkai .
ORC-1384 is resolved and assigned to you.

PR #1432, opened 2023-03-09
ORC-1393: Add `reset(DiskRangeList input, long length)` to `InStream` impl class

difin:

Adding a possibility to re-set the length of the UncompressedStream when calling reset() on it.

In short, in some cases, after resetting an UncompressedStream, its actual length is longer than its initial length.

Before 'ORC-516 - Update InStream for column compression', InStream.UncompressedStream class had 'length' field and the length was modifiable in reset() method. ORC-516 was the root cause of this regression reported in HIVE-27128.

The `reset()` method was used in `SettableUncompressedStream` class in `setBuffers()` method:

```java
public void setBuffers(DiskRangeInfo diskRangeInfo) {
  reset(diskRangeInfo.getDiskRanges(), diskRangeInfo.getTotalLength());
  setOffset(diskRangeInfo.getDiskRanges());
}
```
After Orc version upgrade in Hive to 1.6.7., and since `SettableUncompressedStream` class was removed from Orc code base, Hive manages it own copy of `SettableUncompressedStream` which doesn't pass new length to `UncompressedStream` when calling reset (because `UncompressedStream` doesn't accept new length any more in the reset method):

```java
public void setBuffers(DiskRangeInfo diskRangeList) {
  reset(diskRangeList.getDiskRanges());
  setOffset(diskRangeList.getDiskRanges());
}
```
When investigating the issue reported in HIVE-27128 and comparing the lengths of the InStream.UncompressedStream prior to the upgrade of ORC version in Hive to 1.6.7. (which included ORC-516) and after I noticed that the issue happens with ORC-516 changes because the length of the InStream.UncompressedStream is set once for all row groups, while without those changes the length is dynamic and sometimes is set to bigger value than the initial value.

1. Passing CI pipeline tests
2. Running Hive Q-test https://github.com/difin/hive/commits/orc_read_err_qtest on the local dev env with a change in Hive's SettableUncompressedStream to pass  diskRangeList.getTotalLength() to the reset method of UncompressedStream.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/InStream.java
+++ b/java/core/src/java/org/apache/orc/impl/InStream.java
@@ -42,7 +42,7 @@ public abstract class InStream extends InputStream {
 
   protected final Object name;
   protected final long offset;
-  protected final long length;
+  protected long length;
   protected DiskRangeList bytes;
   // position in the stream (0..length)
   protected long position;
@@ -87,6 +87,26 @@ public abstract class InStream extends InputStream {
     setCurrent(input, true);
   }
 
+  /**
+   * Reset the input to a new set of data with a different length.
+   *
+   * in some cases, after resetting an UncompressedStream, its actual length is longer than its initial length.
+   * Prior to ORC-516, InStream.UncompressedStream class had the 'length' field and the length was modifiable in
+   * the reset() method. It was used in SettableUncompressedStream class in setBuffers() method.
+   * SettableUncompressedStream was passing 'diskRangeInfo.getTotalLength()' as the length to the reset() method.
+   * SettableUncompressedStream had been removed from ORC code base, but it is required for Apache Hive and
+   * Apache Hive manages its own copy of SettableUncompressedStream since upgrading its Apache ORC version to 1.6.7.
+   * ORC-516 was the root cause of the regression reported in HIVE-27128 - EOFException when reading DATA stream.
+   * This wrapper method allows to resolve HIVE-27128.
+   *
+   * @param input the input data
+   * @param length new length of the stream
+   */
+  protected void reset(DiskRangeList input, long length) {
+    this.length = length;
+    reset(input);
+  }
+
   public abstract void changeIv(Consumer<byte[]> modifier);
 
   static int getRangeNumber(DiskRangeList list, DiskRangeList current) {
@@ -122,7 +142,7 @@ public abstract class InStream extends InputStream {
                               long offset,
                               long length) {
       super(name, offset, length);
-      reset(input);
+      reset(input, length);
     }
 
     @Override
--- a/java/core/src/test/org/apache/orc/impl/TestInStream.java
+++ b/java/core/src/test/org/apache/orc/impl/TestInStream.java
@@ -957,4 +957,47 @@ public class TestInStream {
       assertEquals((byte)i, inBuffer[i], "position " + i);
     }
   }
+
+  private static final byte[] uncompressed = input(
+          0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
+  
+  @Test
+  public void testStreamResetWithIncreasedLength() throws IOException {
+    // Set up an initial buffer of PREVIOUS_LENGTH followed by our stream
+    // at START.
+    final long START = 1_000;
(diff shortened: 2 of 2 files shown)


java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -106,7 +111,7 @@ public static class UncompressedStream extends InStream {
     protected ByteBuffer decrypted;
     protected DiskRangeList currentRange;
     protected long currentOffset;
-
+
```
dongjoon-hyun, line 114: Could you make CI happy, @difin ?
```
Error:  src/java/org/apache/orc/impl/InStream.java:[114] (regexp) RegexpSingleline: No trailing white space allowed.
```

difin: Sure, done.

java/core/src/java/org/apache/orc/impl/InStream.java
```diff
@@ -87,6 +87,11 @@ protected void reset(DiskRangeList input) {
     setCurrent(input, true);
   }
 
+  protected void reset(DiskRangeList input, long length) {
```
dongjoon-hyun, line 90: Please add a function description for this method because this method will be used as a wrapper from now.

dongjoon-hyun (approved):
Looks good to me. Although we don't have a test coverage here, I believe this patch is fully tested with Hive tests, @difin .

1. I copied your JIRA description into the PR description to make a rich commit log.
2. I added one minor comment about a function description. You can reuse the most of the existing comments. Please add the background of this new wrapper `reset` method, especially, the increased length.

dongjoon-hyun:
cc @omalley , @williamhyun , @guiyanakuang

dongjoon-hyun:
Although this `impl` should not be used outside, I believe we need this because Apache Hive is the mother of Apache ORC initially.

deshanxiao:
Could we add a simple unit test like this:
Invocation the new `reset(DiskRangeList input, long length)` and test the real avaliable bypes of Instream?

java/core/src/java/org/apache/orc/impl/InStream.java
difin: Thanks for your comment, done.

difin:
I've added unit tests as suggested.

dongjoon-hyun:
Thank you for updating, @difin .

java/core/src/test/org/apache/orc/impl/TestInStream.java
```diff
@@ -957,4 +957,47 @@ public void testMultiRangeCompressHeader() throws IOException {
...
+    final int PREVIOUS_LENGTH = 30;
+    BufferChunkList list = new BufferChunkList();
+    byte[] previous = new byte[PREVIOUS_LENGTH];
+    Arrays.fill(previous, (byte) -1);
+    list.add(new BufferChunk(ByteBuffer.wrap(previous), START - PREVIOUS_LENGTH));
+    list.add(new BufferChunk(ByteBuffer.wrap(uncompressed), START));
+    // Creating a stream of 10 bytes, but with a shorter length of 5
+    InStream inStream = InStream.create("test", list.get(), START, 5, new InStream.StreamOptions());
+    // Resetting the stream without updating its length
+    inStream.reset(list.get());
```
dongjoon-hyun, line 998: So, Hive also has this use case?

dongjoon-hyun (approved):
+1, LGTM. Merged to main/1.8 for Apache ORC 1.8.3.

java/core/src/test/org/apache/orc/impl/TestInStream.java
difin: Yes, and this causes "Can't finish byte read from uncompressed stream DATA position" exception in Hive (HIVE-27128). We plan to fix it in Hive by upgrading its ORC version once a new ORC version is releases with this fix(ORC-1393) and with ORC-1384 which is a blocker for the upgrade. Thanks a lot for merging my PR!

dongjoon-hyun:
Thank you for your contribution. ORC-1393 is resolved and assigned to you.
Welcome to the Apache ORC community.

PR #1449, opened 2023-03-23
ORC-1399 [C++] Return LongVectorBatch when createRowBatch for boolean type with useTightNumericVector option

ffacs:
Return LongVectorBatch when create a RowBatch for BOOLEAN type with useTightNumericVector option, instead of ByteVectorBatch.

If a ByteVectorBatch is returned, BooleanColumnReader/Writer would crash because they only accept LongVectorBatch.

no test

### DIFF
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -171,10 +171,12 @@ namespace orc {
   }
 
   template <typename BatchType>
-  void BooleanColumnReader<BatchType>::next(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) {
+  void BooleanColumnReader<BatchType>::next(ColumnVectorBatch& rowBatch, uint64_t numValues,
+                                            char* notNull) {
     ColumnReader::next(rowBatch, numValues, notNull);
-    // Since the byte rle places the output in a char*
-    // we cheat here and use the other type and then expand it in a second pass.
+    // Since the byte rle places the output in a char* and BatchType here may be
+    // LongVectorBatch with long*. We cheat here in that case and use the long*
+    // and then expand it in a second pass..
     auto* ptr = dynamic_cast<BatchType&>(rowBatch).data.data();
     rle->next(reinterpret_cast<char*>(ptr), numValues,
               rowBatch.hasNulls ? rowBatch.notNull.data() : nullptr);
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -133,6 +133,7 @@ namespace orc {
     }
   }
 
+  template <typename BatchType>
   class BooleanColumnReader : public ColumnReader {
    private:
     std::unique_ptr<orc::ByteRleDecoder> rle;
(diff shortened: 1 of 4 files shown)


c++/src/TypeImpl.cc
```diff
@@ -286,7 +286,9 @@ namespace orc {
                                                               MemoryPool& memoryPool, bool encoded,
                                                               bool useTightNumericVector) const {
     switch (static_cast<int64_t>(kind)) {
-      case BOOLEAN:
+      case BOOLEAN: {
```
dongjoon-hyun, line 289: According to the PR title, do we need to use `useTightNumericVector` variable?

dongjoon-hyun:
cc @wgtmac

c++/src/TypeImpl.cc
wgtmac: I think this PR fixes the bug introduced by `useTightNumericVector` which uses `ByteVectorBatch` for `BOOLEAN` accidentally.

Or should we fix this by supporting `ByteVectorBatch` for `BOOLEAN` type?

ffacs:
I`ve uploaded a new patch which supports reading/writing `ByteVectorBatch` for `BOOLEAN` type, please take a look in your free time. @dongjoon-hyun @wgtmac @coderex2522

c++/src/ColumnWriter.cc
```diff
@@ -667,11 +669,12 @@ namespace orc {
...
 
-  void BooleanColumnWriter::add(ColumnVectorBatch& rowBatch, uint64_t offset, uint64_t numValues,
+  template <typename BatchType>
+  void BooleanColumnWriter<BatchType>::add(ColumnVectorBatch& rowBatch, uint64_t offset, uint64_t numValues,
                                 const char* incomingMask) {
-    LongVectorBatch* byteBatch = dynamic_cast<LongVectorBatch*>(&rowBatch);
+    BatchType* byteBatch = dynamic_cast<BatchType*>(&rowBatch);
     if (byteBatch == nullptr) {
-      throw InvalidArgument("Failed to cast to LongVectorBatch");
+      throw InvalidArgument("Failed to cast to BooleanVectorBatch");
```
wgtmac, line 677: The exception message may be incorrect. You may need to `#include <typeinfo>` to use its name field to build the message.

wgtmac: https://en.cppreference.com/w/cpp/types/type_info

c++/src/ColumnReader.cc
```diff
@@ -148,35 +149,40 @@ namespace orc {
...
 
-  void BooleanColumnReader::next(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) {
+  template <typename BatchType>
+  void BooleanColumnReader<BatchType>::next(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) {
     ColumnReader::next(rowBatch, numValues, notNull);
-    // Since the byte rle places the output in a char* instead of long*,
-    // we cheat here and use the long* and then expand it in a second pass.
-    int64_t* ptr = dynamic_cast<LongVectorBatch&>(rowBatch).data.data();
+    // Since the byte rle places the output in a char*
+    // we cheat here and use the other type and then expand it in a second pass.
```
wgtmac, line 177:
```suggestion
    // Since the byte rle places the output in a char* and BatchType here may be
    // LongVectorBatch with long*. We cheat here in that case and use the long*
    // and then expand it in a second pass.
```

wgtmac (commented):
The overall is looking good. Left some cosmetic comments.

wgtmac (approved):
LGTM. Thanks @ffacs!

dongjoon-hyun:
Thank you, @ffacs and @wgtmac .

PR #1454, opened 2023-03-27
ORC-1385: [C++] Support schema evolution of numeric types

ffacs:
support schema evolution converting between types within {boolean, tinyint, smallint, int, bigint, float, double}

To support shecma evolution in c++

UT passed

### DIFF
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -53,6 +53,10 @@ namespace orc {
     // whether the vector batch is encoded
     bool isEncoded;
 
+    // whether the vector is not a numeric vector
+    // or it's created with the option useTightNumericVector
+    bool isTight = true;
+
     // custom memory pool
     MemoryPool& memoryPool;
 
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -53,10 +53,6 @@ namespace orc {
     // whether the vector batch is encoded
     bool isEncoded;
 
-    // whether the vector is not a numeric vector
-    // or it's created with the option useTightNumericVector
-    bool isTight = true;
-
     // custom memory pool
     MemoryPool& memoryPool;
 
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -1715,16 +1715,10 @@ namespace orc {
                                             bool useTightNumericVector,
                                             bool throwOnSchemaEvolutionOverflow,
                                             bool convertToReadType) {
-    if (convertToReadType && stripe.getSchemaEvolution()) {
-      if (stripe.getSchemaEvolution()->needConvert(type)) {
-        if (!useTightNumericVector) {
-          throw SchemaEvolutionError(
-              "SchemaEvolution only support tight vector, please create ColumnVectorBatch with "
-              "option "
-              "useTightNumericVector");
-        }
-        return buildConvertReader(type, stripe, throwOnSchemaEvolutionOverflow);
-      }
+    if (convertToReadType && stripe.getSchemaEvolution() &&
+        stripe.getSchemaEvolution()->needConvert(type)) {
+      return buildConvertReader(type, stripe, useTightNumericVector,
+                                throwOnSchemaEvolutionOverflow);
     }
 
     switch (static_cast<int64_t>(type.getKind())) {
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -21,7 +21,9 @@
 #include "Adaptor.hh"
 #include "ByteRLE.hh"
 #include "ColumnReader.hh"
+#include "ConvertColumnReader.hh"
 #include "RLE.hh"
+#include "SchemaEvolution.hh"
 #include "orc/Exceptions.hh"
 
 #include <math.h>
@@ -1699,7 +1701,13 @@ namespace orc {
    * Create a reader for the given stripe.
    */
   std::unique_ptr<ColumnReader> buildReader(const Type& type, StripeStreams& stripe,
-                                            bool useTightNumericVector) {
+                                            bool useTightNumericVector, bool convertToReadType) {
+    if (convertToReadType && stripe.getSchemaEvolution()) {
+      if (stripe.getSchemaEvolution()->needConvert(type)) {
+        return buildConvertReader(type, stripe, useTightNumericVector);
+      }
+    }
+
     switch (static_cast<int64_t>(type.getKind())) {
       case SHORT: {
         if (useTightNumericVector) {
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -830,7 +830,8 @@ namespace orc {
     std::vector<std::unique_ptr<ColumnReader>> children;
 
    public:
-    StructColumnReader(const Type& type, StripeStreams& stipe, bool useTightNumericVector = false);
+    StructColumnReader(const Type& type, StripeStreams& stipe, bool useTightNumericVector = false,
+                       bool throwOnSchemaEvolutionOverflow = false);
 
     uint64_t skip(uint64_t numValues) override;
 
@@ -846,7 +847,8 @@ namespace orc {
   };
 
   StructColumnReader::StructColumnReader(const Type& type, StripeStreams& stripe,
-                                         bool useTightNumericVector)
+                                         bool useTightNumericVector,
+                                         bool throwOnSchemaEvolutionOverflow)
       : ColumnReader(type, stripe) {
     // count the number of selected sub-columns
     const std::vector<bool> selectedColumns = stripe.getSelectedColumns();
@@ -855,7 +857,8 @@ namespace orc {
         for (unsigned int i = 0; i < type.getSubtypeCount(); ++i) {
           const Type& child = *type.getSubtype(i);
           if (selectedColumns[static_cast<uint64_t>(child.getColumnId())]) {
-            children.push_back(buildReader(child, stripe, useTightNumericVector));
+            children.push_back(
+                buildReader(child, stripe, useTightNumericVector, throwOnSchemaEvolutionOverflow));
           }
         }
         break;
@@ -915,7 +918,8 @@ namespace orc {
     std::unique_ptr<RleDecoder> rle;
 
    public:
-    ListColumnReader(const Type& type, StripeStreams& stipe, bool useTightNumericVector = false);
+    ListColumnReader(const Type& type, StripeStreams& stipe, bool useTightNumericVector = false,
+                     bool throwOnSchemaEvolutionOverflow = false);
     ~ListColumnReader() override;
 
     uint64_t skip(uint64_t numValues) override;
@@ -932,7 +936,8 @@ namespace orc {
   };
 
   ListColumnReader::ListColumnReader(const Type& type, StripeStreams& stripe,
-                                     bool useTightNumericVector)
+                                     bool useTightNumericVector,
+                                     bool throwOnSchemaEvolutionOverflow)
       : ColumnReader(type, stripe) {
     // count the number of selected sub-columns
     const std::vector<bool> selectedColumns = stripe.getSelectedColumns();
@@ -943,7 +948,7 @@ namespace orc {
     rle = createRleDecoder(std::move(stream), false, vers, memoryPool, metrics);
     const Type& childType = *type.getSubtype(0);
     if (selectedColumns[static_cast<uint64_t>(childType.getColumnId())]) {
-      child = buildReader(childType, stripe, useTightNumericVector);
+      child = buildReader(childType, stripe, useTightNumericVector, throwOnSchemaEvolutionOverflow);
     }
   }
 
@@ -1035,7 +1040,8 @@ namespace orc {
     std::unique_ptr<RleDecoder> rle;
 
    public:
-    MapColumnReader(const Type& type, StripeStreams& stipe, bool useTightNumericVector = false);
+    MapColumnReader(const Type& type, StripeStreams& stipe, bool useTightNumericVector = false,
+                    bool throwOnSchemaEvolutionOverflow = false);
     ~MapColumnReader() override;
 
     uint64_t skip(uint64_t numValues) override;
@@ -1052,7 +1058,7 @@ namespace orc {
   };
 
   MapColumnReader::MapColumnReader(const Type& type, StripeStreams& stripe,
-                                   bool useTightNumericVector)
+                                   bool useTightNumericVector, bool throwOnSchemaEvolutionOverflow)
       : ColumnReader(type, stripe) {
     // Determine if the key and/or value columns are selected
     const std::vector<bool> selectedColumns = stripe.getSelectedColumns();
@@ -1063,11 +1069,13 @@ namespace orc {
     rle = createRleDecoder(std::move(stream), false, vers, memoryPool, metrics);
     const Type& keyType = *type.getSubtype(0);
     if (selectedColumns[static_cast<uint64_t>(keyType.getColumnId())]) {
-      keyReader = buildReader(keyType, stripe, useTightNumericVector);
+      keyReader =
+          buildReader(keyType, stripe, useTightNumericVector, throwOnSchemaEvolutionOverflow);
     }
     const Type& elementType = *type.getSubtype(1);
     if (selectedColumns[static_cast<uint64_t>(elementType.getColumnId())]) {
-      elementReader = buildReader(elementType, stripe, useTightNumericVector);
+      elementReader =
+          buildReader(elementType, stripe, useTightNumericVector, throwOnSchemaEvolutionOverflow);
     }
   }
 
@@ -1177,7 +1185,8 @@ namespace orc {
     uint64_t numChildren;
 
    public:
-    UnionColumnReader(const Type& type, StripeStreams& stipe, bool useTightNumericVector = false);
+    UnionColumnReader(const Type& type, StripeStreams& stipe, bool useTightNumericVector = false,
+                      bool throwOnSchemaEvolutionOverflow = false);
 
     uint64_t skip(uint64_t numValues) override;
 
@@ -1193,7 +1202,8 @@ namespace orc {
   };
 
   UnionColumnReader::UnionColumnReader(const Type& type, StripeStreams& stripe,
-                                       bool useTightNumericVector)
+                                       bool useTightNumericVector,
+                                       bool throwOnSchemaEvolutionOverflow)
       : ColumnReader(type, stripe) {
     numChildren = type.getSubtypeCount();
     childrenReader.resize(numChildren);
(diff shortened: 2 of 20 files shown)


c++/src/ConvertColumnReader.cc
```diff
@@ -0,0 +1,415 @@
...
+  static inline void setNull(ColumnVectorBatch& dstBatch, uint64_t idx) {
+    dstBatch.notNull.data()[idx] = 0;
+    dstBatch.hasNulls = true;
+  }
+
+  // return false if overflow
+  template <typename ReadType>
+  static bool downCastToInteger(ReadType& dstValue, int64_t inputLong) {
+    dstValue = static_cast<ReadType>(inputLong);
+    if (std::is_same<ReadType, int64_t>::value) {
```
wgtmac, line 69:
```suggestion
    if constexpr (std::is_same<ReadType, int64_t>::value) {
```

c++/src/ConvertColumnReader.cc
```diff
@@ -0,0 +1,415 @@
...
+      return true;
+    }
+    if (static_cast<int64_t>(dstValue) != inputLong) {
+      return false;
+    }
+    return true;
+  }
+
+  // set null if overflow
```
wgtmac, line 78: Should we add a mode to support nullify or throw when overflow is detected? Different engines may expect different behaviors.

wgtmac (changes requested):

c++/include/orc/Vector.hh
```diff
@@ -53,6 +53,10 @@ namespace orc {
     // whether the vector batch is encoded
     bool isEncoded;
 
+    // whether the vector is not a numeric vector
+    // or it's created with the option useTightNumericVector
+    bool isTight = true;
```
wgtmac, line 58: Is it possible not to add this? The reader can get this information by the provided `ReaderOption`

c++/src/ColumnReader.cc
```diff
@@ -1699,7 +1712,21 @@ namespace orc {
...
+                                            bool useTightNumericVector,
+                                            bool throwOnSchemaEvolutionOverflow,
+                                            bool convertToReadType) {
+    if (convertToReadType && stripe.getSchemaEvolution()) {
+      if (stripe.getSchemaEvolution()->needConvert(type)) {
+        if (!useTightNumericVector) {
+          throw SchemaEvolutionError(
+              "SchemaEvolution only support tight vector, please create ColumnVectorBatch with "
+              "option "
+              "useTightNumericVector");
```
wgtmac, line 1724:
```suggestion
              "SchemaEvolution only support tight vector, please create ColumnVectorBatch with "
              "option useTightNumericVector");
```

c++/src/ColumnReader.cc
```diff
@@ -1699,7 +1712,21 @@ namespace orc {
    * Create a reader for the given stripe.
    */
   std::unique_ptr<ColumnReader> buildReader(const Type& type, StripeStreams& stripe,
-                                            bool useTightNumericVector) {
+                                            bool useTightNumericVector,
+                                            bool throwOnSchemaEvolutionOverflow,
+                                            bool convertToReadType) {
+    if (convertToReadType && stripe.getSchemaEvolution()) {
```
wgtmac, line 1718: Can we merge these if statements to a single one?

c++/src/ConvertColumnReader.cc
```diff
@@ -0,0 +1,432 @@
...
+
+namespace orc {
+
+  // Assume that we are using tight numeric vector batch
+  using BooleanVectorBatch = ByteVectorBatch;
+
+  ConvertColumnReader::ConvertColumnReader(const Type& _readType, const Type& fileType,
+                                           StripeStreams& stripe, bool _throwOnOverflow)
+      : ColumnReader(_readType, stripe), readType(_readType), throwOnOverflow(_throwOnOverflow) {
+    reader = buildReader(fileType, stripe, /*useTightNumericVector=*/true,
```
wgtmac, line 29: What about add an input parameter `useTightNumericVector` to the function and throw if it is false? In this way, we don't have to check it any more like line 36 does.

c++/src/ConvertColumnReader.cc
```diff
@@ -0,0 +1,432 @@
...
+      throw SchemaEvolutionError(
+          "SchemaEvolution only support tight vector, please create ColumnVectorBatch with option "
+          "useTightNumericVector");
+    }
+    reader->next(*data, numValues, notNull);
+    rowBatch.resize(data->capacity);
+    rowBatch.numElements = data->numElements;
+    rowBatch.hasNulls = data->hasNulls;
+    if (!rowBatch.hasNulls) {
+      memset(rowBatch.notNull.data(), 1, data->notNull.size());
```
wgtmac, line 46: I guess this is not required

c++/src/ConvertColumnReader.cc
```diff
@@ -0,0 +1,432 @@
...
+    constexpr bool isReadTypeFloatingPoint(std::is_floating_point<ReadType>::value);
+    int64_t longValue = static_cast<int64_t>(srcValue);
+    if (isFileTypeFloatingPoint) {
+      if (isReadTypeFloatingPoint) {
+        destValue = static_cast<ReadType>(srcValue);
+      } else {
+        if (!canFitInLong(static_cast<double>(srcValue)) ||
+            !downCastToInteger(destValue, longValue)) {
+          handleOverflow<FileType, ReadType>(destBatch, idx, shouldThrow);
+          return;
```
wgtmac, line 108: This line is redundant.

c++/src/Reader.cc
```diff
@@ -516,6 +518,13 @@ namespace orc {
     return forcedScaleOnHive11Decimal;
   }
 
+  void RowReaderImpl::getReadColumns(const Type* readType, std::set<uint64_t>& readColumns) const {
```
wgtmac, line 521: nit: `getReadColumns` can be a non-member function.

c++/src/Reader.cc
```diff
@@ -1205,8 +1215,32 @@ namespace orc {
   }
 
   std::unique_ptr<ColumnVectorBatch> RowReaderImpl::createRowBatch(uint64_t capacity) const {
-    return getSelectedType().createRowBatch(capacity, *contents->pool, enableEncodedBlock,
-                                            useTightNumericVector);
+    if (schemaEvolution.getReadType() && selectedSchema.get() == nullptr) {
```
wgtmac, line 1218: nit: use `RowReaderImpl::getSelectedType()` to get selectedSchema so we can make sure it always exists.

c++/src/Reader.cc
```diff
@@ -1205,8 +1215,32 @@ namespace orc {
...
+      auto fileSchema = &getSelectedType();
+      auto readType = schemaEvolution.getReadType();
+      std::set<uint64_t> readColumns;
+      getReadColumns(readType, readColumns);
```
wgtmac, line 1222: What about this?
```cpp
getReadColumns(readType, readColumnIds);
getReadColumns(getSelectedType(), selectedColumnIds);
if (readColumnIds != selectedColumnIds) {
  ...
}
```

c++/include/orc/Vector.hh
ffacs: But, what if the ColumnVectorBatch is created by another reader which was createed without the useTightNumericVector option?

wgtmac: Does the current `ColumnReader` have same issue?

ffacs: > Does the current `ColumnReader` have same issue?

Yes, I think there is.

wgtmac: What probably need to fix that as well.

What about adding a new function to `ColumnVectorBatch` like `int32_t fixed_type_size() const` to return the width in bytes for fixed length type? For variable length type or nested type we can return -1.

c++/src/ConvertColumnReader.cc
ffacs: > I guess this is not required

https://github.com/apache/orc/blob/070c1a5dce3fd7f6b6b8938cc3d872a7b67b1b28/c%2B%2B/src/MemoryPool.cc#L120-L127
When there are no nulls and we are going to set null when an overflow in conversion occurs, the `notNull` of other elements should be `1`, but they are `0` after resizing, so the `memset` is needed.

c++/include/orc/Vector.hh
ffacs: > What probably need to fix that as well.
> 
> What about adding a new function to `ColumnVectorBatch` like `int32_t fixed_type_size() const` to return the width in bytes for fixed length type? For variable length type or nested type we can return -1.

I think what we are going to do is handle incorrect usage by users (such as passing a vector that was created by other readers). However, we don’t know what causes the failure of casting. The vector may have been created by another reader with a different read type from the current reader, or it may have been created by another reader with a different `useTightNumericVector` option. `fixed_type_size()` may not help us infer what the user did wrong, but with `isTight`, we could suggest that the user check their options.

wgtmac: OK, that sounds good. 

However, `isTight` here is a little bit confusing since it applies to only numeric types. We need to be careful to add any field to the public api. And it should also be used with its type together. 

In the `ColumnReader`, we always check the result of dynamic_cast of the vector batch. Is it enough if we can use dynamic_cast and check if it is not null in the `ConvertColumnReader`?

ffacs: > OK, that sounds good.
> 
> However, `isTight` here is a little bit confusing since it applies to only numeric types. We need to be careful to add any field to the public api. And it should also be used with its type together.
> 
> In the `ColumnReader`, we always check the result of dynamic_cast of the vector batch. Is it enough if we can use dynamic_cast and check if it is not null in the `ConvertColumnReader`?

I think that`s OK, so we will only verify the `ReaderOption` and check the result of dynamic_cast in `ConvertColumnReader`.

c++/src/ColumnReader.cc
```diff
@@ -828,7 +830,8 @@ namespace orc {
     std::vector<std::unique_ptr<ColumnReader>> children;
 
    public:
-    StructColumnReader(const Type& type, StripeStreams& stipe, bool useTightNumericVector = false);
+    StructColumnReader(const Type& type, StripeStreams& stipe, bool useTightNumericVector = false,
```
wgtmac, line 833: I am thinking if we can use a new ColumnReaderOptions to include `useTightNumericVector` and `throwOnSchemaEvolutionOverflow` in case we will add more options. But that can be a separate patch since this is not public.

c++/src/Reader.cc
```diff
@@ -1204,9 +1207,33 @@ namespace orc {
...
+    }
+  }
+
   std::unique_ptr<ColumnVectorBatch> RowReaderImpl::createRowBatch(uint64_t capacity) const {
-    return getSelectedType().createRowBatch(capacity, *contents->pool, enableEncodedBlock,
-                                            useTightNumericVector);
+    // If the read type is specified, then check that the selected schema matches the read type
+    // on the first call to createRowBatch.
+    if (schemaEvolution.getReadType() && selectedSchema.get() == nullptr) {
+      auto fileSchema = &getSelectedType();
```
wgtmac, line 1221: Move it to above line 1220.

c++/src/Reader.cc
```diff
@@ -1204,9 +1207,33 @@ namespace orc {
...
+      getColumnIds(type->getSubtype(i), columnIds);
+    }
+  }
+
   std::unique_ptr<ColumnVectorBatch> RowReaderImpl::createRowBatch(uint64_t capacity) const {
-    return getSelectedType().createRowBatch(capacity, *contents->pool, enableEncodedBlock,
-                                            useTightNumericVector);
+    // If the read type is specified, then check that the selected schema matches the read type
+    // on the first call to createRowBatch.
+    if (schemaEvolution.getReadType() && selectedSchema.get() == nullptr) {
```
wgtmac, line 1220:
```suggestion
    auto selectedFileType = &getSelectedType();
    if (schemaEvolution.getReadType()) {
```

c++/src/Reader.cc
```diff
@@ -1204,9 +1207,33 @@ namespace orc {
...
+      getColumnIds(fileSchema, fileColumns);
+      if (readColumns != fileColumns) {
+        std::ostringstream ss;
+        ss << "The selected schema " << fileSchema->toString() << " doesn't match read type "
+           << readType->toString();
+        throw SchemaEvolutionError(ss.str());
+      }
+    }
+    const Type& readType =
+        schemaEvolution.getReadType() ? *schemaEvolution.getReadType() : getSelectedType();
```
wgtmac, line 1234:
```suggestion
        schemaEvolution.getReadType() ? *schemaEvolution.getReadType() : *selectedFileType;
```

c++/src/Reader.cc
```diff
@@ -1204,9 +1207,33 @@ namespace orc {
...
-    return getSelectedType().createRowBatch(capacity, *contents->pool, enableEncodedBlock,
-                                            useTightNumericVector);
+    // If the read type is specified, then check that the selected schema matches the read type
+    // on the first call to createRowBatch.
+    if (schemaEvolution.getReadType() && selectedSchema.get() == nullptr) {
+      auto fileSchema = &getSelectedType();
+      auto readType = schemaEvolution.getReadType();
+      std::set<uint64_t> readColumns, fileColumns;
+      getColumnIds(readType, readColumns);
+      getColumnIds(fileSchema, fileColumns);
```
wgtmac, line 1225:
```suggestion
      getColumnIds(selectedFileType, fileColumns);
```

c++/src/ConvertColumnReader.cc
```diff
@@ -0,0 +1,432 @@
...
+  template <typename FileTypeBatch, typename ReadTypeBatch, typename ReadType>
+  class NumericConvertColumnReader : public ConvertColumnReader {
+   public:
+    NumericConvertColumnReader(const Type& _readType, const Type& fileType, StripeStreams& stripe,
+                               bool _throwOnOverflow)
+        : ConvertColumnReader(_readType, fileType, stripe, _throwOnOverflow) {}
+
+    void next(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) override {
+      ConvertColumnReader::next(rowBatch, numValues, notNull);
+      const auto& srcBatch = dynamic_cast<const FileTypeBatch&>(*data);
```
wgtmac, line 130: Can we wrap `dynamic_cast` to throw a clear error message if failed? Something like
```cpp
template <typename TO, typename FROM>
TO* safe_cast(FROM* from) {
  auto to = dynamic_cast<TO*>(from);
  if (to == nullptr) {
     throw ... // maybe a new exception type for cast in Exceptions.hh
  }
  return to;
}
```

wgtmac (commented):
Rest LGTM.

c++/src/Reader.cc
wgtmac: OK, that sounds good. Sorry for not reading carefully.

c++/src/ConvertColumnReader.cc
```diff
@@ -0,0 +1,445 @@
...
```
wgtmac, line 143: It may also throw here, better to apply `SafeCastBatchTo` as well.

c++/src/ConvertColumnReader.cc
```diff
@@ -0,0 +1,445 @@
...
+  class NumericConvertColumnReader<FileTypeBatch, BooleanVectorBatch, bool>
+      : public ConvertColumnReader {
+   public:
+    NumericConvertColumnReader(const Type& _readType, const Type& fileType, StripeStreams& stripe,
+                               bool _throwOnOverflow)
+        : ConvertColumnReader(_readType, fileType, stripe, _throwOnOverflow) {}
+
+    void next(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) override {
+      ConvertColumnReader::next(rowBatch, numValues, notNull);
+      const auto& srcBatch = dynamic_cast<const FileTypeBatch&>(*data);
```
wgtmac, line 172: ditto

c++/src/ConvertColumnReader.cc
```diff
@@ -0,0 +1,445 @@
...
+    if (static_cast<int64_t>(dstValue) != inputLong) {
+      return false;
+    }
+    return true;
+  }
+
+  template <typename DestBatchRefType>
+  static inline DestBatchRefType SafeCastBatchTo(ColumnVectorBatch& batch, const Type& readType) {
+    try {
+      return dynamic_cast<DestBatchRefType>(batch);
```
wgtmac, line 91: Why not casting to pointer type and checks null directly? If would be more efficient to avoid try-catch on every cast.

ffacs: Done

wgtmac (approved):
LGTM. Thanks!

c++/src/Reader.hh
```diff
@@ -237,6 +241,10 @@ namespace orc {
     bool getThrowOnHive11DecimalOverflow() const;
     bool getIsDecimalAsLong() const;
     int32_t getForcedScaleOnHive11Decimal() const;
+
+    const SchemaEvolution& getSchemaEvolution() const {
```
coderex2522, line 245: Is it possible to consider unifying the return type of the getSchemaEvolution() function in class StripeStreams and class RowReaderImpl.

ffacs: Done

c++/src/SchemaEvolution.cc
```diff
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+#include <iostream>
```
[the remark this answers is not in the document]
ffacs: Removed

coderex2522 (approved):
+1 LGTM. Thanks @ffacs

dongjoon-hyun (commented):
Hi, All.

This code seems to cause a build failure on Clang-Ubuntu environment. Not only old Clang 10.0.0 on Ubuntu 20.04 but also Clang 14.0.0 on Ubuntu 22.04 complains about this patch. Could you take a look at this?

dongjoon-hyun:
Could you double check this compilation issue, @ffacs , @wgtmac , @coderex2522 ?

c++/src/ConvertColumnReader.cc
```diff
@@ -0,0 +1,446 @@
...
+      throw SchemaEvolutionError(
+          "SchemaEvolution only support tight vector, please create ColumnVectorBatch with "
+          "option useTightNumericVector");
+    }
+    const auto& _readType = *stripe.getSchemaEvolution()->getReadType(fileType);
+
+    switch (fileType.getKind()) {
+      case BOOLEAN: {
+        switch (_readType.getKind()) {
+          CASE_CREATE_READER(BYTE, BooleanToByte);
```
dongjoon-hyun, line 260: For example, `;` is required in this line?

dongjoon-hyun: This is fixed via https://github.com/apache/orc/pull/1500#discussion_r1238261627

wgtmac: How did you get the failure? Via docker test?

PR #1455, opened 2023-03-29
ORC-1401: [C++] allow writing an intermediate footer

hinxx:
Current C++ code base does not support writing intermediate footers as Java code base does.

It would be beneficial to the C++ based writers to have an ability to write intermediate footer(s) to the opened file while streaming data into it. The benefit is not losing the so far persisted data (after flushing stripes) in case of writer failure in closing the file (application crash).

The writer that wants to support the feature would follow similar approach as Hive does in writing a side file with the latest footer offsets. The reader would be able to determine the location of the last valid footer by using the last footer location from the side file (instead of assuming that the footer is at the end of the file).

The change was tested by creating a small ORC file writer test application that calls newly added `WriterImpl::writeIntermediateFooter()` after adding a batch of rows:

```
        writer->add(*batch);
        writer->writeIntermediateFooter();
```

The resulting ORC file is readable by the provided stock C++ and Java tools. The preliminary C++ test code and results from the tools are attached to this PR here [notes.txt](https://github.com/apache/orc/files/11105096/notes.txt).

Not all aspects of the possible use were tested yet namely unclean writer app shutdown, writing of the side file, loading of the side file, reading of the last valid footer when file was not properly closed.

This is a draft PR. No proper tests are added at the moment because of incomplete PR.

I'm looking for opinions from the ORC experts in order to finish up the PR.

### DIFF
--- a/c++/src/Writer.cc
+++ b/c++/src/Writer.cc
@@ -307,7 +307,6 @@ namespace orc {
     static const char* magicId;
     static const WriterId writerId;
     bool useTightNumericVector;
-    long numStripes;
     long stripesAtLastFlush;
     long lastFlushOffset;
 
@@ -345,7 +344,7 @@ namespace orc {
     columnWriter = buildWriter(type, *streamsFactory, options);
     stripeRows = totalRows = indexRows = 0;
     currentOffset = 0;
-    numStripes = stripesAtLastFlush = lastFlushOffset = 0;
+    stripesAtLastFlush = lastFlushOffset = 0;
 
     useTightNumericVector = opts.getUseTightNumericVector();
 
@@ -410,10 +409,10 @@ namespace orc {
     if (stripeRows > 0) {
       writeStripe();
     }
-    if (stripesAtLastFlush != numStripes) {
+    if (stripesAtLastFlush != fileFooter.stripes_size()) {
       writeMetadata();
       writeFileFooter();
-      stripesAtLastFlush = numStripes;
+      stripesAtLastFlush = fileFooter.stripes_size();
       writePostscript();
       outStream->flush();
       lastFlushOffset = outStream->getLength();
@@ -548,8 +547,6 @@ namespace orc {
     totalRows += stripeRows;
 
     columnWriter->reset();
-
-    numStripes += 1;
 
     initStripe();
   }
--- a/c++/src/Writer.cc
+++ b/c++/src/Writer.cc
@@ -307,8 +307,8 @@ namespace orc {
     static const char* magicId;
     static const WriterId writerId;
     bool useTightNumericVector;
-    long stripesAtLastFlush;
-    long lastFlushOffset;
+    int32_t stripesAtLastFlush;
+    int64_t lastFlushOffset;
 
    public:
     WriterImpl(const Type& type, OutputStream* stream, const WriterOptions& options);
@@ -321,7 +321,7 @@ namespace orc {
 
     void addUserMetadata(const std::string& name, const std::string& value) override;
 
-    long writeIntermediateFooter() override;
+    int64_t writeIntermediateFooter() override;
 
    private:
     void init();
@@ -344,7 +344,8 @@ namespace orc {
     columnWriter = buildWriter(type, *streamsFactory, options);
     stripeRows = totalRows = indexRows = 0;
     currentOffset = 0;
-    stripesAtLastFlush = lastFlushOffset = 0;
+    stripesAtLastFlush = 0;
+    lastFlushOffset = 0;
 
     useTightNumericVector = opts.getUseTightNumericVector();
 
--- a/c++/src/Writer.cc
+++ b/c++/src/Writer.cc
@@ -421,7 +421,7 @@ namespace orc {
       // init stripe now that we adjusted the currentOffset
       initStripe();
     }
-    return stripesAtLastFlush;
+    return lastFlushOffset;
   }
 
   void WriterImpl::addUserMetadata(const std::string& name, const std::string& value) {
--- a/c++/src/Writer.cc
+++ b/c++/src/Writer.cc
@@ -307,6 +307,9 @@ namespace orc {
     static const char* magicId;
     static const WriterId writerId;
     bool useTightNumericVector;
+    long numStripes;
+    long stripesAtLastFlush;
+    long lastFlushOffset;
 
    public:
     WriterImpl(const Type& type, OutputStream* stream, const WriterOptions& options);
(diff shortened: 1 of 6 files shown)


c++/src/Writer.cc
```diff
@@ -307,6 +307,9 @@ namespace orc {
     static const char* magicId;
     static const WriterId writerId;
     bool useTightNumericVector;
+    long numStripes;
```
hinxx, line 310: I guess numStripes could be replaced by the `fileFooter.stripes_size()`

c++/src/Writer.cc
```diff
@@ -542,6 +568,7 @@ namespace orc {
     // update file statistics
     std::vector<proto::ColumnStatistics> colStats;
     columnWriter->getFileStatistics(colStats);
+    fileFooter.clear_statistics();
```
hinxx, line 571: I'm not sure if this is the right place to do this. It seems to give desired result when doing the intermediate footer writes that cause stripe flush/write, but I'm thinking that it might not be OK when stripes are written as part of normal flushing due to their size.

dongjoon-hyun (commented):
Thank you for filing a JIRA and making a PR, @hinxx . 
I clicked the CI approval button.

dongjoon-hyun:
cc @omalley and @wgtmac

wgtmac (commented):
Thanks for the change! I haven't gone through all the change. Will take a deep look over the weekend.

c++/src/Writer.cc
wgtmac: This should be OK, as we have anyway to overwrite file-level stats.

c++/test/TestWriter.cc
```diff
@@ -1994,6 +2002,109 @@ namespace orc {
...
+    for (uint64_t j = 0; j < 10; ++j) {
+      for (uint64_t i = 0; i < 65535; ++i) {
+        longBatch->data[i] = static_cast<int64_t>(i);
+      }
+      structBatch->numElements = 65535;
+      longBatch->numElements = 65535;
+
+      writer->add(*batch);
+      // force writing a stripe
+      writer->writeIntermediateFooter();
```
wgtmac, line 2029: It would be better to save all 10 locations to intermediate footers and open them respectively to verify only partial data can be read correctly.

wgtmac (changes requested):
Nice addition!

cc @stiga-huang @coderex2522

wgtmac (commented):
Could you please make the CI happy?

c++/src/Writer.cc
```diff
@@ -307,6 +307,8 @@ namespace orc {
     static const char* magicId;
     static const WriterId writerId;
     bool useTightNumericVector;
+    long stripesAtLastFlush;
```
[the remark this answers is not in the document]
hinxx: If I use `size_t` my compiler complains:
```
error: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Werror=sign-compare]
  412 |     if (stripesAtLastFlush != fileFooter.stripes_size()) {
      |         ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

`Footer::stripes_size()` returns `int`, instead of casting to `size_t` maybe we could just do: `int stripesAtLastFlush;`?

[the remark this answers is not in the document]
wgtmac: We can apply static_cast here. Something like `if (stripesAtLastFlush != static_cast<size_t>(fileFooter.stripes_size())) `. Or we can make it int32_t here because `fileFooter.stripes_size()` is of type `int`.

wgtmac:
Some CI checks fail due to following reason, please use `static_cast` to fix those errors.

```
/Users/runner/work/orc/orc/c++/src/Writer.cc:324:10: error: virtual function 'writeIntermediateFooter' has a different return type ('long') than the function it overrides (which has return type 'int64_t' (aka 'long long'))
    long writeIntermediateFooter() override;
    ~~~~ ^
/Users/runner/work/orc/orc/c++/include/orc/Writer.hh:312:21: note: overridden virtual function is here
    virtual int64_t writeIntermediateFooter() = 0;
            ~~~~~~~ ^
/Users/runner/work/orc/orc/c++/src/Writer.cc:418:36: error: implicit conversion changes signedness: 'uint64_t' (aka 'unsigned long long') to 'long' [-Werror,-Wsign-conversion]
      lastFlushOffset = outStream->getLength();
                      ~ ~~~~~~~~~~~^~~~~~~~~~~
/Users/runner/work/orc/orc/c++/src/Writer.cc:419:23: error: implicit conversion changes signedness: 'long' to 'uint64_t' (aka 'unsigned long long') [-Werror,-Wsign-conversion]
      currentOffset = lastFlushOffset;
                    ~ ^~~~~~~~~~~~~~~
3 errors generated.
```

In addition, the C++ format check failures can be fixed by applying `clang-format`. @hinxx

hinxx:
I checked the CI logs and I think I solved all the issue regarding the data types and formatting.

wgtmac:
> I checked the CI logs and I think I solved all the issue regarding the data types and formatting.

Seems the build on macos has a different opinion.

```
[ 58%] Building CXX object c++/src/CMakeFiles/orc.dir/Writer.cc.o
/Users/runner/work/orc/orc/c++/src/Writer.cc:419:36: error: implicit conversion changes signedness: 'uint64_t' (aka 'unsigned long long') to 'int64_t' (aka 'long long') [-Werror,-Wsign-conversion]
      lastFlushOffset = outStream->getLength();
                      ~ ~~~~~~~~~~~^~~~~~~~~~~
/Users/runner/work/orc/orc/c++/src/Writer.cc:420:23: error: implicit conversion changes signedness: 'int64_t' (aka 'long long') to 'uint64_t' (aka 'unsigned long long') [-Werror,-Wsign-conversion]
      currentOffset = lastFlushOffset;
                    ~ ^~~~~~~~~~~~~~~
/Users/runner/work/orc/orc/c++/src/Writer.cc:424:12: error: implicit conversion changes signedness: 'int64_t' (aka 'long long') to 'uint64_t' (aka 'unsigned long long') [-Werror,-Wsign-conversion]
    return lastFlushOffset;
    ~~~~~~ ^~~~~~~~~~~~~~~
3 errors generated.
make[2]: *** [c++/src/CMakeFiles/orc.dir/Writer.cc.o] Error 1
make: *** [all] Error 2
Error: Process completed with exit code 2.
```

hinxx:
> Seems the build on macos has a different opinion.

This is becoming embarrassing. I would expect to see these sign related errors in my Linux g++ build, but for some reason I don't.
Changed `lastFlushOffset` data type to `uint64_t`.

wgtmac (approved):
LGTM. Thanks! @hinxx

dongjoon-hyun:
Thank you, @hinxx and @wgtmac !

Issue #1450, opened 2023-03-23
[Docs] Improve the description of stream order in ORC Specification


deshanxiao:
The order in the orc stream is not guaranteed.

PR #1465, opened 2023-04-11
ORC-1409: [Docs] Add stream order description in ORC spec.

deshanxiao:
This PR is aimed to add more description about stream order in ORC spec.

There are many users who are misled by the order of the document table, in fact the stream has no fixed order.

#1450


dongjoon-hyun (commented):
Please file a JIRA, @deshanxiao in Apache ORC community.
Every ASF project has its own policy and this is the way in the Apache ORC community.

site/specification/ORCv1.md
```diff
@@ -886,6 +886,12 @@ uses three streams PRESENT, DATA, and LENGTH, which stores the length
 of each value. The details of each type will be presented in the
 following subsections.
 
+Note that the order of these streams is **not fixed**. In the example
+of the integer column mentioned above, the order of the PRESENT stream
+and the DATA stream cannot be determined in advance. Instead, we need
+to determine the type of stream based on the Stream Kind, rather than
```
wgtmac, line 892: Should we mention the following:
- index and data streams cannot be interleaved.
- the order of streams from different columns is not fixed as well.

deshanxiao:
Mark the PR as draft because I may add more details about the stream order.

deshanxiao:
Hi @dongjoon-hyun @wgtmac 

The PR is ready for review. Please take a look at it. Thank you~

site/specification/ORCv1.md
```diff
@@ -886,6 +886,17 @@ uses three streams PRESENT, DATA, and LENGTH, which stores the length
 of each value. The details of each type will be presented in the
 following subsections.
 
+There are a few points to note about the order of streams:
+
+* For a specific column type, the order of streams is **not fixed**.
+* Index and data streams cannot be **interleaved**.
+* The order of streams from different columns is **not fixed** as well.
```
wgtmac, line 893:
```suggestion
There is a general order for index and data streams:
* Index streams are always placed together in the beginning of the stripe.
* Data streams are placed together after index streams (if any).
* Inside index streams or data streams, the unencrypted streams should be placed first and then followed by streams grouped by each encryption variant.

There is no fixed order within each unencrypted or encryption variant in the index and data streams:
* Different stream kinds of the same column can be placed in any order.
* Streams from different columns can even be placed in any order.

To get the precise information (a.k.a stream kind, column id and location) of a stream within a stripe, the streams field in the StripeFooter described below is the single source of truth.
```

wgtmac: These statements look a little bit vague to me, I have changed them to provide fixed order and flexible order. WDYT?

wgtmac: Also I would prefer to move them to below line 907.

deshanxiao: Thank you @wgtmac . It has been updated.

site/specification/ORCv2.md
```diff
@@ -1257,6 +1277,19 @@ Because dictionaries are accessed randomly, there is not a position to
 record for the dictionary and the entire dictionary must be read even
 if only part of a stripe is being read.
 
+Note that for columns with multiple streams, the order of these
+streams is **fixed**, which is different from data stream we mentioned above.
```
wgtmac, line 1281: Seems the doc has a line limit that we need to follow.

wgtmac, line 1281:
```suggestion
Note that for columns with multiple streams, the order of stream
positions in the RowIndex is **fixed**, which may be different to
the actual data stream placement.
```

wgtmac: We should also mention that the order is the same as what `Column Encodings` section has described for each column type.

site/specification/ORCv2.md
```diff
@@ -1257,6 +1277,19 @@ Because dictionaries are accessed randomly, there is not a position to
...
+
+The following table shows the order of types that contain multiple streams:
+
+ Type      | The First Postion Stream Type   | The Second Postion Stream Type | Contents        |
+:----------|:--------------------------------|:-------------------------------|:----------------|
+| Binary    | DATA                            | LENGTH                         |                 |
+| String    | DATA                            | LENGTH                         | Direct encoding |
+| Decimal   | DATA                            | SECONDARY                      |                 |
+| Timestamp | DATA                            | SECONDARY                      |                 |
```
wgtmac, line 1290: This table is duplicated to what `Column Encodings` has described. And here it doesn't include PRESENT stream. I'd prefer to remove this.

site/specification/ORCv2.md
deshanxiao: It looks resonable and will add more description/remove the chart.

site/specification/ORCv2.md
deshanxiao: Done

dongjoon-hyun:
I also published to Apache ORC website.
- https://orc.apache.org/specification/ORCv0/#column-encoding-section

wgtmac:
Thank you @dongjoon-hyun !

PR #1469, opened 2023-04-13
ORC-1408: [C++] Add `testVectorBatchHasNull` test case and comment

kaka11chen:

Add a new test case `testVectorBatchHasNull` and more comments.

To improve a test coverage.

Pass the CIs.

### DIFF
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -37,6 +37,8 @@ namespace orc {
    * The base class for each of the column vectors. This class handles
    * the generic attributes such as number of elements, capacity, and
    * notNull vector.
+   * Note: Because we don't reset notNull buffer if hasNulls == false for better performance,
+   * it need to check hasNull firstly, and then check notNull buffer.
    */
   struct ColumnVectorBatch {
     ColumnVectorBatch(uint64_t capacity, MemoryPool& pool);
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -37,8 +37,11 @@ namespace orc {
    * The base class for each of the column vectors. This class handles
    * the generic attributes such as number of elements, capacity, and
    * notNull vector.
-   * Note: Because we don't reset notNull buffer if hasNulls == false for better performance,
-   * it need to check hasNull firstly, and then check notNull buffer.
+   * Note: If hasNull is true, the values in the notNull buffer are not required.
+   * On the writer side, it does not read values from notNull buffer so users are
+   * not expected to write notNull buffer if hasNull is true. On the reader side,
+   * it does not set notNull buffer if hasNull is true, meaning that it is undefined
+   * behavior to consume values from notNull buffer in this case by downstream users.
    */
   struct ColumnVectorBatch {
     ColumnVectorBatch(uint64_t capacity, MemoryPool& pool);
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -100,13 +100,16 @@ namespace orc {
           return;
         }
       }
+      rowBatch.hasNulls = false;
     } else if (incomingMask) {
       // If we don't have a notNull stream, copy the incomingMask
       rowBatch.hasNulls = true;
       memcpy(rowBatch.notNull.data(), incomingMask, numValues);
       return;
+    } else {
+      memset(rowBatch.notNull.data(), 1, numValues);
+      rowBatch.hasNulls = false;
     }
-    rowBatch.hasNulls = false;
   }
 
   void ColumnReader::seekToRowGroup(std::unordered_map<uint64_t, PositionProvider>& positions) {
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -100,16 +100,13 @@ namespace orc {
           return;
         }
       }
-      rowBatch.hasNulls = false;
     } else if (incomingMask) {
       // If we don't have a notNull stream, copy the incomingMask
       rowBatch.hasNulls = true;
       memcpy(rowBatch.notNull.data(), incomingMask, numValues);
       return;
-    } else {
-      memset(rowBatch.notNull.data(), 1, numValues);
-      rowBatch.hasNulls = false;
     }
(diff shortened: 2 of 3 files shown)


dongjoon-hyun:
Thank you for adding a test case, @kaka11chen .

c++/src/ColumnReader.cc
```diff
@@ -100,13 +100,16 @@ namespace orc {
...
         }
       }
+      rowBatch.hasNulls = false;
     } else if (incomingMask) {
       // If we don't have a notNull stream, copy the incomingMask
       rowBatch.hasNulls = true;
       memcpy(rowBatch.notNull.data(), incomingMask, numValues);
       return;
+    } else {
+      memset(rowBatch.notNull.data(), 1, numValues);
```
wgtmac, line 110: The contract of `ColumnVectorBatch` is that we should not depend on values from `ColumnVectorBatch.notNull` buffer if `hasNulls == false`. Probably we should update the site to explicitly state this: https://orc.apache.org/docs/core-cpp.html

wgtmac: Adding an extra `memset` here does not break the contract but introduces a little bit performance penalty for non-null values.

c++/test/TestColumnReader.cc
```diff
@@ -4207,6 +4207,101 @@ namespace orc {
...
+      std::unique_ptr<ColumnReader> reader = buildReader(*rowType, streams);
+
+      reader->next(batch, 100, 0);
+      ASSERT_EQ(100, batch.numElements);
+      ASSERT_EQ(true, !batch.hasNulls);
+      ASSERT_EQ(100, longBatch->numElements);
+      ASSERT_EQ(true, !longBatch->hasNulls);
+      int next = 0;
+      for (size_t i = 0; i < batch.numElements; ++i) {
+        EXPECT_EQ(1, longBatch->notNull[i]);
```
wgtmac, line 4299: Here is related to what I mentioned above. We should not read `longBatch->notNull` if `longBatch->hasNulls` is false.

Not sure if the java side follows the same contract. @dongjoon-hyun @guiyanakuang

c++/src/ColumnReader.cc
coderex2522: The problem is caused only when user misuses ColumnVectorBatch.notNull array without consuming hasNulls flag. Need to check if the java side follows the same contract.

c++/src/ColumnReader.cc
```diff
@@ -100,13 +100,16 @@ namespace orc {
...
+      rowBatch.hasNulls = false;
     }
-    rowBatch.hasNulls = false;
```
coderex2522, line 109: In addition, it is no necessary to repeat this line of code twice.

kaka11chen:
@dongjoon-hyun @wgtmac @coderex2522 I
If the contract of ColumnVectorBatch is that we should not depend on values from `ColumnVectorBatch.notNull` buffer if `hasNulls == false`, it is invalid bug. But this two mismatch maybe introduce misunderstanding for client usage.

c++/test/TestColumnReader.cc
guiyanakuang: I searched for cases in the Java ORC project where `ColumnVector.noNulls` and `ColumnVector.isNull` are used. Most of the cases follow this contract, but unfortunately, there are still some counterexamples.

wgtmac:
I'd suggest to explicitly add a comment to ColumnVectorBatch in the Vector.hh for the contract. Otherwise downstream users may be confused in the future. AFAIK, Apache Impala and Apache Arrow respect this contract.

kaka11chen:
> I'd suggest to explicitly add a comment to ColumnVectorBatch in the Vector.hh for the contract. Otherwise downstream users may be confused in the future. AFAIK, Apache Impala and Apache Arrow respect this contract.

@wgtmac I have added a comment to ColumnVectorBatch, please review it, thx.

c++/include/orc/Vector.hh
```diff
@@ -37,6 +37,8 @@ namespace orc {
    * The base class for each of the column vectors. This class handles
    * the generic attributes such as number of elements, capacity, and
    * notNull vector.
+   * Note: Because we don't reset notNull buffer if hasNulls == false for better performance,
+   * it need to check hasNull firstly, and then check notNull buffer.
```
wgtmac, line 41:
```suggestion
   * it need to check hasNull first, and then check notNull buffer only if hasNull is true.
```

wgtmac (commented):
Thanks, I have left a cosmetic comment.

c++/include/orc/Vector.hh
wgtmac: The new comment seems to be reader-oriented. We'd better rephrase it to describe the contract, something like:

```
Note: If hasNull is true, the values in the notNull buffer are not required.
On the writer side, it does not read values from notNull buffer so users are
not expected to write notNull buffer if hasNull is true. On the reader side,
it does not set notNull buffer if hasNull is true, meaning that it is undefined
behavior to consume values from notNull buffer in this case by downstream users.
```

kaka11chen: > The new comment seems to be reader-oriented. We'd better rephrase it to describe the contract, something like:
> 
> ```
> Note: If hasNull is true, the values in the notNull buffer are not required.
> On the writer side, it does not read values from notNull buffer so users are
> not expected to write notNull buffer if hasNull is true. On the reader side,
> it does not set notNull buffer if hasNull is true, meaning that it is undefined
> behavior to consume values from notNull buffer in this case by downstream users.
> ```

Thanks, it is better, I have updated.

kaka11chen:
> LGTM. Thanks @kaka11chen!

Welcome.

dongjoon-hyun (approved):
Thank you, @kaka11chen , @wgtmac , @guiyanakuang , @coderex2522 .

I switched the JIRA issue type from `Bug` to `Test` and revised the PR title and description.

dongjoon-hyun:

BTW, Welcome to the Apache ORC community, @kaka11chen .
I added you to the Apache ORC contributor group and assigned ORC-1408 to you.

PR #1482, opened 2023-04-28
ORC-1413 fix for ORC row level filter issue with ACID table

zratkai:
This PR fixes ORC row level filter with ACID table issue.

Without this Hive can not work with ORC 1.8.3. and ACID table and row level filter enabled.

Unit test added.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ParserUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/ParserUtils.java
@@ -411,16 +411,20 @@ public class ParserUtils {
     private final ColumnVector[] result;
     private int resultIdx = 0;
 
-    ColumnFinder(TypeDescription schema, VectorizedRowBatch batch, int levels) {
+    ColumnFinder(TypeDescription schema, ColumnVector[] columnVectors, int levels) {
       if (schema.getCategory() == TypeDescription.Category.STRUCT) {
-        top = batch.cols;
+        top = columnVectors;
         result = new ColumnVector[levels];
       } else {
         result = new ColumnVector[levels + 1];
-        current = batch.cols[0];
+        current = columnVectors[0];
         top = null;
         addResult(current);
       }
+    }
+
+    ColumnFinder(TypeDescription schema, VectorizedRowBatch vectorizedRowBatch, int levels) {
+      this(schema, vectorizedRowBatch.cols, levels);
     }
 
     private void addResult(ColumnVector vector) {
@@ -459,8 +463,11 @@ public class ParserUtils {
                                                  boolean isCaseSensitive,
                                                  VectorizedRowBatch batch) {
     List<String> names = ParserUtils.splitName(source);
-    ColumnFinder result = new ColumnFinder(schema, batch, names.size());
-    findColumn(removeAcid(schema), names, isCaseSensitive, result);
+    TypeDescription schemaToUse = removeAcid(schema);
+    ColumnVector[] columnVectors = SchemaEvolution.checkAcidSchema(schema)
+                  ? ((StructColumnVector) batch.cols[batch.cols.length - 1]).fields : batch.cols;
+    ColumnFinder result = new ColumnFinder(schemaToUse, columnVectors, names.size());
+    findColumn(schemaToUse, names, isCaseSensitive, result);
     return result.result;
   }
 
--- a/java/core/src/test/org/apache/orc/TestOrcFilterContext.java
+++ b/java/core/src/test/org/apache/orc/TestOrcFilterContext.java
@@ -25,9 +25,23 @@ import org.apache.hadoop.hive.ql.exec.vector.LongColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.MapColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.StructColumnVector;
 import org.apache.hadoop.hive.ql.exec.vector.UnionColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.VectorizedRowBatch;
+import org.apache.hadoop.hive.ql.io.sarg.PredicateLeaf;
+import org.apache.hadoop.hive.ql.io.sarg.SearchArgument;
+import org.apache.hadoop.hive.ql.io.sarg.SearchArgumentFactory;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 import org.apache.orc.impl.OrcFilterContextImpl;
+import org.apache.orc.impl.SchemaEvolution;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Random;
+import java.util.Arrays;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
@@ -61,9 +75,22 @@ public class TestOrcFilterContext {
                                            TypeDescription.createList(TypeDescription.createChar()))
                 )
     );
+  private static Configuration configuration;
+  private static FileSystem fileSystem;
+  private static final Path workDir = new Path(System.getProperty("test.tmp.dir",
+          "target" + File.separator + "test"
+                  + File.separator + "tmp"));
+  private static final Path filePath = new Path(workDir, "orc_filter_file.orc");
+
+  private static final int RowCount = 400;
+
   private final OrcFilterContext filterContext = new OrcFilterContextImpl(schema, false)
     .setBatch(schema.createRowBatch());
-
+  TypeDescription typeDescriptionACID =
+          TypeDescription.fromString("struct<int1:int,string1:string>");
+  TypeDescription acidSchema = SchemaEvolution.createEventSchema(typeDescriptionACID);
+  private final OrcFilterContext filterContextACID = new OrcFilterContextImpl(acidSchema, true)
+          .setBatch(acidSchema.createRowBatch());
   @BeforeEach
   public void setup() {
     filterContext.reset();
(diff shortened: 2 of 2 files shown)


zratkai:
@pavibhai could you please review?

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -459,10 +459,31 @@ public static ColumnVector[] findColumnVectors(TypeDescription schema,
                                                  boolean isCaseSensitive,
                                                  VectorizedRowBatch batch) {
     List<String> names = ParserUtils.splitName(source);
-    ColumnFinder result = new ColumnFinder(schema, batch, names.size());
+    ColumnFinder result = new ColumnFinder(schema, removeAcidColumnsFromVectorizedRowBatch(schema, batch), names.size());
     findColumn(removeAcid(schema), names, isCaseSensitive, result);
     return result.result;
   }
+
```
pavibhai, line 466: I would think we should go a different route. My proposal would be something like.

If the schema is acid schema then when you are looking up column names prefix them with `row.` and just use the full schema. This will ensure that we respect the nesting of the vectors under the struct column vector related to `row`

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -459,10 +459,31 @@ public static ColumnVector[] findColumnVectors(TypeDescription schema,
...
+      StructColumnVector structColumnVector = (StructColumnVector)(batch.cols[acidColumnOffset]);
+      ColumnVector[] columnVectors = new ColumnVector[structColumnVector.fields.length];
+      for (int i = 0; i < columnVectors.length; i++) {
+        columnVectors[i] = structColumnVector.fields[i];
+      }
+      vectorizedRowBatch.cols = columnVectors;
+      return vectorizedRowBatch;
+    }
+    return batch;
```
pavibhai, line 485: I might be missing the logic of where we are updating back the changes that are not referenced example batch.size back to the batch after the filter processing is complete.

I go back to the earlier comment that we should do this differently.

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
```diff
@@ -225,4 +230,12 @@ public void testRepeatingVector() {
     assertTrue(OrcFilterContext.isNull(vectorBranch, 1));
     assertTrue(OrcFilterContext.isNull(vectorBranch, 2));
   }
+  
+  @Test
+  public void testACIDTable(){
```
pavibhai, line 235: I would also add an actual filter test with data to ensure that the behavior is as expected.

pavibhai (changes requested):
Thanks for making the PR.

* Need to made CI happy
* Proposed a different approach on fixing this

java/core/src/java/org/apache/orc/impl/ParserUtils.java
zratkai: Changed according to this comment.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
zratkai: Changed the logic. Please check the PR.

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
zratkai: Could you please give an existing test as an example?

pavibhai: You can take a look at [TestRowFilteringIOSkip](https://github.com/apache/orc/blob/main/java/core/src/test/org/apache/orc/TestRowFilteringIOSkip.java).

I am not sure what the contract wit hive is in this case, but will be good to have the following scenarios:
* File written with ACID schema and read with ACID schema with filter
* File written with ACID schema and read without ACID schema with filter
* File written with no ACID schema and read with ACID schema with filter

pavibhai: @zratkai I can help with the this test too if it is taking a lot of your time. Please do let me know.

zratkai: @pavibhai sorry for the delay, I was on a holiday for some days. I created a filter test for ACID table. I am not sure how the last two case should work:

- File written with ACID schema and read without ACID schema with filter
- File written with no ACID schema and read with ACID schema with filter

 I think the schemas should match or not?

pavibhai: > I created a filter test for ACID table.
I will take a look at the test today and drop an update.

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
```diff
@@ -225,4 +247,101 @@ public void testRepeatingVector() {
...
+            .build();
+    Reader.Options readerOptions = reader.options()
+            .searchArgument(searchArgument, new String[] {"int1"})
+            .useSelected(true)
+            .allowSARGToFilter(true);
+    VectorizedRowBatch vectorizedRowBatch = acidSchema.createRowBatch();
+    long rowCount = 0;
+    try (RecordReader recordReader = reader.rows(readerOptions)) {
+      assertTrue(recordReader.nextBatch(vectorizedRowBatch));
+      rowCount += vectorizedRowBatch.size;
```
pavibhai, line 342: Can we please validate that the entry is matching? i.e. the value of the LongColumnVector value is the `id` that was passed in.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -281,7 +282,7 @@ public void visit(TypeDescription type, int posn) {
   public static TypeDescription findSubtype(TypeDescription schema,
                                             ParserUtils.StringPosition source,
                                             boolean isSchemaEvolutionCaseAware) {
-    TypeFinder result = new TypeFinder(removeAcid(schema));
+    TypeFinder result = new TypeFinder(schema);
```
pavibhai, line 285: I am uncomfortable changing this here. This method is being used in multiple places and we need to ensure in all the places the source includes the prefix of `row`.

These are the following two broad approaches I see here.

1. Ignore the`row` `StructColumnVector` in all processing, which seems to be case with the current code when removing acid schema.
2. Consider the `row` column vector in all of the processing (partially implemented)

I am leaning towards the first option so as not to change the behavior of the APIs. e.g. if previously one was calling with col1, now they are expected to call as row.col1 which is a change in behavior.

This is also feels like it will minimize the amount of code that needs to be changed.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -459,8 +466,8 @@ public static ColumnVector[] findColumnVectors(TypeDescription schema,
                                                  boolean isCaseSensitive,
                                                  VectorizedRowBatch batch) {
     List<String> names = ParserUtils.splitName(source);
-    ColumnFinder result = new ColumnFinder(schema, batch, names.size());
```
pavibhai, line 462: It feels like it would be sufficient to add a new constructor to ColumnFinder that can also take an array of column vectors that in case the case of acid schema is the column vectors from the `row` column vector.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
pavibhai: Let me know what you think.

pavibhai: I tried with the following to get the test to pass.

```diff
diff --git a/java/core/src/java/org/apache/orc/impl/ParserUtils.java b/java/core/src/java/org/apache/orc/impl/ParserUtils.java
index 6491fb8ba..b56fa35b3 100644
--- a/java/core/src/java/org/apache/orc/impl/ParserUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/ParserUtils.java
@@ -411,18 +411,22 @@ public class ParserUtils {
     private final ColumnVector[] result;
     private int resultIdx = 0;

-    ColumnFinder(TypeDescription schema, VectorizedRowBatch batch, int levels) {
+    ColumnFinder(TypeDescription schema, ColumnVector[] cols, int levels) {
       if (schema.getCategory() == TypeDescription.Category.STRUCT) {
-        top = batch.cols;
+        top = cols;
         result = new ColumnVector[levels];
       } else {
         result = new ColumnVector[levels + 1];
-        current = batch.cols[0];
+        current = cols[0];
         top = null;
         addResult(current);
       }
     }

+    ColumnFinder(TypeDescription schema, VectorizedRowBatch batch, int levels) {
+      this(schema, batch.cols, levels);
+    }
+
     private void addResult(ColumnVector vector) {
       result[resultIdx] = vector;
       resultIdx += 1;
@@ -459,7 +463,10 @@ public class ParserUtils {
                                                  boolean isCaseSensitive,
                                                  VectorizedRowBatch batch) {
     List<String> names = ParserUtils.splitName(source);
-    ColumnFinder result = new ColumnFinder(schema, batch, names.size());
+    TypeDescription schemaToUse = removeAcid(schema);
+    ColumnVector[] cols = SchemaEvolution.checkAcidSchema(schema)
+      ? ((StructColumnVector) batch.cols[batch.cols.length - 1]).fields : batch.cols;
+    ColumnFinder result = new ColumnFinder(schemaToUse, cols, names.size());
     findColumn(removeAcid(schema), names, isCaseSensitive, result);
     return result.result;
   }
```

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
zratkai: Added more validation.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
zratkai: @pavibhai thanks for the idea! I agree with you, changed accordingly.

java/core/src/java/org/apache/orc/impl/ParserUtils.java
```diff
@@ -27,6 +27,7 @@
 import org.apache.orc.TypeDescription;
 
 import java.util.ArrayList;
+import java.util.LinkedList;
```
pavibhai, line 30: .nit this should no longer be needed

zratkai: Removed.

pavibhai:
@zratkai Can you please take care of the checks?

zratkai:
@pavibhai fixed the checkstyle issue.

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
```diff
@@ -18,17 +18,25 @@
...
-import org.apache.hadoop.hive.ql.exec.vector.ColumnVector;
-import org.apache.hadoop.hive.ql.exec.vector.ListColumnVector;
-import org.apache.hadoop.hive.ql.exec.vector.LongColumnVector;
-import org.apache.hadoop.hive.ql.exec.vector.MapColumnVector;
-import org.apache.hadoop.hive.ql.exec.vector.StructColumnVector;
-import org.apache.hadoop.hive.ql.exec.vector.UnionColumnVector;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hive.ql.exec.vector.*;
```
pavibhai, line 24: Can we replace this back to individual imports instead of `*`

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
```diff
@@ -61,9 +69,24 @@ public class TestOrcFilterContext {
...
+  private static Configuration configuration;
+  private static FileSystem fileSystem;
+  private static final Path workDir = new Path(System.getProperty("test.tmp.dir",
+          "target" + File.separator + "test"
+                  + File.separator + "tmp"));
+  private static final Path filePath = new Path(workDir, "orc_filter_file.orc");
+
+  private static final int RowCount = 400;
+
+  private static final int scale = 3;
```
pavibhai, line 81: Given that the schema you are using is just int and string, we can remove this.

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
```diff
@@ -225,4 +248,114 @@ public void testRepeatingVector() {
...
+  @Test
+  public void testACIDTable() {
+    ColumnVector[] columnVector = filterContextACID.findColumnVector("string1");
+    assertEquals(1, columnVector.length);
+    assertTrue(columnVector[0] instanceof BytesColumnVector, "Expected a  BytesColumnVector, but found "+ columnVector[0].getClass());
+    columnVector = filterContextACID.findColumnVector("int1");
+    assertEquals(1, columnVector.length);
+    assertTrue(columnVector[0] instanceof LongColumnVector, "Expected a  LongColumnVector, but found "+ columnVector[0].getClass());
+  }
+
```
pavibhai, line 261: .nit please remove the extra new line

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
```diff
@@ -225,4 +248,114 @@ public void testRepeatingVector() {
...
+    }
+  }
+  
+  private void populateColumnValues(TypeDescription typeDescription, ColumnVector[] columnVectors, int index, long value) {
+    for (int columnId = 0; columnId < typeDescription.getChildren().size() ; columnId++) {
+      switch (typeDescription.getChildren().get(columnId).getCategory()) {
+        case INT:
+          ((LongColumnVector)columnVectors[columnId]).vector[index] = value;
+          break;
+        case LONG:
```
pavibhai, line 309: I think it might be cleaner to remove code that is not being used. `LONG` and `DECIMAL`.

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
```diff
@@ -225,4 +248,114 @@ public void testRepeatingVector() {
...
+
+
+  @Test
+  public void testRowFilterWithACIDTable() throws IOException {
+    createAcidORCFile();
+    readSingleRowWithFilter(new Random().nextInt(RowCount));
+    fileSystem.delete(filePath, false);
+    
+  }
+  private void createAcidORCFile() throws IOException {
```
pavibhai, line 270: .nit need a newline

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
```diff
@@ -225,4 +248,114 @@ public void testRepeatingVector() {
...
+          break;
+        case STRUCT:
+          populateColumnValues(typeDescription.getChildren().get(columnId), ((StructColumnVector)columnVectors[columnId]).fields, index, value);
+          break;           
+        default:
+          throw new IllegalArgumentException();
+      }
+    }
+  }
+  private void readSingleRowWithFilter(long id) throws IOException {
```
pavibhai, line 329: .nit need newline

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
```diff
@@ -225,4 +248,114 @@ public void testRepeatingVector() {
...
+      assertTrue(((StructColumnVector) vectorizedRowBatch.cols[5]).fields[0] instanceof LongColumnVector);
+      assertTrue(((StructColumnVector) vectorizedRowBatch.cols[5]).fields[1] instanceof BytesColumnVector);
+      assertEquals(id, ((LongColumnVector) ((StructColumnVector) vectorizedRowBatch.cols[5]).fields[0]).vector[vectorizedRowBatch.selected[0]]);
+      checkStringColumn(id, vectorizedRowBatch);
+      assertFalse(recordReader.nextBatch(vectorizedRowBatch));
+    }
+    assertEquals(1, rowCount);
+  }
+
+  private static void checkStringColumn(long id, VectorizedRowBatch vectorizedRowBatch) {
```
pavibhai, line 354: Can we not just use `BytesColumnVector.toString(rowid)` and compare with the expected value?

pavibhai (commented):
@zratkai Thanks for the changes. Overall this looks good. Couple of minor fixes and we are good to go.

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
zratkai: LONG is used, removed DECIMAL.

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
zratkai: Yes, it can be done. Thank for showing this!

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
zratkai: Fixed.

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
zratkai: Fixed.

java/core/src/test/org/apache/orc/TestOrcFilterContext.java
zratkai: Fixed.

pavibhai:
@zratkai Thanks for the contribution.

If you are able. Can you please raise two PRs that backport this to 1.8 and 1.9 branches?

dongjoon-hyun:
Thank you so much, @zratkai and @pavibhai !

zratkai:
@pavibhai I open PRs for 1.8 and 1.9.

PR #1496, opened 2023-05-16
ORC-1413. Fix for ORC row level filter issue with ACID table

zratkai:
ORC-1413 fix for ORC row level filter issue with ACID table (#1482)

This PR fixes ORC row level filter with ACID table issue.

Without this Hive can not work with ORC 1.8.3. and ACID table and row level filter enabled.

Unit test added.


zratkai:
@pavibhai cherry- picked.

PR #1500, opened 2023-05-16
ORC-1386: [C++] Support schema evolution from numeric to string group/decimal/timestamp

ffacs:
support conversion from numeric to string group/decimal/timestamp

To support schema evolution in c++

UT passed

### DIFF
--- a/c++/src/ConvertColumnReader.cc
+++ b/c++/src/ConvertColumnReader.cc
@@ -240,8 +240,8 @@ namespace orc {
       trueValue.resize(readType.getMaximumLength(), ' ');
       falseValue.resize(readType.getMaximumLength(), ' ');
     } else if (readType.getKind() == VARCHAR) {
-      trueValue = trueValue.substr(0, std::min(4UL, readType.getMaximumLength()));
-      falseValue = falseValue.substr(0, std::min(5UL, readType.getMaximumLength()));
+      trueValue = trueValue.substr(0, std::min(static_cast<uint64_t>(4), readType.getMaximumLength()));
+      falseValue = falseValue.substr(0, std::min(static_cast<uint64_t>(5), readType.getMaximumLength()));
     }
     // cast the bool value to string and truncate to the max length
     for (uint64_t i = 0; i < numValues; ++i) {
--- a/c++/src/ConvertColumnReader.cc
+++ b/c++/src/ConvertColumnReader.cc
@@ -249,8 +249,8 @@ namespace orc {
     std::string falseValue;
   };
 
-  size_t BooleanToStringVariantColumnReader::convertToStrBuffer(ColumnVectorBatch& rowBatch,
-                                                                uint64_t numValues) {
+  uint64_t BooleanToStringVariantColumnReader::convertToStrBuffer(ColumnVectorBatch& rowBatch,
+                                                                  uint64_t numValues) {
     uint64_t size = 0;
     strBuffer.resize(numValues);
     const auto& srcBatch = *SafeCastBatchTo<const BooleanVectorBatch*>(data.get());
@@ -270,7 +270,7 @@ namespace orc {
     NumericToStringVariantColumnReader(const Type& _readType, const Type& fileType,
                                        StripeStreams& stripe, bool _throwOnOverflow)
         : ConvertToStringVariantColumnReader(_readType, fileType, stripe, _throwOnOverflow) {}
-    size_t convertToStrBuffer(ColumnVectorBatch& rowBatch, uint64_t numValues) override;
+    uint64_t convertToStrBuffer(ColumnVectorBatch& rowBatch, uint64_t numValues) override;
   };
 
   template <typename FileTypeBatch>
--- a/c++/src/ConvertColumnReader.cc
+++ b/c++/src/ConvertColumnReader.cc
@@ -186,9 +186,281 @@ namespace orc {
     }
   };
 
+  class ConvertToStringVariantColumnReader : public ConvertColumnReader {
+   public:
+    ConvertToStringVariantColumnReader(const Type& _readType, const Type& fileType,
+                                       StripeStreams& stripe, bool _throwOnOverflow)
+        : ConvertColumnReader(_readType, fileType, stripe, _throwOnOverflow) {}
+
+    void next(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) override {
+      ConvertColumnReader::next(rowBatch, numValues, notNull);
+
+      // cache converted string in the buffer
+      auto totalLength = convertToStrBuffer(rowBatch, numValues, notNull);
+
+      // contact string values to blob buffer of vector batch
+      auto& dstBatch = *SafeCastBatchTo<StringVectorBatch*>(&rowBatch);
+      dstBatch.blob.resize(totalLength);
+      char* blob = dstBatch.blob.data();
+      for (uint64_t i = 0; i < numValues; ++i) {
+        if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+          const auto size = strBuffer[i].size();
+          ::memcpy(blob, strBuffer[i].c_str(), size);
+          dstBatch.data[i] = blob;
+          dstBatch.length[i] = static_cast<int32_t>(size);
+          blob += size;
+        }
+      }
+      strBuffer.clear();
+    }
+
+    virtual size_t convertToStrBuffer(ColumnVectorBatch& rowBatch, uint64_t numValues,
+                                      char* notNull) = 0;
+
+   protected:
+    std::vector<std::string> strBuffer;
+  };
+
+  class BooleanToStringVariantColumnReader : public ConvertToStringVariantColumnReader {
+   public:
+    BooleanToStringVariantColumnReader(const Type& _readType, const Type& fileType,
+                                       StripeStreams& stripe, bool _throwOnOverflow)
+        : ConvertToStringVariantColumnReader(_readType, fileType, stripe, _throwOnOverflow) {}
+
+    size_t convertToStrBuffer(ColumnVectorBatch& rowBatch, uint64_t numValues,
+                              char* notNull) override {
+      size_t size = 0;
+      strBuffer.resize(numValues);
+      const auto& srcBatch = *SafeCastBatchTo<const BooleanVectorBatch*>(data.get());
+      std::string trueValue = "TRUE";
+      std::string falseValue = "FALSE";
+      if (readType.getKind() == CHAR) {
+        trueValue.resize(readType.getMaximumLength(), ' ');
+        falseValue.resize(readType.getMaximumLength(), ' ');
+      } else if (readType.getKind() == VARCHAR) {
+        trueValue = trueValue.substr(0, std::min((uint64_t)4, readType.getMaximumLength()));
+        falseValue = falseValue.substr(0, std::min((uint64_t)5, readType.getMaximumLength()));
+      }
+      // cast the bool value to string and truncate to the max length
+      for (uint64_t i = 0; i < numValues; ++i) {
+        if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+          strBuffer[i] = (srcBatch.data[i] ? trueValue : falseValue);
+          size += strBuffer[i].size();
+        }
+      }
+      return size;
+    }
+  };
+
+  template <typename FileTypeBatch>
+  class NumericToStringVariantColumnReader : public ConvertToStringVariantColumnReader {
+   public:
+    NumericToStringVariantColumnReader(const Type& _readType, const Type& fileType,
+                                       StripeStreams& stripe, bool _throwOnOverflow)
+        : ConvertToStringVariantColumnReader(_readType, fileType, stripe, _throwOnOverflow) {}
+
+    size_t convertToStrBuffer(ColumnVectorBatch& rowBatch, uint64_t numValues,
+                              char* notNull) override {
+      size_t size = 0;
+      strBuffer.resize(numValues);
+      const auto& srcBatch = *SafeCastBatchTo<const FileTypeBatch*>(data.get());
+      if (readType.getKind() == STRING) {
+        for (uint64_t i = 0; i < numValues; ++i) {
+          if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+            strBuffer[i] = std::to_string(srcBatch.data[i]);
+            size += strBuffer[i].size();
+          }
+        }
+      } else if (readType.getKind() == VARCHAR) {
+        const auto maxLength = readType.getMaximumLength();
+        for (uint64_t i = 0; i < numValues; ++i) {
+          if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+            strBuffer[i] = std::to_string(srcBatch.data[i]);
+            if (strBuffer[i].size() > maxLength) {
+              strBuffer[i].resize(maxLength);
+            }
+            size += strBuffer[i].size();
+          }
+        }
+      } else {
+        const auto maxLength = readType.getMaximumLength();
+        for (uint64_t i = 0; i < numValues; ++i) {
+          if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+            strBuffer[i] = std::to_string(srcBatch.data[i]);
+            if (strBuffer[i].size() > maxLength) {
+              strBuffer[i].resize(maxLength);
+            } else {
+              strBuffer[i].resize(maxLength, ' ');
+            }
+            size += strBuffer[i].size();
+          }
+        }
+      }
+      return size;
+    }
+  };
+
+  template <typename FileTypeBatch, typename ReadTypeBatch, bool isFileTypeDouble>
+  class NumericToDecimalColumnReader : public ConvertColumnReader {
+   public:
+    NumericToDecimalColumnReader(const Type& _readType, const Type& fileType, StripeStreams& stripe,
+                                 bool _throwOnOverflow)
+        : ConvertColumnReader(_readType, fileType, stripe, _throwOnOverflow) {
+      precision = static_cast<int32_t>(readType.getPrecision());
+      scale = static_cast<int32_t>(readType.getScale());
+    }
+
+    void next(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) override {
+      ConvertColumnReader::next(rowBatch, numValues, notNull);
+
+      const auto& srcBatch = *SafeCastBatchTo<const FileTypeBatch*>(data.get());
+      auto& dstBatch = *SafeCastBatchTo<ReadTypeBatch*>(&rowBatch);
+      dstBatch.precision = precision;
+      dstBatch.scale = scale;
+      for (uint64_t i = 0; i < numValues; ++i) {
+        if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+          if constexpr (isFileTypeDouble) {
+            convertDoubleToDecimal(dstBatch, i, srcBatch.data[i]);
+          } else {
+            convertIntegerToDecimal(dstBatch, i, srcBatch.data[i]);
+          }
+        }
+      }
+    }
+
+   private:
+    template <typename srcType>
+    void convertDoubleToDecimal(ReadTypeBatch& dstBatch, uint64_t idx, srcType value) {
+      std::string strValue = std::to_string(value);
+      int32_t fromScale = 0;
+      int32_t fromPrecision = static_cast<int32_t>(strValue.length());
+      Int128 i128 = 0;
+      for (size_t i = 0; i < strValue.length(); ++i) {
+        auto c = strValue[i];
+        if (c == '.') {
+          fromScale = static_cast<int32_t>(strValue.length() - i - 1);
+          fromPrecision -= 1;
+          continue;
+        }
+        i128 *= 10;
+        i128 += c - '0';
+      }
+      auto result = convertDecimal(i128, fromPrecision, fromScale, precision, scale);
+      if (result.first) {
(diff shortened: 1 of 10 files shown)


c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+
+    virtual size_t convertToStrBuffer(ColumnVectorBatch& rowBatch, uint64_t numValues) = 0;
+
+   protected:
+    std::vector<std::string> strBuffer;
+  };
+
+  class BooleanToStringVariantColumnReader : public ConvertToStringVariantColumnReader {
+   public:
+    BooleanToStringVariantColumnReader(const Type& _readType, const Type& fileType,
```
wgtmac, line 225: ditto

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+
+    size_t convertToStrBuffer(ColumnVectorBatch& rowBatch, uint64_t numValues) override;
+  };
+
+  size_t BooleanToStringVariantColumnReader::convertToStrBuffer(ColumnVectorBatch& rowBatch,
+                                                                uint64_t numValues) {
+    size_t size = 0;
+    strBuffer.resize(numValues);
+    const auto& srcBatch = *SafeCastBatchTo<const BooleanVectorBatch*>(data.get());
+    std::string trueValue = "TRUE";
```
wgtmac, line 237: Does it follow the Java implementation?

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+    std::string falseValue = "FALSE";
+    if (readType.getKind() == CHAR) {
+      trueValue.resize(readType.getMaximumLength(), ' ');
+      falseValue.resize(readType.getMaximumLength(), ' ');
+    } else if (readType.getKind() == VARCHAR) {
+      trueValue = trueValue.substr(0, std::min(static_cast<uint64_t>(4), readType.getMaximumLength()));
```
wgtmac, line 243: I am not sure if it should throw if truncation happens.

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+          size += strBuffer[i].size();
+        }
+      }
+    } else if (readType.getKind() == VARCHAR) {
+      const auto maxLength = readType.getMaximumLength();
+      for (uint64_t i = 0; i < numValues; ++i) {
+        if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+          strBuffer[i] = std::to_string(srcBatch.data[i]);
+          if (strBuffer[i].size() > maxLength) {
+            strBuffer[i].resize(maxLength);
```
wgtmac, line 284: Should we throw in this case? The result does not make sense at all.

wgtmac: We probably need to throw or nullify if the result needs to be truncated. Otherwise the converted string does not keep original value which is not meaningful.

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+          } else {
+            strBuffer[i].resize(maxLength, ' ');
+          }
+          size += strBuffer[i].size();
+        }
+      }
+    }
+    return size;
+  }
+  template <typename FileTypeBatch, typename ReadTypeBatch, bool isFileTypeDouble>
```
wgtmac, line 305: add one blank line above.

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+    void next(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) override {
+      ConvertColumnReader::next(rowBatch, numValues, notNull);
+
+      const auto& srcBatch = *SafeCastBatchTo<const FileTypeBatch*>(data.get());
+      auto& dstBatch = *SafeCastBatchTo<ReadTypeBatch*>(&rowBatch);
+      dstBatch.precision = precision;
+      dstBatch.scale = scale;
+      for (uint64_t i = 0; i < numValues; ++i) {
+        if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+          if constexpr (isFileTypeDouble) {
```
wgtmac, line 324:
```suggestion
          if constexpr (isFloatingFileType) {
```

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+            convertDoubleToDecimal(dstBatch, i, srcBatch.data[i]);
+          } else {
+            convertIntegerToDecimal(dstBatch, i, srcBatch.data[i]);
+          }
+        }
+      }
+    }
+
+   private:
+    template <typename srcType>
```
wgtmac, line 334:
```suggestion
    template <typename SrcType>
```

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+          } else {
+            dstBatch.values[idx] = result.second.toLong();
+          }
+        } else {
+          dstBatch.values[idx] = result.second;
+        }
+      }
+    }
+
+    template <typename srcType>
```
wgtmac, line 366:
```suggestion
    template <typename SrcType>
```

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+        }
+      }
+    }
+
+   private:
+    template <typename srcType>
+    void convertDoubleToDecimal(ReadTypeBatch& dstBatch, uint64_t idx, srcType value) {
+      std::string strValue = std::to_string(value);
+      int32_t fromScale = 0;
+      int32_t fromPrecision = static_cast<int32_t>(strValue.length());
```
wgtmac, line 338: What if value is a negative number?

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+            convertIntegerToDecimal(dstBatch, i, srcBatch.data[i]);
+          }
+        }
+      }
+    }
+
+   private:
+    template <typename srcType>
+    void convertDoubleToDecimal(ReadTypeBatch& dstBatch, uint64_t idx, srcType value) {
+      std::string strValue = std::to_string(value);
```
wgtmac, line 336: String conversion would be slow. Can we simply convert it into an integer and reuse the integer code path?

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
...
+      if (dstBatch.nanoseconds[idx] < 0) {
+        dstBatch.data[idx] -= 1;
+        dstBatch.nanoseconds[idx] += static_cast<int32_t>(1e9);
+      }
+    } else {
+      dstBatch.data[idx] = value;
+      dstBatch.nanoseconds[idx] = 0;
+    }
+    if (needConvertTimezone) {
+      dstBatch.data[idx] = readerTimezone.convertFromUTC(dstBatch.data[idx]);
```
wgtmac, line 452: Does this follow the Java impl as well?

c++/src/ConvertColumnReader.cc
```diff
@@ -235,10 +514,64 @@ namespace orc {
...
+
 #define CASE_CREATE_READER(TYPE, CONVERT) \
   case TYPE:                              \
     return std::make_unique<CONVERT##ColumnReader>(_readType, fileType, stripe, throwOnOverflow);
 
+  const static int32_t MAX_PRECISION_64 = 18;
+
+#define CASE_CREATE_DECIMAL_READER(FROM)                                                   \
+  case DECIMAL: {                                                                          \
+    if (_readType.getPrecision() <= MAX_PRECISION_64) {                                    \
```
wgtmac, line 566:
```suggestion
    if (_readType.getPrecision() > 0 && _readType.getPrecision() <= MAX_PRECISION_64) {                                    \
```

dongjoon-hyun (commented):
In addition to the above comment, could you make CI happy, @ffacs ?
```
/Users/runner/work/orc/orc/c++/src/ConvertColumnReader.cc:189:9: error: 'ConvertToStringVariantColumnReader' has no out-of-line virtual method definitions; its vtable will be emitted in every translation unit [-Werror,-Wweak-vtables]
  class ConvertToStringVariantColumnReader : public ConvertColumnReader {
        ^
/Users/runner/work/orc/orc/c++/src/ConvertColumnReader.cc:392:9: error: 'ConvertToTimestampColumnReader' has no out-of-line virtual method definitions; its vtable will be emitted in every translation unit [-Werror,-Wweak-vtables]
  class ConvertToTimestampColumnReader : public ConvertColumnReader {
        ^
2 errors generated.
make[2]: *** [c++/src/CMakeFiles/orc.dir/ConvertColumnReader.cc.o] Error 1
make: *** [all] Error 2
```

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,289 @@ namespace orc {
     }
   };
 
+  class ConvertToStringVariantColumnReader : public ConvertColumnReader {
+   public:
+    ConvertToStringVariantColumnReader(const Type& _readType, const Type& fileType,
```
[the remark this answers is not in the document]
ffacs: > `_readType` and `_throwOnOverflow` do not need to have `_` prefix, right?

error: parameter 'readType' shadows member inherited from type 'ConvertColumnReader' [-Werror,-Wshadow-field]

c++/src/Timezone.cc
```diff
@@ -587,6 +587,12 @@ namespace orc {
       return clk + getVariant(clk).gmtOffset;
     }
 
+    int64_t convertFromUTC(int64_t clk) const override {
```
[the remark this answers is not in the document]
ffacs: > Could you add a test for this?

done

c++/src/ConvertColumnReader.cc
ffacs: > Does it follow the Java implementation?

https://github.com/apache/orc/blob/ec2ea9c6aff8b8515452df651f08695639c18cbb/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java#L1118-L1133
https://github.com/apache/orc/blob/ec2ea9c6aff8b8515452df651f08695639c18cbb/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java#L123-L146
I think so, but i am not good at java, please take a look

c++/src/ConvertColumnReader.cc
ffacs: the getPrecision() returns a uint64_t value

c++/src/ConvertColumnReader.cc
ffacs: > Does this follow the Java impl as well?

https://github.com/apache/orc/blob/ec2ea9c6aff8b8515452df651f08695639c18cbb/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java#L1518-L1524

please take a look

c++/src/ConvertColumnReader.cc
wgtmac: Yes, but precision == 0 indicates it should use int128.

ffacs: > Yes, but precision == 0 indicates it should use int128.

done

c++/src/ConvertColumnReader.cc
ffacs: > String conversion would be slow. Can we simply convert it into an integer and reuse the integer code path?

The integer portion of a double may larger then std:: numeric_limits<uint64_t>::max(), so i convert it to Int128 first.

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,327 @@ namespace orc {
...
+          trueValue.resize(readType.getMaximumLength(), ' ');
+          falseValue.resize(readType.getMaximumLength(), ' ');
+        }
+      }
+    }
+
+    size_t convertToStrBuffer(ColumnVectorBatch& rowBatch, uint64_t numValues) override;
+
+    private:
+      std::string trueValue;
```
wgtmac, line 248: trueValue and falseValue can actually be const static variable

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,327 @@ namespace orc {
...
+  }
+
+  class BooleanToStringVariantColumnReader : public ConvertToStringVariantColumnReader {
+   public:
+    BooleanToStringVariantColumnReader(const Type& _readType, const Type& fileType,
+                                       StripeStreams& stripe, bool _throwOnOverflow)
+        : ConvertToStringVariantColumnReader(_readType, fileType, stripe, _throwOnOverflow) {
+      trueValue = "TRUE";
+      falseValue = "FALSE";
+      if (readType.getKind() != STRING) {
```
wgtmac, line 233:
```suggestion
      if (readType.getKind() == CHAR || readType.getKind() == VARCHAR) {
```

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,327 @@ namespace orc {
...
+        if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+          strBuffer[i] = std::to_string(srcBatch.data[i]);
+          if (strBuffer[i].size() > maxLength) {
+            handleOverflow<decltype(srcBatch.data[i]), std::string>(rowBatch, i, throwOnOverflow);
+          } else {
+            size += strBuffer[i].size();
+          }
+        }
+      }
+    } else {
```
wgtmac, line 301:
```suggestion
    } else if (readType.getKind() == CHAR) {
```
And explicitly throw in the else branch

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,327 @@ namespace orc {
...
+  size_t BooleanToStringVariantColumnReader::convertToStrBuffer(ColumnVectorBatch& rowBatch,
+                                                                uint64_t numValues) {
+    size_t size = 0;
+    strBuffer.resize(numValues);
+    const auto& srcBatch = *SafeCastBatchTo<const BooleanVectorBatch*>(data.get());
+    // cast the bool value to string and truncate to the max length
+    for (uint64_t i = 0; i < numValues; ++i) {
+      if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+        strBuffer[i] = (srcBatch.data[i] ? trueValue : falseValue);
+        size += strBuffer[i].size();
```
wgtmac, line 261: Though unlikely, an overflow check is good to be here and all additions below.

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,327 @@ namespace orc {
...
+    void next(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) override {
+      ConvertColumnReader::next(rowBatch, numValues, notNull);
+
+      const auto& srcBatch = *SafeCastBatchTo<const FileTypeBatch*>(data.get());
+      auto& dstBatch = *SafeCastBatchTo<ReadTypeBatch*>(&rowBatch);
+      dstBatch.precision = precision;
+      dstBatch.scale = scale;
+      for (uint64_t i = 0; i < numValues; ++i) {
+        if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+          if constexpr (isFLoatingFileType) {
```
wgtmac, line 343:
```suggestion
          if constexpr (isFloatingFileType) {
```

c++/src/Int128.cc
```diff
@@ -488,4 +488,42 @@ namespace orc {
     return value;
   }
 
+  std::pair<bool, Int128> convertDecimal(Int128 value, int32_t fromScale,
```
wgtmac, line 491: Please add some tests for this new function.

c++/src/ConvertColumnReader.cc
```diff
@@ -186,10 +186,327 @@ namespace orc {
...
+          } else {
+            convertIntegerToDecimal(dstBatch, i, srcBatch.data[i]);
+          }
+        }
+      }
+    }
+
+   private:
+    template <typename SrcType>
+    void convertDoubleToDecimal(ReadTypeBatch& dstBatch, uint64_t idx, SrcType value) {
```
wgtmac, line 354: Can we extract its core logic into Int128.hh and make sure it is covered by some tests?

wgtmac (commented):
The implementation looks good to me. Left some comments mostly for tests.

c++/src/ConvertColumnReader.cc
ffacs: > Can we extract its core logic into Int128.hh and make sure it is covered by some tests?

done

c++/src/ConvertColumnReader.cc
ffacs: done

c++/src/ConvertColumnReader.cc
ffacs: I make it uint64_t then it should not overflow.

c++/src/ConvertColumnReader.cc
ffacs: done

c++/src/ConvertColumnReader.cc
ffacs: done

c++/src/ConvertColumnReader.cc
ffacs: It can't be const static because type 'CHAR' may modify them.

c++/src/Int128.cc
```diff
@@ -488,4 +488,84 @@ namespace orc {
...
   }
 
+  std::pair<bool, Int128> convertDecimal(Int128 value, int32_t fromScale, int32_t toPrecision,
+                                         int32_t toScale, bool round) {
+    std::pair<bool, Int128> result;
+    bool negative = value < 0;
+    result.second = value.abs();
+    result.first = false;
+
+    Int128 upperBound = scaleUpInt128ByPowerOfTen(1, toPrecision, result.first);
```
wgtmac, line 498: nit: check result.first in case of invalid toPrecision

wgtmac: Or assert fromScale, toPrecision, toScale are in the valid ranges.

c++/src/Int128.cc
```diff
@@ -488,4 +488,84 @@ namespace orc {
...
+
+    result.second += roundOffset;
+    if (negative) {
+      result.second *= -1;
+    }
+    return result;
+  }
+
+  template <typename T>
+  std::pair<bool, Int128> convertDecimal(T value, int32_t precision, int32_t scale) {
```
wgtmac, line 531: Should we check if value is nan?

c++/src/Int128.cc
```diff
@@ -488,4 +488,84 @@ namespace orc {
...
+    std::pair<bool, Int128> result = {false, 0};
+    if (value <= -std::ldexp(static_cast<T>(1), 127) ||
```
wgtmac, line 533: Make std::ldexp(static_cast<T>(1), 127) const static as it is repeated used.

c++/test/TestInt128.cc
```diff
@@ -643,4 +643,242 @@ namespace orc {
...
+    pair = convertDecimal(fromDouble, toPrecision, toScale);
+    EXPECT_EQ(pair.first, true);  // overflow
+
+    toPrecision = 8;
+    toScale = 3;
+    pair = convertDecimal(fromDouble, toPrecision, toScale);
+    EXPECT_EQ(pair.first, false);  // no overflow
+    EXPECT_EQ(pair.second, Int128(12345136)) << pair.second.toString();
+
+    fromDouble = -12345.13579;
```
wgtmac, line 786: Could you add more numbers to test? Would be good to test float type as well.

wgtmac: Probably +0.0, -0.0, and NaN should be added

wgtmac: Add large numbers that are beyond 18 digits.

c++/src/Int128.cc
ffacs: > Or assert fromScale, toPrecision, toScale are in the valid ranges.

done

c++/src/Int128.cc
```diff
@@ -488,4 +488,84 @@ namespace orc {
...
+    }
+
+    result.second += roundOffset;
+    if (negative) {
+      result.second *= -1;
+    }
+    return result;
+  }
+
+  template <typename T>
```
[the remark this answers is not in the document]
ffacs: > Should we use std::enable_if to restrict T to be floating type only?

done

c++/src/Int128.cc
ffacs: > Should we check if value is nan?

done

c++/src/Int128.cc
ffacs: > Make std::ldexp(static_cast(1), 127) const static as it is repeated used.

done

wgtmac:
Thanks @ffacs! Overall looks good.

Could you please make all the CIs pass?

wgtmac (approved):
LGTM. Thanks!

c++/src/ConvertColumnReader.cc
```diff
@@ -257,169 +603,169 @@ namespace orc {
...
           case DATE:
-          case VARCHAR:
-          case CHAR:
-          case TIMESTAMP_INSTANT:
             CASE_EXCEPTION
         }
       }
       case LONG: {
         switch (_readType.getKind()) {
-          CASE_CREATE_READER(BOOLEAN, LongToBoolean);
```
dongjoon-hyun, line 356: We need this semicolon removal in `branch-1.9` too.

dongjoon-hyun:
Let me bring this to branch-1.9 too~

wgtmac:
Thanks @dongjoon-hyun!

PR #1521, opened 2023-05-27
ORC-1434: Remove `org.apache.hadoop` from `dependabot.yml`

dongjoon-hyun:

This PR aims to remove `org.apache.hadoop` from `dependabot.yml.`

After ORC-1430, Apache ORC can use te latest Apache Hadoop always.

Since this is a dependabot setting, this is irrelevant to CI (Java/C++).


dongjoon-hyun:
Merged to main for Apache ORC 2.0.

PR #1529, opened 2023-06-04
ORC-1440: Check for protobuf config based module

nehaljwani:
The recent versions of `libprotobuf` (>=4.22.0) provide their own CMake configs. We should check for the existence of the config and use the hand-rolled version as a fallback as it does not capture all deps.

See also: https://gitlab.kitware.com/cmake/cmake/-/issues/24321

### DIFF
--- a/cmake_modules/FindProtobuf.cmake
+++ b/cmake_modules/FindProtobuf.cmake
@@ -32,36 +32,58 @@ endif()
 
 message (STATUS "PROTOBUF_HOME: ${PROTOBUF_HOME}")
 
-find_path (PROTOBUF_INCLUDE_DIR google/protobuf/io/zero_copy_stream.h HINTS
-  ${_protobuf_path}
-  NO_DEFAULT_PATH
-  PATH_SUFFIXES "include")
+find_package (Protobuf CONFIG)
+if (Protobuf_FOUND)
+    set (PROTOBUF_LIBRARY protobuf::libprotobuf)
+    set (PROTOBUF_STATIC_LIB PROTOBUF_STATIC_LIB-NOTFOUND)
+    set (PROTOC_LIBRARY protobuf::libprotoc)
+    set (PROTOC_STATIC_LIB PROTOC_STATIC_LIB-NOTFOUND)
+    set (PROTOBUF_EXECUTABLE protobuf::protoc)
 
-find_path (PROTOBUF_INCLUDE_DIR google/protobuf/io/coded_stream.h HINTS
-  ${_protobuf_path}
-  NO_DEFAULT_PATH
-  PATH_SUFFIXES "include")
+    get_target_property (target_type protobuf::libprotobuf TYPE)
+    if (target_type STREQUAL "STATIC_LIBRARY")
+        set(PROTOBUF_STATIC_LIB protobuf::libprotobuf)
+    endif ()
 
-find_library (PROTOBUF_LIBRARY NAMES protobuf HINTS
-  ${_protobuf_path}
-  PATH_SUFFIXES "lib")
+    get_target_property (target_type protobuf::libprotoc TYPE)
+    if (target_type STREQUAL "STATIC_LIBRARY")
+        set (PROTOC_STATIC_LIB protobuf::libprotoc)
+    endif ()
 
-find_library (PROTOBUF_STATIC_LIB NAMES ${CMAKE_STATIC_LIBRARY_PREFIX}protobuf${CMAKE_STATIC_LIBRARY_SUFFIX} HINTS
-  ${_protobuf_path}
-  PATH_SUFFIXES "lib")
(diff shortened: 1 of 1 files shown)


cmake_modules/FindProtobuf.cmake
```diff
@@ -26,6 +26,24 @@
 # PROTOBUF_EXECUTABLE: location of protoc
 
 
+find_package(Protobuf CONFIG)
+if(Protobuf_FOUND)
```
dongjoon-hyun, line 30: Thank you for making a PR. Could you follow the existing coding style?

https://github.com/apache/orc/blob/ec2ea9c6aff8b8515452df651f08695639c18cbb/cmake_modules/FindProtobuf.cmake#L74

dongjoon-hyun:
Also, cc @wgtmac

cmake_modules/FindProtobuf.cmake
nehaljwani: I've updated the commit.

wgtmac:
Thanks @nehaljwani and @dongjoon-hyun!

These changes look fine to me. However, we are still on a very old version of protobuf (3.5.1). I am not sure if it will work properly with higher versions of protobuf. I have seen issues when Apache Arrow tried to compile ORC with a different protobuf version.

nehaljwani:
I also maintain the conda-forge recipe for orc. The latest build there was successful with this patch and protobuf v4.23.2

cmake_modules/FindProtobuf.cmake
```diff
@@ -26,6 +26,24 @@
...
+    set (PROTOC_STATIC_LIB PROTOC_STATIC_LIB-NOTFOUND)
+    get_target_property (target_type protobuf::libprotobuf TYPE)
+    if (target_type STREQUAL "STATIC_LIBRARY")
+        set(PROTOBUF_STATIC_LIB protobuf::libprotobuf)
+    endif ()
+    get_target_property (target_type protobuf::libprotoc TYPE)
+    if (target_type STREQUAL "STATIC_LIBRARY")
+        set (PROTOC_STATIC_LIB protobuf::libprotoc)
+    endif ()
+    get_target_property (PROTOBUF_INCLUDE_DIR protobuf::libprotoc INTERFACE_INCLUDE_DIRECTORIES)
```
stiga-huang, line 43: `PROTOBUF_EXECUTABLE` is missing here. Is it set in protobuf's own CMake configs?

cmake_modules/FindProtobuf.cmake
```diff
@@ -26,6 +26,24 @@
...
+    return ()
```
stiga-huang, line 44: Before `return`, it'd be nice to print the variables as we did at line 93 - 102. Or will the protobuf's config print some info?

wgtmac:
Probably it is a good time to upgrade protobuf version as well. @stiga-huang

cmake_modules/FindProtobuf.cmake
nehaljwani: Updated in the revised commit.

cmake_modules/FindProtobuf.cmake
nehaljwani: Updated in the revised commit. Now it prints:
```
-- PROTOBUF_HOME: /path/to/libprotobuf
-- Found the Protobuf headers: /path/to/libprotobuf/include
-- Found the Protobuf library: protobuf::libprotobuf
-- Found the Protoc library: protobuf::libprotoc
-- Found the Protoc executable: protobuf::protoc
```

nehaljwani:
>Do you build the entire orc lib (including tools) or just e.g. liborc.a? Just curious if all the functionality work with protobuf v4.23.2. If so, we can add tests to guard them.

I used the following steps locally (not sure if it covers all targets):
```
cmake -S/path/to/orc -B/tmp/build -GNinja \
   -DCMAKE_PREFIX_PATH=$CONDA_PREFIX \
   -DPROTOBUF_HOME=$CONDA_PREFIX \
   -DBUILD_JAVA=OFF \
   -DZLIB_HOME=$CONDA_PREFIX \
   -DSNAPPY_HOME=$CONDA_PREFIX \
   -DLZ4_HOME=$CONDA_PREFIX \
   -DZSTD_HOME=$CONDA_PREFIX \
   -DSTOP_BUILD_ON_WARNING=OFF \
   -DBUILD_CPP_TESTS=OFF

cmake --build /tmp/build
```

wgtmac:
Could you please have a try with `-DBUILD_CPP_TESTS=ON -DBUILD_TOOLS=ON`?

nehaljwani:
Done, the build succeeds with both options turned on with my patch applied.

wgtmac (approved):
Thanks! LGTM

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Thank you, @nehaljwani , @wgtmac , @stiga-huang . Merged to main for Apache ORC 2.0.

PR #1538, opened 2023-06-19
ORC-1446: Publish snapshot from branch-1.9

dongjoon-hyun:

This PR aims to publish snapshot from `branch-1.9`.

We missed this when we create `branch-1.9`.

We need to verify this after merging.


dongjoon-hyun:
Let me merge this because this is an infra PR which is irrelevant to PR Builder.

Issue #1532, opened 2023-06-11
Release Apache ORC 1.9.0


dongjoon-hyun:
- [x] GitHub Action CI Check (Java 8/11/17/21-ea) and `Ubuntu`/`MacOS`/`Windows`.
   - https://github.com/apache/orc/actions/workflows/build_and_test.yml?query=branch%3Abranch-1.9
- [x] Docker Tests (CentOS7, Debian10/11, Ubuntu20/22, Fedora37)
  - [x] https://github.com/apache/orc/pull/1548

```bash
$ tail -n5 logs/*test*                                                                            
==> logs/centos7-test.log <==
100% tests passed, 0 tests failed out of 8

Total Test time (real) = 772.17 sec
Built target test-out
Finished centos7 at Fri Jun 23 02:03:16 PM PDT 2023

==> logs/debian10-test.log <==
100% tests passed, 0 tests failed out of 8

Total Test time (real) = 697.48 sec
Built target test-out
Finished debian10 at Fri Jun 23 02:03:56 PM PDT 2023

==> logs/debian11_jdk=11-test.log <==
100% tests passed, 0 tests failed out of 8

Total Test time (real) = 810.38 sec
Built target test-out
Finished debian11_jdk=11 at Fri Jun 23 02:03:38 PM PDT 2023

==> logs/debian11-test.log <==
100% tests passed, 0 tests failed out of 8

Total Test time (real) = 725.80 sec
Built target test-out
Finished debian11 at Fri Jun 23 02:01:37 PM PDT 2023

==> logs/fedora37-test.log <==
100% tests passed, 0 tests failed out of 8

Total Test time (real) = 767.15 sec
Built target test-out
Finished fedora37 at Fri Jun 23 02:03:47 PM PDT 2023

==> logs/ubuntu20-test.log <==
100% tests passed, 0 tests failed out of 8

Total Test time (real) = 552.28 sec
Built target test-out
Finished ubuntu20 at Fri Jun 23 02:05:10 PM PDT 2023

==> logs/ubuntu22_jdk=11_cc=clang-test.log <==
100% tests passed, 0 tests failed out of 8

Total Test time (real) = 223.52 sec
Built target test-out
Finished ubuntu22_jdk=11_cc=clang at Fri Jun 23 02:40:32 PM PDT 2023

==> logs/ubuntu22_jdk=11-test.log <==
100% tests passed, 0 tests failed out of 8

Total Test time (real) = 727.59 sec
Built target test-out
[...]

PR #1420, opened 2023-02-21
ORC-1377: [C++] Enforce override keyword

wgtmac:
### What changes were proposed in this pull request?
Enforce adding `override` keyword to virtual functions.

### Why are the changes needed?
A lot of warnings are related to missing override keyword in the Github Actions.

### How was this patch tested?
Make sure all tests pass.

PR #1545, opened 2023-06-22
ORC-1450: Stop enforcing `override` keyword rule

dongjoon-hyun:

This PR aims to recover Clang 10 dev environment by stopping enforcing `override` keyword.

- This is a partial revert of #1420 because we still keep the changed other code.
- GitHub Action missed this.
  - `Clang on Ubuntu 20.04` job missed this because it uses Clang 11.0.0.
  - `Clang on MacOS 11` job missed this because it uses AppleClang 13.0.0.

During `Apache ORC 1.9.0 Preparation`, I found that this broke `ubuntu20_jdk=11_cc=clang` docker test due to the `unknown warning group` errors.
- #1532

Pass the CI and do the manual testing with Docker environment.


dongjoon-hyun:
cc @williamhyun and @wgtmac

dongjoon-hyun:
Let me merge this in order to proceed Apache ORC 1.9.0.

PR #1552, opened 2023-06-25
ORC-1455: [C++] Fix build with unused macro in CpuInfoUtil.cc

wgtmac:
Move UNUSED macro under the definition of CPUINFO_ARCH_X86 in order to avoid unused macro error.

Without this patch, the C++ build fails on MacOS and other non-x86 machines.

It builds successfully and passes all tests on my laptop.


guiyanakuang (approved):

dongjoon-hyun:
Thank you, @wgtmac and @guiyanakuang .

PR #1561, opened 2023-07-05
ORC-1460: Update ORC spec to clarify how dictionary entries are sorted

progval:

The spec is updated to clarify that dictionary entries are sorted based on the UTF-8 encoding, not on any of the Unicode Collation algorithms.

Not strictly needed, but I had a moment of doubt when reading the spec, and had to check the implementation.

### How was this patch tested

This matches the C++ implementation:

https://github.com/apache/orc/blob/294a5e28f7f0420eb1fdc76dffc33608692c1b20/c%2B%2B/src/ColumnWriter.cc#L913-L923

### DIFF
--- a/site/specification/ORCv0.md
+++ b/site/specification/ORCv0.md
@@ -626,7 +626,8 @@ the length of each value is written into the LENGTH stream. In direct
 encoding, if the values were ["Nevada", "California"]; the DATA
 would be "NevadaCalifornia" and the LENGTH would be [6, 10].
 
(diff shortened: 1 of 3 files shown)


dongjoon-hyun:
Could you make an Apache ORC JIRA and use it for this PR? Or, we can use `MINOR` instead of JIRA ID like the following.
```
MINOR: Clarify how dictionary entries are sorted in ORC specs
```

progval:
Done

cxzl25 (approved):

wgtmac (approved):

Issue #1543, opened 2023-06-20
ORC-1531: Create `orc-format` module and repo


dongjoon-hyun:
- https://github.com/apache/orc-format
- https://repository.apache.org/content/groups/snapshots/org/apache/orc/orc-format/1.0.0-SNAPSHOT/

deshanxiao:
Thank you for adding the task. @dongjoon-hyun

mystic-lama:
Is the ask to create a repo very similar to [parquet-format](https://github.com/apache/parquet-format), that shall host the specification and [orc_proto.proto](https://github.com/mystic-lama/orc/blob/main/proto/orc_proto.proto). Possibly generating bindings too.

dongjoon-hyun:
Yes, that's the model, @mystic-lama .

mystic-lama:
Did a quick take at this and created the [orc-format](https://github.com/mystic-lama/orc-format-init) repo. If this is going in right direction, can continue to work on it. Feedback appreaciated

mystic-lama:
Create JIRA [ORC-1531](https://issues.apache.org/jira/browse/ORC-1531) to track the work and discussion

dongjoon-hyun:
Thank you, @mystic-lama . I took a look. 

For the versioning,

https://github.com/mystic-lama/orc-format-init/blob/1f88cb6cc9fd47f430e78aa212923635550fd710/pom.xml#L25C7-L25C7

The following is a little too much for me because we are still on `ORCv1` and will not introduce any format-wise change.
```
2.0.0-SNAPSHOT
```

dongjoon-hyun:
In addition, we don't need `nohive` and `shaded-protobuf` classifier for now.

```
$ ls -al
total 9176
drwxr-xr-x  8 dongjoon  staff      256 Dec  3 15:15 .
drwxr-xr-x  4 dongjoon  staff      128 Dec  3 15:15 ..
-rw-r--r--  1 dongjoon  staff      291 Dec  3 15:15 _remote.repositories
-rw-r--r--  1 dongjoon  staff     1127 Dec  3 15:15 maven-metadata-local.xml
-rw-r--r--  1 dongjoon  staff  2109804 Dec  3 15:15 orc-format-2.0.0-SNAPSHOT-nohive.jar
-rw-r--r--  1 dongjoon  staff  2109804 Dec  3 15:15 orc-format-2.0.0-SNAPSHOT-shaded-protobuf.jar
[...]

PR #1565, opened 2023-07-24
[DRAFT] ORC-1463: [Java] Support Brotli codec

### LINKED ISSUE #1563: ORC-1463: Support brotli compression codec

deshanxiao:
This PR is aimed to add Brotli codec defined by [RFC7932](https://www.ietf.org/rfc/rfc7932.txt).

To support more codec in ORC. 

UT and CI pipeline.

This closes #1563


deshanxiao:
`com.github.rdblue:brotli-codec` has no binary for aarch64. We can skip it.

java/pom.xml
```diff
@@ -271,6 +272,12 @@
         <version>1.1.1</version>
         <optional>true</optional>
       </dependency>
+      <dependency>
+        <groupId>com.github.rdblue</groupId>
+        <artifactId>brotli-codec</artifactId>
+        <version>${brotli-codec.version}</version>
+        <optional>true</optional>
```
dongjoon-hyun, line 279: Oh, got it. Apache ORC community has only an `Optional` dependency.

java/shims/pom.xml
```diff
@@ -42,6 +42,10 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.github.rdblue</groupId>
+      <artifactId>brotli-codec</artifactId>
+    </dependency>
```
dongjoon-hyun, line 48: Sorry, but this looks wrong to me. Please do not add the compression dependency to `shims` module.

dongjoon-hyun (commented):
Thank you for making a PR. However, I have a concern on the dependency management. Like Apache ORC 1.9, `shim` module should have only `Hadoop`-shim related dependency and eventually we want to slim or remove `shim` module by using a single Hadoop client library.

- https://mvnrepository.com/artifact/org.apache.orc/orc-shims/1.9.0

java/core/pom.xml
```diff
@@ -72,6 +72,10 @@
       <groupId>org.threeten</groupId>
       <artifactId>threeten-extra</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.github.rdblue</groupId>
```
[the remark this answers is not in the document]
deshanxiao: I tried looking for more common repositories, but couldn't find them. Here is parquet pom file: https://github.com/apache/parquet-mr/blob/d2c3c6d2e761a17b75d56fd356a37a2f754072f7/parquet-hadoop/pom.xml#L207

I'm glad to update it if we can find a better one.

java/shims/pom.xml
deshanxiao: Got it, let me try to re-implement this part.

deshanxiao:
It seems that pom file has some foramt errors. Using `mvn spotless:apply` can auto fix it.

deshanxiao:
Disable unit test when arch is `aarch64`. 

Hi @dongjoon-hyun  Could you help me trigger the Workflow? Thanks~

dongjoon-hyun:
I checked but it seems the workflow is triggered already, @deshanxiao .

deshanxiao:
Thanks @guiyanakuang for trigger the workflow.

deshanxiao:
Hi @dongjoon-hyun, the pipeline has passed. If you have any comments for the PR. Please ping me, thanks~

dongjoon-hyun (dismissed):
@deshanxiao We have two issues still.
1. ARM64 Issue: https://issues.apache.org/jira/browse/PARQUET-1975
2. Apache Parquet has it as `runtime` dependency. I guess we need to do in the same way.

dongjoon-hyun:
BTW, do you know why [HADOOP-13126](https://issues.apache.org/jira/browse/HADOOP-13126) is still open?

And, Hadoop JIRA seems to say the following. Could you double-check if there is no license issue or not?
> [Brotli is licensed with the MIT license](https://github.com/google/brotli/blob/master/LICENSE), and the [JNI library jbrotli is ALv2](https://github.com/MeteoGroup/jbrotli/blob/master/LICENSE).

deshanxiao:
> BTW, do you know why [HADOOP-13126](https://issues.apache.org/jira/browse/HADOOP-13126) is still open?
> 

It seems that the reviewer think this modification is too large to cause the Hadoop PR freeze.
See https://github.com/apache/hadoop/pull/2723.

> And, Hadoop JIRA seems to say the following. Could you double-check if there is no license issue or not?
> 
> > [Brotli is licensed with the MIT license](https://github.com/google/brotli/blob/master/LICENSE), and the [JNI library jbrotli is ALv2](https://github.com/MeteoGroup/jbrotli/blob/master/LICENSE).

From the dependencies:
https://github.com/rdblue/brotli-codec/blob/9bbc2bcec1e7307e84552f28e90f20cb3099f189/build.gradle#L58

We are only indirectly referencing `jbrotli`. So the license requirement you mentioned is right.

BTW,  `brotli-codec` is just an adapter to meet with hadoop codec framework. We can use jbrotli directly. What do you think? @dongjoon-hyun

java/shims/src/java/org/apache/orc/impl/HadoopShims.java
```diff
@@ -35,6 +35,7 @@ enum DirectCompressionType {
     ZLIB_NOHEADER,
     ZLIB,
     SNAPPY,
+    BROTLI,
```
wgtmac, line 38: IIUC, shims here are used to workaround codec provided by Hadoop libraries of different versions. It seems that we do not need this line.

proto/orc_proto.proto
```diff
@@ -386,6 +386,7 @@ enum CompressionKind {
   LZO = 3;
   LZ4 = 4;
   ZSTD = 5;
+  BROTLI = 6;
```
wgtmac, line 389: In order to check this in, we'd better update the website and the C++ code (at least throw an unimplemented exception on the reader side)

wgtmac (commented):
Before diving into the detail implementation, I just want to make sure if brotli is widely used compared to others. Google itself does not provide an encoder implementation in pure Java. If the benefit it brings does not worth the effort of supporting it, I would rather not do this.

proto/orc_proto.proto
deshanxiao: Thanks @wgtmac I will do it.

deshanxiao:
> I found this one but not sure if it is mature enough: https://github.com/hyperxpro/Brotli4j

From 
https://issues.apache.org/jira/browse/PARQUET-1975?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=17281798#comment-17281798

Looks like someone else has had a similar idea. It seems that someone has mentioned a similar idea, but for some reason, parquet still uses a temporary maven library.

I checked that netty is using this library in:
https://github.com/netty/netty/blob/1c5053740dccc4ec952b4cfb8c4bcf697b49d9b6/pom.xml#L763

We could try to re-implement with it @wgtmac

java/shims/src/java/org/apache/orc/impl/HadoopShims.java
deshanxiao: Thanks I will remove it.

deshanxiao:
> Before diving into the detail implementation, I just want to make sure if brotli is widely used compared to others. Google itself does not provide an encoder implementation in pure Java. If the benefit it brings does not worth the effort of supporting it, I would rather not do this.

From the test results of this [Jira](https://issues.apache.org/jira/browse/HADOOP-13126), I am positive about supporting brotli.

I add an [Issue](https://github.com/google/brotli/issues/1060) to track it.

deshanxiao:
As mentioned in this https://github.com/google/brotli/issues/1060, we could wait for official support to come out.

I will pause this PR until its dependencies are ready, what do you think? @dongjoon-hyun @wgtmac

dongjoon-hyun:
+1 for pausing. Thank you, @deshanxiao .

dongjoon-hyun (commented):
To @deshanxiao and @wgtmac , here are my thoughts.

1. For `Parquet` feature parity, we need this.
2. Since this PR needs to change `orc_proto.proto`, we need to proceed before #1543 (orc-format v1.0.0) if we want to have this.
3. If there is no upstream change, the only way could be the `Parquet` way because this is for `Parquet`-feature parity.

- Parquet doesn't support this in arm64 architecture. Also, it's `optional` and `runtime` dependency. We should follow it if we want to proceed this.

```
  <profiles>
    <!-- PARQUET-1975 Do not add brotli-codec for ARM64 architectures -->
    <profile>
      <id>non-aarch64</id>
      <activation>
        <os>
          <arch>!aarch64</arch>
        </os>
      </activation>
      <dependencies>
        <dependency>
          <groupId>com.github.rdblue</groupId>
          <artifactId>brotli-codec</artifactId>
          <version>${brotli-codec.version}</version>
          <scope>runtime</scope>
          <optional>true</optional>
        </dependency>
      </dependencies>
    </profile>
  </profiles>
```

WDYT?

deshanxiao:
+1 for add this to orc-format repo.

BTW, before implementing the feature, please give me some time to summarize the pros and cons of all third-party libraries.

wgtmac:
Do we need a PoC implementation on the C++ side before making a decision? AFAIK, official C++ brotli library is complete and it is used in the Apache Arrow for years.

dongjoon-hyun:
@wgtmac No~ This PR (Java) implementation is enough to add BROTLI to `orc-format`. C++ implementation can follow later.
> Do we need a PoC implementation on the C++ side before making a decision?

dongjoon-hyun (commented):
FYI, `orc_proto.proto` will be deleted via the following.
- https://github.com/apache/orc/pull/1684

wgtmac:
BTW, as parquet feature parity has been mentioned here, I have a couple of questions:
- Do we need to formally start the action?
- Should we also keep an eye on the lance format?
- Is it a good time to revisit the TODO items on ORC v2 (https://orc.apache.org/specification/ORCv2/)?

@dongjoon-hyun @mystic-lama

dongjoon-hyun:
Could you send an email to dev@orc mailing list, @wgtmac ?

dongjoon-hyun:
Okay, let me make a PR for you, @deshanxiao .

dongjoon-hyun:
This is the PR.
- https://github.com/apache/orc-format/pull/7

deshanxiao:
Thank you @dongjoon-hyun !

dongjoon-hyun:
Do you think you can resume this, @deshanxiao ?

deshanxiao:
> May I ask if you think you can resume this, @deshanxiao ?

Sure, this work will resume soon.

deshanxiao:
The following is a comparison of the various implementation options we can take:

### Option 1: (Parquet mode):
Use the same way with parquet to support brotli codec, which will use a private maven repository with optional and runtime dependency.

#### Pros:
1.This method has been tested in practice and has been adopted by parquet.

#### Cons:
1. aarch64 is not supported.
2. We will have to use private maven repository. The code in this repository has not been updated in many years.

### Option 2: brotli4j (Netty mode):
#### Pros:
1. brotli4j is still under maintenance and supports all platforms.
2. license check is passed.

image


#### Cons:
1.This solution was mentioned in the parquet community, but in the end, Option 1 was adopted. The specific reason is unknown.



### Option 3: google/brotli:
#### Pros:
1.Official support.

#### Cons:
1.No brotli encoder. We need to wait for official support.


At present, I prefer `Option 2`. What do you think? @dongjoon-hyun @mystic-lama @wgtmac

wgtmac:
Option 2 looks best on my side. Thanks @deshanxiao!

dongjoon-hyun:
Thank you, @deshanxiao .  I agree with you. +1 for Option 2.

deshanxiao:
Thank you @dongjoon-hyun @wgtmac . Please let me re-implement a version based on brotli4j.

Close this PR and will create a new one.

dongjoon-hyun:
Thank you for leading this area, @deshanxiao .

dongjoon-hyun:
cc @williamhyun too for the above context.

PR #1575, opened 2023-08-09
ORC-1474: replaced deprecated `getMinimum/Maximum` in TestColumnStatistics

mystic-lama:
As part of ORC-661, getMinimum/getMaximum were deprecated.
TestColumnStatistics uses these deprecated methods.
We need to replace them as
getMinimum -> getMinimumLocalDate
getMaximum -> getMaximumLocalDate

Updated deprecated methods to keep tests in sync with implementation and preventing failure when deprecated methods are removed from dependencies

The Unit test case passes

### DIFF
--- a/java/core/src/test/org/apache/orc/TestColumnStatistics.java
+++ b/java/core/src/test/org/apache/orc/TestColumnStatistics.java
@@ -19,6 +19,7 @@
 package org.apache.orc;
 
 import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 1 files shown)


java/core/src/test/org/apache/orc/TestColumnStatistics.java
```diff
@@ -28,6 +27,7 @@
 import org.apache.hadoop.hive.serde2.io.DateWritable;
 import org.apache.hadoop.hive.serde2.io.HiveDecimalWritable;
 import org.apache.hadoop.io.Text;
+import org.apache.hadoop.shaded.org.apache.commons.text.StringEscapeUtils;
```
wgtmac, line 30: Is this change required?

mystic-lama: @wgtmac StringEscapeUtils is deprecated. I am split on this change, hence left a note in the PR description above, as commons-text is not available as direct dependency in ORC, and not inclined to add it just for one class.
Open to suggestions

wgtmac: @dongjoon-hyun @williamhyun Any suggestion? I am not that familiar with that dependency.

dongjoon-hyun: `*.shaded.` is supposed to be used by the original project only. Depending on other project's shaded dependency is worse than using a deprecated dependency. I agree with @wgtmac 's concern. Let's remove this.

mystic-lama: @dongjoon-hyun Ack, let me remove this change

mystic-lama: Rolled back the change, and updated PR description and JIRA to reflect the same, to avoid confusion in future

mystic-lama: @dongjoon-hyun Thanks for the feedback. Updated the description. Does it look better?

dongjoon-hyun (approved):
+1, LGTM. Thank you, @mystic-lama and @wgtmac .

PR #1579, opened 2023-08-13
ORC-1475: [C++] ConvertColumnReader.TestConvertNumericToStringVariant fails when compiled with unsigned char

ffacs:
Fix the failure of ConvertColumnReader.TestConvertNumericToStringVariant when char is unsigned.

The UT fails when compiled with unsigned char.

The UT is passed when compiled both with and without the flag -funsigned-char.

### DIFF
--- a/c++/test/TestConvertColumnReader.cc
+++ b/c++/test/TestConvertColumnReader.cc
@@ -246,8 +246,8 @@ namespace orc {
       for (int j = 0; j < TEST_CASES; j++) {
         int flag = j % 2 == 0 ? -1 : 1;
         uint64_t idx = static_cast<uint64_t>(j);
-        col0.data[idx] = static_cast<char>(flag * (j % 128));
-        col1.data[idx] = static_cast<short>(flag * (j % 32768));
+        col0.data[idx] = static_cast<int8_t>(flag * (j % 128));
+        col1.data[idx] = static_cast<int16_t>(flag * (j % 32768));
         col2.data[idx] = flag * j;
         col3.data[idx] = flag * j;
         col4.data[idx] = static_cast<float>(flag * j) * 1.234f;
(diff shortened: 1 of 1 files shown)


wgtmac (approved):
LGTM. Thanks @ffacs for a quick fix!

mystic-lama:
+1 Thanks @ffacs

guiyanakuang (approved):
+1 LGTM

guiyanakuang:
Merged to main, Thank you @ffacs @wgtmac and @mystic-lama

dongjoon-hyun:
Please don't forget to set `Milestone` field, @guiyanakuang .

dongjoon-hyun:
BTW, @ffacs and all.
- There is not much detail in JIRA and here about the environment and how to reproduce this. It's always better to make the git commit message include the enough information to understand to be complete.
- When all GitHub Action jobs are green, we need an evidence how to reproduce this.
  - https://github.com/apache/orc/actions/runs/5822748667/job/15788083903

dongjoon-hyun:
Thank you, @ffacs , @wgtmac , @mystic-lama , @guiyanakuang . 
I know that this PR is valid.

As a release manager, I have one question because this is reported during RC0 vote like the following.

- https://lists.apache.org/thread/7k735b134j7h20ln13b0ztw7kyd2xtd2
```
bash-5.1$ ./run-all.sh local branch-1.9
Launching centos7 as 85438
...
FAILED centos7
...
```

I'm wondering how we can reproduce this failure in **Docker** testing because it was reported in the above via `./run-all.sh local branch-1.9` without any change. Apache ORC `branch-1.9` doesn't have that flag, `unsigned-char`, in our source code or `Dockerfiles`.

wgtmac:
gcc/clang uses signed-char by default on most architectures. But on ARM the default is unsigned-char. I suspect this is the difference. I cannot reproduce this on Apple M1 Pro.

dongjoon-hyun:
According to @wgtmac 's comment, https://github.com/apache/orc/pull/1582#issuecomment-1680003473.
> gcc/clang uses signed-char by default on most platforms. But on ARM the default is unsigned-char.

I reproduced it on M1 Max. Previously, I ran Docker tests on old Intel Mac while I'm running UTs in M1 Max. So I missed the difference. Thank you, @wgtmac .

dongjoon-hyun:
This is the way I reproduced.
```
$ git checkout v1.9.1-rc0
$ git clean -fdx
$ docker system prune -a
$ cd docker
$ ./reinit.sh centos7
$ git branch 0815
$ sh run-one.sh local 0815 centos7
```

guiyanakuang:
> Please don't forget to set `Milestone` field, @guiyanakuang .

Thank you for the reminder! I've noted it .

mystic-lama:
@dongjoon-hyun I also found the issue during release validation on M1

dongjoon-hyun:
To @mystic-lama , of course, ORC-1475 is based on your initial report on dev@orc. We appreciated your contribution (testing and reporting).
> @dongjoon-hyun I also found the issue during release validation on M1

What I did here is to reproduce your initial report. :)

PR #1584, opened 2023-08-15
ORC-1478: Adding Unit Tests for org.apache.orc.impl.DynamicIntArray

mystic-lama:
The PR adds Unit Test for org.apache.orc.impl.DynamicIntArray class

Unit tests helps catching issues early and acts as safety net while refactoring code

The build passes and the added test cases pass.


dongjoon-hyun (approved):

PR #1586, opened 2023-08-15
ORC-1479: Enhance examples usage message to use uber jar

mystic-lama:
The PR aims to enhance the message printed when using examples.
From examples/target directory, if we run the command it gives an error

```
bash-5.1$ java -jar orc-examples-*.jar --help
no main manifest attribute, in orc-examples-2.0.0-SNAPSHOT-sources.jar
```

If we replace the usage with following, the error goes away

```
$java -jar orc-examples-*-uber.jar [--help]
```

The change shall make it more user friendly, avoiding the error message

Manually tested. Following is the result post

example_usage_fix

### DIFF
--- a/java/examples/src/java/org/apache/orc/examples/Driver.java
+++ b/java/examples/src/java/org/apache/orc/examples/Driver.java
@@ -67,7 +67,7 @@ public class Driver {
     if (options.command == null) {
       System.err.println("ORC Java Examples");
       System.err.println();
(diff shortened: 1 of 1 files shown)


dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Merged to main for Apache ORC 2.0. Thank you, @mystic-lama .

PR #1587, opened 2023-08-16
ORC-1481: [C++] Better error message when TZDB is unavailable

wgtmac:
Check the availability of TZDB and throw helpful error message before it crashes.

When local IANA TZDB is unavailable or TZDIR env is not properly set, getTimezoneByName() simply fails without much helpful guidance on what to do.

Added a test case TestTimezone.testMissingTZDB to make sure error message is as expected.

### DIFF
--- a/c++/src/Timezone.cc
+++ b/c++/src/Timezone.cc
@@ -678,8 +678,7 @@ namespace orc {
     if (!std::filesystem::exists(std::filesystem::path(filename))) {
       std::stringstream ss;
       ss << "Time zone file " << filename << " does not exist."
-         << " Please install IANA time zone database and set TZDIR env"
-         << " if it is not installed at /usr/share/zoneinfo";
+         << " Please install IANA time zone database and set TZDIR env.";
       throw TimezoneError(ss.str());
     }
     try {
--- a/c++/src/Timezone.cc
+++ b/c++/src/Timezone.cc
@@ -678,8 +678,8 @@ namespace orc {
     if (!std::filesystem::exists(std::filesystem::path(filename))) {
       std::stringstream ss;
       ss << "Time zone file " << filename << " does not exist."
-         << " Please install IANA time zone database and set TZDIR env properly"
-         << " if not at /usr/share/zoneinfo";
+         << " Please install IANA time zone database and set TZDIR env"
+         << " if it is not installed at /usr/share/zoneinfo";
       throw TimezoneError(ss.str());
     }
     try {
--- a/c++/src/Timezone.cc
+++ b/c++/src/Timezone.cc
@@ -24,7 +24,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
-#include <iostream>
+#include <filesystem>
 #include <map>
 #include <sstream>
 
@@ -675,6 +675,13 @@ namespace orc {
     if (itr != timezoneCache.end()) {
       return *(itr->second).get();
     }
+    if (!std::filesystem::exists(std::filesystem::path(filename))) {
+      std::stringstream ss;
+      ss << "Time zone file " << filename << " does not exist."
+         << " Please install IANA time zone database and set TZDIR env properly"
+         << " if not at /usr/share/zoneinfo";
+      throw TimezoneError(ss.str());
+    }
     try {
       std::unique_ptr<InputStream> file = readFile(filename);
       size_t size = static_cast<size_t>(file->getLength());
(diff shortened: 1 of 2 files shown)


c++/src/Timezone.cc
```diff
@@ -675,6 +675,13 @@ namespace orc {
     if (itr != timezoneCache.end()) {
       return *(itr->second).get();
     }
+    if (!std::filesystem::exists(std::filesystem::path(filename))) {
+      std::stringstream ss;
+      ss << "Time zone file " << filename << " does not exist."
+         << " Please install IANA time zone database and set TZDIR env properly"
```
mystic-lama, line 681: Just a suggestion: Can rephrase as 

"Please install IANA time zone database and set TZDIR env property to point to /usr/share/zoneinfo" or something like that

wgtmac: What about this: `Please install IANA time zone database and set TZDIR env if it is not installed at /usr/share/zoneinfo`

c++/src/Timezone.cc
```diff
@@ -675,6 +675,13 @@ namespace orc {
     if (itr != timezoneCache.end()) {
       return *(itr->second).get();
     }
+    if (!std::filesystem::exists(std::filesystem::path(filename))) {
+      std::stringstream ss;
+      ss << "Time zone file " << filename << " does not exist."
+         << " Please install IANA time zone database and set TZDIR env"
+         << " if it is not installed at /usr/share/zoneinfo";
```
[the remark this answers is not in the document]
wgtmac: The default location `/usr/share/zoneinfo` does not apply to Windows. I think the error message works for Windows, too.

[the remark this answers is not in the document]
dongjoon-hyun: Yes, it's not applied to `Windows OS` because the path style is invalid.

This PR might lead a user to try to check the location of invalid path information by mapping from `/usr/share/zoneinfo` to `C:\\usr\share\zoneinfo`. Although the user ends up to conclude that the condition statement (`if it is not installed at /usr/share/zoneinfo`) is `false`, this message might confuse him by suggesting him to install it into `C:\\usr\share\zoneinfo`.

When the error message happens, I'm wondering if we don't need the message (`if it is not installed at /usr/share/zoneinfo`) at all. 

The first two lines look sufficient to me.
```c++
ss << "Time zone file " << filename << " does not exist."
   << " Please install IANA time zone database and set TZDIR env.";
```

dongjoon-hyun (approved):
+1, LGTM (except one comment, https://github.com/apache/orc/pull/1587/files#r1298940081).

c++/src/Timezone.cc
```diff
@@ -675,6 +675,13 @@ namespace orc {
     if (itr != timezoneCache.end()) {
       return *(itr->second).get();
     }
+    if (!std::filesystem::exists(std::filesystem::path(filename))) {
```
coderex2522, line 678: +1, LGTM! @wgtmac  However, the logic of determining whether a file exists can be integrated into the readLocalFile function by throwing a file not found exception. This exception can be caught in the getTimezoneByFilename function and further thrown as a Timezone error.

coderex2522 (approved):

c++/src/Timezone.cc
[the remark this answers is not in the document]
wgtmac: I agree with you, let me change it.

c++/src/Timezone.cc
wgtmac: Yes, I just want to check proactively before reading it so the error message is clear. It may fail reading due to other reasons.

wgtmac:
I've merged it. Thanks @mystic-lama @dongjoon-hyun @coderex2522!

dongjoon-hyun:
Thank you so much, @wgtmac and all.

rafal-c:
Sorry for writing in a closed thread but I think the question fits here. I downloaded the database from https://data.iana.org/time-zones/releases/tzdata2023c.tar.gz unpacked it, set the env variable TZDIR to the unpacked directory and still can't get ORC to work as it looks for a file called "UTC" but files in the database are mostly continent names like "africa", "asia", etc.
What am I missing?

scw:
> Sorry for writing in a closed thread but I think the question fits here. I downloaded the database from https://data.iana.org/time-zones/releases/tzdata2023c.tar.gz unpacked it, set the env variable TZDIR to the unpacked directory and still can't get ORC to work as it looks for a file called "UTC" but files in the database are mostly continent names like "africa", "asia", etc. What am I missing?

You still need to build the package, while that has data is is still a source distribution. For example, see how conda-forge does it here: https://github.com/conda-forge/tzdata-feedstock/blob/main/recipe/build.sh

Once you've built it, you should see the expected files that can be used for `TZDIR`

PR #1589, opened 2023-08-18
ORC-1484: Use `JIRA_ACCESS_TOKEN` in `merge_orc_pr.py`

dongjoon-hyun:

This PR aims to use `JIRA_ACCESS_TOKEN` instead of a pair of JIRA_USERNAME and JIRA_PASSWORD.

To reduce the security vulnerability for the committers, we had better use `Access Tokens` instead of user ID and password.

Personal Access Tokens are available here.
- https://issues.apache.org/jira/secure/ViewProfile.jspa

Manual.

I used this script when I merge #1586 .


dongjoon-hyun:
cc @williamhyun

dongjoon-hyun:
Since this is a infra code instead of ORC library, I'll merge this.

PR #1592, opened 2023-08-21
ORC-1488: [Java] Mark getFileMetadata as deprecated

deshanxiao:
This PR is aimed to mark `getFileMetadata `as deprecated.
This PR targets for ORC 1.9.x.
For ORC 2.0, please see #1588

1.FileMetadata has been marked as deprecated in ORC 1.6.0 [Jira](https://issues.apache.org/jira/browse/ORC-520) [PR](https://github.com/apache/orc/pull/423/files#diff-7873f5bef8fbe71c6f99481532462b30a899f6425867e561df09982aa361b9d5R27)
2.FileMetadata does not have any class to implement it, and inside ORC apache1.9, getFileMetadata always returns null.

Exist UT

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcFile.java
+++ b/java/core/src/java/org/apache/orc/OrcFile.java
@@ -358,6 +358,9 @@ public class OrcFile {
       return this;
     }
 
+    /**
+     * @deprecated Use {@link #orcTail(OrcTail)} instead.
+     */
     public FileMetadata getFileMetadata() {
       return fileMetadata;
     }


deshanxiao:
cc @mystic-lama @dongjoon-hyun

mystic-lama (approved):
+1 LGTM

dongjoon-hyun (changes requested):
Hi, @deshanxiao . We need a new JIRA issue for this because the deprecation should happen in `1.9.x` and the removal might happen in `2.0.x`.

dongjoon-hyun:
Let me repeat once more.
- `Deprecation` is not the same with `Removal` and `Depreciation` should happen before the actual removal to give the users enough preparation time .
- `Deprecation` doesn't mean an automatic `Removal`. We don't know when Apache ORC become able to remove it yet.

deshanxiao:
Yes I agree with you @dongjoon-hyun . Thank you for your reminder. I have created a new JIRA [ORC-1488](https://issues.apache.org/jira/browse/ORC-1488)

dongjoon-hyun (approved):
+1, LGTM . Thank you, @deshanxiao .

dongjoon-hyun (commented):
Oops. The PR description seems to be misleading in another way. What do you want to mean by the following?

> 2.There is not any class implementing FileMetadata

Apache ORC 1.9 is still using it here, isn't it?
https://github.com/apache/orc/blob/d3d3b60d3b5e7b161f634b3ac44627acfccd00f6/java/core/src/java/org/apache/orc/impl/ReaderImpl.java#L542

deshanxiao:
> Oops. The PR description seems to be misleading in another way. What do you want to mean by the following?
> 
> > 2.There is not any class implementing FileMetadata
> 
> Apache ORC 1.9 is still using it here, isn't it?
> 
> https://github.com/apache/orc/blob/d3d3b60d3b5e7b161f634b3ac44627acfccd00f6/java/core/src/java/org/apache/orc/impl/ReaderImpl.java#L542

`FileMetadata `does not have any class to implement it, and in ORC apache1.9, `getFileMetadata `always returns empty

dongjoon-hyun:
What do you mean by `empty`?
> FileMetadata does not have any class to implement it, and in ORC apache1.9, getFileMetadata always returns empty

If it means `NULL`, could you revise the PR description in that way? In addition, we can change the code like the following.

```java
public FileMetadata getFileMetadata() {
   return null;
}
```

deshanxiao:
@dongjoon-hyun I have another point of view, if the downstream uses this method to set filemeta, we should not reimplement it for compatibility reasons.

dongjoon-hyun:
If we are not sure, we had better not touch this until we figure out it, @deshanxiao .

dongjoon-hyun:
Could you answer my question, https://github.com/apache/orc/pull/1592#issuecomment-1687311110?

deshanxiao:
> Could you answer my question, [#1592 (comment)](https://github.com/apache/orc/pull/1592#issuecomment-1687311110)?

Yes, I mean `getFileMetadata` always returns `null` inside ORC 1.9 .

deshanxiao:
Unfortunately, hive implements `FileMetadata `interface which is not implemented in ORC, and also sets it:

Judging from its implementation and comments, our `FileMetadata`extends `ConsumerFileMetadata` because FileMetadata lacks some information. But ORC have all of information it needs. This reminds us that we may need to reconsider the implementation of FileMetadata, or provide a new interface. @dongjoon-hyun 

``` java
/** ORC file metadata. Currently contains some duplicate info due to how different parts
 * of ORC use different info. Ideally we would get rid of protobuf structs in code beyond reading,
 * or instead use protobuf structs everywhere instead of the mix of things like now.
 */
public final class OrcFileMetadata implements FileMetadata, ConsumerFileMetadata {
  private final List<StripeInformation> stripes;
  private final List<Integer> versionList;
  private final List<OrcProto.StripeStatistics> stripeStats;
  private final List<OrcProto.Type> types;
```

``` java
public interface ConsumerFileMetadata {
  int getStripeCount();
  CompressionKind getCompressionKind();
  List<Type> getTypes();
  TypeDescription getSchema() throws FileFormatException;
  OrcFile.Version getFileVersion();
  CalendarKind getCalendar();
}
```

https://github.com/apache/hive/blob/3d3acc7a19399d749a39818573a76a0dbbaf2598/llap-server/src/java/org/apache/hadoop/hive/llap/io/metadata/OrcFileMetadata.java#L39

deshanxiao:
Overall, main downstreams of Apache ORC that depend on `FileMetadata` are:

1.Apache  Hive (`OrcFileMetadata implements FileMetadata`)
2.Apache Spark (None)
3.Apache Iceberg (None)

Apache ORC v1.9 and higher do not implement the interface `FileMetadata` and the reference in [ReaderImpl.java](https://github.com/apache/orc/blob/d3d3b60d3b5e7b161f634b3ac44627acfccd00f6/java/core/src/java/org/apache/orc/impl/ReaderImpl.java#L542) will always be null.

dongjoon-hyun:
Thank you for the investigation and sharing, @deshanxiao .

PR #1594, opened 2023-08-22
ORC-1489: Assign a writer id to CUDF

guiyanakuang:
### What changes were proposed in this pull request?

This pr is aimed at assigning a writer id to the CUDF.

### Why are the changes needed?

This helps to locate the writer of a specific orc file, and it also helps the reader to do some special reads for files created by different writers.

### How was this patch tested?

Added UT

PR #1595, opened 2023-08-22
ORC-1482: Adaptation to read ORC files created by CUDF

guiyanakuang:

This pr is aimed at adapting to read ORC files created by CUDF, which may have missing statistics in their DOUBLE/FLOAT columns.

Official ORC readers can't read CUDF-created ORC files properly. 

Added UT.


guiyanakuang:
Marked as a draft for now, with additional unit tests later on

guiyanakuang:
@jlowe Can you help verify if this pr fixes the bug?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -1249,7 +1249,7 @@ public boolean[] pickRowGroups(StripeInformation stripe,
      * Get the count of exceptions for testing.
      * @return
      */
-    long[] getExceptionCount() {
+    public long[] getExceptionCount() {
```
dongjoon-hyun, line 1252: I'm curious why we need to change this. According to the function description, this is used in the test code already, isn't it?
> Get the count of exceptions for testing.

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -695,8 +695,8 @@ static TruthValue evaluatePredicateProto(OrcProto.ColumnStatistics statsProto,
                    " include ORC-517. Writer version: {}",
           predicate.getColumnName(), writerVersion);
       return TruthValue.YES_NO_NULL;
-    } else if (category == TypeDescription.Category.DOUBLE ||
-        category == TypeDescription.Category.FLOAT) {
+    } else if ((category == TypeDescription.Category.DOUBLE ||
+        category == TypeDescription.Category.FLOAT) && cs instanceof DoubleColumnStatistics) {
```
dongjoon-hyun, line 699: This doesn't match with the following goal of PR.
> This pr is aimed at adapting to read ORC files created by CUDF, which may have missing statistics in their DOUBLE/FLOAT columns.

Previously, we assume that FLOAT type also has `DoubleColumnStatistics`. Now, it breaks the assumption in case of `FLOAT` type with `FloatColumnStatistics`.

In order to match with the PR description, do we need a null check here?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
guiyanakuang: Unit test that uses this method, but the test file is not in the same package as RecordReaderImpl.java

java/core/src/test/org/apache/orc/TestColumnStatistics.java
```diff
@@ -699,6 +703,43 @@ public void testDecimalMinMaxStatistics() throws Exception {
         "Incorrect minimum value");
   }
 
+  @Test
+  public void testDoubleColumnWithoutDoubleStatistics() throws Exception {
```
dongjoon-hyun, line 707: Shall we move `testDoubleColumnWithoutDoubleStatistics` to `TestRecordReaderImpl`?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
guiyanakuang: When specific statistics are missing, cs is a `ColumnStatistics` object.

java/core/src/test/org/apache/orc/TestColumnStatistics.java
guiyanakuang: Let's refactor this pr later

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
dongjoon-hyun: Oh, got it.

jlowe (approved):
> @jlowe Can you help verify if this pr fixes the bug?

I built a new version of the ORC 1.8.4 jars with this patch, dropped them into Spark 3.4.1 and verified that I no longer get the exception when trying to do predicate pushdown on a cudf-generated ORC file that happens to have no double stats.  Thanks for the quick turnaround on this!

dongjoon-hyun (approved):
+1, LGTM if CI passes. Thank you, @guiyanakuang and @jlowe .

dongjoon-hyun:
cc @wgtmac 

Do you want to have this at your release?

wgtmac:
Yes, I think this should be useful. @dongjoon-hyun 

@guiyanakuang Please backport it to branch-1.8 once merged. Thanks!

dongjoon-hyun:
Thank you so much, @wgtmac . Then, let's target #1594 to your release too.

dongjoon-hyun:
Since #1598 exists for 1.8.5, let me switch this to 1.9.2.

PR #1596, opened 2023-08-23
ORC-1490: Fix bug in `getWriterVersion` due to unchecked `footer.hasWriter()`

guiyanakuang:
In this pull request, a conditional check for `footer.hasWriter()` is introduced before calling `footer.getWriter()` in the `getWriterVersion()` method. If the writer field is not present in the footer, the method will default to `UNKNOWN`.

Previously, the `getWriterVersion()` method did not check for the presence of the writer field in the footer. This led to a default value of 0 being used for `footer.getWriter()`, resulting in an incorrect WriterVersion being returned.

Added UT.


java/core/src/java/org/apache/orc/impl/OrcTail.java
```diff
@@ -31,6 +31,7 @@
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
+import static org.apache.orc.OrcFile.WriterImplementation.UNKNOWN;
```
wgtmac, line 34: nit: should we avoid this static import?

dongjoon-hyun (changes requested):

Although I understand the issue and intention, I'm not sure if this PR makes Apache ORC safer or more robust because I'm not sure that the AS-IS code works correctly when the writer version is `FUTURE`. There are a few conditional code based on the writer versions, aren't there?

https://github.com/apache/orc/blob/873d00c1f42753e9522cc488db59b1c979299eea/java/core/src/java/org/apache/orc/OrcFile.java#L1056-L1059


https://github.com/apache/orc/blob/84cefd50be4651c37d31fab7581a9a14786d1237/java/core/src/java/org/apache/orc/OrcFile.java#L236-L239

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4133,7 +4133,7 @@ public void testZeroByteOrcFile(Version fileFormat) throws Exception {
     assertEquals(0, reader.getStripes().size());
     assertEquals(0, reader.getStatistics().length);
     assertEquals(0, reader.getMetadataKeys().size());
-    assertEquals(OrcFile.CURRENT_WRITER, reader.getWriterVersion());
+    assertEquals(OrcFile.WriterVersion.FUTURE, reader.getWriterVersion());
```
dongjoon-hyun, line 4136: Also, this looks like a wrong behavior change to me.

java/core/src/java/org/apache/orc/impl/OrcTail.java
```diff
@@ -121,8 +121,10 @@ public OrcProto.PostScript getPostScript() {
   public OrcFile.WriterVersion getWriterVersion() {
     OrcProto.PostScript ps = fileTail.getPostscript();
     OrcProto.Footer footer = fileTail.getFooter();
-    OrcFile.WriterImplementation writer =
-        OrcFile.WriterImplementation.from(footer.getWriter());
+    OrcFile.WriterImplementation writer = OrcFile.WriterImplementation.UNKNOWN;
+    if (footer.hasWriter()) {
+      writer = OrcFile.WriterImplementation.from(footer.getWriter());
+    }
```
dongjoon-hyun, line 127: Technically, this read-path change doesn't match with our writer behavior because we throws `IllegalArgumentException` explicitly.

https://github.com/apache/orc/blob/873d00c1f42753e9522cc488db59b1c979299eea/java/core/src/java/org/apache/orc/OrcFile.java#L749-L752

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
guiyanakuang: Shouldn't the writer version of a zero byte orc file be unknown?

java/core/src/java/org/apache/orc/impl/OrcTail.java
guiyanakuang: I don't think so; A low version writer certainly can't specify that the orc written is a futrue version. But a reader of a low version should be able to read an orc file created by a higher version, even if that higher version is unknown in the low-code implementation.

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
dongjoon-hyun: No, it's defined as a valid file always. So, it should be the up-to-date version, which is `CURRENT_WRITER`.

java/core/src/java/org/apache/orc/impl/OrcTail.java
dongjoon-hyun: No, it seems that you are confused with the literal `FUTURE` version. The literal `FUTURE` version means `infinite` in Math, @guiyanakuang .

dongjoon-hyun: I believe UNSTABLE_PRE_2_0 and FUTURE are for developer testing, and not valid for the production ORC files.
https://github.com/apache/orc/blob/873d00c1f42753e9522cc488db59b1c979299eea/java/core/src/java/org/apache/orc/OrcFile.java#L71-L85

guiyanakuang:
> Thank you for making a PR, @guiyanakuang .
> 
> Although I understand the issue and intention, I'm not sure if this PR makes Apache ORC safer or more robust because I'm not sure that the AS-IS code works correctly when the writer version is `FUTURE`. There are a few conditional code based on the writer versions, aren't there?
> 
> https://github.com/apache/orc/blob/873d00c1f42753e9522cc488db59b1c979299eea/java/core/src/java/org/apache/orc/OrcFile.java#L1056-L1059
> 
> https://github.com/apache/orc/blob/84cefd50be4651c37d31fab7581a9a14786d1237/java/core/src/java/org/apache/orc/OrcFile.java#L236-L239

When the writer's version is `FUTURE`. there are actually two possibilities

1: The orc file doesn't specify a writer, so it's unknown.

2: The orc file file specifies a writer, but the version is outside the scope of the current impl (i.e., the orc file was created by a higher version of the impl).

In order to prevent loss of info (e.g., new features in statistics), we prevent the use of a lower version of an impl to merge an orc file created by a higher version.

java/core/src/java/org/apache/orc/impl/OrcTail.java
guiyanakuang: That being the case, close this PR and keep the status quo

PR #1604, opened 2023-08-30
ORC-1496: Use iterator to suggest backporting branches

dongjoon-hyun:

This PR aims to use iterator to suggest backporting branches.

Previously, only `branch-1.9` was suggested. After this PR, the script will suggest `branch-1.9` and `branch-1.8` and so on.

Manually.


dongjoon-hyun:
I tested this manually when I merged #1602 .

Since this is a dev infra script for committers only, let me merge this.

PR #1607, opened 2023-08-31
ORC-1315: [C++] Fix byte to integer conversions fail on platforms with unsigned char type

wgtmac:

This patch adds static_cast for the char type cast in RLE/ColumnReader.

For the context of [ORC-1315](https://issues.apache.org/jira/browse/ORC-1315), this patch fixes signed char to unsigned char conversions fail where char is by default unsigned.

No new tests were added.


coderex2522 (approved):
+1 LGTM.

PR #1609, opened 2023-09-04
ORC-1500: [C++] The partition field does not support English special characters

kumonlin:
The URI encoding is deleted during the initialization of OrcHdfsFile

For the context of [ORC-1500](https://issues.apache.org/jira/browse/ORC-1500), the C++ orc reader does not support the partition field containing English special characters. As a result, the SQL job fails to be executed and no execution result is displayed.

No new tests were added.

### DIFF
--- a/c++/src/OrcHdfsFile.cc
+++ b/c++/src/OrcHdfsFile.cc
@@ -115,17 +115,17 @@ namespace orc {
       }
 
       hdfs::FileHandle* file_raw = nullptr;
-      status = file_system->Open(uri.get_path(), &file_raw);
+      status = file_system->Open(uri.get_path(true), &file_raw);
       if (!status.ok()) {
-        throw ParseError("Can't open " + uri.get_path() + ". " + status.ToString());
+        throw ParseError("Can't open " + uri.get_path(true) + ". " + status.ToString());
       }
       // Wrapping file_raw into a unique pointer to guarantee deletion
       file.reset(file_raw);
 
       hdfs::StatInfo stat_info;
-      status = file_system->GetFileInfo(uri.get_path(), stat_info);
+      status = file_system->GetFileInfo(uri.get_path(true), stat_info);
       if (!status.ok()) {
-        throw ParseError("Can't stat " + uri.get_path() + ". " + status.ToString());
+        throw ParseError("Can't stat " + uri.get_path(true) + ". " + status.ToString());
       }
       totalLength = stat_info.length;
     }


wgtmac:
Thanks for submitting a PR!

BTW, could you please keep the PR template?

dongjoon-hyun (commented):
Thank you for making a PR, @kumonlin .

I have the same comment with @wgtmac . In addition, do you think you can make a unit test case for this `encoded_output` change? 

https://github.com/apache/hadoop/blame/a6c2526c6c7ccfc4483f58695aedf0cb892b4c4d/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/include/hdfspp/uri.h#L69

```cpp
  std::string get_path(bool encoded_output=false) const;
```

wgtmac:
Actually I am a little bit confused with the title. Do you mean the hive partition which is a directory in the HDFS file path? In terms of the ORC reader, it does not know any partition field.

kumonlin:
If the partition field is set in the Hive data set, the data content stored in the partition field is an independent path in the HDFS. Therefore, if the partition field contains special characters, The orc hdfs reader invokes libhdfspp to parse the URI. However, during the parsing, encoding causes the URI to change.

kumonlin:
> Thanks for submitting a PR!
> 
> BTW, could you please keep the PR template?

done

kumonlin:
> 

The URI is the open-source libhdfspp interface of the Hadoop community, and the corresponding unit test is available. Therefore, you do not need to add the URI unit test to the ORC.

kumonlin:
> Thank you for making a PR, @kumonlin .
> 
> I have the same comment with @wgtmac . In addition, do you think you can make a unit test case for this `encoded_output` change?
> 
> https://github.com/apache/hadoop/blame/a6c2526c6c7ccfc4483f58695aedf0cb892b4c4d/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/include/hdfspp/uri.h#L69
> 
> ```c++
>   std::string get_path(bool encoded_output=false) const;
> ```
Thinks for your suggest.
The URI is the open-source libhdfspp interface of the Hadoop community, and the corresponding unit test is available. Therefore, you do not need to add the URI unit test to the ORC.

dongjoon-hyun:
No, we need a test coverage for your new code in this PR, not some existing code in the other repository, @kumonlin .
> The URI is the open-source libhdfspp interface of the Hadoop community, and the corresponding unit test is available. Therefore, you do not need to add the URI unit test to the ORC.

dongjoon-hyun (changes requested):
If you cannot add a unit test, could you describe how did you get those Hive tables which causes this problem? Then, at least, we can verify it manually.

wgtmac:
It seems that we don't have any test coverage for HDFS. Could you add one to call `readHdfsFile` by reading a local file and verify the issue is fixed?

dongjoon-hyun:
Gentle ping, @kumonlin . Can I get some reproducible procedures ?
> If you cannot add a unit test, could you describe how did you get those Hive tables which causes this problem? Then, at least, we can verify it manually.

wgtmac:
I have tried to add a simple unit test based on local filesystem by mocking a `core-site.xml` file and setting `fs.defaultFs` to `file:///`. However, it seems that the C++ hdfs client couldn't accept that and complained `Error connecting to the cluster: defaultFS is empty. defaultFS of [file:///] is not supported`. Instead, mocking a HDFS mini cluster to test this would be an overkill.

I checked the code and the `url.get_path(true)` has been tested here: https://github.com/apache/hadoop/blob/a6c2526c6c7ccfc4483f58695aedf0cb892b4c4d/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/tests/uri_test.cc#L180-L189.

dongjoon-hyun:
Thank you for checking. Do you want to merge this, @wgtmac ?

dongjoon-hyun:
Feel free to merge if you want.

wgtmac:
Thanks @dongjoon-hyun and @kumonlin! I've merged it.

PR #1619, opened 2023-09-20
ORC-1506: Replacing deprecated valueOf() with recommended forNumber()

mystic-lama:
The PR aims at replacing deprecated valueOf() with recommended forNumber().
valueOf() has been deprecated for a while and shall be removed in future versions. We are only removing it from the code we have written, it still exists in auto-generated code. Autogenerate code shall be compliant with version upgrade in future.

Couldn't find the exact PR for the change, however the reason seems to be performance related from search about the change

The change helps protect against the removal of deprecated API in future releases.

No new tests were added, current tests pass

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/WriterImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/WriterImpl.java
@@ -612,7 +612,7 @@ public class WriterImpl implements WriterInternal, MemoryManager.Callback {
     HadoopShims.KeyMetadata meta = key.getMetadata();
     result.setKeyName(meta.getKeyName());
     result.setKeyVersion(meta.getVersion());
-    result.setAlgorithm(OrcProto.EncryptionAlgorithm.valueOf(
+    result.setAlgorithm(OrcProto.EncryptionAlgorithm.forNumber(
         meta.getAlgorithm().getSerialization()));
     return result;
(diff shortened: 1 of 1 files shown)


dongjoon-hyun (commented):
Is this PR fixing all instances of `ProtoBuffer valueOf usage?

mystic-lama:
@dongjoon-hyun Let me check once again. I found it as I was reading code and the inspection didn't yield anything else. I shall validate once again and update code if needed.

dongjoon-hyun:
Thank you, @mystic-lama .

mystic-lama:
@dongjoon-hyun I check the code and didn't find any other usage. Auto-generate code does use the valueOf(), however not planing on changing it here. Not sure what all it may break, if someone is using it outside of this repo. The valueOf() is internally calling forNumber().

```java
/**
  * @param value The numeric wire value of the corresponding enum entry.
  * @return The enum associated with the given numeric wire value.
  * @deprecated Use {@link #forNumber(int)} instead.
  */
  @java.lang.Deprecated
  public static EncryptionAlgorithm valueOf(int value) {
    return forNumber(value);
  }
```

Let me know if I need to investigate more.

dongjoon-hyun (approved):
+1, LGTM.

PR #1622, opened 2023-09-23
ORC-1510: Fixes package for TestOrcUtils package and adds more test cases

mystic-lama:
The PR proposes following changes
* update package for core/src/test/org/apache/orc/util/TestOrcUtils.java -> core/src/test/org/apache/orc/TestOrcUtils.java to match the Source file package
* Adds more test cases

* Test class package should match the file under test package to avoid package imports and as best practice. Mismatch in package was also creating confusion while browsing the code
* Additional test cases were added to have more test coverage

Added unit test cases pass. Does not have any effect on source, as changes are limited to test classes


dongjoon-hyun (approved):

PR #1629, opened 2023-09-27
ORC-1387: [C++] Support schema evolution from decimal to numeric/decimal

ffacs:
Support schema evolution from decimal to {boolean, byte, short, int, long, float, double, decimal}

to support schema evolution

UT passed


c++/src/Int128.cc
```diff
@@ -435,6 +435,13 @@ namespace orc {
     return buf.str();
   }
 
+  double Int128::toDouble() const {
+    if (fitsInLong()) {
+      return static_cast<double>(toLong());
+    }
+    return static_cast<double>(lowbits) + std::ldexp(static_cast<double>(highbits), 64);
```
wgtmac, line 442: Should we do some error handling? What if the double value does not precisely store the original value?

c++/src/ConvertColumnReader.cc
```diff
@@ -466,6 +466,136 @@ namespace orc {
...
+        }
+      }
+    }
+
+   private:
+    void convertDecimalToInteger(ReadTypeBatch& dstBatch, uint64_t idx,
+                                 const FileTypeBatch& srcBatch) {
+      using FileType = decltype(srcBatch.values[idx]);
+      Int128 result = scaleDownInt128ByPowerOfTen(srcBatch.values[idx], scale);
+      int64_t longValue = result.toLong();
```
wgtmac, line 504: This would be better to move to below line 508.

c++/src/ConvertColumnReader.cc
```diff
@@ -466,6 +466,136 @@ namespace orc {
...
+      }
+      convertNumericElement<ReadType, int64_t>(longValue, dstBatch.data[idx], dstBatch, idx,
+                                               throwOnOverflow);
+    }
+
+    void convertDecimalToDouble(ReadTypeBatch& dstBatch, uint64_t idx,
+                                const FileTypeBatch& srcBatch) {
+      double doubleValue = 0;
+      Int128 i128 = srcBatch.values[idx];
+      doubleValue = i128.toDouble();
```
wgtmac, line 517:
```suggestion
      double doubleValue = srcBatch.values[idx].toDouble();
```

wgtmac (commented):
Thanks! Mostly LGTM. I've left some comments.

c++/src/Int128.cc
ffacs: > 
I didn't check the precision because I want Int128::toDouble() to act like 
static_csat<double>(__int128_t). Do you think is that OK?

wgtmac: Does the java code do the same thing?

ffacs: > Does the java code do the same thing?

The java side uses BigDecimal.doubleValue() to convert a decimal to a double.This operation would lose information, but it seems that the java ORC didn't do any checks too. 
```java
    if(intCompact != -9223372036854775808) {
            if (scale == 0) {
                return (double)intCompact;
            } else {
                /*
                 * If both intCompact and the scale can be exactly
                 * represented as double values, perform a single
                 * double multiply or divide to compute the (properly
                 * rounded) result.
                 */
                if (Math.abs(intCompact) < 1L<<52 ) {
                    // Don't have too guard against
                    // Math.abs(MIN_VALUE) because of outer check
                    // against INFLATED.
                    if (scale > 0 && scale < double10pow.length) {
                        return (double)intCompact / double10pow[scale];
                    } else if (scale < 0 && scale > -double10pow.length) {
                        return (double)intCompact * double10pow[-scale];
                    }
                }
            }
        }
        // Somewhat inefficient, but guaranteed to work.
        return Double.parseDouble(this.toString());
```

wgtmac: It is OK to lose precision as IEEE754 is unable to store all integers as is.

wgtmac (approved):

dongjoon-hyun (approved):
+1, LGTM.

PR #1636, opened 2023-10-15
ORC-1519: Removed LineLength and star imports rules from checkstyle suppressions and updated code to remove warnings

mystic-lama:
The PR aims at removing checkstyle suppressions for LineLength and AvoidStarImport rules.
The LineLength max property has been updated to 120 from 100 (checked the value in Apache Hive and Apache Arrow, both use 120)
A suppression for src/test/org/apache/orc/tools/TestFileDump.java has been added. It has 2 places with long json content and it was not feasible to break it without affecting the readability.

The changes are need to keep code uniform for future commits

Changes are limited to Unit Test code, and all test cases pass


dongjoon-hyun (changes requested):
Sorry but I'm not sure that we need this, @mystic-lama .

mystic-lama:
Closing the PR, as it's not needed

PR #1627, opened 2023-09-26
ORC-1512: Drop Java 8/11 and make Java 17 by default

dongjoon-hyun:
### What changes were proposed in this pull request?

This PR aims to drop Java 8 and 11 support and make Java 17 as the default and minimum Java version for Apache ORC 2.0.0.

This PR also upgrade GitHub Action test pipelines and Dockefiles.
- Update GitHub Action Java Versions to Java 17 and 21.
- Change GitHub Action Java Distribution to `Azul Zulu` from `Eclipse Temurin` because `temurin` doesn't have Java 21 yet.
- Update Dockerfiles to use Java 17 and 21
- Delete CentOS7 and Debian10 which requires non-trivial Java installations.

### Why are the changes needed?

Apache ORC 2.0 is aligned with Apache Spark 4.0 which dropped Java 8 and 11.
- https://issues.apache.org/jira/browse/SPARK-45315 (Drop JDK 8/11 and make JDK 17 by default)

In general, there are some extended support period for Java 11. However, Oracle's Premier Support ends this month, `September 2023`.
- https://www.oracle.com/java/technologies/java-se-support-roadmap.html

### How was this patch tested?

Pass the CIs

PR #1639, opened 2023-10-17
ORC-1520: Removed JDK 8 settings from pom

mystic-lama:
The PR proposes to remove jdk8 related parameters from pom, since we moved to JDK 17 as part of #1627 

Since we moved to JDK 17, JDK 8 parameters are not relevant any more.
This shall keep files up to date with changes

Build runs successfully with the changes.


dongjoon-hyun (approved):

PR #1662, opened 2023-11-24
ORC-1528: Fix readBytes potential overflow in RecordReaderUtils.ChunkReader#create

yebukong:

- I have adjusted the calculation of `reqBytes` and `readBytes` ranges in the `org.apache.orc.impl.RecordReaderUtils.ChunkReader#create` method from `Integer` to `Long`, and added validation for `Integer` overflow

- Adds more test cases

This PR aims to fix [ORC-1528](https://issues.apache.org/jira/browse/ORC-1528) 

org.apache.orc.impl.TestRecordReaderUtils#testBufferChunkOffsetExceedsMaxInt

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/IOUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/IOUtils.java
@@ -31,6 +31,11 @@ public final class IOUtils {
   public static final int DEFAULT_BUFFER_SIZE = 8192;
 
   /**
+   * The maximum size of array to allocate, value being the same as {@link java.util.Hashtable}
+   */
+  public static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
+
+  /**
    * Returns a new byte array of size {@link #DEFAULT_BUFFER_SIZE}.
    *
    * @return a new byte array of size {@link #DEFAULT_BUFFER_SIZE}.
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
@@ -792,11 +792,11 @@ public class RecordReaderUtils {
         current = (BufferChunk) current.next;
       }
       reqBytes += ef - cf;
-      if (reqBytes >= IOUtils.MAX_ARRAY_SIZE) {
+      if (reqBytes > IOUtils.MAX_ARRAY_SIZE) {
         throw new IllegalArgumentException("invalid reqBytes value " + reqBytes + ",out of bounds " + IOUtils.MAX_ARRAY_SIZE);
       }
       long readBytes = e - f;
-      if (readBytes >= IOUtils.MAX_ARRAY_SIZE) {
+      if (readBytes > IOUtils.MAX_ARRAY_SIZE) {
         throw new IllegalArgumentException("invalid readBytes value " + readBytes + ",out of bounds " + IOUtils.MAX_ARRAY_SIZE);
       }
       return new ChunkReader(from, to, (int) readBytes, (int) reqBytes);
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
@@ -770,32 +770,32 @@ public class RecordReaderUtils {
     }
 
     static ChunkReader create(BufferChunk from, BufferChunk to) {
-      long f = Long.MAX_VALUE;
-      long e = Long.MIN_VALUE;
+      long start = Long.MAX_VALUE;
+      long end = Long.MIN_VALUE;
 
-      long cf = Long.MAX_VALUE;
-      long ef = Long.MIN_VALUE;
+      long currentStart = Long.MAX_VALUE;
+      long currentEnd = Long.MIN_VALUE;
       long reqBytes = 0L;
 
       BufferChunk current = from;
       while (current != to.next) {
-        f = Math.min(f, current.getOffset());
-        e = Math.max(e, current.getEnd());
-        if (ef == Long.MIN_VALUE || current.getOffset() <= ef) {
-          cf = Math.min(cf, current.getOffset());
-          ef = Math.max(ef, current.getEnd());
+        start = Math.min(start, current.getOffset());
+        end = Math.max(end, current.getEnd());
+        if (currentEnd == Long.MIN_VALUE || current.getOffset() <= currentEnd) {
+          currentStart = Math.min(currentStart, current.getOffset());
+          currentEnd = Math.max(currentEnd, current.getEnd());
         } else {
-          reqBytes += ef - cf;
-          cf = current.getOffset();
-          ef = current.getEnd();
+          reqBytes += currentEnd - currentStart;
+          currentStart = current.getOffset();
+          currentEnd = current.getEnd();
         }
         current = (BufferChunk) current.next;
       }
-      reqBytes += ef - cf;
+      reqBytes += currentEnd - currentStart;
       if (reqBytes >= Integer.MAX_VALUE) {
         throw new IllegalArgumentException("invalid reqBytes value " + reqBytes + ",out of bounds " + Integer.MAX_VALUE);
       }
-      long readBytes = e - f;
+      long readBytes = end - start;
       if (readBytes >= Integer.MAX_VALUE) {
         throw new IllegalArgumentException("invalid readBytes value " + readBytes + ",out of bounds " + Integer.MAX_VALUE);
       }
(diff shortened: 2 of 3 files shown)


guiyanakuang:
Thank you for making a PR, @yebukong

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -770,18 +770,18 @@ void readRanges(FSDataInputStream file, boolean allocateDirect, double extraByte
     }
 
     static ChunkReader create(BufferChunk from, BufferChunk to) {
-      long f = Integer.MAX_VALUE;
-      long e = Integer.MIN_VALUE;
+      long f = Long.MAX_VALUE;
```
mystic-lama, line 773: Recommendation: This is already existing so please feel free to ignore. 
Since we are touching the code, Can we rename the variables for easy reading?
Couldn't understand the meaning of  f, e, cf, ce 😢

yebukong: Great, let me try.

java/core/src/test/org/apache/orc/impl/TestRecordReaderUtils.java
```diff
@@ -177,6 +178,37 @@ public void execute() throws Throwable {
     assertTrue(dis.isAllReleased());
   }
 
+  @Test
+  public void testBufferChunkOffsetExceedsMaxInt() {
+    List<long[]> mockData = Arrays.asList(
+            new long[]{15032282586L, 15032298848L}
+            , new long[]{15032298848L, 15032299844L}
```
dongjoon-hyun, line 185: nit. Could you put `,` at the end of the previous line instead of the first of line?

java/core/src/test/org/apache/orc/impl/TestRecordReaderUtils.java
```diff
@@ -177,6 +178,37 @@ public void execute() throws Throwable {
...
+            , new long[]{15058484422L, 15058484862L}
+            , new long[]{15058484862L, 15058484878L}
+    );
+    TestOrcLargeStripe.RangeBuilder rangeBuilder = new TestOrcLargeStripe.RangeBuilder();
+    mockData.forEach(e -> rangeBuilder.range(e[0], (int) (e[1] - e[0])));
+    BufferChunkList rangeList = rangeBuilder.build();
+
+    RecordReaderUtils.ChunkReader chunkReader =
+            RecordReaderUtils.ChunkReader.create(rangeList.get(),
+                    134217728);
```
dongjoon-hyun, line 205: It seems that we don't need to split lint for this. Shall we change like the following?
```java
-            RecordReaderUtils.ChunkReader.create(rangeList.get(),
-                    134217728);
+            RecordReaderUtils.ChunkReader.create(rangeList.get(), 134217728);
```

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
dongjoon-hyun: Actually, I'm opposite to @mystic-lama 's comment.

We don't want to mix a bug fix and a style fix (or refactorying) in the same PR. It's because it increases the complexity indeed.

If we want to rename this, it should be done before or after this PR.

Please don't change the existing variable name in this PR.

dongjoon-hyun (changes requested):

dongjoon-hyun:
Thank you for reporting a bug and making a PR, @yebukong .
Please make a renaming PR separately. I can merge that renaming PR first before this main PR.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -770,29 +770,36 @@ void readRanges(FSDataInputStream file, boolean allocateDirect, double extraByte
...
+          reqBytes += currentEnd - currentStart;
+          currentStart = current.getOffset();
+          currentEnd = current.getEnd();
         }
         current = (BufferChunk) current.next;
       }
-      reqBytes += ef - cf;
-      return new ChunkReader(from, to, (int) (e - f), reqBytes);
+      reqBytes += currentEnd - currentStart;
+      if (reqBytes >= Integer.MAX_VALUE) {
```
guiyanakuang, line 795: https://github.com/apache/orc/blob/22a06b4e7575f539be970ec9658f5b597ef7c83d/java/core/src/java/org/apache/orc/impl/StringHashTableDictionary.java#L61-L65
I think it might be more accurate to extract and reuse this `MAX_ARRAY_SIZE` constant, the java maximum array length is not actually `Integer.MAX_VALUE`.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
mystic-lama: Ack, @dongjoon-hyun has a point. @yebukong we can do that in follow up PR or may be leave it as is :)
Apologies for the noise.

yebukong:
@dongjoon-hyun I've already maked a renaming PR, #1664 , PTAL

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
yebukong: Should I define a `MAX_ARRAY_SIZE` constant in `RecordReaderUtils` ? I noticed it is marked as `private` in `StringHashTableDictionary`.

guiyanakuang: I tend to extract it into a utility class (e.g. IOUtils) and reuse the same constant.

dongjoon-hyun:
> I've already maked a renaming PR, #1664. @dongjoon-hyun PTAL. After the completion of renaming PR, i will push the changes to the main PR.

To @yebukong , I repled on your new PR about why we cannot proceed that renaming PR first.
- https://github.com/apache/orc/pull/1664#issuecomment-1826990271

dongjoon-hyun:
In short, 
- ORC-1528 should land at `main/branch-1.9/branch-1.8` branches as a bug fix.
- ORC-1530 should land at only `main` branch as an improvement or task PR

yebukong:
I have pushed the latest code modifications. @dongjoon-hyun @guiyanakuang  @mystic-lama PTAL.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -792,7 +792,14 @@ static ChunkReader create(BufferChunk from, BufferChunk to) {
         current = (BufferChunk) current.next;
       }
       reqBytes += ef - cf;
-      return new ChunkReader(from, to, (int) (e - f), reqBytes);
+      if (reqBytes >= IOUtils.MAX_ARRAY_SIZE) {
+        throw new IllegalArgumentException("invalid reqBytes value " + reqBytes + ",out of bounds " + IOUtils.MAX_ARRAY_SIZE);
```
guiyanakuang, line 796: UT don't have a use case for throwing an IllegalArgumentException, can you add it?

yebukong: Thank you for the reminder. I will refine it later

yebukong: I have added new UT related to readBytes and reqBytes validation.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
yebukong: I agree "the same concept should be kept in a single source".
         
Should we include the changes to remove the `MAX_ARRAY_SIZE` definition from `StringHashTableDictionary.java` in this PR?

java/core/src/java/org/apache/orc/impl/IOUtils.java
```diff
@@ -30,6 +30,11 @@ public final class IOUtils {
 
   public static final int DEFAULT_BUFFER_SIZE = 8192;
 
+  /**
+   * The maximum size of array to allocate, value being the same as {@link java.util.Hashtable}
```
mystic-lama, line 34: Can we add info like max array to allocate for what purpose? Please only add if the specific context is known, else we leave it as is.

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Merged to main/1.9/1.8.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
yebukong: Considering that this PR has been closed, is it still necessary to continue the discussion on the issues mentioned above?

java/core/src/java/org/apache/orc/impl/IOUtils.java
yebukong: https://github.com/openjdk/jdk/blob/jdk-17%2B35/src/java.base/share/classes/java/util/Hashtable.java#L391-L397

Because the comment for `java.util.Hashtable#MAX_ARRAY_SIZE` is detailed, I did not add excessive description.

yebukong:
You're welcome, it's my pleasure.

PR #1684, opened 2023-12-06
ORC-1546: [C++] Use `orc-format` 1.0.0-alpha and remove `proto` module

dongjoon-hyun:

This PR aims to use `orc-format` 1.0.0-alpha and remove `proto` module completely.
- https://github.com/apache/orc-format/releases/tag/v1.0.0-alpha

To verity C++ module to use `orc-format` successfully.

Pass the CIs.


dongjoon-hyun:
cc @williamhyun , @wgtmac , @guiyanakuang, @mystic-lama

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+set(ORC_FORMAT_VERSION "1.0.0-alpha")
```
dongjoon-hyun, line 18: This should be `1.0.0` soon.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -68,6 +69,17 @@ if (DEFINED ENV{GTEST_HOME})
   set (GTEST_HOME "$ENV{GTEST_HOME}")
 endif ()
 
+# ----------------------------------------------------------------------
+# ORC Format
+ExternalProject_Add (orc-format_ep
+  URL "https://github.com/apache/orc-format/archive/refs/tags/v${ORC_FORMAT_VERSION}.tar.gz"
```
dongjoon-hyun, line 75: This will be ASF Download site after `orc-format 1.0.0` release.

dongjoon-hyun:
Let me merge this first to unblock the migration.

dongjoon-hyun:
FYI, Apache Spark 4.0.0 integration test is running now.
- https://github.com/dongjoon-hyun/spark/actions/runs/7120736334

cmake_modules/ThirdpartyToolchain.cmake
wgtmac: How do we use an unreleased proto file? This may happen when we want to implement a feature of new format definition. Otherwise we need to hack somewhere to do PoC implementation and only check in the code after the format is released.

dongjoon-hyun: To @wgtmac , it's a C++ question, right?
- For Java, SNAPSHOT is available.
- For C++, we can create a tag like the above.

PR #1688, opened 2023-12-09
ORC-1551: Use `orc-format` `1.0.0-beta`

dongjoon-hyun:

This PR aims to use `orc-format` `1.0.0-beta`.

`1.0.0-beta` has the following changes.
- https://github.com/apache/orc-format/pull/5
- https://github.com/apache/orc-format/pull/4
- https://github.com/apache/orc-format/pull/7
- https://github.com/apache/orc-format/pull/9

Pass the CIs.


dongjoon-hyun:
cc @williamhyun , @wgtmac , @guiyanakuang, @stiga-huang , @coderex2522 , @deshanxiao

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
```diff
@@ -4157,9 +4157,9 @@ public void testFutureOrcFile(Version fileFormat) throws Exception {
       String m = e.getMessage();
       assertTrue(m.contains("version1999.orc was written by a future ORC version 19.99."));
       assertTrue(m.contains("This file is not readable by this version of ORC."));
-      assertTrue(m.contains("Postscript: footerLength: 19 compression: NONE " +
-          "compressionBlockSize: 65536 version: 19 version: 99 metadataLength: 0 " +
-          "writerVersion: 1"));
+      assertTrue(m.contains("Postscript: footer_length: 19 compression: NONE " +
```
wgtmac, line 4160: I assume this file should not be affected.

wgtmac (approved):
Thanks @dongjoon-hyun!

dongjoon-hyun:
Thank you, @wgtmac and @guiyanakuang .

To start and maintain `Apache ORC Format 1.0.0` in the official ProtocolBuffer way, this is the minimum required change due to the following.
- https://github.com/apache/orc-format/pull/5

As we know, `ProtocolBuffers` is a safe protocol for field renaming in terms of binary compatibility. So, Apache ORC 1.x libraries (Java/C++) can read and write `ORC Format v1.0` files without any problems. However, the generated C++ code and dump meta info is different like we see in this PR.

The new C++ API and dump meta data follows an official ProtocolBuffers recommendation. So, I believe we can help and persuade the down-streams like Apache Arrow easily without many blockers.

java/core/src/test/org/apache/orc/TestVectorOrcFile.java
dongjoon-hyun: Ya, the error message contains the protobuf field names.

dongjoon-hyun:
Let me merge this~

PR #1691, opened 2023-12-12
Add a copyright messaging to the BpackingAvx512.hh

guangzegu:
###  What changes were proposed in this pull request?
 Add copyright messaging to BpackingAvx512.hh

The vector unpacking functions  in this PR[ https://github.com/apache/orc/pull/1375]() is derived from Intel's QPL library,  hence,  Maybe copyright messaging should be added.

QPL Link: [https://github.com/intel/qpl]()

N/A

### DIFF
--- a/c++/src/BpackingAvx512.hh
+++ b/c++/src/BpackingAvx512.hh
@@ -16,6 +16,46 @@
  * limitations under the License.
  */
 
+/*******************************************************************************
+ * Copyright (C) 2022 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ ******************************************************************************/
+
+/**
+ * @brief Contains implementation of functions for unpacking 1..32-bit data to bytes
+ *
+ * @details Function list:
+ *          - @ref vectorUnpack1
+ *          - @ref vectorUnpack2
+ *          - @ref vectorUnpack3
+ *          - @ref vectorUnpack4
+ *          - @ref vectorUnpack5
+ *          - @ref vectorUnpack6
+ *          - @ref vectorUnpack7
+ *          - @ref vectorUnpack9
+ *          - @ref vectorUnpack10
+ *          - @ref vectorUnpack11
+ *          - @ref vectorUnpack12
+ *          - @ref vectorUnpack13
+ *          - @ref vectorUnpack14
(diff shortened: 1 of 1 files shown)


c++/src/BpackingAvx512.hh
```diff
@@ -16,6 +16,46 @@
  * limitations under the License.
  */
 
+/*******************************************************************************
+ * Copyright (C) 2022 Intel Corporation
```
dongjoon-hyun, line 20: Could you give us the original source link?

wgtmac:
If we need to regard the code as third-party and remove ASF license header, some refactoring work is required to isolate code from Intel QPL and then add the copyright. It seems that `BpackingAvx512.hh` is mix of ORC RLE decoder interface and AVX512 implementation.

CalvinKirs:
> If we need to regard the code as third-party and remove ASF license header, some refactoring work is required to isolate code from Intel QPL and then add the copyright. It seems that `BpackingAvx512.hh` is mix of ORC RLE decoder interface and AVX512 implementation.↳
It would be even better if you could provide the original source of the code.
 I haven't checked if the referenced code has been modified on top of it, but if so, we can also add the corresponding AL2 header. FYI https://issues.apache.org/jira/browse/LEGAL-626

wpleonardo:
> > If we need to regard the code as third-party and remove ASF license header, some refactoring work is required to isolate code from Intel QPL and then add the copyright. It seems that `BpackingAvx512.hh` is mix of ORC RLE decoder interface and AVX512 implementation.↳
> 
> It would be even better if you could provide the original source of the code. I haven't checked if the referenced code has been modified on top of it, but if so, we can also add the corresponding AL2 header. FYI https://issues.apache.org/jira/browse/LEGAL-626

The source code link: https://github.com/intel/qpl/blob/c2ced94c53c1ee22191201a59878e9280bc9b9b8/sources/core-sw/src/filtering/opt/qplc_unpack_be_8u_k0.h#L158
This implementation of this part is referred to Intel/qpl code.
@wgtmac @dongjoon-hyun So sorry for late response. Something happened in my family in the past half a year.

wgtmac:
@wpleonardo Sorry to hear that

PR #1692, opened 2023-12-21
ORC-1553: Row grouping reads should be skipped when the statistics are written without any values for the SArg column

guiyanakuang:
This PR aims to fix an issue where the column statistics were incorrectly evaluated in scenarios where no values were written, resulting in the inability to skip row groups.

The fix improves the evaluation logic of statistics, enabling the skipping of row groups that don't need to be read, thus enhancing performance.

Unit tests have been added to validate the changes.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
@@ -667,6 +667,14 @@ public class RecordReaderImpl implements RecordReader {
                                            TypeDescription type,
                                            boolean writerUsedProlepticGregorian,
                                            boolean useUTCTimestamp) {
+
+    // When statsProto is EMPTY_COLUMN_STATISTICS
+    // this column does not actually provide statistics
+    // we cannot make any assumptions, so we return YES_NO_NULL.
+    if (statsProto == EMPTY_COLUMN_STATISTICS) {
+      return TruthValue.YES_NO_NULL;
+    }
+
     ColumnStatistics cs = ColumnStatisticsImpl.deserialize(
         null, statsProto, writerUsedProlepticGregorian, true);
     ValueRange range = getValueRange(cs, predicate, useUTCTimestamp);
@@ -702,11 +710,6 @@ public class RecordReaderImpl implements RecordReader {
                 predicate.getColumnName());
         return dstas.hasNull() ? TruthValue.YES_NO_NULL : TruthValue.YES_NO;
       }
-    } else if (statsProto == EMPTY_COLUMN_STATISTICS) {
-      // When statsProto is EMPTY_COLUMN_STATISTICS
-      // this column does not actually provide statistics
-      // we cannot make any assumptions, so we return YES_NO_NULL.
-      return TruthValue.YES_NO_NULL;
     }
     return evaluatePredicateRange(predicate, range,
         BloomFilterIO.deserialize(kind, encoding, writerVersion, type.getCategory(),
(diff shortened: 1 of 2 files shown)


guiyanakuang:
@neopaf Can you test if this pr works for your data?

guiyanakuang:
cc @dongjoon-hyun @wgtmac

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -702,6 +702,11 @@ static TruthValue evaluatePredicateProto(OrcProto.ColumnStatistics statsProto,
                 predicate.getColumnName());
         return dstas.hasNull() ? TruthValue.YES_NO_NULL : TruthValue.YES_NO;
       }
+    } else if (statsProto == EMPTY_COLUMN_STATISTICS) {
```
wgtmac, line 705: Should this be put ahead of line 674?

java/core/src/java/org/apache/orc/impl/RecordReaderImpl.java
```diff
@@ -747,8 +752,10 @@ static TruthValue evaluatePredicateRange(PredicateLeaf predicate,
                                            ValueRange range,
                                            BloomFilter bloomFilter,
                                            boolean useUTCTimestamp) {
+    // If range is invalid, that means that no value (including null) is written to this column
+    // we should return TruthValue.NO for any predicate.
     if (!range.isValid()) {
-      return TruthValue.YES_NO_NULL;
+      return TruthValue.NO;
```
wgtmac, line 758: I'm not sure if this is safe. IMHO, we'd better to be conservative by returning all data if we're uncertain about any filter.

guiyanakuang: @wgtmac , Thanks for the review, but I think the change here is the key to fixing the bug.
The schema in the unit test is `struct<x:struct<z:double>,y:int>`
I've constructed 1024 rows of data, and `x` is null. This means that the statistics for `x.z` are kept in the initial state `numberOfValues = 0 hasNull = false`. 

Any filtering of `x.z` at this point should return `TruthValue.NO`

Here's the enumeration operator: for data that doesn't exist (which distinguishes it from having a value and being Null) I think it's valid to return `TruthValue.NO`. If there is a counterexample, I'd be happy to modify it.

```java
public interface PredicateLeaf {

  /**
   * The possible operators for predicates. To get the opposites, construct
   * an expression with a not operator.
   */
  public static enum Operator {
    EQUALS,
    NULL_SAFE_EQUALS,
    LESS_THAN,
    LESS_THAN_EQUALS,
    IN,
    BETWEEN,
    IS_NULL
  }
  ....
}
```

neopaf: I will check if it helped in my situation.

But am also concerned about this particular line.

As I understand https://issues.apache.org/jira/browse/ORC-1075, this line was **correct**.

guiyanakuang: This pr uses the following judgment instead of ORC-1075, which is more accurate.
```java
if (statsProto == EMPTY_COLUMN_STATISTICS) {
}
```

neopaf: Right!

dongjoon-hyun (approved):
+1. This PR looks fine to me.

dongjoon-hyun:
Do you still have concerns, @wgtmac ?

dongjoon-hyun:
Let me merge this to be considered as a part of Apache ORC 1.9.x and 2.0.0. We can revert this if there is any issues during the release cycles~

PR #1714, opened 2023-12-28
ORC-1463: [Java] Support Brotli codec

### LINKED ISSUE #1563: ORC-1463: Support brotli compression codec

deshanxiao:
This PR is aimed to add Brotli codec defined by [RFC7932](https://www.ietf.org/rfc/rfc7932.txt).
This is a new PR implemented based on [Brotli4j](https://github.com/hyperxpro/Brotli4j). Previous PR: https://github.com/apache/orc/pull/1565

To support more codec in ORC. 

UT and CI pipeline.

Test brotli in example `CompressionWriter`. It will create a 10.2kb file while snappy produces a 44.2kb file.


java/core/src/java/org/apache/orc/impl/BrotliCodec.java
```diff
@@ -0,0 +1,215 @@
...
+    return DEFAULT_OPTIONS;
+  }
+
+  @Override
+  public boolean compress(
+      ByteBuffer in,
+      ByteBuffer out,
+      ByteBuffer overflow,
+      Options options) throws IOException {
+    BrotliOptions brotliOptions = (BrotliOptions) options;
```
deshanxiao, line 115: After a lot of testing I found out there is a problem that if the input data does not start from 0, this will cause the decompression to fail. So here is an extra copy when dealing with this situation.

deshanxiao: Fixed by  brotli4j v1.15.0

williamhyun (approved):
+1 LGTM, Pending CI.

Thank you, @deshanxiao

dongjoon-hyun (approved):
+1, LGTM, too.

deshanxiao:
Thank you @williamhyun @dongjoon-hyun !

dongjoon-hyun:
Oh, this seems to cause more transitive dependencies in addition to `brotli4j.jar`. Are these expected and required ones, @deshanxiao ? Can we remove `service-1.15.0.jar`?

- https://github.com/dongjoon-hyun/spark/pull/9

  - native-osx-aarch64/1.15.0//native-osx-aarch64-1.15.0.jar
  - service/1.15.0//service-1.15.0.jar

dongjoon-hyun:
Thank you! For Spark community, I'll exclude it officially and add a documentation later to guide the users to add that dependency from application side.

cxzl25:
> For Spark community, I'll exclude it officially and add a documentation later to guide the users to add that dependency from application side.


Currently, the master branch of Spark has used ORC 2.0.0 version and supports brotli compression encoding. (SPARK-44115)
However, due to some hardcode checks in Spark, brotli cannot be used normally. 

Can I propose a PR that Spark supports ORC brotli ? 
 https://github.com/cxzl25/spark/commit/0cd5d03116709ec86d9f96e269200f2d05f799ca

```sql
set spark.sql.orc.compression.codec=BROTLI;
```
```
java.lang.IllegalArgumentException: The value of spark.sql.orc.compression.codec should be one of uncompressed, lz4, lzo, snappy, zlib, none, zstd, but was brotli
```

By the way, `netty-codec` declares brotli-related dependencies for multiple OSs.

https://github.com/netty/netty/blob/3cd364107167600e8eb4b0b85553ed895519e2ed/codec/pom.xml#L91-L125

dongjoon-hyun:
Yes, it was intentional, @cxzl25 , because it's a new feature from Spark's perspective. We need to focus on Spark perspective to do that like Spark documentation and configs.

Of course, you are welcome always. 😄 Go for it~
> Can I propose a PR that Spark supports ORC brotli ?

Issue #1430, opened 2023-03-07
[C++] DataBuffer Constructor (or a non default constructor) uses reserve instead of resize?


XinyuZeng:
The default constructor of DataBuffer uses resize:

https://github.com/apache/orc/blob/764257a7fb7ade4de9fed3813c67cb36629d5b00/c%2B%2B/src/MemoryPool.cc#L54-L58

which has extra initialization and free operations than reserve:

https://github.com/apache/orc/blob/764257a7fb7ade4de9fed3813c67cb36629d5b00/c%2B%2B/src/MemoryPool.cc#L81-L94

However, in some use cases, for example, create a row batch (ColumnVectorBatch) and read data from ORC file, the resize operation is redundant and reserve should be enough. Not sure this is a problem but I found a significant amount of time will be spent on creating batch if the batch size is large.

dongjoon-hyun:
cc @wgtmac , @stiga-huang , @coderex2522

wgtmac:
`reserve` only allocates the memory but does not call constructor of the objects of the buffer.  This is what `resize` does and will be undefined behavior if not doing so. Could you share your use case or some profiling data (e.g. flame graph) so we can discuss optimization opportunities in detail? @XinyuZeng

XinyuZeng:
We can refer to the FileScan utility: https://github.com/apache/orc/blob/main/tools/src/FileScan.cc#L32. When the batch_size is set to large (e.g., the number of rows in the whole file), the scan time increases from ~0.5 to ~0.7, and the additional time is on the ColumnVectorBatch creation, specifically `new (buf + i) T()` operation, but this is not necessary.

I am doing this because the method of scanning to ColumnVectorBatch first and then transforming to another in-memory format (e.g., arrow) is not zero-copy. There is an opportunity to transfer the memory ownership of ColumnVectorBatch to Arrow ([link](https://github.com/apache/arrow/issues/21238)) (although it is hard given ORC's ColumnVectorBatch's API), but that requires the allocation step of ColumnVectorBatch to be efficient.

XinyuZeng:
I am happy to work on this and probably also the zero-copy to arrow in April if this is not in a hurry. Sorry but I got busy this month.

wgtmac:
No problem. Just take your time.

Smith-Cruise:
I faced the same problem, DataBuffer's constructor is too heavy.
[...]

PR #1739, opened 2024-01-10
ORC-1580: Change default DataBuffer constructor to use reserve instead of resize

XinyuZeng:

See https://github.com/apache/orc/issues/1430 - default DataBuffer constructor now uses `reserve`.

I also add a `zeroOut` function to manually memset the buffer. For example, `MapColumnWrite::add` may use uninitialized `offsets` value (at `uint64_t elemOffset = static_cast<uint64_t>(offsets[0]);`). i.e., not memset the `offsets` buffer will result in segfault because the code implicitly assume the value should be 0. Therefore I still `zeroOut` the buffer for Map, List and Union, even though List and Union pass the unit tests without memset.

Reduce overhead in `ColumnVectorBatch` construction and compression/decompression buffer construction

Correctness via `make test-out`. I further use `orc-scan` utility, setting `batchSize` to 1 million and then scan a file with 1 million rows and 20 columns. After this PR, the time reduces from 2.9s to 2.2s. The compression/decompression buffer overhead reported in #1430 can be verified by the reporter if necessary.

No

### DIFF
--- a/c++/include/orc/MemoryPool.hh
+++ b/c++/include/orc/MemoryPool.hh
@@ -35,6 +35,11 @@ namespace orc {
   };
   MemoryPool* getDefaultPool();
 
+  struct no_memset_tag_t {
+    explicit no_memset_tag_t() = default;
+  };
+  inline constexpr no_memset_tag_t no_memset_tag{};
+
   template <class T>
   class DataBuffer {
    private:
@@ -51,7 +56,7 @@ namespace orc {
 
    public:
     DataBuffer(MemoryPool& pool, uint64_t _size = 0);
-    DataBuffer(MemoryPool& pool, uint64_t _size, bool noMemSet);
+    DataBuffer(MemoryPool& pool, uint64_t _size, no_memset_tag_t no_memset_tag);
 
     DataBuffer(DataBuffer<T>&& buffer) noexcept;
 
--- a/c++/include/orc/MemoryPool.hh
+++ b/c++/include/orc/MemoryPool.hh
@@ -19,11 +19,9 @@
 #ifndef MEMORYPOOL_HH_
 #define MEMORYPOOL_HH_
 
+#include <memory>
 #include "orc/Int128.hh"
 #include "orc/orc-config.hh"
-
-#include <memory>
-
 namespace orc {
 
   class MemoryPool {
@@ -56,7 +54,7 @@ namespace orc {
 
    public:
     DataBuffer(MemoryPool& pool, uint64_t _size = 0);
-    DataBuffer(MemoryPool& pool, uint64_t _size, no_memset_tag_t no_memset_tag);
+    DataBuffer(MemoryPool& pool, uint64_t _size, no_memset_tag_t);
 
     DataBuffer(DataBuffer<T>&& buffer) noexcept;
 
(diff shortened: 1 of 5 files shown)


c++/src/MemoryPool.cc
```diff
@@ -57,6 +57,15 @@ namespace orc {
     resize(newSize);
   }
 
+  // This constructor is used to create a DataBuffer that does not need to be memset to 0.
+  // bool noMemSet is not used in the code, but is kept here to avoid run-time branching.
+  template <class T>
+  DataBuffer<T>::DataBuffer(MemoryPool& pool, uint64_t newSize, [[maybe_unused]] bool noMemSet)
```
XinyuZeng, line 63: I feel the current way to differentiate the two constructors is ugly. Feel free to propose better one.

c++/include/orc/MemoryPool.hh
```diff
@@ -51,6 +51,7 @@ namespace orc {
 
    public:
     DataBuffer(MemoryPool& pool, uint64_t _size = 0);
+    DataBuffer(MemoryPool& pool, uint64_t _size, bool noMemSet);
```
dongjoon-hyun, line 54: When do we use `noMemSet=false`?

XinyuZeng: No. The problem is that it is possible to provide explicit `template <bool noMemSet>` to a constructor. Perhaps we can allow this branch in constructor? That makes the code better but introduce a branch that should be resolved at compile time.

XinyuZeng: In the new commit I tried to use an empty class tag type to differentiate the constructor

c++/src/MemoryPool.cc
```diff
@@ -57,6 +57,15 @@ namespace orc {
     resize(newSize);
   }
 
+  // This constructor is used to create a DataBuffer that does not need to be memset to 0.
+  // memset_tag is not used in the code, but is kept here to avoid run-time branching.
+  template <class T>
+  DataBuffer<T>::DataBuffer(MemoryPool& pool, uint64_t newSize, no_memset_tag_t)
```
wgtmac, line 63: Is it much cleaner if we directly change the original constructor to use this implementation without memset? For places where memset is required, we explicitly call it to initialize the values? Thus we do not bother to add a new constructor.

XinyuZeng: The new commit changes to this way. It can pass test. But I did not check the logic of every place where DataBuffer is constructed to ensure they do not implicitly assume the buffer is set to zero. The only place I found so far is `MapColumnWriter::add`

wgtmac:
cc @stiga-huang @ffacs @coderex2522

stiga-huang (approved):
LGTM. There are some other places that we can replace resize() with reserve(), e.g. in BlockDecompressionStream::NextDecompress()
https://github.com/apache/orc/blob/ffb0c87c94c8e1b5e0a7d722586585a1e1c23cfa/c%2B%2B/src/Compression.cc#L762-L765

We can further optimize them in follow-up tasks.

ffacs:
+1, LGTM

wgtmac (approved):

dongjoon-hyun (commented):
@XinyuZeng , could you revise the PR title and description? For example, PR title seems to become inconsistent from the AS-IS code.

XinyuZeng:
> @XinyuZeng , could you revise the PR title and description? For example, PR title seems to become inconsistent from the AS-IS code.

Done

dongjoon-hyun (approved):
+1, LGTM. Thank you, @XinyuZeng and all!

PR #1756, opened 2024-01-17
ORC-XXX: Support orc.compression.zstd.workers

dongjoon-hyun:

### DIFF
--- a/java/core/src/java/org/apache/orc/OrcConf.java
+++ b/java/core/src/java/org/apache/orc/OrcConf.java
@@ -80,6 +80,9 @@ public enum OrcConf {
       "hive.exec.orc.compression.zstd.windowlog", 0,
       "Set the maximum allowed back-reference distance for "
           + "ZStandard codec, expressed as power of 2."),
+  COMPRESSION_ZSTD_WORKERS("orc.compression.zstd.workers",
+          "hive.exec.orc.compression.zstd.workers", 0,
+          "Define the number of workers to use with ZStandard codec while writing data."),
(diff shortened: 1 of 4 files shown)


dongjoon-hyun:
Evaluating if this has any benefits in ORC.

cxzl25:
Thanks @dongjoon-hyun for doing this, this is also what I want to introduce this configuration after zstd-jni merge, like Spark and Parquet also have similar configurations.

dongjoon-hyun:
Ya, indeed.

BTW, it seems that there is no perf gain with this so far. Interesting.

cxzl25:
> it seems that there is no perf gain with this so far

Based on the product environment verification of this PR, I tested `orc.compression.zstd.workers` 0, 6, 15, and 16, and there seems to be no difference.

Although Paruqet also provides options for the number of zstd workers

https://github.com/apache/parquet-mr/blob/c82d5b471a558124b03e67759038661a046f5938/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/codec/ZstandardCodec.java#L53-L54

dongjoon-hyun:
Thank you for double-check. Ya, it seems that our implementation has some limitations or bug.

Apache Spark also has the ZStandardCodec implementation based on this `zstd-jni` and it shows 30% or 40% improvement in the micro-bencharmk.
- https://github.com/apache/spark/blob/39f8e1a5953b5897f893151d24dc585a80c0c8a0/core/benchmarks/ZStandardBenchmark-results.txt#L27-L47

I'm still digging this because I believe this should be a part of Apache ORC 2.0.0

PR #1761, opened 2024-01-18
ORC-1388: [C++] Support schema evolution from decimal to timestamp/string group

ffacs:
Support conversion from
{decimal}
to
{string, char, varchar, timestamp, timestamp_instant}

Support schema evolution at cpp side.

UT passed.

NO

### DIFF
--- a/c++/src/ConvertColumnReader.cc
+++ b/c++/src/ConvertColumnReader.cc
@@ -593,6 +593,105 @@ namespace orc {
     int32_t toScale;
   };
 
+  template <typename FileTypeBatch>
+  class DecimalToTimestampColumnReader : public ConvertToTimestampColumnReader {
+   public:
+    DecimalToTimestampColumnReader(const Type& _readType, const Type& fileType,
+                                   StripeStreams& stripe, bool _throwOnOverflow)
+        : ConvertToTimestampColumnReader(_readType, fileType, stripe, _throwOnOverflow),
+          precision(static_cast<int32_t>(fileType.getPrecision())),
+          scale(static_cast<int32_t>(fileType.getScale())) {}
+
+    void next(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) override {
+      ConvertColumnReader::next(rowBatch, numValues, notNull);
+      const auto& srcBatch = *SafeCastBatchTo<const FileTypeBatch*>(data.get());
+      auto& dstBatch = *SafeCastBatchTo<TimestampVectorBatch*>(&rowBatch);
+      for (uint64_t i = 0; i < rowBatch.numElements; ++i) {
+        if (!rowBatch.hasNulls || rowBatch.notNull[i]) {
+          convertDecimalToTimestamp(dstBatch, i, srcBatch);
+        }
+      }
+    }
+
+   private:
+    void convertDecimalToTimestamp(TimestampVectorBatch& dstBatch, uint64_t idx,
+                                   const FileTypeBatch& srcBatch) {
+      constexpr int SecondToNanoFactor = 9;
+      // '-1000000000-01-01T00:00Z'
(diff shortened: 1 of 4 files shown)


wgtmac:
Thanks for fixing the CI. I will take a look this week.

c++/src/ConvertColumnReader.cc
```diff
@@ -593,6 +593,105 @@ namespace orc {
...
+          convertDecimalToTimestamp(dstBatch, i, srcBatch);
+        }
+      }
+    }
+
+   private:
+    void convertDecimalToTimestamp(TimestampVectorBatch& dstBatch, uint64_t idx,
+                                   const FileTypeBatch& srcBatch) {
+      constexpr int SecondToNanoFactor = 9;
+      // '-1000000000-01-01T00:00Z'
```
wgtmac, line 620: Where are the min/max values defined? It would be good to add a comment for record.

c++/src/ConvertColumnReader.cc
```diff
@@ -593,6 +593,105 @@ namespace orc {
...
+      constexpr int64_t MIN_EPOCH_SECONDS = -31557014167219200L;
+      // '1000000000-12-31T23:59:59.999999999Z'
+      constexpr int64_t MAX_EPOCH_SECONDS = 31556889864403199L;
+      // dummy variable, there's no risk of overflow
```
wgtmac, line 624: Where is it?

c++/src/ConvertColumnReader.cc
```diff
@@ -593,6 +593,105 @@ namespace orc {
...
+        fractionPortion =
+            scaleUpInt128ByPowerOfTen(fractionPortion, SecondToNanoFactor - scale, overflow);
+      } else {
+        fractionPortion = scaleDownInt128ByPowerOfTen(fractionPortion, scale - SecondToNanoFactor);
+      }
+      if (fractionPortion < 0) {
+        fractionPortion += 1e9;
+        integerPortion -= 1;
+      }
+      dstBatch.data[idx] = integerPortion.toLong();
```
wgtmac, line 645: line 630 has guaranteed `toLong()` will not overflow?

c++/src/ConvertColumnReader.cc
```diff
@@ -593,6 +593,105 @@ namespace orc {
...
+    const int32_t precision;
+    const int32_t scale;
+  };
+
+  template <typename FileTypeBatch>
+  class DecimalToStringVariantColumnReader : public ConvertToStringVariantColumnReader {
+   public:
+    DecimalToStringVariantColumnReader(const Type& _readType, const Type& fileType,
+                                       StripeStreams& stripe, bool _throwOnOverflow)
+        : ConvertToStringVariantColumnReader(_readType, fileType, stripe, _throwOnOverflow),
```
wgtmac, line 662: Unrelated to this PR: we have many weird `_xxx` variable names appearing in the function signature. I'm thinking to do something like apache/arrow, which only use `xxx_` as the name of private class member variables. In this way, we can get rid of the weird case like this.

wgtmac (commented):
Mostly LGTM. I have left some questions.

c++/src/ConvertColumnReader.cc
ffacs: > Unrelated to this PR: we have many weird `_xxx` variable names appearing in the function signature. I'm thinking to do something like apache/arrow, which only use `xxx_` as the name of private class member variables. In this way, we can get rid of the weird case like this.

Maybe we can use clang-tidy with option "readability-identifier-naming.ClassMemberSuffix" to do this. I can make a PR to fix the identifiers naming style this or next week.

wgtmac: Thank you!

wgtmac (approved):

PR #1767, opened 2024-01-22
ORC-1458: Reduce HDFS NameNode getFileInfo RPC 

cxzl25:
If the file in HDFS is in a completed state, avoid calling the HDFS getFileInfo RPC.

Provide `orc.file.length.fast` configuration to enable this behavior.

Now reading an ORC file in HDFS will generate at least one `open` and `getFileInfo` RPC. This optimization can remove getFileInfo RPC as much as possible, improve ORC reading efficiency, and reduce the number of HDFS RPCs.

The production environment has been running stably for several months

No

### DIFF
--- a/java/core/src/java/org/apache/orc/util/OrcInputStreamUtil.java
+++ b/java/core/src/java/org/apache/orc/util/OrcInputStreamUtil.java
@@ -27,8 +27,6 @@ import java.lang.reflect.Method;
 
 public class OrcInputStreamUtil {
 
-  private static final String DFS_CLASS = "org.apache.hadoop.hdfs.DFSInputStream";
-
   private static Method shortCircuitForbiddenMethod;
 
   static {
@@ -43,9 +41,8 @@ public class OrcInputStreamUtil {
   }
 
   private static void initInt() throws ClassNotFoundException, NoSuchMethodException {
-    Class<?> dfsClass = Class.forName(DFS_CLASS);
     // org.apache.hadoop.hdfs.DFSInputStream.shortCircuitForbidden Method is not public
-    shortCircuitForbiddenMethod = dfsClass.getDeclaredMethod("shortCircuitForbidden");
+    shortCircuitForbiddenMethod = DFSInputStream.class.getDeclaredMethod("shortCircuitForbidden");
     shortCircuitForbiddenMethod.setAccessible(true);
   }
 
--- a/java/core/src/java/org/apache/orc/util/OrcInputStreamUtil.java
+++ b/java/core/src/java/org/apache/orc/util/OrcInputStreamUtil.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


java/core/src/java/org/apache/orc/util/OrcInputStreamUtil.java
```diff
@@ -0,0 +1,79 @@
...
+
+import org.apache.hadoop.fs.FSDataInputStream;
+
+import java.io.InputStream;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+public class OrcInputStreamUtil {
+
+  private static final String DFS_CLASS = "org.apache.hadoop.hdfs.DFSInputStream";
```
dongjoon-hyun, line 29: Why do we reflect this when Apache ORC can access this directly?

```
$ javap -cp hadoop-client-api-3.3.6.jar org.apache.hadoop.hdfs.DFSInputStream | head -n3
Compiled from "DFSInputStream.java"
public class org.apache.hadoop.hdfs.DFSInputStream extends org.apache.hadoop.fs.FSInputStream implements org.apache.hadoop.fs.ByteBufferReadable,org.apache.hadoop.fs.CanSetDropBehind,org.apache.hadoop.fs.CanSetReadahead,org.apache.hadoop.fs.HasEnhancedByteBufferAccess,org.apache.hadoop.fs.CanUnbuffer,org.apache.hadoop.fs.StreamCapabilities,org.apache.hadoop.fs.ByteBufferPositionedReadable {
  public static boolean tcpReadsDisabledForTesting;
```

dongjoon-hyun (changes requested):
This is a high-risky way which Apache Hadoop community is also not recommended. 

May I ask if you have any reference in ASF communities which adopts this hack, @cxzl25 ?

dongjoon-hyun:
Gentle ping, @cxzl25 .

java/core/src/java/org/apache/orc/util/OrcInputStreamUtil.java
cxzl25: `org.apache.hadoop.hdfs.DFSInputStream.shortCircuitForbidden` Method is not public.

cxzl25: Before [ORC-1430](https://issues.apache.org/jira/browse/ORC-1430), we could not access hdfs-related classes in the core module, so I used reflection.
Now I have done some refactoring, but the method of `shortCircuitForbidden` is not public, so I still need to use reflection.

cxzl25:
> This is a high-risky way which Apache Hadoop community is also not recommended.
> 
> May I ask if you have any reference in ASF communities which adopts this hack

@Hexiaoqiao Can you help review this PR?

Hexiaoqiao:
Thanks involve me here. I have left my concerns [here](https://issues.apache.org/jira/browse/ORC-1458?focusedCommentId=17739786&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17739786), I would like to reassert: I believe this could reduce load to NameNode when using ORC on HDFS but I am worried it involve any potential issues from HDFS view.

Back to this PR, IIUC this is used only for extract footer, right? Technically, for one immutable file, the result from `DFSInputStream#getFileLength()` and `FileStatus#getLen()` will be same, but it is not one recommended usage as @dongjoon-hyun mentioned above. 

Thanks again.

dongjoon-hyun:
Happy New Year. Do you still want to introduce this feature in Apache ORC for your use case, @cxzl25 ?

cxzl25:
Happy New Year.  @dongjoon-hyun 

> Do you still want to introduce this feature in Apache ORC for your use case

Let me close this PR first, this is a bit tricky, BTW S3 has a similar problem.

dongjoon-hyun:
Thank you for closing, @cxzl25 .  

For the following, could you file a JIRA Issue too? If we are aware of issues, the community can make alternative solution .
> BTW S3 has a similar problem.

cxzl25:
> For the following, could you file a JIRA Issue too? If we are aware of issues, the community can make alternative solution

Sorry for the late reply.

Created https://issues.apache.org/jira/browse/ORC-1845

dongjoon-hyun:
Thank you, @cxzl25 .

PR #1769, opened 2024-01-22
ORC-1601: Reduce get HadoopShims sync block in HadoopShimsFactory

cxzl25:
Improve calling efficiency of `org.apache.orc.impl.HadoopShimsFactory#get`.

`org.apache.orc.impl.HadoopShimsFactory#get` is a static method with synchronized. It needs to be locked every time it is called. This is unnecessary.

GA

No


dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Thank you, @cxzl25 .

PR #1784, opened 2024-02-04
ORC-1609: Fix the compilation problem of `TestJsonFileDump` in branch 1.9

cxzl25:
Fix `branch_1.9` compile problem

https://github.com/apache/orc/actions/runs/7769676224/job/21189006989

GA

No

### DIFF
--- a/java/tools/src/test/org/apache/orc/tools/TestJsonFileDump.java
+++ b/java/tools/src/test/org/apache/orc/tools/TestJsonFileDump.java
@@ -159,7 +159,7 @@ public class TestJsonFileDump {
     FileOutputStream myOut = new FileOutputStream(workDir + File.separator + outputFilename);
 
     // replace stdout and run command
-    System.setOut(new PrintStream(myOut, true, StandardCharsets.UTF_8));
+    System.setOut(new PrintStream(myOut, true, StandardCharsets.UTF_8.toString()));
     FileDump.main(new String[]{testFilePath.toString(), "-j", "-p"});
     System.out.flush();
     System.setOut(origOut);


java/tools/src/test/org/apache/orc/tools/TestJsonFileDump.java
```diff
@@ -159,7 +159,7 @@ public void testDoubleNaNAndInfinite() throws Exception {
     FileOutputStream myOut = new FileOutputStream(workDir + File.separator + outputFilename);
 
     // replace stdout and run command
-    System.setOut(new PrintStream(myOut, true, StandardCharsets.UTF_8));
+    System.setOut(new PrintStream(myOut, true, StandardCharsets.UTF_8.toString()));
```
paliwalashish, line 162: Looks good, does it make sense to have the same fix here and in main branch? The above API is available in both JDK 8 and 17, and the Charset one is only available in JDK17. Just a suggestion, else fix looks good

dongjoon-hyun:
Thanks. Please use a new Jira ID.

java/tools/src/test/org/apache/orc/tools/TestJsonFileDump.java
dongjoon-hyun: No need for main/branch-2.0 because the minimum is Java 17

cxzl25:
> Please use a new Jira ID

Thanks. Done.

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Merged to branch-1.9.

PR #1789, opened 2024-02-07
ORC-1613: Zstd decompression supports direct buffer

cxzl25:
`ZstdCodec` implements the `DirectDecompressionCodec` interface.

`zstd-jni` supports direct buffer decompression, which can reduce Buffer copying.

add UT

No

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ZstdCodec.java
+++ b/java/core/src/java/org/apache/orc/impl/ZstdCodec.java
@@ -25,7 +25,7 @@ import org.apache.orc.CompressionKind;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 
-public class ZstdCodec implements CompressionCodec {
+public class ZstdCodec implements CompressionCodec, DirectDecompressionCodec {
   private ZstdOptions zstdOptions = null;
   private ZstdCompressCtx zstdCompressCtx = null;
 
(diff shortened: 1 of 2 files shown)


java/core/src/java/org/apache/orc/impl/ZstdCodec.java
```diff
@@ -214,6 +214,11 @@ public boolean compress(ByteBuffer in, ByteBuffer out,
 
   @Override
   public void decompress(ByteBuffer in, ByteBuffer out) throws IOException {
+    if (in.isDirect() && out.isDirect()) {
```
dongjoon-hyun, line 217: Just a question. When is this used in Apache ORC code base?

cxzl25: If we use zerocopy `orc.use.zerocopy=true`, we may use the direct buffer, which can fallback to the original way of reading if the codec doesn't implement `DirectDecompressionCodec`.

https://github.com/apache/orc/blob/4eff23aa3ff6180f00cab472b91c3f67f82d6683/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java#L600-L607

dongjoon-hyun: Got it.

dongjoon-hyun (approved):

dongjoon-hyun:
Merged to main/2.0. Thank you, @cxzl25 .

PR #1790, opened 2024-02-07
ORC-1614: Use `directOut.put(out)` instead of `directOut.put(out.array())` in `TestBrotli` test

cxzl25:
Set ByteBuffer limit in `TestBrotli` test

`TestBrotli#testDirectDecompress` attempts to put the compressed result in direct ByteBuffer.
When calling `decompress`, we will find that the input buffer length is still `10000` not `217` since we put the array without `limit` on the length.

```java
      directOut.put(out.array());
      directOut.flip();
```

This PR is aimed to use `directOut.put(out)` instead of `directOut.put(out.array())`

GA

No

### DIFF
--- a/java/core/src/test/org/apache/orc/impl/TestBrotli.java
+++ b/java/core/src/test/org/apache/orc/impl/TestBrotli.java
@@ -136,10 +136,12 @@ public class TestBrotli {
       // write bytes to heap buffer.
       assertTrue(brotliCodec.compress(in, out, null,
           brotliCodec.getDefaultOptions()));
+      int position = out.position();
       out.flip();
       // copy heap buffer to direct buffer.
       directOut.put(out.array());
       directOut.flip();
+      directOut.limit(position);
 
       brotliCodec.decompress(directOut, directResult);
 
--- a/java/core/src/test/org/apache/orc/impl/TestBrotli.java
+++ b/java/core/src/test/org/apache/orc/impl/TestBrotli.java
@@ -136,12 +136,10 @@ public class TestBrotli {
       // write bytes to heap buffer.
       assertTrue(brotliCodec.compress(in, out, null,
           brotliCodec.getDefaultOptions()));
-      int position = out.position();
       out.flip();
       // copy heap buffer to direct buffer.
-      directOut.put(out.array());
+      directOut.put(out);
       directOut.flip();
-      directOut.limit(position);
 
       brotliCodec.decompress(directOut, directResult);
 


java/core/src/test/org/apache/orc/impl/TestBrotli.java
```diff
@@ -136,10 +136,12 @@ public void testDirectDecompress() {
       // write bytes to heap buffer.
       assertTrue(brotliCodec.compress(in, out, null,
           brotliCodec.getDefaultOptions()));
+      int position = out.position();
       out.flip();
       // copy heap buffer to direct buffer.
       directOut.put(out.array());
       directOut.flip();
+      directOut.limit(position);
```
paliwalashish, line 144: May be I didn't understand the issue correctly, but from [documentation](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/ByteBuffer.html#flip()) of flip(), limit is set by it.

"Flips this buffer. The limit is set to the current position and then the position is set to zero"

cxzl25: `out` is compressed by `brotliCodec`, and the length after compression is 217, but the `out.array()` used here is still the 10000 length of out, so the length that can be read after `directOut.flip()` is 10000, not 217

deshanxiao:
Good catch, Thanks!

In fact, I made a wrong decision to directly put the array into `directOut` since I misjudged the length of the array.

I prefer to change `directOut.put(out.array());` to `directOut.put(out);`

What do you think? @cxzl25

java/core/src/test/org/apache/orc/impl/TestBrotli.java
paliwalashish: Makes sense. Thanks !

dongjoon-hyun:
- Could you update the PR title and description accordingly, @cxzl25 ?
- Could you review this once more, @deshanxiao ?

deshanxiao (approved):

deshanxiao:
Thanks @dongjoon-hyun , I have updated the PR title and description.

deshanxiao:
Merged to main/2.0. Thank you, @cxzl25 @paliwalashish @dongjoon-hyun !

dongjoon-hyun:
Great! Thank you all!

dongjoon-hyun:
Hi, @deshanxiao . `branch-2.0` didn't have this patch. Did you use [this merge script](https://github.com/apache/orc/blob/main/dev/merge_orc_pr.py)?

Screenshot 2024-02-09 at 19 29 40

dongjoon-hyun:
Please note that `main` is Apache ORC 2.1.0.

deshanxiao:
Thanks @dongjoon-hyun , I have changed the comment and jira.

dongjoon-hyun:
Oh, we had better backport this to branch-2.0, didn't we?

deshanxiao:
Thanks @dongjoon-hyun Cherry-pick the fix to `branch-2.0`

Issue #1791, opened 2024-02-08
orc CXX fail to build if libgtest-dev is installed (debian-like systems)


douardda:
Trying to compile the cxx tools/libs, I have had linking errors. Here is a minimal dockerfile to reproduce the issue:

```
FROM debian:bookworm

RUN export DEBIAN_FRONTEND=noninteractive && \
  apt-get update && apt-get upgrade -y && \
  apt-get install -y \
  build-essential \
  cmake \
  gcc \
  g++ \
  git \
  libsasl2-dev \
  libssl-dev \
  make 

WORKDIR /root/
RUN git clone https://github.com/apache/orc /root/orc

CMD mkdir build && \
    cd build && \
    cmake ../orc -DBUILD_JAVA=OFF && \
    make package && \
    echo "OK without libgtest" && \
    mkdir /root/build_err && \
    cd /root/build_err && \
    DEBIAN_FRONTEND=noninteractive apt install -y libgtest-dev  && \
    cmake ../orc -DBUILD_JAVA=OFF && \
    make package && \
    echo "OK WITH libgtest"
```

Then:

```
$ docker build -t test_orc .
$ docker run --rm -ti  test_orc
[...]
OK without libgtest
[...]
Setting up googletest (1.12.1-0.2) ...
Setting up libgtest-dev:amd64 (1.12.1-0.2) ...
[...]
[  8%] Performing configure step for 'snappy_ep'
-- snappy_ep configure command succeeded.  See also /root/build_err/snappy_ep-prefix/src/snappy_ep-stamp/snappy_ep-configure-*.log
[  8%] Performing build step for 'snappy_ep'
CMake Error at /root/build_err/snappy_ep-prefix/src/snappy_ep-stamp/snappy_ep-build-RELWITHDEBINFO.cmake:49 (message):
  Command failed: 2

   'make'

  See also

    /root/build_err/snappy_ep-prefix/src/snappy_ep-stamp/snappy_ep-build-*.log


make[2]: *** [CMakeFiles/snappy_ep.dir/build.make:86: snappy_ep-prefix/src/snappy_ep-stamp/snappy_ep-build] Error 1
make[1]: *** [CMakeFiles/Makefile2:259: CMakeFiles/snappy_ep.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

The second cmake session fails with a link error while processing the snappy_ep build step. The log file contains lots of 'undefined reference' errors like:

```
[...]

PR #1792, opened 2024-02-09
ORC-1618: Disable building tests for snappy

douardda:
Disabling the building of tests is a workaround for issue #1791 and is probably not an issue since snappy tests are not executed as part of the build process of orc c++ libs.

Note that snappy 1.1.9 and 1.1.10 cannot be used as is because source assets on snappy's gh page are not buildable out of the box, so stick to 1.1.8 for now.

Disable  building snappy tests as a workaround for issue #1791.

compilation broken in some situations otherwise (see issue)

compile & unit tests

no

### DIFF
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -99,7 +99,7 @@ else ()
 
   ExternalProject_Add (snappy_ep
     URL "https://github.com/google/snappy/archive/${SNAPPY_VERSION}.tar.gz"
-    CMAKE_ARGS ${SNAPPY_CMAKE_ARGS}
+    CMAKE_ARGS ${SNAPPY_CMAKE_ARGS} -DSNAPPY_BUILD_TESTS=OFF
     ${THIRDPARTY_LOG_OPTIONS}
     BUILD_BYPRODUCTS "${SNAPPY_STATIC_LIB}")
 


cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -99,7 +99,7 @@ else ()
 
   ExternalProject_Add (snappy_ep
     URL "https://github.com/google/snappy/archive/${SNAPPY_VERSION}.tar.gz"
-    CMAKE_ARGS ${SNAPPY_CMAKE_ARGS}
+    CMAKE_ARGS ${SNAPPY_CMAKE_ARGS} -DSNAPPY_BUILD_TESTS=OFF
```
dongjoon-hyun, line 102: Thank you for suggestion, but why do we need to upgrade when we disable like this?

wgtmac: Same question here.

douardda: I did it because it seems not to have negative impact, but I don't think it's needed to fix the issue. I can change the PR not to include this version bump if you prefer.

douardda:
I've updated the PR to only keep the 'disable build tests' part.

cmake_modules/ThirdpartyToolchain.cmake
wgtmac: Thanks for the explanation! Do you mind creating a JIRA for this?

dongjoon-hyun:
> I've updated the PR to only keep the 'disable build tests' part.

Thank you for update, @douardda . Please update the PR title and description accordingly.

cmake_modules/ThirdpartyToolchain.cmake
douardda: Sorry, I don't have a Jira account (and I'd rather not have to create one just for this very simple MR).

dongjoon-hyun: Understood. Let me do that for you, @douardda .

douardda: Thanks!

PR #1813, opened 2024-02-20
ORC-1629: Make VectoredIO configurable

cxzl25:
Provide configuration item `orc.use.vectoredio` to decide whether to enable Hadoop's VectoredIO reading.

Conveniently verify the performance difference between turning on and not turning on VectoredIO.

add UT

No


dongjoon-hyun (changes requested):
Sorry but this is intentionally designed like that, @cxzl25 .

cxzl25:
> but this is intentionally designed like that

Oh, I see. Thanks.

dongjoon-hyun:
Sorry for not accepting this proposal. The original PR also had a similar proposal too.

PR #1816, opened 2024-02-26
ORC-1631: Support `summary` output in `sizes` command

cxzl25:
Add support for summarizing the number of files, file sizes and file lines in the sizes command.

When we count the size of each field, we only know the percentage and the average size of each row, but we do not know the overall value.

local test

```bash
java -jar orc-tools-2.1.0-SNAPSHOT-uber.jar sizes -h
usage: sizes
 -h,--help              Print help message
 -i,--ignoreExtension   Ignore ORC file extension
 -s,--summary           Summarize the number of files, file sizes, and
                        file rows
```

```
java -jar orc-tools-2.1.0-SNAPSHOT-uber.jar sizes -s
```

```
Total Files: 5
Total Sizes: 4803687270
Total Rows: 39820045
Percent  Bytes/Row  Name
  26.41  31.86
```

No

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/ColumnSizes.java
+++ b/java/tools/src/java/org/apache/orc/tools/ColumnSizes.java
@@ -126,7 +126,7 @@ public class ColumnSizes {
     }
   }
 
-  private void printResults(PrintStream out) {
+  private void printResults(PrintStream out, boolean summary) {
     List<StringLongPair> sizes = new ArrayList<>(columnSizes.length + 5);
     for(int column = 0; column < columnSizes.length; ++column) {
       if (columnSizes[column] > 0) {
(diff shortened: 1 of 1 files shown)


java/tools/src/java/org/apache/orc/tools/ColumnSizes.java
```diff
@@ -217,6 +227,11 @@ private static Options createOptions() {
         .desc("Ignore ORC file extension")
         .build());
 
+    result.addOption(Option.builder("s")
+        .longOpt("summary")
+        .desc("Summarize the number of files, file size, and number of file lines")
```
dongjoon-hyun, line 232: `file lines`? `Line` is not Apache ORC format. Maybe, did you want to say `Rows`?

cxzl25: Thanks. Updated.

dongjoon-hyun (approved):
Merged to main/2.0.

PR #1828, opened 2024-02-28
ORC-1639: [C++] Simplify flags in CMake and fix corresponding compiling errors

wgtmac:
- Remove unnecessary compiler flags and replace -Weverything with -Wall.
- Fix warnings and errors that emerge after the flag change.

The C++ library has added -Weverything which results in a bunch of -Wno-xxx flags. We should reduce them as many as possible.

- Tested it locally with clang13, clang17 and gcc13.
- Make sure all CIs are passed.

No.


dongjoon-hyun (approved):
+1, LGTM (if CI passes).

dongjoon-hyun:
Merged to branch-2.0. Thank you!

PR #1834, opened 2024-03-01
ORC-1644: Add `merge` tool to merge multiple ORC files into a single ORC file

cxzl25:
This PR aims to add merge tool to merges multiple ORC files to produce a single ORC file.

In the ORC 1.3.0 version, the `OrcFile#mergeFiles` method was introduced by [ORC-132](https://issues.apache.org/jira/browse/ORC-132) , which supports merging multiple ORC files into one ORC file. 
However, when merging, we need to write Java code to call it. 
There is no simple command that can be called directly.

Add UT

No

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/Driver.java
+++ b/java/tools/src/java/org/apache/orc/tools/Driver.java
@@ -91,10 +91,10 @@ public class Driver {
       System.err.println("   data - print the data from the ORC file");
       System.err.println("   json-schema - scan JSON files to determine their schema");
       System.err.println("   key - print information about the keys");
+      System.err.println("   merge - merge multiple ORC files into a single ORC file");
       System.err.println("   meta - print the metadata about the ORC file");
       System.err.println("   scan - scan the ORC file");
       System.err.println("   sizes - list size on disk of each column");
-      System.err.println("   merge - Merges multiple ORC files to produce a single ORC file");
       System.err.println("   version - print the version of this ORC tool");
       System.err.println();
       System.err.println("To get more help, provide -h to the command");
@@ -121,6 +121,9 @@ public class Driver {
       case "key":
         KeyTool.main(conf, options.commandArgs);
         break;
+      case "merge":
+        MergeFiles.main(conf, options.commandArgs);
+        break;
       case "meta":
         FileDump.main(conf, options.commandArgs);
         break;
@@ -129,9 +132,6 @@ public class Driver {
         break;
       case "sizes":
         ColumnSizes.main(conf, options.commandArgs);
-        break;
-      case "merge":
-        MergeFiles.main(conf, options.commandArgs);
         break;
       case "version":
         PrintVersion.main(conf, options.commandArgs);
--- a/java/tools/src/java/org/apache/orc/tools/Driver.java
+++ b/java/tools/src/java/org/apache/orc/tools/Driver.java
@@ -94,6 +94,7 @@ public class Driver {
       System.err.println("   meta - print the metadata about the ORC file");
       System.err.println("   scan - scan the ORC file");
       System.err.println("   sizes - list size on disk of each column");
+      System.err.println("   merge - Merges multiple ORC files to produce a single ORC file");
       System.err.println("   version - print the version of this ORC tool");
       System.err.println();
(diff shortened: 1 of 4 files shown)


java/tools/src/java/org/apache/orc/tools/Driver.java
```diff
@@ -94,6 +94,7 @@ public static void main(String[] args) throws Exception {
       System.err.println("   meta - print the metadata about the ORC file");
       System.err.println("   scan - scan the ORC file");
       System.err.println("   sizes - list size on disk of each column");
+      System.err.println("   merge - Merges multiple ORC files to produce a single ORC file");
```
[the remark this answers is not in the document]
dongjoon-hyun: Also, please revise the description like the following; (1) Use a lower case at the first char, (2) drop `s` at the end of `Merges`, (3) simplify from `to produce` -> `into`.
```
- Merges multiple ORC files to produce a single ORC file
- merge multiple ORC files into a single ORC file
```

java/tools/src/java/org/apache/orc/tools/MergeFiles.java
```diff
@@ -0,0 +1,132 @@
...
+      System.exit(1);
+    }
+  }
+
+  private static Options createOptions() {
+    Options result = new Options();
+
+    result.addOption(Option.builder("id")
+        .longOpt("inputDir")
+        .desc("Input orc directory to be merged")
```
dongjoon-hyun, line 111: This looks like a limitation. Do you mean we cannot simply merge two files without putting them into a new artificial directory?

java/tools/src/java/org/apache/orc/tools/MergeFiles.java
```diff
@@ -0,0 +1,132 @@
...
+        inputDir, outputPath, inputFiles.size(), mergedFiles.size());
+    if (!unSuccessMergedFiles.isEmpty()) {
+      System.exit(1);
+    }
+  }
+
+  private static Options createOptions() {
+    Options result = new Options();
+
+    result.addOption(Option.builder("id")
```
dongjoon-hyun, line 109: `id`? Is this consistent inside Apache ORC args? This looks very misleading to me.

java/tools/src/java/org/apache/orc/tools/MergeFiles.java
```diff
@@ -0,0 +1,132 @@
...
+        .longOpt("inputDir")
+        .desc("Input orc directory to be merged")
+        .hasArg()
+        .build());
+
+    result.addOption(Option.builder("op")
+        .longOpt("outputPath")
```
dongjoon-hyun, line 116: Why do we need to chars, `op`, when there is no conflicts with other options?

java/tools/src/java/org/apache/orc/tools/Driver.java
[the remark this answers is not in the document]
cxzl25: Thanks, I fixed it.

java/tools/src/java/org/apache/orc/tools/MergeFiles.java
cxzl25: I removed the parameter of the input path. Now we can support multiple directories and files without this limitation.

java/tools/src/java/org/apache/orc/tools/MergeFiles.java
cxzl25: Now I use the `-o` and `--output` options, consistent with other tools commands.

https://github.com/apache/orc/blob/01ebb961ba30f25efd33777a5220225feedc45c2/java/tools/src/java/org/apache/orc/tools/KeyTool.java#L101-L103

https://github.com/apache/orc/blob/01ebb961ba30f25efd33777a5220225feedc45c2/java/tools/src/java/org/apache/orc/tools/convert/ConvertTool.java#L257-L259

williamhyun (approved):
+1 LGTM

Thank you @cxzl25 @dongjoon-hyun, I will be merging to main and 2.0 for Apache ORC 2.0.1

dongjoon-hyun (commented):
+1, LGTM, too.

dongjoon-hyun:
Thank you, @williamhyun .

PR #1837, opened 2024-03-06
ORC-1646: Close the reader when reading the schema with the `convert` command

cxzl25:
This PR aims to close the reader when reading the schema with the `convert` command.

When using the `convert` command, when the input file format is ORC, the ORC schema is read, but not the close ORC reader.

local test

No


dongjoon-hyun (approved):
+1, LGTM.

Issue #1845, opened 2024-03-14
[C++] uniform identifiers naming style.


ffacs:
Unrelated to this PR: we have many weird `_xxx` variable names appearing in the function signature. I'm thinking to do something like apache/arrow, which only use `xxx_` as the name of private class member variables. In this way, we can get rid of the weird case like this.

_Originally posted by @wgtmac in https://github.com/apache/orc/pull/1761#discussion_r1468513025_

ffacs:
The following `.clang-tidy` file can help us to check naming styles mismatch.  And we can use `clang-tidy` to fix them.
```
Checks: "-*,
        readability-identifier-naming,
"

CheckOptions:
  [
      { key: readability-identifier-naming.PrivateMemberSuffix,   value: "_"  },
  ]

WarningsAsErrors: ''
HeaderFilterRegex: ''
FormatStyle:     none
```
But there is a problem that some protected/public variables need to be initialized on construction, what should the naming of 
corresponding parameter should be? What do you think? @wgtmac @dongjoon-hyun

wgtmac:
Could we add `NOLINT` to those public/protected variables?

ffacs:
> Could we add `NOLINT` to those public/protected variables?

Sorry for my confusing expression. What I mean just like below
https://github.com/apache/orc/blob/e1f185eedd7dc0d9d60339117c9299f5b26ddeaf/c%2B%2B/src/ConvertColumnReader.cc#L26-L33

The `readType` is a protected member of `ConvertColumnReader`. if we still use `readType` as the name, we can`t uniform the naming of constructor parameter.

wgtmac:
What about this?
- Use `_` suffix for private member variable.
- Do not use `_` suffix for protected and public member variable.
- Avoid any `_` prefix and suffix in any function parameter.

dongjoon-hyun:
Can we borrow some existing styles from other open source projects?
IIRC, initially, @wgtmac mentioned Arrow community. Are the above rules came from Arrow community?

dongjoon-hyun:
For the record, I welcome any linter-based rules. Thank you for making this efforts, @ffacs and @wgtmac .

wgtmac:
We could leverage https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html for variable naming. Apache Arrow used cpplint approach which we cannot borrow: https://github.com/apache/arrow/blob/main/cpp/build-support/cpplint.py

PR #1442, opened 2023-03-16
ORC-1395: Enforce license check via github action

wgtmac:
### What changes were proposed in this pull request?
Add a Gitbub action to enforce license check of c++ source files.

### Why are the changes needed?
https://github.com/apache/orc/pull/1427 was accidentally merged without license headers added to new files. We need a tool to automatically check it before merging.

### How was this patch tested?
Make sure the new action works as expected.

PR #1855, opened 2024-03-18
ORC-1658: [C++] uniform identifiers naming style.

ffacs:
1. Add run_clang_tidy.py which comes from LLVM and clang-tidy configuration file.
2. Use clang-tidy to format the C++ code.

To close https://github.com/apache/orc/issues/1845

UT passed

NO


ffacs:
This patch I`ve done the following things:
1. run clang-tidy `cd builf && python3 ../run_clang_tidy.py -p . -fix`
3. format codes
```
git ls-files --exclude-standard | grep ".*\(cc\|hh\)$" | xargs clang-format -i -style=file --dump-config 
```
3. add licence to run_clang_tidy.py and clang-tidy config file

Currently the patch can be built on my macOS(arm Apple clang version 15.0.0) and ubuntu(x64 clang-18) and passes UTs with `mkdir build -p && cd build && cmake .. -DBUILD_JAVA=OFF -DCMAKE_BUILD_TYPE=DEBUG -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang -DCMAKE_EXPORT_COMPILE_COMMANDS=1 && make -j20 && make test` .

c++/include/orc/sargs/Literal.hh
```diff
@@ -158,13 +158,13 @@ namespace orc {
...
 
    private:
-    LiteralVal mValue;        // data value for this literal if not null
-    PredicateDataType mType;  // data type of the literal
-    size_t mSize;             // size of mValue if it is Buffer
-    int32_t mPrecision;       // precision of decimal type
-    int32_t mScale;           // scale of decimal type
-    bool mIsNull;             // whether this literal is null
-    size_t mHashCode;         // precomputed hash code for the literal
+    LiteralVal mValue_;        // data value for this literal if not null
```
wgtmac, line 161: Is it possible to remove the `m` prefix?

c++/src/BloomFilter.hh
```diff
@@ -90,7 +90,7 @@ namespace orc {
     bool operator==(const BitSet& other) const;
 
    private:
-    std::vector<uint64_t> mData;
+    std::vector<uint64_t> mData_;
```
wgtmac, line 93: ditto, would be good to remove m prefix here and below.

c++/src/sargs/ExpressionTree.hh
```diff
@@ -74,10 +74,10 @@ namespace orc {
     TruthValue evaluate(const std::vector<TruthValue>& leaves) const;
 
    private:
-    Operator mOperator;
-    std::vector<TreeNode> mChildren;
-    size_t mLeaf;
-    TruthValue mConstant;
+    Operator mOperator_;
```
wgtmac, line 77: ditto

c++/src/sargs/PredicateLeaf.hh
```diff
@@ -143,13 +143,13 @@ namespace orc {
...
 
    private:
-    Operator mOperator;
-    PredicateDataType mType;
-    std::string mColumnName;
-    bool mHasColumnName;
-    uint64_t mColumnId;
-    std::vector<Literal> mLiterals;
-    size_t mHashCode;
+    Operator mOperator_;
```
wgtmac, line 146: ditto

c++/src/sargs/SargsApplier.hh
```diff
@@ -125,27 +125,27 @@ namespace orc {
     static uint64_t findColumn(const Type& type, const std::string& colName);
 
    private:
-    const Type& mType;
-    const SearchArgument* mSearchArgument;
-    const SchemaEvolution* mSchemaEvolution;
-    uint64_t mRowIndexStride;
-    WriterVersion mWriterVersion;
+    const Type& mType_;
```
wgtmac, line 128: ditto

c++/src/sargs/SearchArgument.hh
```diff
@@ -66,8 +66,8 @@ namespace orc {
     std::string toString() const override;
 
    private:
-    std::shared_ptr<ExpressionTree> mExpressionTree;
-    std::vector<PredicateLeaf> mLeaves;
+    std::shared_ptr<ExpressionTree> mExpressionTree_;
```
wgtmac, line 69: same here

run_clang_tidy.py
```diff
@@ -0,0 +1,343 @@
...
+      print('Error applying fixes.\n', file=sys.stderr)
+      traceback.print_exc()
+      return_code=1
+
+  if tmpdir:
+    shutil.rmtree(tmpdir)
+  sys.exit(return_code)
+
+if __name__ == '__main__':
+  main()
```
wgtmac, line 343: Is it possible to run this in the CI?

wgtmac (commented):
Thanks for the work! I've left some comments. Most are about the requests to remove `m` prefix. This can be a separate patch.

run_clang_tidy.py
ffacs: > Is it possible to run this in the CI?

Absolutely. We can leverage [clang-tidy/tool/clang-tidy-diff.py](https://github.com/llvm-mirror/clang-tools-extra/blob/5c40544fa40bfb85ec888b6a03421b3905e4a4e7/clang-tidy/tool/clang-tidy-diff.py#L3)

c++/include/orc/Exceptions.hh
```diff
@@ -28,8 +28,8 @@ namespace orc {
 
   class NotImplementedYet : public std::logic_error {
    public:
-    explicit NotImplementedYet(const std::string& what_arg);
-    explicit NotImplementedYet(const char* what_arg);
+    explicit NotImplementedYet(const std::string& whatArg);
+    explicit NotImplementedYet(const char* whatArg);
```
stiga-huang, line 32: Do we consider using snake case instead of mixed case (camel case) for variable names? Impala C++ Code follows a modified version of the Google C++ Style Guide at:
https://google.github.io/styleguide/cppguide.html#Variable_Names

I think either way is ok. Just remembered that when I first read the ORC C++ code, they look like Java code. :)

ffacs: There are lots of public/protected variables using camel case, if we change them into snake case, the library codes would be not backwards compatible. Backwards compatibility is important, I think.

wgtmac: +1 for keeping the compatibility.

run_clang_tidy.py
wgtmac: Perhaps we can leverage Github Action like https://github.com/marketplace/actions/c-c-linter or similar actions from here: https://github.com/marketplace?category=&type=actions&verification=&query=sort%3Apopularity-desc+clang-tidy

I have added two actions in the past:
- apache/skywalking-eyes@main for license header check
https://github.com/apache/orc/pull/1442
- jidicula/clang-format-action@v4.9.0 for clang-format check
https://github.com/apache/orc/pull/1308

It seems that we can use a single linter action to run both clang-tidy and clang-format.

c++/include/orc/Exceptions.hh
stiga-huang: I see. That makes sense to me.

ffacs:
@dongjoon-hyun Hi, do you have any suggestions about this patch?

dongjoon-hyun:
> @dongjoon-hyun Hi, do you have any suggestions about this patch?

Although I've been monitoring this PR and discussion, there is no other objection for the AS-IS direction including backward compatibility consideration.

dongjoon-hyun:
Thank you!

PR #1856, opened 2024-03-21
ORC-1663: [C++] Enable TestTimezone.testMissingTZDB on Windows

wgtmac:

Enable TestTimezone.testMissingTZDB unit test to run on Windows.

When /usr/share/zoneinfo is unavailable and TZDIR env is unset, creating C++ ORC reader will crash on Windows. We need to better deal with this case. See context from the Apache Arrow community: https://github.com/apache/arrow/issues/36026 and https://github.com/apache/arrow/issues/40633

Make sure the test passes on Windows.

No.

### DIFF
--- a/c++/test/TestTimezone.cc
+++ b/c++/test/TestTimezone.cc
@@ -428,15 +428,18 @@ namespace orc {
 
   TEST(TestTimezone, testMissingTZDB) {
     const char* tzDirBackup = std::getenv("TZDIR");
-    setEnv("TZDIR", "/path/to/wrong/tzdb");
+    if (tzDirBackup != nullptr) {
+      ASSERT_TRUE(delEnv("TZDIR"));
+    }
+    ASSERT_TRUE(setEnv("TZDIR", "/path/to/wrong/tzdb"));
     EXPECT_THAT([]() { getTimezoneByName("America/Los_Angeles"); },
                 testing::ThrowsMessage<TimezoneError>(testing::HasSubstr(
                     "Time zone file /path/to/wrong/tzdb/America/Los_Angeles does not exist."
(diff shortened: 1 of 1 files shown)


c++/test/TestTimezone.cc
```diff
@@ -403,20 +410,37 @@ namespace orc {
...
+#endif
+  }
+
+  bool delEnv(const char* name) {
+#ifdef _MSC_VER
+    return SetEnvironmentVariableA(name, nullptr);
+#else
+    return unsetenv(name) == 0;
+#endif
+  }
```
wgtmac, line 427:
```suggestion
  bool setEnv(const char* name, const char* value) {
#ifdef _MSC_VER
    return _putenv_s(name, value) == 0;
#else
    return setenv(name, value, 1) == 0;
#endif
  }

  bool delEnv(const char* name) {
#ifdef _MSC_VER
    return _putenv_s(name, "") == 0;
#else
    return unsetenv(name) == 0;
#endif
  }
```

c++/test/TestTimezone.cc
```diff
@@ -21,6 +21,13 @@
 #include "wrap/gmock.h"
 #include "wrap/gtest-wrapper.h"
 
+#ifdef _MSC_VER
+// clang-format off
+#include "windows.h"
+#include "processenv.h"
+// clang-format on
+#endif
+
```
wgtmac, line 30:
```suggestion
#include <cstdlib>
```

dongjoon-hyun (approved):

dongjoon-hyun:
Merged to main/2.0.

wgtmac:
Thanks @dongjoon-hyun !

PR #1857, opened 2024-03-21
ORC-1662: [C++] Upgrade protobuf to 3.21.12

ffacs:
Upgrade protobuf to 3.21.12

To fix compilation failure on vs2022

UT passed

NO

### DIFF
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -20,7 +20,7 @@ set(LZ4_VERSION "1.9.3")
 set(SNAPPY_VERSION "1.1.7")
 set(ZLIB_VERSION "1.2.11")
 set(GTEST_VERSION "1.12.1")
-set(PROTOBUF_VERSION "3.5.1")
+set(PROTOBUF_VERSION "3.21.12")
 set(ZSTD_VERSION "1.5.5")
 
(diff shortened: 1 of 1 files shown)


wgtmac:
Thanks for the PR! Unfortunately the new protobuf version is not compatible to compile libhdfspp. Let me investigate this.

wgtmac:
Update: I tried to upgrade libhdfspp. But unfortunately, I ran this command https://github.com/apache/orc/blob/main/c%2B%2B/libs/libhdfspp/pull_hdfs.sh#L25 without check. It wiped out my current directory (which is the root user profile).

dongjoon-hyun:
Oh, too bad for the accident.

dongjoon-hyun:
Could you rebase this PR, @ffacs ?

wgtmac:
@dongjoon-hyun I think it is a good idea to bump protobuf to a higher version. FTR, we internally upgraded to protobuf/5.27.0 for C++.

wgtmac:
Arrow is still using v21.3: https://github.com/apache/arrow/blob/main/cpp/thirdparty/versions.txt#L95

dongjoon-hyun:
Given that Arrow community is our big customer for ORC C++, I'm fine with the AS-IS PR also. 

Then, I'll bump the protobuf-java only because Spark community is our big customer for ORC java.

Thank you!

dongjoon-hyun:
Merged to main for Apache ORC 3.0.0. Thank you again ,@ffacs and @wgtmac .

PR #1859, opened 2024-03-26
ORC-1664: Migrate to `removeUnusedImports` of `spotless-maven-plugin`

cxzl25:
This PR aims to enable the `removeUnusedImports` function in `spotless-maven-plugin`.

Now we can use the following command to automatically remove unused imports.
```bash
mvn spotless:apply -Panalyze
```

We now checkstyle configured `<module name="UnusedImports"/>`, but it will only check and will not automatically fix the problem.

`spotless-maven-plugin` provides a useful function to automatically remove unused imports, which can provide the efficiency of java developers.

local test

No

### DIFF
--- a/java/checkstyle.xml
+++ b/java/checkstyle.xml
@@ -38,7 +38,6 @@
       <property name="allowClassImports" value="false"/>
       <property name="allowStaticMemberImports" value="false"/>
     </module>
-    <module name="UnusedImports"/>
     <module name="RedundantImport"/>
     <!-- https://checkstyle.sourceforge.io/config_imports.html#ImportOrder IntelliJ default example -->
     <module name="CustomImportOrder">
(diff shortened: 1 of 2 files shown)


dongjoon-hyun (changes requested):
+1 for this direction but we need to avoid duplication.

So, please remove the following.

https://github.com/apache/orc/blob/fa627ec6d7c72289c8a83632e6a43ae48603fc4b/java/checkstyle.xml#L41

In addition, can we migrate more rules from `checkstyle` to `spotless`?

cxzl25:
> please remove the following

done

> In addition, can we migrate more rules from `checkstyle` to `spotless`?

This is a good idea, I can continue to try it in the new JIRA.

dongjoon-hyun (approved):
Thank you for updating.

PR #1860, opened 2024-03-26
ORC-1665: Migrate to `importOrder` of `spotless-maven-plugin`

cxzl25:
This PR aims to enable the `importOrder` function in `spotless-maven-plugin`.

```bash
mvn spotless:apply -Panalyze
```

Although there is now checkstyle to limit the order of imports, it cannot be automatically adjusted. `spotless-maven-plugin` can check and automatically adjust.

https://github.com/apache/orc/pull/1859#pullrequestreview-1959266086

local test

No

### DIFF
--- a/java/checkstyle.xml
+++ b/java/checkstyle.xml
@@ -38,15 +38,6 @@
       <property name="allowClassImports" value="false"/>
       <property name="allowStaticMemberImports" value="false"/>
     </module>
-    <module name="RedundantImport"/>
-    <!-- https://checkstyle.sourceforge.io/config_imports.html#ImportOrder IntelliJ default example -->
-    <module name="CustomImportOrder">
-      <property name="customImportOrderRules" value="THIRD_PARTY_PACKAGE###SPECIAL_IMPORTS###STANDARD_JAVA_PACKAGE###STATIC"/>
-      <property name="specialImportsRegExp" value="^javax\."/>
-      <property name="standardPackageRegExp" value="^java\."/>
-      <property name="sortImportsInGroupAlphabetically" value="true"/>
-      <property name="separateLineBetweenGroups" value="false"/>
-    </module>
     <module name="ModifierOrder"/>
     <module name="RedundantModifier"/>
     <module name="NoLineWrap"/>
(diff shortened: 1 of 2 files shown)


java/core/src/java/org/apache/orc/InMemoryKeystore.java
```diff
@@ -330,4 +330,3 @@ private byte[] getMaterial() {
   }
 
 }
-
```
cxzl25, line 333: Because these files have two newlines at the end, one newline is automatically deleted after spotless apply.

java/bench/core/src/java/org/apache/orc/bench/core/RecordCounters.java
```diff
@@ -49,4 +49,3 @@ public void print() {
     System.out.println("Invocations: " + invocations);
   }
 }
-
```
dongjoon-hyun, line 52: Could you revert this change?

java/core/src/java/org/apache/orc/InMemoryKeystore.java
dongjoon-hyun: > Because these files have two newlines at the end, one newline is automatically deleted after spotless apply.

Please make a new PR for the end line change separately before all of these migration.

dongjoon-hyun: Also, we need a way to prevent any future regression in terms of the end line.

cxzl25: > Please make a new PR for the end line change separately before all of these migration.

OK

> we need a way to prevent any future regression in terms of the end line

Now `spotless-maven-plugin` will check this behavior, so the beginning of the CI failed at the beginning.

https://github.com/apache/orc/actions/runs/8437582330/job/23107719304#step:5:8408

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Could you resolve the conflicts, @cxzl25 ?

PR #1862, opened 2024-03-27
ORC-1667: Add `check` tool to check the index of the specified column

cxzl25:
This PR aims to check the index of the specified column.

We can test the filtering effect by specifying different types.

`check --type stat`  -  Only use column statistics.
`check --type bloom-filter` -  Only use bloom filter.
`check --type predicate`  - Used in combination with column statistics and bloom filter.

ORC supports specifying multiple columns to generate bloom filter indexes, but it lacks a convenient tool to verify the effect of bloom filter.

Parquet also has similar commands.
[PARQUET-2138](https://issues.apache.org/jira/browse/PARQUET-2138): Add ShowBloomFilterCommand to parquet-cli

Add UT

No

### DIFF
--- a/site/_docs/java-tools.md
+++ b/site/_docs/java-tools.md
@@ -11,7 +11,7 @@ supports both the local file system and HDFS.
 
 The subcommands for the tools are:
 
-  * bloom-filter (since ORC 2.1) - check the bloom filter of the specified column
+  * check (since ORC 2.0.1) - check the index of the specified column
   * convert (since ORC 1.4) - convert CSV/JSON/ORC files to ORC
   * count (since ORC 1.6) - recursively find *.orc and print the number of rows
   * data - print the data of an ORC file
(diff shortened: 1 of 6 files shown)


site/_docs/java-tools.md
```diff
@@ -11,6 +11,7 @@ supports both the local file system and HDFS.
 
 The subcommands for the tools are:
 
+  * bloom-filter (since ORC 2.1) - check the bloom filter of the specified column
```
dongjoon-hyun, line 14: - Since we can backport `tools` module patch, let's change this to `ORC 2.0.1`.
- The command name, `bloom-filter`, is a little ambiguous. Could you provide more intuitive name?

cxzl25: > let's change this to ORC 2.0.1.

OK

> The command name, bloom-filter, is a little ambiguous.

How about using `check-bloom-filter`?

dongjoon-hyun: Sounds better to me for this context.

BTW, if you have more plan to extend this commend, you can use `check` as subcommand or argument, too.

cxzl25: Thank you very much for your suggestion. 
Let us use the `check` command. Based on this, we can add min and max statistics to check later.

dongjoon-hyun (approved):
+1, LGTM. Thank you!

PR #1867, opened 2024-03-31
ORC-1671: update timestamp doc in specfication

Jefffrey:

Updating timestamp specification documentation to be more accurate to the implementation.

Timestamp specification is not clear on some nuances of it's implementation. Pulling knowledge from the following for these updates:

- https://github.com/apache/arrow/issues/34590
- https://issues.apache.org/jira/browse/ORC-10
- https://issues.apache.org/jira/browse/ORC-784
- https://issues.apache.org/jira/browse/ORC-763
- https://issues.apache.org/jira/browse/ORC-189
- C++ implementation

No

### DIFF
--- a/site/_docs/types.md
+++ b/site/_docs/types.md
@@ -69,7 +69,9 @@ create table Foobar (
 
 ORC includes two different forms of timestamps from the SQL world:
 
-* **Timestamp** is a date and time without a time zone, which does not change based on the time zone of the reader.
+* **Timestamp** is a date and time without a time zone, where the timestamp value is stored in the writer timezone
+encoded at the stripe level, if present. ORC readers will read this value back into the reader's timezone. Usually
+both writer and reader timezones default to UTC, however older ORC files may contain non-UTC writer timezones
 * **Timestamp with local time zone** is a fixed instant in time, which does change based on the time zone of the reader.
 
 Unless your application uses UTC consistently, **timestamp with
@@ -78,7 +80,7 @@ use cases. When users say an event is at 10:00, it is always in
 reference to a certain timezone and means a point in time, rather than
 10:00 in an arbitrary time zone.
 
-| Type        | Value in America/Los_Angeles | Value in America/New_York |
-| ----------- | ---------------------------- | ------------------------- |
-| **timestamp** | 2014-12-12 6:00:00           | 2014-12-12 6:00:00        |
-| **timestamp with local time zone** | 2014-12-12 9:00:00 | 2014-12-12 6:00:00 |
+| Type                               | Value in America/Los_Angeles | Value in America/New_York |
+| ---------------------------------- | ---------------------------- | ------------------------- |
+| **timestamp**                      | 2014-12-12 6:00:00           | 2014-12-12 6:00:00        |
+| **timestamp with local time zone** | 2014-12-12 9:00:00           | 2014-12-12 6:00:00        |
--- a/site/specification/ORCv1.md
+++ b/site/specification/ORCv1.md
@@ -1155,6 +1155,9 @@ records non-null values, a DATA stream that records the number of
 seconds after 1 January 2015, and a SECONDARY stream that records the
 number of nanoseconds.
 
+* Note that if writer timezone is set, 1 January 2015 is according to
+this timezone and not according to UTC
+
 Because the number of nanoseconds often has a large number of trailing
 zeros, the number has trailing decimal zero digits removed and the
 last three bits are used to record how many zeros were removed. if the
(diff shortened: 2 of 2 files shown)


site/specification/ORCv1.md
```diff
@@ -1155,6 +1155,9 @@ records non-null values, a DATA stream that records the number of
 seconds after 1 January 2015, and a SECONDARY stream that records the
 number of nanoseconds.
 
+* Note that if writer timezone is set, 1 January 2015 is according to
+this timezone and not according to UTC
```
Jefffrey, line 1159: According to implementation:

https://github.com/apache/orc/blob/3b5b2a6286df48a4ab471aece74bc7b7947042ad/c%2B%2B/src/Timezone.hh#L66-L72

https://github.com/apache/orc/blob/9b79de995430b240cb68f225c22bef69ebbd8b07/c%2B%2B/src/ColumnReader.cc#L335

Also, I'm not certain if the writer timezone is mandatory in the stripe if a TIMESTAMP column is present in the file. Might need some clarification on this?

site/specification/ORCv1.md
```diff
@@ -1170,6 +1173,35 @@ DIRECT_V2     | PRESENT         | Yes      | Boolean RLE
               | DATA            | No       | Signed Integer RLE v2
               | SECONDARY       | No       | Unsigned Integer RLE v2
 
+Due to ORC-763, values before the UNIX epoch which have nanoseconds greater
+than 999,999 are adjusted to have 1 second less.
```
Jefffrey, line 1177: According to

https://github.com/apache/orc/blob/9b79de995430b240cb68f225c22bef69ebbd8b07/c%2B%2B/src/ColumnReader.cc#L350-L352

site/specification/ORCv1.md
```diff
@@ -1170,6 +1173,35 @@ DIRECT_V2     | PRESENT         | Yes      | Boolean RLE
...
+from 1 January 1970 00:00:00 UTC to 1 January 2015 08:00:00 UTC is 1,420,099,200. This is
+added to the DATA value to produce a value of -20,751,903. As this is before the
+UNIX epoch (since it is negative), and the SECONDARY value, 199,900,000, is
+greater than 999,999, then this DATA value is adjusted to become -20,751,904
+(1 second subtracted).
+
+This value by itself represents 5 May 1969 19:34:56.1999, which now needs to be adjusted
+from US/Pacific (the writer's timezone) to UTC (the reader's timezone). As the value is
+within daylight savings for US/Pacific, 7 hours are subtracted to give the final value
+of 5 May 1969 12:34:56.1999.
```
Jefffrey, line 1196: I'm kinda unclear on how this exactly works, I'm just going off the C++ implementation here:

https://github.com/apache/orc/blob/9b79de995430b240cb68f225c22bef69ebbd8b07/c%2B%2B/src/ColumnReader.cc#L335-L348

Happy to be corrected if my understanding or wording is inaccurate anywhere :+1:

site/_docs/types.md
```diff
@@ -69,7 +69,9 @@ create table Foobar (
 
 ORC includes two different forms of timestamps from the SQL world:
 
-* **Timestamp** is a date and time without a time zone, which does not change based on the time zone of the reader.
+* **Timestamp** is a date and time without a time zone, where the timestamp value is stored in the writer timezone
+encoded at the stripe level, if present. ORC readers will read this value back into the reader's timezone. Usually
+both writer and reader timezones default to UTC, however older ORC files may contain non-UTC writer timezones
```
wgtmac, line 74: > Usually
both writer and reader timezones default to UTC, however older ORC files may contain non-UTC writer timezones`

The main purpose of this timestamp type is to restore the wall clock time regardless of the reader time zone. IIRC, the Java implementation uses the writer local time zone instead of UTC. Due to insufficient C++ time zone utility support in earlier days, the C++ code by default uses UTC as the writer time zone to avoid complexity.

site/specification/ORCv1.md
wgtmac: IIUC, writer timezone field is mandatory if there is any timestamp type: https://github.com/apache/orc/blob/513922af6c1d1a07e632a03231d5b546eeafb26e/java/core/src/java/org/apache/orc/impl/WriterImpl.java#L559-L565

Jefffrey: Thanks, I think I'll add this to the spec as well (maybe proto too, in the other repo?)

site/_docs/types.md
Jefffrey: I see. I was confused as from the discussion/comments here: https://github.com/apache/arrow/pull/34591

It suggested that the reader's timezone was important when deserializing the values

wgtmac: IMHO, we'd better not expose the implementation detail here. The original summary and the table below well explains what to expect from users.

site/specification/ORCv1.md
wgtmac: Similarly, I'm not inclined to add these details to the specs if we are not 100% sure about it.

site/_docs/types.md
Jefffrey: I'm not sure I agree with that, as this timestamp type has been a great source of confusion for me when attempting to implement an ORC to Arrow reader in Rust.

For example, in the test file [TestOrcFile.testDate2038.orc](https://github.com/apache/orc/blob/d193d130eb94cd2ee737423cc2f09a00d0035ebb/examples/TestOrcFile.testDate2038.orc), according to PyArrow (which uses the ORC C++ reader underneath), the first value is:

```python
>> from pyarrow import orc
>>> orc.read_table("/home/jeffrey/Code/orc/examples/TestOrcFile.testDate2038.orc")[0][0]
<pyarrow.TimestampScalar: '2038-05-05T12:34:56.100000000'>
```

The underlying integer value for the DATA stream is `736_601_696` seconds (let's ignore nanoseconds). When adding this to 1 Jan 2015 00:00:00, regardless of timezone, we get 5 May 2038 11:34:56. Which seems... surprising.

- https://www.timeanddate.com/date/dateadded.html?d1=1&m1=1&y1=2015&type=add&ay=&am=&aw=&ad=&h1=0&i1=0&s1=0&ah=&ai=&as=736601696&rec=

This isn't at all what I would expect hence my confusion (perhaps I just don't understand properly how DST factors into this). But diving into the internals and seeing reader & writer timezone offsets being considered when decoding the timestamp value just adds to the confusion, especially as that contradicts the specification.

I was hoping to open up some discussion on clarifying the exact definition for this timestamp type as part of this PR.

wgtmac: Sorry for the late reply. If you find any inconsistent behavior between C++ and Java sides, please refer to the Java implementation as reference. We need to fix any issue on the C++ side.

Jefffrey: I'll try check more into this to confirm my understanding :+1: 

Moving to draft in the meantime

PR #1868, opened 2024-03-31
ORC-642: update PatchedBase doc with patch ceiling in spec

Jefffrey:

Update PatchedBase specification doc to include details about the behaviour of padding the patch gap + patch width bits to nearest fixed btis.

Ensure spec is accurate to implementation details

No

### DIFF
--- a/site/specification/ORCv1.md
+++ b/site/specification/ORCv1.md
@@ -824,14 +824,31 @@ the index values and the additional value bits.
   bit is set, the entire value is negated.
 * Data values (W * L bits padded to the byte) - A sequence of W bit positive
   values that are added to the base value.
-* Patch list (PLL * (PGW + PW) bytes) - A list of patches for values
+* Patch list (PLL * ceil(PGW + PW) bits) - A list of patches for values
   that didn't fit within W bits. Each entry in the list consists of a
   gap, which is the number of elements skipped from the previous
   patch, and a patch value. Patches are applied by logically or'ing
   the data values with the relevant patch shifted W bits left. If a
   patch is 0, it was introduced to skip over more than 255 items. The
   combined length of each patch (PGW + PW) must be less or equal to
-  64.
+  64. (PGW + PW) is padded to the nearest fixed bit size according to the
+  below table before being encoded in the patch list.
+
+(PGW + PW)    | ceil(PGW + PW)
+:------------ | :-------------
+1 <= x <= 24  | x
+25            | 26
+26            | 26
+27            | 28
+28            | 28
+29            | 30
+30            | 30
+31            | 32
+32            | 32
+33 <= x <= 40 | 40
+41 <= x <= 48 | 48
+49 <= x <= 56 | 56
+57 <= x <= 64 | 64
 
 The unsigned sequence of [2030, 2000, 2020, 1000000, 2040, 2050, 2060, 2070,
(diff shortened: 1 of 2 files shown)


site/specification/ORCv1.md
```diff
@@ -824,14 +824,31 @@ the index values and the additional value bits.
...
+27            | 28
+28            | 28
+29            | 30
+30            | 30
+31            | 32
+32            | 32
+33 <= x <= 40 | 40
+41 <= x <= 48 | 48
+49 <= x <= 56 | 56
+57 <= x <= 64 | 64
```
Jefffrey, line 851: From

https://github.com/apache/orc/blob/9b79de995430b240cb68f225c22bef69ebbd8b07/c%2B%2B/src/RLEV2Util.cc#L29-L33

site/specification/ORCv1.md
```diff
@@ -824,14 +824,31 @@ the index values and the additional value bits.
...
+* Patch list (PLL * ceil(PGW + PW) bits) - A list of patches for values
   that didn't fit within W bits. Each entry in the list consists of a
   gap, which is the number of elements skipped from the previous
   patch, and a patch value. Patches are applied by logically or'ing
   the data values with the relevant patch shifted W bits left. If a
   patch is 0, it was introduced to skip over more than 255 items. The
   combined length of each patch (PGW + PW) must be less or equal to
-  64.
+  64. (PGW + PW) is padded to the nearest fixed bit size according to the
+  below table before being encoded in the patch list.
```
Jefffrey, line 835: According to https://github.com/apache/orc/blob/9b79de995430b240cb68f225c22bef69ebbd8b07/c%2B%2B/src/RleDecoderV2.cc#L332

wgtmac: Could you please rename `ceil` to `closestFixedBits` or `cfb`? `ceil` does not seems to have its common meaning here.

site/specification/ORCv1.md
wgtmac: Is it better to follow the java code to add 0 as well? https://github.com/apache/orc/blob/513922af6c1d1a07e632a03231d5b546eeafb26e/java/core/src/java/org/apache/orc/impl/SerializationUtils.java#L362-L391

wgtmac (changes requested):
Thanks for the fix! Could you please also add them to ORCv2.md as well?

site/specification/ORCv1.md
Jefffrey: Sure thing, done

site/specification/ORCv1.md
Jefffrey: I don't think it makes sense to have 0 as the input is PGW + PW, and the spec states that PGW is 1 to 8 bits and PW is 1 to 64 bits, so it can never be 0 (or 1, for that matter, I guess :thinking: )

wgtmac (approved):
LGTM +1

wgtmac:
cc @dongjoon-hyun @deshanxiao

deshanxiao (approved):

dongjoon-hyun (approved):
+1, LGTM.

deshanxiao:
Thanks @Jefffrey @wgtmac @dongjoon-hyun . Merged to main.

PR #1870, opened 2024-03-31
ORC-1672: Remove test packages `o.a.o.tools.check`

cxzl25:
This PR aims to remove test packages `o.a.o.tools.check`.

ORC-1667 was originally designed for bloom-filter, so the package name has not been modified.

GA

No

### DIFF
--- a/java/tools/src/test/org/apache/orc/tools/check/TestCheckTool.java
+++ b/java/tools/src/test/org/apache/orc/tools/check/TestCheckTool.java
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-package org.apache.orc.tools.bloomfilter;
+package org.apache.orc.tools.check;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
--- a/java/tools/src/test/org/apache/orc/tools/TestCheckTool.java
+++ b/java/tools/src/test/org/apache/orc/tools/TestCheckTool.java
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-package org.apache.orc.tools.check;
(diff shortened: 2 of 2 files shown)


java/tools/src/test/org/apache/orc/tools/check/TestCheckTool.java
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-package org.apache.orc.tools.bloomfilter;
+package org.apache.orc.tools.check;
```
dongjoon-hyun, line 19: May I ask why do we need this new test package when the original main code is in `package org.apache.orc.tools`?

dongjoon-hyun: BTW, this design is not this PR's issue. I checked the existing code and found that we need to revise the existing test code package too.

dongjoon-hyun (changes requested):
I made a PR to fix the underlying test package issue. I hope we can revise this PR in the same direction, @cxzl25 .
- #1871

java/tools/src/test/org/apache/orc/tools/check/TestCheckTool.java
cxzl25: > why do we need this new test package when the original main code is in `package org.apache.orc.tools`?

I'm sorry about that.

I tried to recall why it was designed like this at the beginning, but I kind of forgot.

cxzl25:
> I made a PR to fix the underlying test package issue. I hope we can revise this PR in the same direction

Thanks for your help, I also removed check package name.

dongjoon-hyun (approved):
Merged to main/2.0.

PR #1877, opened 2024-04-06
ORC-1677: [C++] Remove `m` prefix of class member variables

ffacs:
Remove the `m` prefix of identifiers.

To unify naming style of identifiers.

UT passed.

NO


wgtmac (approved):
+1 

Thanks @ffacs!

PR #1882, opened 2024-04-10
ORC-1684: [C++] Find tzdb without TZDIR when in conda-environments

h-vetinari:

Find tzdb without having to set `TZDIR` when in a conda-environment (where `tzdata` [has](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fnoarch%2Ftzdata-2024a-h0c530f3_0.conda) a uniform location of `$CONDA_PREFIX/share/zoneinfo` across all platforms).

This is due to issues in arrow (see https://github.com/apache/arrow/issues/36026) that cannot really be fixed there, as it assumes that orc >=2.0 knows how to find the tzdb. Having to set `TZDIR` in all user environments is an intrusive change that should be avoided, and since the cost here is checking a single environment variable, it's hopefully not too onerous for consideration.

CI here

No

CC @wgtmac

See also: #1587

### DIFF
--- a/c++/src/Timezone.cc
+++ b/c++/src/Timezone.cc
@@ -659,7 +659,15 @@ namespace orc {
   const char* getTimezoneDirectory() {
     const char* dir = getenv("TZDIR");
     if (!dir) {
-      dir = DEFAULT_TZDIR;
+      // this is present if we're in an activated conda environment
+      const char* conda_prefix = getenv("CONDA_PREFIX");
+      if (conda_prefix) {
+        std::string conda_dir(conda_prefix);
+        conda_dir += "/share/zoneinfo";
+        dir = conda_dir.c_str();
+      } else {
+        dir = DEFAULT_TZDIR;
+      }
     }
     return dir;
   }
--- a/c++/src/Timezone.cc
+++ b/c++/src/Timezone.cc
@@ -656,7 +656,7 @@ namespace orc {
     epoch_ = utcEpoch - getVariant(utcEpoch).gmtOffset;
   }
 
-  const char* getTimezoneDirectory() {
+  std::string getTimezoneDirectory() {
     const char* dir = getenv("TZDIR");
     if (!dir) {
       // this is present if we're in an activated conda environment
@@ -664,12 +664,13 @@ namespace orc {
       if (condaPrefix) {
         std::string condaDir(condaPrefix);
         condaDir += "/share/zoneinfo";
-        dir = condaDir.c_str();
+        return condaDir;
       } else {
         dir = DEFAULT_TZDIR;
       }
     }
-    return dir;
+    std::string tzDir(dir);
+    return tzDir;
   }
 
   /**
--- a/c++/src/Timezone.cc
+++ b/c++/src/Timezone.cc
@@ -669,8 +669,7 @@ namespace orc {
         dir = DEFAULT_TZDIR;
       }
     }
-    std::string tzDir(dir);
-    return tzDir;
+    return dir;
   }
 
   /**
--- a/c++/src/Timezone.cc
+++ b/c++/src/Timezone.cc
@@ -660,11 +660,11 @@ namespace orc {
     const char* dir = getenv("TZDIR");
     if (!dir) {
       // this is present if we're in an activated conda environment
-      const char* conda_prefix = getenv("CONDA_PREFIX");
-      if (conda_prefix) {
-        std::string conda_dir(conda_prefix);
-        conda_dir += "/share/zoneinfo";
-        dir = conda_dir.c_str();
+      const char* condaPrefix = getenv("CONDA_PREFIX");
+      if (condaPrefix) {
+        std::string condaDir(condaPrefix);
+        condaDir += "/share/zoneinfo";
+        dir = condaDir.c_str();
       } else {
         dir = DEFAULT_TZDIR;
       }
--- a/c++/test/TestTimezone.cc
+++ b/c++/test/TestTimezone.cc
@@ -476,13 +476,13 @@ namespace orc {
       // CONDA_PREFIX contains backslashes on windows; test that this doesn't blow up
       std::replace(condaPrefix.begin(), condaPrefix.end(), '/', '\\');
       ASSERT_TRUE(setEnv("CONDA_PREFIX", condaPrefix.c_str()));
-      const Timezone* la1 = &getTimezoneByName("America/Los_Angeles");
-      const Timezone* la2 = &getTimezoneByName("America/Los_Angeles");
-      EXPECT_EQ(la1, la2);
-      EXPECT_EQ("PST", getVariantFromZone(*la1, "1974-01-06 09:59:59"));
-      EXPECT_EQ("PDT", getVariantFromZone(*la1, "1974-01-06 10:00:00"));
-      EXPECT_EQ("PDT", getVariantFromZone(*la1, "1974-10-27 08:59:59"));
-      EXPECT_EQ("PST", getVariantFromZone(*la1, "1974-10-27 09:00:00"));
+      const Timezone* la3 = &getTimezoneByName("America/Los_Angeles");
+      const Timezone* la4 = &getTimezoneByName("America/Los_Angeles");
+      EXPECT_EQ(la3, la4);
+      EXPECT_EQ("PST", getVariantFromZone(*la3, "1974-01-06 09:59:59"));
+      EXPECT_EQ("PDT", getVariantFromZone(*la3, "1974-01-06 10:00:00"));
+      EXPECT_EQ("PDT", getVariantFromZone(*la3, "1974-10-27 08:59:59"));
+      EXPECT_EQ("PST", getVariantFromZone(*la3, "1974-10-27 09:00:00"));
 
       ASSERT_TRUE(delEnv("CONDA_PREFIX"));
       ASSERT_TRUE(setEnv("TZDIR", tzDirBackup));
--- a/c++/test/TestTimezone.cc
+++ b/c++/test/TestTimezone.cc
@@ -466,7 +466,7 @@ namespace orc {
       std::replace(condaPrefix.begin(), condaPrefix.end(), '/', '\\');
       ASSERT_TRUE(setEnv("CONDA_PREFIX", condaPrefix.c_str()));
 
-      // as above
+      // as above, but different timezone to avoid hitting cache
       const Timezone* syd = &getTimezoneByName("Australia/Sydney");
       EXPECT_EQ("AEDT", getVariantFromZone(*syd, "2024-04-06 15:59:59"));
       EXPECT_EQ("AEST", getVariantFromZone(*syd, "2024-04-06 16:00:00"));
--- a/c++/test/TestTimezone.cc
+++ b/c++/test/TestTimezone.cc
@@ -421,21 +421,13 @@ namespace orc {
 #endif
   }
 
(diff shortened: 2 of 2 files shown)


h-vetinari:
If desired, I can test this patch on conda-forge infrastructure (build a patched orc that's published in a separate channel, then run https://github.com/conda-forge/arrow-cpp-feedstock/pull/1086 against that).

wgtmac:
Thanks for the fix! Do you have a JIRA account? I can assign this to you.

c++/src/Timezone.cc
```diff
@@ -659,7 +659,15 @@ namespace orc {
   const char* getTimezoneDirectory() {
     const char* dir = getenv("TZDIR");
     if (!dir) {
-      dir = DEFAULT_TZDIR;
+      // this is present if we're in an activated conda environment
+      const char* conda_prefix = getenv("CONDA_PREFIX");
```
wgtmac, line 663: Usually we use variable naming style like `condaPrefix`

c++/src/Timezone.cc
```diff
@@ -659,7 +659,15 @@ namespace orc {
...
+      if (conda_prefix) {
+        std::string conda_dir(conda_prefix);
+        conda_dir += "/share/zoneinfo";
+        dir = conda_dir.c_str();
```
wgtmac, line 667: This unfortunately returns a dangling pointer.

wgtmac (changes requested):

c++/src/Timezone.cc
h-vetinari: Not sure I understand (but then, my C++ is limited...); `dir` is `const char *`, and `std::string` doesn't auto-convert. How should I set `dir` with the correct value? I can also construct the `conda_dir` path differently if you want, I just followed the example from `getTimezoneByName` that was using `std::string`.

c++/src/Timezone.cc
h-vetinari: sure!

h-vetinari (commented):
Thanks for the quick review!

> Thanks for the fix! Do you have a JIRA account? I can assign this to you.

https://issues.apache.org/jira/secure/ViewProfile.jspa?name=h-vetinari

c++/src/Timezone.cc
wgtmac: c_str() returns the pointer in the backing string object which is a temporary object and is destructed when the function returns.

wgtmac: It seems we need to change the function to return std::string instead

h-vetinari: > It seems we need to change the function to return std::string instead

I tried that now; not sure how you'd like to do this stylistically (i.e. the early return for `condaDir`); for now I didn't want to touch the `getenv` bits, because that always returns `char *`.

h-vetinari:
Would it be possible/acceptable to create a symlink in CI? I believe we could write a test that exercises the branch for `CONDA_PREFIX` here similar to
https://github.com/apache/orc/blob/b75dcaa060ce4e9cb5bc4600b00bfb8769efc50b/c%2B%2B/test/TestTimezone.cc#L423
but it would need a working tzdb at the end of `$CONDA_PREFIX/share/zoneinfo`.

h-vetinari:
Interestingly, I cannot inspect any CI logs here. Unfolding any section just gives a red `Error:`.

wgtmac:
> Ah, just found
> 
> https://github.com/apache/orc/blob/b75dcaa060ce4e9cb5bc4600b00bfb8769efc50b/cmake_modules/ThirdpartyToolchain.cmake#L340
> 
> 
> So I guess it should be there?

That is only used in WIN32 to make test pass on Windows...

h-vetinari:
> That is only used in WIN32 to make test pass on Windows...

Well, then we could still test that the CONDA_PREFIX mechanics work on windows, no? See my last commit, which tries to do this.

c++/test/TestTimezone.cc
```diff
@@ -437,4 +437,29 @@ namespace orc {
...
+  TEST(TestTimezone, testTzdbFromCondaEnv) {
+    const char* tzDirBackup = std::getenv("TZDIR");
+    // test only makes sense if TZDIR exists
+    if (tzDirBackup != nullptr) {
+      ASSERT_TRUE(delEnv("TZDIR"));
+
+      // remove "/share/zoneinfo" from TZDIR to get (equivalent of) CONDA_PREFIX
+      std::string condaPrefix(tzDirBackup);
+      condaPrefix += "/../..";
+      ASSERT_TRUE(setEnv("CONDA_PREFIX", condaPrefix.c_str()));
```
h-vetinari, line 449: Hm, so I thought I'd avoid the dangling pointer problem here as `condaPrefix` is alive for the duration of the `setEnv` call, but it seems that it doesn't work and produces a garbage path.

h-vetinari: It's possible that the pointer is just empty because we deleted TZDIR too soon. Let's try reversing that.

h-vetinari:
OK, so the test I added passed now:
```
[ RUN      ] TestTimezone.testTzdbFromCondaEnv
[       OK ] TestTimezone.testTzdbFromCondaEnv (0 ms)
```
but restoring `TZDIR` still fails:
```
 unknown file: error: C++ exception with description "Time zone file ������������������������������������������������������������������������������/UTC does not exist. 
```

Since this is copied from `testMissingTZDB`, I'm surprised the same error doesn't happen there, because that test is saving `tzDirBackup` in exactly the same way - i.e. just a pointer to an environment variable which then gets deleted. It seems to me that the restoration of that environment variable cannot succeed (without a deepcopy before deleting `TZDIR`), and so far that test may have only passed due to checking `if (tzDirBackup != nullptr)`, but - potentially - never actually restored the environment variable...? 🤔

h-vetinari:
Tests are passing on windows now! 🥳

h-vetinari:
This is all green now and should be ready for closer review. :)

c++/src/Timezone.cc
```diff
@@ -656,12 +656,21 @@ namespace orc {
...
+        std::string condaDir(condaPrefix);
+        condaDir += "/share/zoneinfo";
+        return condaDir;
+      } else {
+        dir = DEFAULT_TZDIR;
+      }
     }
-    return dir;
+    std::string tzDir(dir);
+    return tzDir;
```
wgtmac, line 673:
```suggestion
    return dir;
```
This is enough

c++/test/TestTimezone.cc
```diff
@@ -420,9 +421,21 @@ namespace orc {
...
+  }
+
   TEST(TestTimezone, testMissingTZDB) {
-    const char* tzDirBackup = std::getenv("TZDIR");
-    if (tzDirBackup != nullptr) {
+    const char* tzDir = std::getenv("TZDIR");
+    char* tzDirBackup = nullptr;
+    if (tzDir != nullptr) {
+      // avoid that unsetting environment variable wrecks pointer to tzDir
+      tzDirBackup = deepcopy(tzDir);
```
wgtmac, line 438: Could we simply use `std::string tzDirBackup = tzDir;` In this way, we don't need to free it in the end.

ffacs:
Thank you, @h-vetinari . This patch looks good to me if all comments from @wgtmac are fixed.

c++/src/Timezone.cc
h-vetinari: Yeah, going from `char*` to `std::string` works; I was trying to make the conversion obvious, but happy to do it like this.

c++/test/TestTimezone.cc
h-vetinari: Done.

h-vetinari:
Thanks for the reviews! I made the requested changes.

c++/test/TestTimezone.cc
h-vetinari: Sorry for the newbie mistakes; I had checked that `std::string` can be initialized with `nullptr` (even though I now get that it's against the design), but not that it cannot be compared. 😅 

Anyway, using `s.empty()` [works](https://en.cppreference.com/w/cpp/string/basic_string/empty) even without an explicit initialization of `std::string`.

h-vetinari:
I have a last question - I see that there's some sort of caching for the tzdb going on (or at least, there's a test for it):
https://github.com/apache/orc/blob/695e0f33609adae0114708c3be756bb7b77f4d79/c%2B%2B/test/TestTimezone.cc#L346-L347

Can we delete that cache somehow to rule out that this is what's causing `testTzdbFromCondaEnv` to pass?

wgtmac:
> I have a last question - I see that there's some sort of caching for the tzdb going on (or at least, there's a test for it):
> 
> https://github.com/apache/orc/blob/695e0f33609adae0114708c3be756bb7b77f4d79/c%2B%2B/test/TestTimezone.cc#L346-L347
> 
> Can we delete that cache somehow to rule out that this is what's causing `testTzdbFromCondaEnv` to pass?

We can either use a different timezone to avoid caching or add a function to invalidate the cache.

h-vetinari:
Sorry for the force-pushes; I had to figure out the right dates/times (which is a bit harder because we apparently need to specify them in UTC rather than local time; in the latter it would always be around 2/3am).

c++/test/TestTimezone.cc
```diff
@@ -421,20 +422,61 @@ namespace orc {
...
+      std::string condaPrefix(tzDirBackup);
+      condaPrefix += "/../..";
+      ASSERT_TRUE(setEnv("CONDA_PREFIX", condaPrefix.c_str()));
+
+      // small test sample to ensure tzbd loads with CONDA_PREFIX, even without TZDIR
+      const Timezone* zrh = &getTimezoneByName("Europe/Zurich");
+      EXPECT_EQ("CET", getVariantFromZone(*zrh, "2024-03-31 00:59:59"));
+      EXPECT_EQ("CEST", getVariantFromZone(*zrh, "2024-03-31 01:00:00"));
+      EXPECT_EQ("CEST", getVariantFromZone(*zrh, "2024-10-26 23:59:59"));
+      EXPECT_EQ("CET", getVariantFromZone(*zrh, "2024-10-27 01:00:00"));
```
h-vetinari, line 463: This initially looked like an inconsistency in `getVariantFromZone`, at least w.r.t. how the testing is done for Los Angeles / New York. In particular, in the switch from daylight savings time (DST) to standard time (STD), the time from 02:00-02:59am is passed twice, first as the last hour of DST, and then as the first hour of STD.

Strictly speaking it is well-defined (in terms of UTC), which is the last minute of DST before passing to STD, so I'll match this here as well. What I ended up stumbling over is that current DST changes always happen between 2am and 3am, whereas the DST change in 1974 in the US [happened](https://www.timeanddate.com/time/change/usa/los-angeles?year=1974) from 1am to 2am.

h-vetinari:
OK, I think I'm done now from my side. This tests all the things I could think of as relevant.

wgtmac (approved):
Thanks @h-vetinari for all your effort!

h-vetinari:
Thanks a lot for the review and help here @wgtmac! 🙏 

Would you consider including this in the 2.0.1 release? I'm probably going to backport this in conda-forge if not, but it would be nicer if it could be included in the release here directly; seeing that it's just about checking a single environment variable, I think the risk here is very low.

wgtmac:
Yes, I have already ported it to branch-2.0 and will be released in 2.0.1.

h-vetinari:
Amazing, thanks a lot! 🤩

PR #1883, opened 2024-04-10
ORC-1683: Fix `instanceof` of BinaryStatisticsImpl merge method

cxzl25:
This PR aims to fix `instanceof` of `BinaryStatisticsImpl` merge method.

In [ORC-1542](https://issues.apache.org/jira/browse/ORC-1542), we modified part of instanceof, but `BinaryStatisticsImpl` was not modified because the merge method was written differently.

However, the current code is instanceof `BinaryColumnStatistics` and then explicitly cast `BinaryStatisticsImpl`, so it should be replaced by the new instanceof writing method (`Pattern Matching for instanceof`).

```java
if (other instanceof BinaryColumnStatistics) {
        BinaryStatisticsImpl bin = (BinaryStatisticsImpl) other;
```

GA

No

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
+++ b/java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
@@ -103,9 +103,7 @@ public class ColumnStatisticsImpl implements ColumnStatistics {
       if (other instanceof BooleanStatisticsImpl bkt) {
         trueCount += bkt.trueCount;
       } else {
-        if (isStatsExists() && trueCount != 0) {
-          throw new IllegalArgumentException("Incompatible merging of boolean column statistics");
-        }
+        throw new IllegalArgumentException("Incompatible merging of boolean column statistics");
       }
       super.merge(other);
     }
@@ -209,7 +207,6 @@ public class ColumnStatisticsImpl implements ColumnStatistics {
     @Override
     public void merge(ColumnStatisticsImpl other) {
       if (other instanceof CollectionColumnStatisticsImpl otherColl) {
-
         if(count == 0) {
           minimum = otherColl.minimum;
           maximum = otherColl.maximum;
@@ -223,9 +220,7 @@ public class ColumnStatisticsImpl implements ColumnStatistics {
         }
         sum += otherColl.sum;
       } else {
-        if (isStatsExists()) {
-          throw new IllegalArgumentException("Incompatible merging of collection column statistics");
-        }
+        throw new IllegalArgumentException("Incompatible merging of collection column statistics");
       }
       super.merge(other);
     }
@@ -398,9 +393,7 @@ public class ColumnStatisticsImpl implements ColumnStatistics {
           }
         }
       } else {
-        if (isStatsExists() && hasMinimum) {
-          throw new IllegalArgumentException("Incompatible merging of integer column statistics");
-        }
+        throw new IllegalArgumentException("Incompatible merging of integer column statistics");
       }
       super.merge(other);
     }
@@ -561,9 +554,7 @@ public class ColumnStatisticsImpl implements ColumnStatistics {
         }
         sum += dbl.sum;
       } else {
-        if (isStatsExists() && hasMinimum) {
-          throw new IllegalArgumentException("Incompatible merging of double column statistics");
-        }
+        throw new IllegalArgumentException("Incompatible merging of double column statistics");
       }
       super.merge(other);
     }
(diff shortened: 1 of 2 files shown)


dongjoon-hyun (changes requested):
Thank you, @cxzl25 . Could you find more like the previous PR? We don't want to make many PRs like this online style change.
- #1681

dongjoon-hyun:
In addition, this is not `Fixing anything`. We use `Fix` for bugs.

cxzl25:
> Could you find more like the previous PR? We don't want to make many PRs like this online style change.

I have checked all related instanceof and only ORC-1683 and ORC-1685 are left.

> In addition, this is not `Fixing anything`. We use `Fix` for bugs.

Since there could be a potential `ClassCastException` here, I've split it out to raise a separate PR.

dongjoon-hyun:
Oh, I got what you mean.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -993,8 +993,7 @@ public void updateBinary(byte[] bytes, int offset, int length,
 
     @Override
     public void merge(ColumnStatisticsImpl other) {
-      if (other instanceof BinaryColumnStatistics) {
-        BinaryStatisticsImpl bin = (BinaryStatisticsImpl) other;
```
dongjoon-hyun, line 997: Ya, it's interesting.

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -993,8 +993,7 @@ public void updateBinary(byte[] bytes, int offset, int length,
...
+      if (other instanceof BinaryStatisticsImpl bin) {
```
dongjoon-hyun, line 996: Actually, this PR is still not robust. This could be worse than before.

Instead of CCE, the corner case slips though `else` statement. And, If `isStatsExists() && sum != 0` is false, it propagates to `super.merge(other)`.

dongjoon-hyun (changes requested):
If this is a bug fix, we need a test coverage. Do you think you can add one?

Also, please address the following comment because we should avoid adding a new bug here.
- https://github.com/apache/orc/pull/1883/files?diff=unified&w=0#r1558985093

BTW, thank you!

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
cxzl25: It seems that other types of merge methods have similar problems. When itself is empty, it will merge statistics of incompatible types. I want to open a new JIRA to fix this problem. What do you think?

Here's a test that passes the test.
```java
    TypeDescription stringSchema = TypeDescription.createString();
    ColumnStatisticsImpl stringStats = ColumnStatisticsImpl.create(stringSchema);

    TypeDescription doubleSchema = TypeDescription.createDouble();
    ColumnStatisticsImpl doubleStats = ColumnStatisticsImpl.create(doubleSchema);

    stringStats.increment(3);
    stringStats.updateString(new Text("bob"));
    stringStats.updateString(new Text("david"));
    stringStats.updateString(new Text("charles"));

    doubleStats.merge(stringStats);
    assertEquals(3, ((DoubleColumnStatistics) doubleStats).getNumberOfValues());
```

dongjoon-hyun: Thank you for looking at that. How many JIRAs do you want to handle in total? If everything is due to the same pattern, I'd put it together in ORC-1683 (this).

java/core/src/java/org/apache/orc/impl/ColumnStatisticsImpl.java
```diff
@@ -993,8 +993,7 @@ public void updateBinary(byte[] bytes, int offset, int length,
...
         sum += bin.sum;
       } else {
```
dongjoon-hyun, line 998: BTW, to be clear, this PR changes like the following so that `else` statement need to handle more silently in the AS-IS form. I believe it's wrong. Please handle the type incompatible case before reaching out  this `else` statement. In other words, I guess we may need another `else` statement before this `else`.
```
- if (other instanceof BinaryColumnStatistics) {
+ if (other instanceof BinaryStatisticsImpl bin) {
```

cxzl25: How about just removing the if judgment in else?
For example, removed `if (isStatsExists() && sum != 0) {`

```java
      if (other instanceof BinaryStatisticsImpl bin) {
        sum += bin.sum;
      } else {
        throw new IllegalArgumentException("Incompatible merging of binary column statistics");
      }
```

dongjoon-hyun: Ur, could you elaborate your theoretic background of removal? How do you know that is not a regression and safe?

dongjoon-hyun (approved):

dongjoon-hyun:
Sorry for being late. Please feel free to merge if you are confident, @cxzl25 .

PR #1886, opened 2024-04-10
ORC-1686: [C++] Avoid using std::filesystem

wgtmac:
Remove std::filesystem and use OS API instead.

Apache Arrow C++ library is required to support old compilers like clang8. Since Apache ORC C++ library has used std::filesystem to check TZDB availability since 2.0.0, Apache Arrow requires to add more linking options for std::filesystem. See https://github.com/apache/arrow/pull/41023 for detail.

Passing CIs.

No.


dongjoon-hyun (approved):
According to the PR description, is this a regression? Do we need this at Apache ORC 2.0.1, @wgtmac ?

dongjoon-hyun:
I added `2.0.1` as the milestone. Thank you for doing this follow-up, @wgtmac .

dongjoon-hyun:
Merged to main/2.0.

wgtmac:
Thanks @dongjoon-hyun! Yes, it should be ported to 2.0.1 as well.

PR #1893, opened 2024-04-15
ORC-1688: [C++] Do not access TZDB if there is no timestamp type

ffacs:
1. Don't write writer's Time Zone name when there are no timestamp types.
2. Do not access TZDB if there is no timestamp type.

This patch could accelerate reading by reducing unnecessary IO.

All UT passed.

NO

### DIFF
--- a/c++/src/Timezone.cc
+++ b/c++/src/Timezone.cc
@@ -24,6 +24,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#include <atomic>
 #include <map>
 #include <sstream>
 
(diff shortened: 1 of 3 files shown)


c++/test/TestType.cc
```diff
@@ -490,4 +490,28 @@ namespace orc {
     *(footer.add_types()) = intType;
     checkProtoTypes(footer);
   }
+
+  TEST(TestType, testExistSubType) {
```
wgtmac, line 494: Could you also add a roundtrip test to write and read ORC file without timestamp type with an invalid TZDIR env set? Something like we did in `TEST(TestTimezone, testMissingTZDB)` and be careful about the timezone cache.

wgtmac (commented):
Thanks for adding the fix quickly! I've left some comments.

ffacs:
This implementation should be better than prior one.

wgtmac:
Nice fix! Will take a deeper look at it tomorrow.

c++/src/Timezone.cc
```diff
@@ -691,12 +683,68 @@ namespace orc {
...
+    std::string filename_;
+    mutable std::unique_ptr<TimezoneImpl> impl_;
+    mutable std::atomic<bool> initialized_;
+
+    TimezoneImpl* getImpl() const {
+      if (!initialized_) {
+        std::lock_guard<std::mutex> timezone_lock(timezone_mutex);
+        if (!initialized_) {
+          auto buffer = loadTZDB(filename_);
+          impl_.reset(new TimezoneImpl(filename_, std::move(buffer)));
```
wgtmac, line 705:
```suggestion
          impl_ = std::make_unique<TimezoneImpl>(filename_, std::move(buffer));
```

c++/src/Timezone.cc
```diff
@@ -691,12 +683,68 @@ namespace orc {
...
+
+  class LazyTimezone : public Timezone {
+   private:
+    std::string filename_;
+    mutable std::unique_ptr<TimezoneImpl> impl_;
+    mutable std::atomic<bool> initialized_;
+
+    TimezoneImpl* getImpl() const {
+      if (!initialized_) {
+        std::lock_guard<std::mutex> timezone_lock(timezone_mutex);
```
wgtmac, line 702: `timezone_mutex` is used to guard `TimezoneImpl::timezoneCache`. Should we use a much finer-grained mutex which is per LazyTimezone instance?

wgtmac (approved):
Nice fix! I have left some minor comments.

PR #1897, opened 2024-04-19
ORC-1695: Upgrade gson to 2.10.1

cxzl25:
This PR aims to upgrade `Gson` to `2.10.1`.

Jan 7, 2023

https://github.com/google/gson/releases/tag/gson-parent-2.10.1

Oct 25, 2022

https://github.com/google/gson/releases/tag/gson-parent-2.10

Aug 1, 2022

https://github.com/google/gson/releases/tag/gson-parent-2.9.1

GA

No

### DIFF
--- a/java/tools/src/test/resources/orc-file-dump.json
+++ b/java/tools/src/test/resources/orc-file-dump.json
@@ -2,7 +2,7 @@
   "fileName": "TestFileDump.testDump.orc",
   "fileVersion": "0.12",
   "writerVersion": "ORC_14",
-  "softwareVersion": "ORC Java 1.9.0-SNAPSHOT",
+  "softwareVersion": "ORC Java 2.1.0-SNAPSHOT",
   "numberOfRows": 21000,
   "compression": "ZLIB",
   "compressionBufferSize": 4096,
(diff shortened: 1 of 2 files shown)


java/tools/src/test/resources/orc-file-dump.json
```diff
@@ -461,48 +461,48 @@
               "numHashFunctions": 7,
               "bitCount": 9600,
               "popCount": 238,
-              "loadFactor": 0.024791667237877846,
-              "expectedFpp": 5.756256582500896E-12
+              "loadFactor": 0.024791667,
+              "expectedFpp": 5.7562566E-12
```
cxzl25, line 465: The `com.google.gson.stream.JsonWriter#value(float)` method is not provided in gson 1.9.0, but is provided in gson 1.9.1.

https://github.com/google/gson/pull/2130

This causes the test to fail after upgrading gson, because in version 2.9.0, the generated json float will be implicitly converted to double.

### json format
https://github.com/apache/orc/blob/bd2af54a7aa8865f220d862eab024c27983ab85e/java/tools/src/java/org/apache/orc/tools/JsonFileDump.java#L426-L432

### not json format
https://github.com/apache/orc/blob/bd2af54a7aa8865f220d862eab024c27983ab85e/java/tools/src/java/org/apache/orc/tools/FileDump.java#L719-L726

For other test files in non-json format, we can see that the loadFactor is float.

https://github.com/apache/orc/blob/bd2af54a7aa8865f220d862eab024c27983ab85e/java/tools/src/test/resources/orc-file-dump-bloomfilter2.out#L68-L69

dongjoon-hyun (approved):
+1, LGTM. Although it changes the value, it looks reasonable to me.

dongjoon-hyun:
Merged to main/2.0

PR #1916, opened 2024-04-28
ORC-1365: [C++] Use BlockBuffer to replace dataBuffer of  rawInputBuffer

luffy-zh:
The purpose of this PR is to replace the DataBuffer with BlockBuffer. This way the orc compressor can start with a small initial size and grow automatically as needed.

This patch uses BlockBuffer to replace the DataBuffer of class CompressionStream in order to solve the [issue](https://issues.apache.org/jira/browse/ORC-1365).

The UTs in TestBufferedOutputStream.cc and TestCompression.cc can cover this patch.

### DIFF
--- a/c++/src/BlockBuffer.cc
+++ b/c++/src/BlockBuffer.cc
@@ -44,7 +44,7 @@ namespace orc {
     if (blockIndex >= getBlockNumber()) {
       throw std::out_of_range("Block index out of range");
     }
-    return Block(blocks_[blockIndex], std::min(currentSize_ - blockIndex * blockSize_, blockSize_));
+    return Block(blocks_[blockIndex], getBlockSize(blockIndex));
   }
 
   BlockBuffer::Block BlockBuffer::getNextBlock() {
@@ -128,4 +128,37 @@ namespace orc {
       metrics->IOCount.fetch_add(ioCount);
     }
   }
+
+  uint64_t BlockBuffer::getBlockSize(uint64_t blockId) const {
+    return std::min(currentSize_ - blockId * blockSize_, blockSize_);
+  }
+
+  bool BlockBuffer::getBlockData(uint64_t blockIndex, void** data, uint64_t* size) const {
+    if (blockIndex >= getBlockNumber()) {
+      return false;
+    }
+    *size = getBlockSize(blockIndex);
+    if (*size == 0) {
+      return false;
+    }
+    *data = &(blocks_[blockIndex][0]);
+    return true;
+  }
+
+  bool BlockBuffer::requestBuffer(void** buffer, int* size) {
+    if (currentSize_ == currentCapacity_) {
+      char* memptr = memoryPool_.malloc(blockSize_);
+      blocks_.push_back(memptr);
+      currentCapacity_ += blockSize_;
+      currentSize_ += blockSize_;
+      *buffer = blocks_.back();
+      *size = static_cast<int>(blockSize_);
+    } else {
+      *buffer = &(blocks_[currentSize_ / blockSize_][currentSize_ % blockSize_]);
+      *size = static_cast<int>(blockSize_ - currentSize_ % blockSize_);
+      currentSize_ += static_cast<uint64_t>(*size);
+    }
+    return true;
+  }
+
 }  // namespace orc
--- a/c++/src/BlockBuffer.cc
+++ b/c++/src/BlockBuffer.cc
@@ -128,5 +128,4 @@ namespace orc {
       metrics->IOCount.fetch_add(ioCount);
     }
   }
-
 }  // namespace orc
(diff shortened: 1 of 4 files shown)


c++/src/Compression.cc
```diff
@@ -260,31 +270,53 @@ namespace orc {
       throw std::runtime_error("Failed to reset inflate.");
     }
 
-    strm_.avail_in = static_cast<unsigned int>(bufferSize);
-    strm_.next_in = rawInputBuffer.data();
+    // reset output buffer
+    if (outputPosition != 0) {
+      BufferedOutputStream::BackUp(outputSize - outputPosition);
```
ffacs, line 275: Sorry but I don't get the intention we reset output buffer here.

c++/src/Compression.cc
```diff
@@ -187,45 +161,81 @@ namespace orc {
...
                                        WriterMetrics* metrics)
-      : CompressionStreamBase(outStream, compressionLevel, capacity, blockSize, pool, metrics) {
+      : CompressionStreamBase(outStream, compressionLevel, capacity, blockSize, pool, metrics),
+        rawInputBuffer(pool, blockSize) {
     // PASS
   }
 
   bool CompressionStream::Next(void** data, int* size) {
-    if (bufferSize != 0) {
+    if (rawInputBuffer.size() != 0) {
```
ffacs, line 213: Maybe we can leverage the flexibility of BlockBuffer Here. That is, we can do compress only when the size of rawInputBuffer reachs compression block buffer size.

c++/src/Compression.cc
```diff
@@ -161,6 +133,8 @@ namespace orc {
       offset += currentSize;
       outputPosition += currentSize;
     }
+    BufferedOutputStream::BackUp(outputSize - outputPosition);
```
ffacs, line 136: Why should we do this?

c++/src/BlockBuffer.hh
```diff
@@ -106,18 +106,39 @@ namespace orc {
...
+     */
+    bool getBlockData(uint64_t blockIndex, void** data, uint64_t* size) const;  // add ut
+
+    /**
+     * Get next availiable memory block
+     * @param data the pointer to the block data
+     * @param size the size of the block data
+     * @return true if the block data is successfully retrieved
+     */
+    bool requestBuffer(void** buffer, int* size);  // add ut
```
ffacs, line 141: ditto

c++/src/BlockBuffer.hh
```diff
@@ -106,18 +106,39 @@ namespace orc {
...
     void writeTo(OutputStream* output, WriterMetrics* metrics);
+
+    /**
+     * Get the block data by block index
+     * @param blockIndex the index of blocks
+     * @param data the pointer to the block data
+     * @param size the size of the block data
+     * @return true if the block data is successfully retrieved
+     */
+    bool getBlockData(uint64_t blockIndex, void** data, uint64_t* size) const;  // add ut
```
ffacs, line 133: Please remove the comment at the end of the line.

ffacs (changes requested):
Thank you @luffy-zh, I've left some comments.

c++/src/BlockBuffer.hh
wgtmac: Isn't `Block getBlock(uint64_t blockIndex) const;` at line 81 doing the same job?

c++/src/BlockBuffer.hh
wgtmac: Isn't `Block getNextBlock();` at line 91 doing the same thing?

c++/src/BlockBuffer.hh
```diff
@@ -106,18 +106,39 @@ namespace orc {
     }
 
     void resize(uint64_t size);
+
+    uint64_t getBlockSize(uint64_t blockId) const;
```
wgtmac, line 110: Please add some comments

c++/src/Compression.cc
luffy-zh: In the steam compressor, we will call writeData() multiple times, which requires the BufferedOutputStream to handle any remaining bytes correctly。

c++/src/BlockBuffer.hh
luffy-zh: never mind, I'll delete it.

c++/src/Compression.cc
luffy-zh: Great advice, I'll address it in my next work project.

c++/src/BlockBuffer.hh
luffy-zh: I'll delete this redundant function.

c++/src/Compression.cc
luffy-zh: never mind, I've delete it

c++/src/BlockBuffer.cc
```diff
@@ -128,4 +128,5 @@ namespace orc {
       metrics->IOCount.fetch_add(ioCount);
     }
   }
+
```
wgtmac, line 131:
```suggestion
```

ffacs:
I have no extra comments other than removing the blank line.

PR #1931, opened 2024-05-12
ORC-1389: [C++] Support schema evolution from string group to numeric/string group

ffacs:
Support conversion from

{string, char, varchar}
to
{boolean, byte, short, int, long, float, double, string, char, varchar}

To support schema evolution on c++ side.

UT passed

NO

### DIFF
--- a/c++/src/ConvertColumnReader.cc
+++ b/c++/src/ConvertColumnReader.cc
@@ -722,12 +722,17 @@ namespace orc {
     void convertToInteger(ReadTypeBatch& dstBatch, const StringVectorBatch& srcBatch,
                           uint64_t idx) {
       int64_t longValue = 0;
-      try {
+      if (throwOnOverflow) {
         longValue = std::stoll(std::string(srcBatch.data[idx], srcBatch.length[idx]));
-      } catch (...) {
-        handleOverflow<std::string, ReadType>(dstBatch, idx, throwOnOverflow);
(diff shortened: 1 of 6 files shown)


c++/src/ConvertColumnReader.cc
```diff
@@ -694,6 +695,112 @@ namespace orc {
...
+      }
+    }
+
+   private:
+    void convertToInteger(ReadTypeBatch& dstBatch, const StringVectorBatch& srcBatch,
+                          uint64_t idx) {
+      int64_t longValue = 0;
+      try {
+        longValue = std::stoll(std::string(srcBatch.data[idx], srcBatch.length[idx]));
+      } catch (...) {
```
wgtmac, line 727: Should we explicitly catch `std::invalid_argument` and `std::out_of_range` so that we can get a clearer error message?

wgtmac (approved):
+1

Left a minor comment

c++/src/ConvertColumnReader.cc
ffacs: I throw the original exception when `throwOnOverflow` is true, is that ok?

wgtmac: It seems that the original code looks better ...

wgtmac:
Feel free to merge this when you are ready.

PR #1932, opened 2024-05-13
ORC-1711:[C++] Introduce a memory block size parameter for write option

luffy-zh:
What changes were proposed in this pull request?
1.  Add the memory block size parameter to the writer option, which initializing the compressed input buffer block size
2. The compressed stream will retain the input buffer until the input buffer size reaches the compression block size, allowing the compressed stream to start with a minimal initial memory footprint.

Why are the changes needed?
This code segment distinguishes between the compression block size and the input buffer size to solve the [issue](https://issues.apache.org/jira/browse/ORC-1711).

How was this patch tested?
The UTs in TestCompression.cc and TestWriter.cc can cover this patch.

### DIFF
--- a/c++/include/orc/Writer.hh
+++ b/c++/include/orc/Writer.hh
@@ -266,17 +266,17 @@ namespace orc {
     bool getUseTightNumericVector() const;
 
     /**
-     * Set the initial capacity of output buffer in the class BufferedOutputStream.
+     * Set the initial block size of input buffer in the class CompressionStream.
      * Each column contains one or more BufferOutputStream depending on its type,
      * and these buffers will automatically expand when more memory is required.
      */
-    WriterOptions& setOutputBufferCapacity(uint64_t capacity);
+    WriterOptions& setMemoryBlockSize(uint64_t capacity);
 
     /**
-     * Get the initial capacity of output buffer in the class BufferedOutputStream.
-     * @return if not set, return default value which is 1 MB.
+     * Get the initial block size of input buffer in the class CompressionStream.
+     * @return if not set, return default value which is 64 KB.
      */
-    uint64_t getOutputBufferCapacity() const;
+    uint64_t getMemoryBlockSize() const;
   };
 
   class Writer {
--- a/c++/src/Compression.cc
+++ b/c++/src/Compression.cc
@@ -65,6 +65,7 @@ namespace orc {
       return true;
     }
     virtual uint64_t getSize() const override;
+    virtual uint64_t getRawInputBufferSize() const override = 0;
 
    protected:
     void writeData(const unsigned char* data, int size);
@@ -96,6 +97,9 @@ namespace orc {
     // Compression block header pointer array
     static const uint32_t HEADER_SIZE = 3;
     std::array<char*, HEADER_SIZE> header;
+
+    // Compression block size
+    uint64_t compressBlockSize;
   };
 
   CompressionStreamBase::CompressionStreamBase(OutputStream* outStream, int compressionLevel,
@@ -106,7 +110,8 @@ namespace orc {
         outputBuffer(nullptr),
         bufferSize(0),
         outputPosition(0),
-        outputSize(0) {
+        outputSize(0),
+        compressBlockSize(capacity) {
     // init header pointer array
     header.fill(nullptr);
   }
@@ -162,6 +167,9 @@ namespace orc {
     virtual void BackUp(int count) override;
     virtual void suppress() override;
     virtual uint64_t flush() override;
+    uint64_t getRawInputBufferSize() const override {
+      return rawInputBuffer.size();
+    }
 
    protected:
     // return total compressed size
@@ -169,6 +177,9 @@ namespace orc {
 
     // Buffer to hold uncompressed data until user calls Next()
     BlockBuffer rawInputBuffer;
+
+    // compress with raw fallback
+    void compressWithRawFallback();
   };
 
   void CompressionStream::BackUp(int count) {
@@ -181,11 +192,7 @@ namespace orc {
   }
 
   uint64_t CompressionStream::flush() {
-    void* data;
-    int size;
-    if (!Next(&data, &size)) {
-      throw std::runtime_error("Failed to flush compression buffer.");
-    }
+    compressWithRawFallback();
     BufferedOutputStream::BackUp(outputSize - outputPosition);
     rawInputBuffer.resize(0);
     outputSize = outputPosition = 0;
@@ -207,8 +214,8 @@ namespace orc {
     // PASS
   }
 
-  bool CompressionStream::Next(void** data, int* size) {
-    if (rawInputBuffer.size() != 0) {
+  void CompressionStream::compressWithRawFallback() {
+     if (rawInputBuffer.size() != 0) {
       ensureHeader();
 
       uint64_t preSize = getSize();
@@ -231,6 +238,13 @@ namespace orc {
         writeHeader(totalCompressedSize, false);
       }
       rawInputBuffer.resize(0);
+    }
+  }
+
+  bool CompressionStream::Next(void** data, int* size) {
+    // triggle compress when rawInputBuffer is reach the capacity
+    if (rawInputBuffer.size() == compressBlockSize) {
+      compressWithRawFallback();
     }
 
     auto block = rawInputBuffer.getNextBlock();
@@ -915,6 +929,9 @@ namespace orc {
     virtual void BackUp(int count) override;
     virtual uint64_t flush() override;
     virtual std::string getName() const override = 0;
+    uint64_t getRawInputBufferSize() const override {
+      return rawInputBuffer.size();
+    }
 
    protected:
     // compresses a block and returns the compressed size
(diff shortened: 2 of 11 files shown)


ffacs (commented):
@luffy-zh Thank you for making the PR.
Could you explain why we need `outputStream->getRawInputBufferSize()` to record position? it seems that `flushedSize` and `bufferPosition` wasn't changed.

luffy-zh:
> @luffy-zh Thank you for making the PR. Could you explain why we need `outputStream->getRawInputBufferSize()` to record position? it seems that `flushedSize` and `bufferPosition` were't changed.

When we call recordPosition(), we need to record the output buffer length and input buffer length. We have multiple blocks in the input buffer, 'bufferPosition' only records the effective length of the last block, that's why we need to use getRawInputBufferSize() in the outputStream.

c++/src/Compression.cc
```diff
@@ -96,6 +97,9 @@ namespace orc {
     // Compression block header pointer array
     static const uint32_t HEADER_SIZE = 3;
     std::array<char*, HEADER_SIZE> header;
+
+    // Compression block size
+    uint64_t compressBlockSize;
```
ffacs, line 102: compressionBlockSize

c++/src/Compression.cc
```diff
@@ -207,7 +214,7 @@ namespace orc {
     // PASS
   }
```
ffacs, line 215:
```cpp
 CompressionStream::CompressionStream(OutputStream* outStream, int compressionLevel,
                                       uint64_t compressionBlockSize, uint64_t memoryBlockSize, MemoryPool& pool,
                                       WriterMetrics* metrics) 
```
capacity -> compressionBlockSize, blockSize->memoryBlockSize
Same for other compression streams.

c++/src/Compression.cc
```diff
@@ -232,6 +239,13 @@ namespace orc {
       }
       rawInputBuffer.resize(0);
     }
+  }
+
+  bool CompressionStream::Next(void** data, int* size) {
+    // triggle compress when rawInputBuffer is reach the capacity
+    if (rawInputBuffer.size() == compressBlockSize) {
```
ffacs, line 246: We should make sure `compressionBlockSize` is a multiple of `memoryBlockSize`.

c++/src/Writer.cc
```diff
@@ -45,11 +45,11 @@ namespace orc {
     std::string timezone;
     WriterMetrics* metrics;
     bool useTightNumericVector;
-    uint64_t outputBufferCapacity;
+    uint64_t memoryBlockSize;
 
     WriterOptionsPrivate() : fileVersion(FileVersion::v_0_12()) {  // default to Hive_0_12
       stripeSize = 64 * 1024 * 1024;                               // 64M
-      compressionBlockSize = 64 * 1024;                            // 64K
+      compressionBlockSize = 256 * 1024;                           // 256K
```
ffacs, line 52: Why do we make a change here?

c++/src/Writer.cc
```diff
@@ -355,7 +355,7 @@ namespace orc {
     // compression stream for stripe footer, file footer and metadata
     compressionStream_ =
         createCompressor(options_.getCompression(), outStream_, options_.getCompressionStrategy(),
-                         options_.getOutputBufferCapacity(), options_.getCompressionBlockSize(),
+                         options_.getMemoryBlockSize(), options_.getCompressionBlockSize(),
```
ffacs, line 358: Parameters do not match declaration.

c++/src/io/OutputStream.cc
```diff
@@ -98,6 +98,12 @@ namespace orc {
     dataBuffer_->resize(0);
   }
 
+  uint64_t BufferedOutputStream::getRawInputBufferSize() const {
+    // we're unable to determine the size of the raw input buffer
+    // simply return 0
+    return 0;
```
ffacs, line 104: Throw a exception would be better.

ffacs (changes requested):

ffacs:
> > @luffy-zh Thank you for making the PR. Could you explain why we need `outputStream->getRawInputBufferSize()` to record position? it seems that `flushedSize` and `bufferPosition` were't changed.
> 
> When we call recordPosition(), we need to record the output buffer length and input buffer length. We have multiple blocks in the input buffer, 'bufferPosition' only records the effective length of the last block, that's why we need to use getRawInputBufferSize() in the outputStream.

Make sense.

c++/include/orc/Writer.hh
```diff
@@ -266,17 +266,15 @@ namespace orc {
     bool getUseTightNumericVector() const;
 
     /**
-     * Set the initial capacity of output buffer in the class BufferedOutputStream.
-     * Each column contains one or more BufferOutputStream depending on its type,
-     * and these buffers will automatically expand when more memory is required.
+     * Set the initial block size of input buffer in the class CompressionStream.
      */
-    WriterOptions& setOutputBufferCapacity(uint64_t capacity);
+    WriterOptions& setMemoryBlockSize(uint64_t capacity);
```
wgtmac, line 271: We cannot remove or rename any public API. This will immediately break downstream code once they upgrade to this version. If we think these APIs are not used any more, we'd better add `[[deprecated]]` attribute.

c++/src/Writer.cc
wgtmac: We should not change the default behavior if unnecessary.

c++/src/io/OutputStream.cc
wgtmac: +1

wgtmac (changes requested):

c++/src/Compression.cc
luffy-zh: I'll add some checks in the writer init function.

c++/src/Writer.cc
luffy-zh: I will cancel these changes

c++/src/Writer.cc
```diff
@@ -352,11 +363,18 @@ namespace orc {
 
     useTightNumericVector_ = opts.getUseTightNumericVector();
 
+    if (options_.getCompressionBlockSize() % options_.getMemoryBlockSize() != 0) {
+      std::cerr << options_.getCompressionBlockSize() << " : " << options_.getMemoryBlockSize()
```
ffacs: Del this

ffacs (changes requested):

ffacs (approved):
LGTM. Thank you @luffy-zh .

c++/src/Compression.hh
```diff
@@ -45,12 +45,10 @@ namespace orc {
...
    */
-  std::unique_ptr<BufferedOutputStream> createCompressor(CompressionKind kind,
-                                                         OutputStream* outStream,
-                                                         CompressionStrategy strategy,
-                                                         uint64_t bufferCapacity,
-                                                         uint64_t compressionBlockSize,
-                                                         MemoryPool& pool, WriterMetrics* metrics);
+  std::unique_ptr<BufferedOutputStream> createCompressor(
+      CompressionKind kind, OutputStream* outStream, CompressionStrategy strategy,
+      uint64_t bufferCapacity, uint64_t compressionBlockSize, uint64_t memoryBlockSize,
```
wgtmac, line 50: Please update the comment above to be in sync.

wgtmac: Now it seems that we have different concepts mixed here. It would be helpful to explain what are the distinctions among these buffers and blocks.

c++/src/Compression.cc
```diff
@@ -1203,8 +1237,8 @@ namespace orc {
       case CompressionKind_ZLIB: {
         int level =
             (strategy == CompressionStrategy_SPEED) ? Z_BEST_SPEED + 1 : Z_DEFAULT_COMPRESSION;
-        return std::make_unique<ZlibCompressionStream>(outStream, level, bufferCapacity,
-                                                       compressionBlockSize, pool, metrics);
+        return std::make_unique<ZlibCompressionStream>(
+            outStream, level, bufferCapacity, compressionBlockSize, memoryBlockSize, pool, metrics);
```
wgtmac, line 1241: If `memoryBlockSize` is only used for streaming compressor, we should rename it to be clear. However, I do think we need to support block compressor as well as ZSTD (block-based) is the default one. This can be a separate PR though.

c++/src/Compression.cc
```diff
@@ -232,6 +244,22 @@ namespace orc {
...
       rawInputBuffer.resize(0);
     }
+  }
+
+  bool CompressionStream::Next(void** data, int* size) {
+    if (rawInputBuffer.size() > compressionBlockSize) {
+      std::stringstream ss;
+      ss << "uncompressed data size " << rawInputBuffer.size() << " is larger than block size "
+         << compressionBlockSize
+         << ". compressionBlockSize should be set equal to multiply of "
```
wgtmac, line 254:
```suggestion
         << ". compressionBlockSize should be set equal to multiple of "
```

c++/src/Compression.cc
```diff
@@ -232,6 +244,22 @@ namespace orc {
...
+            "memoryBlockSize";
+      throw std::logic_error(ss.str());
```
wgtmac, line 256: Should we add a new Exception type?

c++/src/Compression.cc
```diff
@@ -200,14 +210,16 @@ namespace orc {
...
-      : CompressionStreamBase(outStream, compressionLevel, capacity, blockSize, pool, metrics),
-        rawInputBuffer(pool, blockSize) {
+      : CompressionStreamBase(outStream, compressionLevel, capacity, compressionBlockSize,
+                              memoryBlockSize, pool, metrics),
+        rawInputBuffer(pool, memoryBlockSize) {
     // PASS
   }
 
-  bool CompressionStream::Next(void** data, int* size) {
+  void CompressionStream::compressWithRawFallback() {
```
wgtmac, line 222: What does raw fallback mean? Should we just name it `compressInternal`?

c++/src/Compression.cc
```diff
@@ -232,6 +244,22 @@ namespace orc {
...
+    if (rawInputBuffer.size() > compressionBlockSize) {
+      std::stringstream ss;
+      ss << "uncompressed data size " << rawInputBuffer.size() << " is larger than block size "
+         << compressionBlockSize
+         << ". compressionBlockSize should be set equal to multiply of "
+            "memoryBlockSize";
+      throw std::logic_error(ss.str());
+    }
+
+    // triggle compress when rawInputBuffer is reach the capacity
```
wgtmac, line 259:
```suggestion
    // compress data in the rawInputBuffer when it is full
```

c++/src/Compression.cc
wgtmac: BTW, we should only put useful error message here. It seems the error happens when the size is greater than the threshold and we should not say anything about the multiple thing. That is something we need to check when creating the compression stream.

c++/src/Compression.cc
```diff
@@ -155,20 +162,27 @@ namespace orc {
...
+    }
 
    protected:
     // return total compressed size
     virtual uint64_t doStreamingCompression() = 0;
 
     // Buffer to hold uncompressed data until user calls Next()
     BlockBuffer rawInputBuffer;
+
+    // compress with raw fallback
```
wgtmac, line 184: The comment seems unnecessary

c++/include/orc/Writer.hh
```diff
@@ -277,6 +277,17 @@ namespace orc {
      * @return if not set, return default value which is 1 MB.
      */
     uint64_t getOutputBufferCapacity() const;
+
+    /**
+     * Set the initial block size of input buffer in the class CompressionStream.
```
wgtmac, line 282: Could we add more comment to let readers know what's the difference with `setOutputBufferCapacity` above? Perhaps we also need to modify the comment in setOutputBufferCapacity. Users rely on these public methods to tune their use cases.

wgtmac (changes requested):

c++/src/Compression.cc
luffy-zh: Make sense, I'll delete those irrelevant comments

c++/src/Compression.cc
luffy-zh: Good advice. I will do this in my next work if necessary

c++/src/Compression.cc
luffy-zh: Maybe we can replace all std::logic_error in CompressionStream with new exception types.  I'll address this issue in my next work item.

c++/include/orc/Writer.hh
luffy-zh: Thanks for the reminder, more comments have been added here.

PR #1944, opened 2024-05-30
ORC-1720: [C++] Unified compressor/decompressor exception types

luffy-zh:
Add new exception types to the decompressor/compressor.

The current implementation of the compressor/decompressor uses a variety of std::exception types, which can perplex users. We can enhance users' debugging efficiency by adopting a unified exception class.

The tests in TestWriter.cc can cover this patch.

NO

### DIFF
--- a/c++/include/orc/Exceptions.hh
+++ b/c++/include/orc/Exceptions.hh
@@ -67,6 +67,28 @@ namespace orc {
     SchemaEvolutionError(const SchemaEvolutionError&);
     SchemaEvolutionError& operator=(const SchemaEvolutionError&) = delete;
   };
+
+  class CompressionError : public std::runtime_error {
+   public:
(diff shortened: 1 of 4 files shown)


c++/include/orc/Exceptions.hh
```diff
@@ -67,6 +67,28 @@ namespace orc {
...
+    explicit CompressionError(const std::string& whatArg);
+    explicit CompressionError(const char* whatArg);
+    ~CompressionError() noexcept override;
+    CompressionError(const CompressionError&);
+
+   private:
+    CompressionError& operator=(const CompressionError&);
+  };
+
+  class DecompressionError : public std::runtime_error {
```
wgtmac, line 82: What about just using `CompressionError` for both compression and decompression?

ffacs (approved):
LGTM. Thank you @luffy-zh

PR #1946, opened 2024-05-31
ORC-1724: JsonFileDump utility should print user metadata

cxzl25:
This PR aims to implement JsonFileDump to output user metadata.

ORC-223 implements the output of user metadata in non-json format, but the json format does not output user metadata.

add UT

No


dongjoon-hyun (approved):
+1, LGTM. Thank you for this tool improvement, @cxzl25 .
Merged to main/2.0.

PR #1960, opened 2024-06-24
ORC-1741: Respect decimal reader isRepeating flag

cxzl25:
Decimal type, when `isRepeating` itself is false, do not try to change it.

https://github.com/apache/hive/pull/5218#discussion_r1647367003

[ORC-1266](https://issues.apache.org/jira/browse/ORC-1266): DecimalColumnVector resets the isRepeating flag in the nextVector method

Add UT

No

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
@@ -1551,6 +1551,7 @@ public class TreeReaderFactory {
       HiveDecimalWritable[] vector = result.vector;
       HiveDecimalWritable decWritable;
       if (result.noNulls) {
+        boolean preIsRepeating = result.isRepeating;
         result.isRepeating = true;
         for (int r = 0; r < batchSize; ++r) {
           decWritable = vector[r];
@@ -1562,7 +1563,11 @@ public class TreeReaderFactory {
           }
           setIsRepeatingIfNeeded(result, r);
         }
+        if (!preIsRepeating && result.isRepeating) {
+          result.isRepeating = preIsRepeating;
+        }
       } else if (!result.isRepeating || !result.isNull[0]) {
+        boolean preIsRepeating = result.isRepeating;
         result.isRepeating = true;
         for (int r = 0; r < batchSize; ++r) {
           if (!result.isNull[r]) {
@@ -1575,6 +1580,9 @@ public class TreeReaderFactory {
             }
           }
           setIsRepeatingIfNeeded(result, r);
+        }
+        if (!preIsRepeating && result.isRepeating) {
+          result.isRepeating = preIsRepeating;
         }
       }
     }
@@ -1595,6 +1603,7 @@ public class TreeReaderFactory {
       HiveDecimalWritable[] vector = result.vector;
       HiveDecimalWritable decWritable;
       if (result.noNulls) {
+        boolean preIsRepeating = result.isRepeating;
         result.isRepeating = true;
         int previousIdx = 0;
         for (int r = 0; r != filterContext.getSelectedSize(); ++r) {
@@ -1612,8 +1621,12 @@ public class TreeReaderFactory {
           setIsRepeatingIfNeeded(result, idx);
           previousIdx = idx + 1;
         }
+        if (!preIsRepeating && result.isRepeating) {
+          result.isRepeating = preIsRepeating;
+        }
         skipStreamRows(batchSize - previousIdx);
       } else if (!result.isRepeating || !result.isNull[0]) {
+        boolean preIsRepeating = result.isRepeating;
         result.isRepeating = true;
         int previousIdx = 0;
         for (int r = 0; r != filterContext.getSelectedSize(); ++r) {
@@ -1633,6 +1646,9 @@ public class TreeReaderFactory {
           setIsRepeatingIfNeeded(result, idx);
           previousIdx = idx + 1;
         }
+        if (!preIsRepeating && result.isRepeating) {
+          result.isRepeating = preIsRepeating;
+        }
         skipStreamRows(countNonNullRowsInRange(result.isNull, previousIdx, batchSize));
       }
     }
@@ -1651,13 +1667,18 @@ public class TreeReaderFactory {
       // read the scales
       scaleReader.nextVector(result, scratchScaleVector, batchSize);
       if (result.noNulls) {
+        boolean preIsRepeating = result.isRepeating;
         result.isRepeating = true;
         for (int r = 0; r < batchSize; ++r) {
           final long scaleFactor = powerOfTenTable[scale - scratchScaleVector[r]];
           result.vector[r] = SerializationUtils.readVslong(valueStream) * scaleFactor;
           setIsRepeatingIfNeeded(result, r);
         }
+        if (!preIsRepeating && result.isRepeating) {
+          result.isRepeating = preIsRepeating;
+        }
       } else if (!result.isRepeating || !result.isNull[0]) {
+        boolean preIsRepeating = result.isRepeating;
         result.isRepeating = true;
         for (int r = 0; r < batchSize; ++r) {
           if (!result.isNull[r]) {
@@ -1665,6 +1686,9 @@ public class TreeReaderFactory {
             result.vector[r] = SerializationUtils.readVslong(valueStream) * scaleFactor;
           }
           setIsRepeatingIfNeeded(result, r);
+        }
+        if (!preIsRepeating && result.isRepeating) {
+          result.isRepeating = preIsRepeating;
         }
       }
       result.precision = (short) precision;
@@ -1686,6 +1710,7 @@ public class TreeReaderFactory {
       // Read all the scales
       scaleReader.nextVector(result, scratchScaleVector, batchSize);
       if (result.noNulls) {
+        boolean preIsRepeating = result.isRepeating;
         result.isRepeating = true;
         int previousIdx = 0;
         for (int r = 0; r != filterContext.getSelectedSize(); r++) {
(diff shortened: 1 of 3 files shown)


difin:
Hi @cxzl25, I did a test on my local env: backported this patch into Orc branch 1.9, built Hive with it then run the Hive query test that had been failing before and it passed. Can you please create a new Orc release from branch 1.9 so that we could use it in Hive? This is because we can't use Orc 2.0 or later because it requires Java 17+ and Hive still support Java 8.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
```diff
@@ -1562,7 +1563,11 @@ private void nextVector(DecimalColumnVector result,
           }
           setIsRepeatingIfNeeded(result, r);
         }
+        if (!preIsRepeating && result.isRepeating) {
```
deniskuzZ, line 1566: why don't you move this logic inside of `setIsRepeatingIfNeeded` instead of scattering it all over the code?

deniskuzZ: also a question from the Hive Orc PR:
where do we handle repeated nulls? it seems that setIsRepeatingIfNeeded is only called when result.noNulls || !result.isNull[0]

williamhyun (commented):
If this PR is ready, I can merge it to 1.9.4 as a release manager.

cc: @wgtmac @guiyanakuang @deshanxiao

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
cxzl25: This is just a draft PR, it should not be considered a bug.

When we convert the same value of the same batch size into `isRepeating=true`, the two should be equivalent.

I'm not sure why this behavior will affect the calculation results of Hiive. I repeatedly debugged this in Hive, but I didn't find the root cause.

```java
    DecimalColumnVector v1 = new DecimalColumnVector(1024, 10, 2);
    v1.isRepeating = true;
    v1.vector[0] = new HiveDecimalWritable("1.234");

    DecimalColumnVector v2 = new DecimalColumnVector(1024, 10, 2);
    for (int i = 0; i < 1024; i++) {
      v2.vector[i] = new HiveDecimalWritable("1.234");
    }

    StringBuilder sb1 = new StringBuilder();
    for (int i = 0; i < 1024; i++) {
      v1.stringifyValue(sb1, i);
    }
    StringBuilder sb2 = new StringBuilder();
    for (int i = 0; i < 1024; i++) {
      v2.stringifyValue(sb2, i);
    }
    System.out.println(sb1.toString().equals(sb2.toString()));
```

guiyanakuang:
@williamhyun, Thank you for the reminder. I reviewed the code again and found the root cause of the issue. I believe we need to update this patch to ensure a thorough fix for this bug.

guiyanakuang:
```
    private void setIsRepeatingIfNeeded(DecimalColumnVector result, int index) {
      if (result.isRepeating && index > 0 && (!result.vector[0].equals(result.vector[index]) ||
          result.isNull[0] != result.isNull[index])) {
        result.isRepeating = false;
      }
    }
```

`result.vector[0].equals(result.vector[index])`  We cannot directly perform this check because:

```
public DecimalColumnVector(int size, int precision, int scale) {
  super(Type.DECIMAL, size);
  this.precision = (short) precision;
  this.scale = (short) scale;
  vector = new HiveDecimalWritable[size];
  for (int i = 0; i < size; i++) {
    vector[i] = new HiveDecimalWritable(0);  // Initially zero.
  }
}
```

The initial value `new HiveDecimalWritable(0)` will be used, and we can only use `result.vector[0]` for comparison after ensuring `result.isNull[0] == false`.

cc @cxzl25

dongjoon-hyun:
Which version is this PR targeting, @cxzl25 ?

dongjoon-hyun:
If this is not a bug, please set the milestone of this PR to ORC 2.1.0. Otherwise, please set the milestone to 1.9.5 and change the JIRA issue type from `Improvement` to `Bug`.

> This is just a draft PR, it should not be considered a bug.

Screenshot 2024-07-18 at 22 02 27

cxzl25:
> Which version is this PR targeting

If the introduction of ORC-1266 causes Hive calculation results to be inaccurate, in order to ensure that Hive can use the 1.9.x version, I hope to merge it into the 1.9 branch.

java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java
cxzl25: `DoubleTreeReader` and `FloatTreeReader` also have similar behavior. If the entire batch size is repeated, it will set `isRepeating=true`, which is consistent with the behavior of ORC-1266.

---


DoubleTreeReader

https://github.com/apache/orc/blob/5e6a4cddf8599e351338030a2ceae919308f4848/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java#L1034-L1046

FloatTreeReader

https://github.com/apache/orc/blob/5e6a4cddf8599e351338030a2ceae919308f4848/java/core/src/java/org/apache/orc/impl/TreeReaderFactory.java#L818-L830

---
Some tests

```java

  @Test
  public  void testDoubleIsRepeatingFlag() throws IOException {
    Configuration conf = new Configuration();
    FileSystem fs = FileSystem.get(conf);
    Path testFilePath = new Path(workDir, "testDoubleIsRepeatingFlag.orc");
    fs.delete(testFilePath, true);

    Configuration doubleConf = new Configuration(conf);
    doubleConf.set(OrcConf.STRIPE_ROW_COUNT.getAttribute(), "1024");
    doubleConf.set(OrcConf.ROWS_BETWEEN_CHECKS.getAttribute(), "1");
    String typeStr = "double";
    TypeDescription schema = TypeDescription.fromString("struct<col1:" + typeStr + ">");
    Writer w = OrcFile.createWriter(testFilePath, OrcFile.writerOptions(doubleConf).setSchema(schema));

    VectorizedRowBatch b = schema.createRowBatch();
    DoubleColumnVector f1 = (DoubleColumnVector) b.cols[0];
    for (int i = 0; i < 1024; i++) {
      f1.vector[i] = -119.4594594595D;
    }
    b.size = 1024;
    w.addRowBatch(b);

    b.reset();
    for (int i = 0; i < 1024; i++) {
      f1.vector[i] = 9318.4351351351D;
    }
    b.size = 1024;
    w.addRowBatch(b);

    b.reset();
    for (int i = 0; i < 1024; i++) {
      f1.vector[i] = -4298.1513513514D;
    }
    b.size = 1024;
    w.addRowBatch(b);

    b.reset();
    w.close();

    Reader.Options options = new Reader.Options();
    try (Reader reader = OrcFile.createReader(testFilePath, OrcFile.readerOptions(conf));
         RecordReader rows = reader.rows(options)) {
      VectorizedRowBatch batch = schema.createRowBatch();

      rows.nextBatch(batch);
      assertEquals(1024, batch.size);
      assertTrue(batch.cols[0].isRepeating);

      rows.nextBatch(batch);
      assertEquals(1024, batch.size);
      assertTrue(batch.cols[0].isRepeating);

      rows.nextBatch(batch);
      assertEquals(1024, batch.size);
      assertTrue(batch.cols[0].isRepeating);

    }
  }

  @Test
  public  void testFloatIsRepeatingFlag() throws IOException {
    Configuration conf = new Configuration();
    FileSystem fs = FileSystem.get(conf);
    Path testFilePath = new Path(workDir, "testFloatIsRepeatingFlag.orc");
    fs.delete(testFilePath, true);

    Configuration floatConf = new Configuration(conf);
    floatConf.set(OrcConf.STRIPE_ROW_COUNT.getAttribute(), "1024");
    floatConf.set(OrcConf.ROWS_BETWEEN_CHECKS.getAttribute(), "1");
    String typeStr = "float";
    TypeDescription schema = TypeDescription.fromString("struct<col1:" + typeStr + ">");
    Writer w = OrcFile.createWriter(testFilePath, OrcFile.writerOptions(floatConf).setSchema(schema));

    VectorizedRowBatch b = schema.createRowBatch();
    DoubleColumnVector f1 = (DoubleColumnVector) b.cols[0];
    for (int i = 0; i < 1024; i++) {
      f1.vector[i] = -119.4594594595D;
    }
    b.size = 1024;
    w.addRowBatch(b);

    b.reset();
    for (int i = 0; i < 1024; i++) {
      f1.vector[i] = 9318.4351351351D;
    }
    b.size = 1024;
    w.addRowBatch(b);

    b.reset();
    for (int i = 0; i < 1024; i++) {
      f1.vector[i] = -4298.1513513514D;
    }
    b.size = 1024;
    w.addRowBatch(b);

    b.reset();
    w.close();

    Reader.Options options = new Reader.Options();
    try (Reader reader = OrcFile.createReader(testFilePath, OrcFile.readerOptions(conf));
         RecordReader rows = reader.rows(options)) {
      VectorizedRowBatch batch = schema.createRowBatch();

      rows.nextBatch(batch);
      assertEquals(1024, batch.size);
      assertTrue(batch.cols[0].isRepeating);

      rows.nextBatch(batch);
      assertEquals(1024, batch.size);
      assertTrue(batch.cols[0].isRepeating);

      rows.nextBatch(batch);
      assertEquals(1024, batch.size);
      assertTrue(batch.cols[0].isRepeating);
    }
  }
```

dongjoon-hyun:
If that is your intention, please switch it to `Bug` instead of `Improvement` because we don't backport `Improvement` in `core` module.

Anyway, that's too bad because we missed [1.9.4](https://github.com/apache/orc/milestone/30) release train.

dongjoon-hyun (approved):
+1, LGTM. Thank you.

PR #1963, opened 2024-06-27
ORC-1732: [C++] Fix detecting Homebrew-installed Protobuf on MacOS

luffy-zh:
Detect Protobuf installed by Homebrew on macOS.

Deal with the issue discussed [here](https://github.com/apache/arrow/issues/42149)

Test it locally.

NO


tools/test/TestFileScan.cc
```diff
@@ -211,9 +211,22 @@ void checkForError(const std::string& filename, const std::string& errorMsg) {
...
+  EXPECT_EQ("", output);
+  if (error.find(errorMsg1) == std::string::npos && error.find(errorMsg2) == std::string::npos) {
+    FAIL() << error;
+  }
+}
+
 TEST(TestFileScan, testErrorHandling) {
   checkForError(findExample("corrupt/stripe_footer_bad_column_encodings.orc"),
-                "bad number of ColumnEncodings in StripeFooter: expected=6, actual=0");
+                "bad number of ColumnEncodings in StripeFooter: expected=6, actual=0",
```
wgtmac, line 228: Why do we need to modify this test?

wgtmac:
@kou Could you help review this? Thanks!

cmake_modules/FindProtobuf.cmake
```diff
@@ -55,15 +56,34 @@ if (Protobuf_FOUND)
...
+      endif ()
+    endif ()
+  else ()
+    set (PROTOBUF_LIBRARY ${Protobuf_LIBRARIES})
+    set (PROTOBUF_INCLUDE_DIR ${Protobuf_INCLUDE_DIRS})
+    if (NOT PROTOBUF_INCLUDE_DIR)
+      message(FATAL_ERROR "Cannot determine Protobuf include directory.")
+    endif ()
+
+    if (Protobuf_LIBRARIES MATCHES "\\.a$")
```
kou, line 81: We may be able to use https://cmake.org/cmake/help/latest/variable/CMAKE_STATIC_LIBRARY_SUFFIX.html instead of `.a`.

cmake_modules/FindProtobuf.cmake
```diff
@@ -55,15 +56,34 @@ if (Protobuf_FOUND)
...
-        set (PROTOC_STATIC_LIB protobuf::libprotoc)
+      set (PROTOC_STATIC_LIB protobuf::libprotoc)
     endif ()
 
-    get_target_property (PROTOBUF_INCLUDE_DIR protobuf::libprotoc INTERFACE_INCLUDE_DIRECTORIES)
+    get_target_property (PROTOBUF_INCLUDE_DIR protobuf::libprotobuf INTERFACE_INCLUDE_DIRECTORIES)
+    if (NOT PROTOBUF_INCLUDE_DIR)
+      set (PROTOBUF_INCLUDE_DIR ${Protobuf_INCLUDE_DIRS})
+      if (NOT PROTOBUF_INCLUDE_DIR)
+        message(FATAL_ERROR "Cannot determine Protobuf include directory.")
```
kou, line 71: How about showing detection sources too for easy to debug?

```suggestion
        message(FATAL_ERROR "Cannot determine Protobuf include directory from protobuf::libprotobuf and Protobuf_INCLUDE_DIRS.")
```

kou (commented):
How about adding a CI job on macOS?
https://github.com/apache/orc/compare/main...kou:orc:cpp-protobuf?expand=1#diff-48c0ee97c53013d18d6bbae44648f7fab9af2e0bf5b0dc1ca761e18ec5c478f2R101-R133 is a sample configuration.

tools/test/TestFileScan.cc
luffy-zh: Google::MessageLite::ParseFromZeroCopyStream behavior has changed since pb 3.7.x.

wgtmac:
The new github action doesn't run. I found the error message below:

```
Invalid workflow file: .github/workflows/build_and_test.yml#L224
The workflow is not valid. .github/workflows/build_and_test.yml (Line: 224, Col: 7): 'run' is already defined .github/workflows/build_and_test.yml (Line: 225, Col: 7): 'run' is already defined
```

https://github.com/apache/orc/actions/runs/9746489872

cmake_modules/FindProtobuf.cmake
luffy-zh: Thanks for looking at that, '.a' has been replaced with CMAKE_STATIC_LIBRARY_SUFFIX

.github/workflows/build_and_test.yml
```diff
@@ -197,3 +197,29 @@ jobs:
         with:
           config: .github/.licenserc.yaml
 
+  macos-c++-check:
```
wgtmac, line 200:
```suggestion
  macos-cpp-check:
```

.github/workflows/build_and_test.yml
```diff
@@ -197,3 +197,29 @@ jobs:
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Install dependencies
+      run: |
+        brew update
+        brew install protobuf
+
+    - name: Test
```
wgtmac, line 218:
```suggestion
    name: "C++ Test on macOS"
    strategy:
      fail-fast: false
      matrix:
        version: [12, 14]
    runs-on: macos-${{ matrix.version }}
    steps:
    - name: Checkout repository
      uses: actions/checkout@v4
    - name: Install dependencies
      run: |
        brew update
        brew install protobuf
    - name: Test
```

Let's make it compact

wgtmac (approved):
+1

Thanks @luffy-zh for the fix and @kou for reviewing it!

.github/workflows/build_and_test.yml
```diff
@@ -197,3 +197,25 @@ jobs:
         with:
           config: .github/.licenserc.yaml
 
+  macos-cpp-check:
+    name: "C++ Test on macOS"
+    strategy:
+      fail-fast: false
+      matrix:
+        version: [12, 14]
```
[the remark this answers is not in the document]
luffy-zh: There is no significant difference between 13 and 12 for C++ compilation.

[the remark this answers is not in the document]
dongjoon-hyun: Then, please add 13 too.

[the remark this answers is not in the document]
luffy-zh: done

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Thank you, @luffy-zh , @wgtmac , @kou .
Merged to main/2.0.

PR #1964, opened 2024-06-27
MINOR: Fix syntax of DOAP

raboof:

I'm having some trouble finding an authoritative reference, but it looks like a release (which is a property) may not contain multiple Versions (which is a node) in RDF/XML. The python rdflib at least doesn't support it, and whatever generates https://projects.apache.org/json/projects/orc.json and https://projects.apache.org/project.html?orc also doesn't seem to process it correctly.

So infrastructure that uses the information for the DOAP does the right thing.

n/a

no

### DIFF
--- a/site/doap_orc.rdf
+++ b/site/doap_orc.rdf
@@ -48,6 +48,8 @@ the values that are required for the current query.</description>
         <created>2022-12-02</created>
         <revision>1.8.1</revision>
       </Version>
+    </release>
+    <release>
       <Version>
         <name>Stable release</name>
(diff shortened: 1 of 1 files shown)


dongjoon-hyun (commented):
Thank you for making a contribution, @raboof . 

Could you update the versions at this PR together?

You can find the latest version and date here.
- https://orc.apache.org/releases/

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Merged to main.

PR #1974, opened 2024-07-11
ORC-1742: Suppor print the id, name and type of each column in dump tool

cxzl25:
This PR aims to suppor print the id, name and type of each column in dump tool.

When we dump an ORC with a complex structure, we only output the column id of each column, but we do not know the column name and type corresponding to the column id.
If we use json format, it will output the id, name and type of each column.

Add UT

No


dongjoon-hyun (approved):

dongjoon-hyun:
Merged to main/2.0.

PR #1990, opened 2024-07-25
ORC-1749: Fix `supportVectoredIO` for hadoop version string with optional patch labels

otorreno:
Parse correctly semantic versioning when there is the optional labels in the patch are present.

There are cases where the hadoop version info may not be respecting the semantic versioning. It is the case for the hadoop version provided in some of the AWS managed services. This causes a ExceptionInInitializerError while trying to instantiate and ORC file reader.

Included unit tests. It required a small refactor to be able to test it in an easier way

No

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
+++ b/java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
@@ -23,6 +23,7 @@ import org.apache.hadoop.fs.FileRange;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hive.common.io.DiskRangeList;
+import org.apache.hadoop.util.VersionInfo;
 import org.apache.orc.CompressionCodec;
 import org.apache.orc.DataReader;
 import org.apache.orc.OrcProto;
@@ -48,7 +49,8 @@ import java.util.function.Supplier;
  */
 public class RecordReaderUtils {
   private static final HadoopShims SHIMS = HadoopShimsFactory.get();
-  private static final boolean supportVectoredIO = SHIMS.supportVectoredIO();
+  private static final boolean supportVectoredIO =
+      SHIMS.supportVectoredIO(VersionInfo.getVersion());
   private static final Logger LOG = LoggerFactory.getLogger(RecordReaderUtils.class);
 
   private static class DefaultDataReader implements DataReader {
--- a/java/shims/src/java/org/apache/orc/impl/HadoopShims.java
+++ b/java/shims/src/java/org/apache/orc/impl/HadoopShims.java
@@ -20,7 +20,6 @@ package org.apache.orc.impl;
 
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
-import org.apache.hadoop.util.VersionInfo;
 import org.apache.orc.EncryptionAlgorithm;
 
 import java.io.Closeable;
@@ -132,9 +131,9 @@ public interface HadoopShims {
    */
   boolean endVariableLengthBlock(OutputStream output) throws IOException;
 
-  default boolean supportVectoredIO() {
+  default boolean supportVectoredIO(String version) {
     // HADOOP-18103 is available since Apache Hadoop 3.3.5+
-    String[] versionParts = VersionInfo.getVersion().split("[.]");
+    String[] versionParts = version.split("[.-]");
     int major = Integer.parseInt(versionParts[0]);
     int minor = Integer.parseInt(versionParts[1]);
     int patch = Integer.parseInt(versionParts[2]);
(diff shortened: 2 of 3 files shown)


dongjoon-hyun (changes requested):
Thank you for making a PR, @otorreno .

Do you think you can provide a test case to give us more context, @otorreno ?
> // There could be cases where we are not able to parse the version
> // Here we are defaulting to the non vectoredIO if that is the case

otorreno:
> Thank you for making a PR, @otorreno .
> 
> Do you think you can provide a test case to give us more context, @otorreno ?
> 
> > // There could be cases where we are not able to parse the version
> > // Here we are defaulting to the non vectoredIO if that is the case

The version I am getting from AWS for hadoop is 3.3.6-amzn-3, so we fail parsing the patch

otorreno:
@dongjoon-hyun I can certainly write a test case, but I would have to pass down the VersionInfo and do a little refactor. Let me know if you would like me to do that

otorreno:
@dongjoon-hyun just pushed the tests

dongjoon-hyun (changes requested):
`3.3.6-amzn-3` looks like reasonable requests which we had better handle because it follows semantic versioning.

Can we make this proposal more compatible with [Semantic Versioning](https://semver.org)? For example, `3.3.6-amzn-3` has HADOOP-18103 if they follows semantic versioning. So, I believe `supportVectoredIO` should return true.

Could you make a change in a way to consider `Semantic Versioning`?

java/shims/src/java/org/apache/orc/impl/HadoopShims.java
```diff
@@ -132,13 +131,19 @@ ByteBuffer readBuffer(int maxLength,
...
   boolean endVariableLengthBlock(OutputStream output) throws IOException;
 
-  default boolean supportVectoredIO() {
-    // HADOOP-18103 is available since Apache Hadoop 3.3.5+
-    String[] versionParts = VersionInfo.getVersion().split("[.]");
-    int major = Integer.parseInt(versionParts[0]);
-    int minor = Integer.parseInt(versionParts[1]);
-    int patch = Integer.parseInt(versionParts[2]);
-    return major == 3 && (minor > 3 || (minor == 3 && patch > 4));
+  default boolean supportVectoredIO(String version) {
```
dongjoon-hyun, line 134: This change is okay for me.

otorreno:
> `3.3.6-amzn-3` looks like reasonable requests which we had better handle because it follows semantic versioning.
> 
> Can we make this proposal more compatible with [Semantic Versioning](https://semver.org)? For example, `3.3.6-amzn-3` has HADOOP-18103 if they follows semantic versioning. So, I believe `supportVectoredIO` should return true.
> 
> Could you make a change in a way to consider `Semantic Versioning`?

Done @dongjoon-hyun

dongjoon-hyun:
As a last piece, please update the PR title and description according to your latest code change.

Here, there is no `bad hadoop version string`.

otorreno:
> As a last piece, please update the PR title and description according to your latest code change.
> 
> Here, there is no `bad hadoop version string`.

you caught me doing that :). Done

dongjoon-hyun (approved):
+1, LGTM (Pending CIs).

BTW, I created ORC-1749 to track this. Do you happen to have Apache JIRA ID, @otorreno ?

otorreno:
> +1, LGTM (Pending CIs).
> 
> BTW, I created [ORC-1749](https://issues.apache.org/jira/browse/ORC-1749) to track this. Do you happen to have Apache JIRA ID, @otorreno ?

I do not. I guess I should create one

dongjoon-hyun:
Yes, please apply one. I can approve the ID creation quickly for you.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
```diff
@@ -48,7 +49,7 @@
  */
 public class RecordReaderUtils {
   private static final HadoopShims SHIMS = HadoopShimsFactory.get();
-  private static final boolean supportVectoredIO = SHIMS.supportVectoredIO();
+  private static final boolean supportVectoredIO = SHIMS.supportVectoredIO(VersionInfo.getVersion());
```
dongjoon-hyun, line 52: Could you make CI happy?
```
[INFO] There is 1 error reported by Checkstyle 10.17.0 with checkstyle.xml ruleset.
Error:  src/java/org/apache/orc/impl/RecordReaderUtils.java:[52] (sizes) LineLength: Line is longer than 100 characters (found 101).
```

otorreno:
> Yes, please apply one. I can approve the ID creation quickly for you.

I have requested one with `oscartorreno` as username.

dongjoon-hyun:
Ya, I approved you request.

And, I changed the reporter field of ORC-1749 to `oscartorreno`. :)

When this PR is merged, the assignee field will be filled with `oscartorreno` too.

java/core/src/java/org/apache/orc/impl/RecordReaderUtils.java
otorreno: Let's see now.

dongjoon-hyun:
Welcome to the Apache ORC community, @otorreno .

You are added to the Apache ORC contributor group and ORC-1749 is assigned to you.

This is a part of Apache ORC 2.0.2 release milestone and will be released in 3 weeks before August 15th. In addition, this will be a part of Apache Spark 4.0.0-preview2 in a near future.
- https://github.com/apache/orc/milestone/32

Thank you again and congratulations for your first commit to ASF repository.

Issue #1183, opened 2022-07-20
Add C++ API for Columnar Encryption


dongjoon-hyun:
This is created based on the following dev@orc mail on June 11st.

https://lists.apache.org/thread/pkp6ffh9pqok7v618zxtox708mv26sz0

dongjoon-hyun:
cc @wgtmac , @williamhyun , @guiyanakuang , @stiga-huang

guiyanakuang:
I looked at the commit history of the Java api. It seems like a lot of work. Maybe we should follow the steps of the Java implementation and divide it into smaller topics.
<img width="686" alt="image" src="https://user-images.githubusercontent.com/4069905/180588407-7b8a9c1d-db3e-4f66-b273-eee05e8ff906.png">

dongjoon-hyun:
I agree with you, @guiyanakuang . It will take lots of efforts.

wgtmac:
+1 @guiyanakuang @dongjoon-hyun.

This is a large feature. We can follow the java implementation and break it down into smaller work items first.

dongjoon-hyun:
Thank you, @wgtmac .

dongjoon-hyun:
I set a milestone 1.9.0 as a goal.

deshanxiao:
Hi @dongjoon-hyun Do we have jira to track the work?

wgtmac:
> Hi @dongjoon-hyun Do we have jira to track the work?

@deshanxiao Not yet. @coderex2522 is working on it already. Will create JIRAs later.

zxf216:
> > Hi @dongjoon-hyun Do we have jira to track the work?
> 
> @deshanxiao Not yet. @coderex2522 is working on it already. Will create JIRAs later.

@coderex2522  Has the ORC file encryption and decryption functionality been implemented?

zxf216:
@dongjoon-hyun  I want to try implementing ORC encryption and decryption feature in C++. Are there any other developers working on this feature?

zxf216:
@coderex2522 Thank you for your support. I would like to know more about the design plan for ORC encryption function. Do you have any materials that you can provide me with? Below is a document I have summarized by reading the Java source code. I'm not sure if it's correct.
document：https://www.processon.com/view/link/6507c37f464f0d50af3e2ae8

coderex2522:
[...]

PR #1993, opened 2024-08-04
read encrypted ORC file

zxf216:
you can use C++ version code to read encrypted ORC files

The C++ version is not yet able to read encrypted ORC files now

    orc::RowReaderOptions rowReaderOptions;
    orc::ReaderOptions readerOpts;
    std::shared_ptr<orc::InMemoryKeystore> keyStore = std::make_shared<orc::InMemoryKeystore>();
    std::string passKey ="secret123";
    keyStore->addKey("pii", 0, orc::EncryptionAlgorithm::AES_CTR_128,passKey);
    readerOpts.setKeyProvider(keyStore);
    printContents("/root/starrocks/encryption.orc",readerOpts, rowReaderOptions);

### DIFF
--- a/c++/include/orc/Reader.hh
+++ b/c++/include/orc/Reader.hh
@@ -39,7 +39,7 @@ namespace orc {
   // classes that hold data members so we can maintain binary compatibility
   struct ReaderOptionsPrivate;
   struct RowReaderOptionsPrivate;
-
+  class KeyProvider;
   /**
    * Expose the reader metrics including the latency and
    * number of calls of the decompression/decoding/IO modules.
@@ -138,6 +138,10 @@ namespace orc {
      * Get the reader metrics.
      */
     ReaderMetrics* getReaderMetrics() const;
+
+    ReaderOptions& setKeyProvider(std::shared_ptr<KeyProvider> keyProvider);
+
+    std::shared_ptr<KeyProvider> getKeyProvider() const;
   };
 
   /**
--- a/c++/src/TypeImpl.cc
+++ b/c++/src/TypeImpl.cc
@@ -837,5 +837,4 @@ namespace orc {
     }
     return nullptr;
   }
-
 }  // namespace orc
--- a/c++/src/security/InMemoryKeystore.cc
+++ b/c++/src/security/InMemoryKeystore.cc
@@ -1,0 +1,247 @@
+// Copyright 2010-present vivo, Inc. All rights reserved.
+//
+//Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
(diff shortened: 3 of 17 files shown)


dongjoon-hyun (commented):
Thank you for making a PR, @zxf216 .

We have been tracking this via 
- #1183

BTW, could you make the CI happy by fixing all failures?

wgtmac:
Thanks @zxf216! Let me take a first pass. cc @coderex2522 @ffacs

zxf216:
> Thank you for making a PR, @zxf216 .
> 
> We have been tracking this via
> 
> * [Add C++ API for Columnar Encryption #1183](https://github.com/apache/orc/issues/1183)
> 
> BTW, could you make the CI happy by fixing all failures?
Okay

dongjoon-hyun:
I converted it to `Draft` because CIs fails still.

c++/src/TypeImpl.cc
```diff
@@ -837,5 +837,4 @@ namespace orc {
     }
     return nullptr;
   }
-
```
wgtmac, line 840: Please revert the unnecessary change.

c++/include/orc/Reader.hh
```diff
@@ -39,7 +39,7 @@ namespace orc {
   // classes that hold data members so we can maintain binary compatibility
   struct ReaderOptionsPrivate;
   struct RowReaderOptionsPrivate;
-
+  class KeyProvider;
```
wgtmac, line 42: The `c++/include/orc` folder is public and files in it will be installed. If the `KeyProvider` is also public, please move its definition to the include folder as well. Otherwise, we should remove the related function from here.

c++/src/security/InMemoryKeystore.cc
```diff
@@ -0,0 +1,247 @@
+// Copyright 2010-present vivo, Inc. All rights reserved.
```
wgtmac, line 1: Please note that we have a license check which does not allow custom modification to the license header in this repo. I'm not sure if this violates any ASF rule to add something like this. @dongjoon-hyun Do you have any concern?

c++/src/security/ReaderEncryption.hh
```diff
@@ -0,0 +1,215 @@
...
+    Type* getRoot();
+
+    int getVariantId();
+
+    std::shared_ptr<SecretKeySpec> getFileFooterKey();
+
+    std::shared_ptr<SecretKeySpec> getStripeKey(long stripe);
+  };
+
+  // 加密变体，每个列一个
```
wgtmac, line 44: Please use English for comment.

wgtmac (changes requested):
Thanks @zxf216 again for the excellent work! I just took a preliminary look and have left some inline comments. As you can see that we have various CI jobs for test cases, code format, license check, etc. I strongly recommend splitting the PR into smaller ones to make the review process easier.

dongjoon-hyun:
Gentle ping, @zxf216 .

PR #2005, opened 2024-08-16
ORC-1264: [C++] Add a writer option to align compression block with row group

luffy-zh:
## What changes were proposed in this pull request?
Add support for the ORC writer to ensure that the compression block is aligned with the row group boundary。

## Why are the changes needed?
To reduce unnecessary I/O and decompression when PPD is in effect, we can enforce the compression block to be aligned with the row group boundary. For more detail, see [link](https://issues.apache.org/jira/projects/ORC/issues/ORC-1264?filter=allopenissues)

## How was this patch tested?
Uts in TestWriter.cc can convert this patch.

## Was this patch authored or co-authored using generative AI tooling?
NO


c++/include/orc/Reader.hh
```diff
@@ -26,6 +26,7 @@
 #include "orc/Vector.hh"
 #include "orc/orc-config.hh"
 #include "orc/sargs/SearchArgument.hh"
+#include "wrap/orc-proto-wrapper.hh"
```
wgtmac, line 29: We should not expose protobuf to the public header.

c++/include/orc/Reader.hh
```diff
@@ -657,6 +658,13 @@ namespace orc {
      * @param rowNumber the next row the reader should return
      */
     virtual void seekToRow(uint64_t rowNumber) = 0;
+
+    /**
+     * Get the row index entries of the current stripe.
+     * @return the row index entries
+     */
+    virtual std::unordered_map<uint64_t, proto::RowIndex>
```
wgtmac, line 666: Ditto

wgtmac (changes requested):

c++/include/orc/Writer.hh
```diff
@@ -290,6 +290,17 @@ namespace orc {
      * @return if not set, return default value which is 64 KB.
      */
     uint64_t getMemoryBlockSize() const;
+
+    /**
+     * Set whether the compression block should be aligned to row group boundary.
```
wgtmac, line 295: Add a comment to provide the expectation that boolean type might not be well aligned due to the `Boolean RLE encoding` requires packing bits into bytes. Otherwise the spec will be broken.

c++/include/orc/Reader.hh
```diff
@@ -657,6 +657,12 @@ namespace orc {
      * @param rowNumber the next row the reader should return
      */
     virtual void seekToRow(uint64_t rowNumber) = 0;
+
+    /**
+     * Get the current stripe position entries for the specified column.
+     * @return the position entries for the specified column.
+     */
+    virtual std::vector<std::vector<int>> getCurrentStripePositionEntries(uint64_t columnId) = 0;
```
wgtmac, line 665: Why this is placed in `RowReader` but not `Reader` which is more appropriate for reading metadata? In addition, it would be more useful if we can query the positions for any combination of columns from any stripe, which may help I/O optimization for index streams.

c++/src/ColumnWriter.cc
```diff
@@ -1222,6 +1269,18 @@ namespace orc {
     }
   }
 
+  void StringColumnWriter::finishStreams() {
+    ColumnWriter::finishStreams();
+    if (useDictionary) {
+      dictDataEncoder->finishEncode();
+      dictLengthEncoder->finishEncode();
+      dictStream->finishStream();
```
wgtmac, line 1277: Why calling this on dictionary stream? It is shared by all row groups so alignment looks unnecessary to me.

c++/src/ColumnWriter.cc
```diff
@@ -750,6 +781,12 @@ namespace orc {
     rleEncoder_->recordPosition(rowIndexPosition.get());
   }
 
+  template <typename BatchType>
+  void BooleanColumnWriter<BatchType>::finishStreams() {
+    ColumnWriter::finishStreams();
+    rleEncoder_->finishEncode();
```
wgtmac, line 787: Add a comment for boolean type. Some trailing boolean bits will be in the next compression block.

c++/test/TestWriter.cc
```diff
@@ -84,21 +86,67 @@ namespace orc {
...
+              break;
+            }
+            case TIMESTAMP_INSTANT:
+            case TIMESTAMP: {
+              if (posIndex != 0 && posIndex != 3) {
+                EXPECT_EQ(posVector[posIndex], 0);
+              }
+              break;
+            }
+            default: {
```
wgtmac, line 116: Does it cover the boolean type?

luffy-zh: Boolean case is covered here, but the rowIndexStripe is set to 1024, a multiple of 8.

c++/include/orc/Reader.hh
wgtmac, line 665: nit: I was thinking if we should design a better data structure, something like:
```
struct RowGroupPositions {
  uint64_t columnId;
  std::vector<int32_t> positions;
}
```
Then it can be reused if we want to add `std::vector<RowGroupPositions> getPositionEntries(int stripe, std::vector<int> columnIds)` in the future.

wgtmac (commented):
Generally LGTM.

c++/include/orc/Reader.hh
luffy-zh: I'll mark it in the next work.

wgtmac: I'd suggest to do this all at once. If we don't change it in time, the public API will get released and we cannot change it any more.

c++/include/orc/Reader.hh
```diff
@@ -657,6 +662,12 @@ namespace orc {
      * @param rowNumber the next row the reader should return
      */
     virtual void seekToRow(uint64_t rowNumber) = 0;
+
+    /**
+     * Get the row group positions of the specified column in the current stripe.
+     * @return the position entries for the specified columns.
+     */
+    virtual std::vector<RowGroupPositions> getPositionEntries(int columnId) = 0;
```
wgtmac, line 670: Sorry that I may have provided the wrong suggestion.

After reading the code base, we should move this function to `Reader` instead of `RowReader`. The main reason is that `RowReader` is used to read data and `Reader` is used to read metadata. A common use case is to read row group positions before creating the RowReader during the planning phase. I have also added `std::map<uint32_t, BloomFilterIndex> Reader::getBloomFilters(uint32_t stripeIndex, const std::set<uint32_t>& included) const` in the past, we may follow the same thing.

From `orc_proto.proto` file, we can find the exact type used for columnId and positions:
```
message RowIndexEntry {
  repeated uint64 positions = 1 [packed=true];
}

message Stream {
  optional uint32 column = 2;
}
```

Therefore my final suggestion would be as follows:
```
// Row group index of a single column in a stripe.
struct RowGroupIndex {
  // Positions are represented as a two-dimensional array where the first
  // dimension is row group index and the second dimension is the position
  // list of the row group. The size of the second dimension should be equal
  // among all row groups.
  std::vector<std::vector<uint64_t>> positions;
};

/**
 * Get row group index of all selected columns in the specified stripe
 * @param stripeIndex index of the stripe to be read for row group index.
 * @param included index of selected columns to return (if not specified,
 *        all columns will be returned).
 * @return map of row group index keyed by its column index.
 */
virtual std::vector<uint32_t, RowGroupIndex> Reader::getRowGroupIndex(
    uint32_t stripeIndex, const std::set<uint32_t>& included) const = 0;
```

In the future, we may also add `std::vector<std::unique_ptr<ColumnStatistics>>` to RowGroupIndex without breaking public API.

c++/src/Reader.cc
```diff
@@ -1491,6 +1467,59 @@ namespace orc {
...
+          (included.empty() || included.find(column) != included.end())) {
+        std::unique_ptr<SeekableInputStream> pbStream =
+            createDecompressor(contents_->compression,
+                               std::make_unique<SeekableFileInputStream>(
+                                   contents_->stream.get(), offset, length, *contents_->pool),
+                               contents_->blockSize, *(contents_->pool), contents_->readerMetrics);
+
+        proto::RowIndex pbRowIndex;
+        if (!pbRowIndex.ParseFromZeroCopyStream(pbStream.get())) {
+          throw ParseError("Failed to parse RowIndex");
```
wgtmac, line 1506: nit: provide more debug info, like stripeIndex and columnId?

c++/src/Reader.cc
```diff
@@ -1491,6 +1467,59 @@ namespace orc {
...
+    std::map<uint32_t, RowGroupIndex> ret;
+    uint64_t offset;
+    auto currentStripeFooter = loadCurrentStripeFooter(stripeIndex, offset);
+
+    // iterate stripe footer to get stream of row_index
+    for (int i = 0; i < currentStripeFooter.streams_size(); i++) {
+      const proto::Stream& stream = currentStripeFooter.streams(i);
+      uint32_t column = static_cast<uint32_t>(stream.column());
+      uint64_t length = static_cast<uint64_t>(stream.length());
+      RowGroupIndex rowGroupIndex = ret[column];
```
wgtmac, line 1494:
```suggestion
      RowGroupIndex& rowGroupIndex = ret[column];
```
This should break the test, or we need to add at least a test case to cover this new API.

wgtmac (changes requested):

c++/src/Reader.cc
luffy-zh: I added the RowGroupIndex size check in TestWriter.cc:97.

wgtmac (approved):
LGTM. Thanks @luffy-zh for making the changes!

dongjoon-hyun (approved):
+1, LGTM. Thank you so much, @luffy-zh and @wgtmac .
It's great to have this in Apache ORC 2.1.0.

dongjoon-hyun:
Merged to main for 2.1.0. Thank you again.

taiyang-li:
@luffy-zh I wonder how to distinguish if row group boundry is aligned to compression block while reading an ORC file.

Issue #1658, opened 2023-11-21
ORC-1124: Add C++ benchmark


dongjoon-hyun:

dongjoon-hyun:
This is removed from `Milestone 2.0.0`.

PR #2010, opened 2024-08-22
ORC-1767: [C++] Improve writing performance of encoded string column and support EncodedStringVectorBatch for StringColumnWriter

taiyang-li:
 

Improve writing performance of encoded string column and support EncodedStringVectorBatch for StringColumnWriter. 
Performance was measured in https://github.com/ClickHouse/orc/pull/15

original tests.

### DIFF
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -248,10 +248,15 @@ namespace orc {
     ~EncodedStringVectorBatch() override;
     std::string toString() const override;
     void resize(uint64_t capacity) override;
+
+    // Calculate data and length in StringVectorBatch from dictionary and index
+    void calculateDataAndLength();
+
     std::shared_ptr<StringDictionary> dictionary;
 
     // index for dictionary entry
     DataBuffer<int64_t> index;
+    std::atomic<bool> calculated{false};
   };
 
   struct StructVectorBatch : public ColumnVectorBatch {
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -25,6 +25,7 @@
 
 #include <cstdlib>
 #include <cstring>
+#include <list>
 #include <memory>
 #include <sstream>
 #include <stdexcept>
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -263,7 +263,7 @@ namespace orc {
     // index for dictionary entry
     DataBuffer<int64_t> index;
 
-    std::atomic<bool> dictionaryEncoded{false};
+    std::atomic<bool> dictionaryDecoded{false};
   };
 
   struct StructVectorBatch : public ColumnVectorBatch {
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -23,6 +23,7 @@
 #include "MemoryPool.hh"
 #include "orc/orc-config.hh"
 
+#include <atomic>
 #include <cstdlib>
 #include <cstring>
 #include <list>
@@ -30,7 +31,6 @@
 #include <sstream>
 #include <stdexcept>
 #include <vector>
-#include <atomic>
 
 namespace orc {
 
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -58,6 +58,8 @@ namespace orc {
     bool hasNulls;
     // whether the vector batch is encoded
     bool isEncoded;
+    // whther the dictionary is decoded into vector batch
+    bool dictionaryDecoded;
 
     // custom memory pool
     MemoryPool& memoryPool;
@@ -90,9 +92,12 @@ namespace orc {
     virtual bool hasVariableLength();
 
     /**
-     * Decode dictionary into vector batch.
+     * Decode possible dictionary into vector batch.
      */
-    virtual void decodeDictionary() {}
+    void decodeDictionary();
+
+   protected:
+    virtual void decodeDictionaryImpl() {}
 
    private:
     ColumnVectorBatch(const ColumnVectorBatch&);
@@ -256,14 +261,12 @@ namespace orc {
     void resize(uint64_t capacity) override;
 
     // Calculate data and length in StringVectorBatch from dictionary and index
-    void decodeDictionary() override;
+    void decodeDictionaryImpl() override;
 
     std::shared_ptr<StringDictionary> dictionary;
 
     // index for dictionary entry
     DataBuffer<int64_t> index;
-
-    std::atomic<bool> dictionaryDecoded{false};
   };
 
   struct StructVectorBatch : public ColumnVectorBatch {
@@ -276,6 +279,9 @@ namespace orc {
     bool hasVariableLength() override;
 
     std::vector<ColumnVectorBatch*> fields;
+
+   protected:
+    void decodeDictionaryImpl() override;
   };
 
   struct ListVectorBatch : public ColumnVectorBatch {
@@ -295,6 +301,9 @@ namespace orc {
 
     // the concatenated elements
     std::unique_ptr<ColumnVectorBatch> elements;
+
+   protected:
+    void decodeDictionaryImpl() override;
   };
 
   struct MapVectorBatch : public ColumnVectorBatch {
@@ -316,6 +325,9 @@ namespace orc {
     std::unique_ptr<ColumnVectorBatch> keys;
     // the concatenated elements
     std::unique_ptr<ColumnVectorBatch> elements;
+
+   protected:
+    void decodeDictionaryImpl() override;
   };
 
   struct UnionVectorBatch : public ColumnVectorBatch {
@@ -339,6 +351,9 @@ namespace orc {
 
     // the sub-columns
     std::vector<ColumnVectorBatch*> children;
+
+   protected:
+    void decodeDictionaryImpl() override;
   };
 
   struct Decimal {
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -23,10 +23,8 @@
 #include "MemoryPool.hh"
 #include "orc/orc-config.hh"
 
-#include <atomic>
 #include <cstdlib>
 #include <cstring>
-#include <list>
 #include <memory>
 #include <sstream>
 #include <stdexcept>
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -263,7 +263,7 @@ namespace orc {
     // index for dictionary entry
     DataBuffer<int64_t> index;
 
-    std::atomic<bool> dictionary_encoded{false};
+    std::atomic<bool> dictionaryEncoded{false};
   };
 
   struct StructVectorBatch : public ColumnVectorBatch {
--- a/c++/include/orc/Vector.hh
+++ b/c++/include/orc/Vector.hh
@@ -89,6 +89,11 @@ namespace orc {
      */
     virtual bool hasVariableLength();
 
+    /**
+     * Decode dictionary into vector batch.
+     */
+    virtual void decodeDictionary() {}
+
    private:
     ColumnVectorBatch(const ColumnVectorBatch&);
     ColumnVectorBatch& operator=(const ColumnVectorBatch&);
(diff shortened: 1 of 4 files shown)


wgtmac:
Thanks for opening a PR! Could you please create a JIRA for it?

wgtmac:
cc @ffacs @luffy-zh

taiyang-li:
@wgtmac could you help review this pr, thanks very much

ffacs:
@taiyang-li Hi, thank you for opening a PR! 
It seems that the time complexity not changed. Do writing cost a lot on sorting? After all, the size of dictionary shouldn't be large when we try to use dictionary encoding.

c++/src/ColumnWriter.cc
```diff
@@ -922,8 +931,8 @@ namespace orc {
       }
     };
 
-    std::map<DictEntry, size_t, LessThan> dict_;
-    std::vector<std::vector<char>> data_;
+    mutable std::vector<DictEntryWithIndex> flat_dict;
```
ffacs, line 934: Please follow the naming style.
[The format check didn't fail as expected.](https://github.com/apache/orc/actions/runs/10503087327/job/29095780106?pr=2010#step:4:66)

c++/src/ColumnWriter.cc
```diff
@@ -1103,6 +1115,12 @@ namespace orc {
       throw InvalidArgument("Failed to cast to StringVectorBatch");
     }
 
+    EncodedStringVectorBatch* encoded_string_batch =
+        dynamic_cast<EncodedStringVectorBatch*>(&rowBatch);
+    if (encoded_string_batch) {
```
ffacs, line 1120: Maybe let the caller call calculateDataAndLength() is better?

taiyang-li:
> @taiyang-li Hi, thank you for opening a PR! It seems that the time complexity not changed. Do writing cost a lot on sorting? After all, the size of dictionary shouldn't be large when we try to use dictionary encoding.

@ffacs I had measured about this. From flamegraph we could see that `std::map::insert` takes most of time. Notice that `std::map::insert` is executed even if key is duplicated.

c++/src/ColumnWriter.cc
taiyang-li: I'm afrait it will increase user's mental burden. Do you think make it transparent for user is ok?

c++/src/ColumnWriter.cc
taiyang-li: I'll fix it.

ffacs:
> > @taiyang-li Hi, thank you for opening a PR! It seems that the time complexity not changed. Do writing cost a lot on sorting? After all, the size of dictionary shouldn't be large when we try to use dictionary encoding.
> 
> @ffacs I had measured about this. From flamegraph we could see that `std::map::insert` takes most of time. Notice that `std::map::insert` is executed even if key is duplicated.

Oh yes, std::map takes $$O(n \ \log(key))$$, but sort takes only $$O(n + key\ \log(Key))$$

c++/include/orc/Vector.hh
```diff
@@ -248,10 +249,15 @@ namespace orc {
...
     void resize(uint64_t capacity) override;
+
+    // Calculate data and length in StringVectorBatch from dictionary and index
+    void calculateDataAndLength();
+
     std::shared_ptr<StringDictionary> dictionary;
 
     // index for dictionary entry
     DataBuffer<int64_t> index;
+    std::atomic<bool> calculated{false};
```
luffy-zh, line 260: Do we need atomic type here?

taiyang-li: It is needed to avoid duplicated invokes from caller.

taiyang-li:
@ffacs do you know where can I add google benchmark  in orc ?

ffacs:
> @ffacs do you know where can I add google benchmark in orc ?

There are no C++ benchmark currently. FYI. https://github.com/apache/orc/issues/1658

taiyang-li:
I added a benchmark (https://github.com/ClickHouse/ClickHouse/blob/1011f8ef9c2511654937c96b7bea7693a4c9715f/src/Common/benchmarks/orc_string_dictionary.cpp) to measure performance changes after this pr. Here is the result: 

```
$ ./build_gcc/src/Common/benchmarks/orc_string_dictionary 
2024-08-26T14:43:53+08:00
Running ./build_gcc/src/Common/benchmarks/orc_string_dictionary
Run on (32 X 2100 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x16)
  L1 Instruction 32 KiB (x16)
  L2 Unified 1024 KiB (x16)
  L3 Unified 11264 KiB (x2)
Load Average: 5.79, 5.70, 5.88
----------------------------------------------------------------------------------------------
Benchmark                                                    Time             CPU   Iterations
----------------------------------------------------------------------------------------------
BM_writeStringDictionary<OldSortedStringDictionary>  148905832 ns    148900910 ns            5
BM_writeStringDictionary<NewSortedStringDictionary>   86750854 ns     86747873 ns            8
```

c++/include/orc/Vector.hh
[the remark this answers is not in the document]
taiyang-li: I think they are equivalent..

c++/include/orc/Vector.hh
```diff
@@ -251,13 +256,14 @@ namespace orc {
...
-    void calculateDataAndLength();
+    void decodeDictionary() override;
 
     std::shared_ptr<StringDictionary> dictionary;
 
     // index for dictionary entry
     DataBuffer<int64_t> index;
-    std::atomic<bool> calculated{false};
+
+    std::atomic<bool> dictionary_encoded{false};
```
ffacs, line 266:
```suggestion
    std::atomic<bool> dictionaryDecoded{false};
```

c++/src/ColumnWriter.cc
```diff
@@ -931,8 +933,8 @@ namespace orc {
       }
     };
 
-    mutable std::vector<DictEntryWithIndex> flat_dict;
-    std::unordered_map<std::string, size_t> key_to_index;
+    mutable std::vector<DictEntryWithIndex> flatDict_;
+    std::unordered_map<std::string, size_t> keyToIndex;
```
ffacs, line 937:
```suggestion
    std::unordered_map<std::string, size_t> keyToIndex_;
```

c++/src/ColumnWriter.cc
```diff
@@ -952,9 +959,12 @@ namespace orc {
   // write dictionary data & length to output buffer
   void SortedStringDictionary::flush(AppendOnlyBufferedStream* dataStream,
                                      RleEncoder* lengthEncoder) const {
-    for (auto it = dict_.cbegin(); it != dict_.cend(); ++it) {
-      dataStream->write(it->first.data, it->first.length);
-      lengthEncoder->write(static_cast<int64_t>(it->first.length));
+    std::sort(flatDict_.begin(), flatDict_.end(), LessThan());
+
+    for (const auto& entry_with_index : flatDict_) {
```
ffacs, line 964:
```suggestion
    for (const auto& entryWithIndex : flatDict_) {
```

ffacs (changes requested):

c++/include/orc/Vector.hh
luffy-zh: Sorry for the late comment. All members of the ColumnVectorBatch class are non-atomic. It is assumed that the ColumnVectorBatch object will not be shared or reused across multiple threads. Could you share the code in ClickHouse that invokes decodeDictionary()?

c++/include/orc/Vector.hh
```diff
@@ -263,7 +263,7 @@ namespace orc {
     // index for dictionary entry
     DataBuffer<int64_t> index;
 
-    std::atomic<bool> dictionary_encoded{false};
+    std::atomic<bool> dictionaryEncoded{false};
```
ffacs, line 266: Ain't this should be `dictionaryDecoded`?

ffacs (changes requested):

c++/include/orc/Vector.hh
taiyang-li: fixed

c++/include/orc/Vector.hh
```diff
@@ -88,6 +89,11 @@ namespace orc {
      */
     virtual bool hasVariableLength();
 
+    /**
+     * Decode dictionary into vector batch.
+     */
+    virtual void decodeDictionary() {}
```
wgtmac, line 95: We need to implement this for struct, map, list and union type by calling it recursively.

c++/include/orc/Vector.hh
```diff
@@ -248,10 +254,16 @@ namespace orc {
...
+
+    // Calculate data and length in StringVectorBatch from dictionary and index
+    void decodeDictionary() override;
+
     std::shared_ptr<StringDictionary> dictionary;
 
     // index for dictionary entry
     DataBuffer<int64_t> index;
+
+    std::atomic<bool> dictionaryDecoded{false};
```
wgtmac, line 266: I think it is an overkill to use atomic here. It should be a misuse to call member functions of ColumnVectorBatch concurrently.

c++/src/ColumnWriter.cc
```diff
@@ -130,6 +130,8 @@ namespace orc {
         hasNullValue = true;
       }
     }
+
+    batch.decodeDictionary();
```
wgtmac, line 134: My intention to add `decodeDictionary()` is that caller should call `decodeDictionary()` before calling `ColumnWriter::add()`. This can make the implementation of `ColumnWriter ` simpler.

c++/include/orc/Vector.hh
wgtmac: BTW, please add a test case to make sure it works as expected.

wgtmac (changes requested):

taiyang-li:
> Thanks for opening a PR! Could you please create a JIRA for it?

https://issues.apache.org/jira/browse/ORC-1767

c++/include/orc/Vector.hh
taiyang-li: It's a personal coding habit in case there might be a multi-threaded call scenario in the future. Besides, `std::atomic<bool>` is so trivial that it would not influence the performance.

taiyang-li: Anyway, I'll remove it to keep code simple.

c++/src/Vector.cc
```diff
@@ -88,6 +88,20 @@ namespace orc {
     }
   }
 
+  void EncodedStringVectorBatch::decodeDictionary() {
+    if (dictionaryDecoded) return;
```
ffacs, line 92:
```suggestion
    if (dictionaryDecoded.exchange(true)) return;
```

c++/src/Vector.cc
```diff
@@ -88,6 +88,20 @@ namespace orc {
...
+
+    size_t n = index.size();
+    resize(n);
+
+    for (size_t i = 0; i < n; ++i) {
+      if (!hasNulls || notNull[i]) {
+        dictionary->getValueByIndex(index[i], data[i], length[i]);
+      }
+    }
+    dictionaryDecoded = true;
```
ffacs, line 102:
```suggestion
```

ffacs (changes requested):

c++/src/Vector.cc
taiyang-li: I had removed `std::atomic`

c++/src/Vector.cc
taiyang-li: dictionaryDecoded had been moved to `ColumnVectorBatch`

taiyang-li:
@ffacs any new requested changes ?

c++/include/orc/Vector.hh
```diff
@@ -57,6 +57,8 @@ namespace orc {
     bool hasNulls;
     // whether the vector batch is encoded
     bool isEncoded;
+    // whther the dictionary is decoded into vector batch
```
ffacs, line 60: whether

ffacs (commented):
This path almost LGTM other than the typo. Thank you @taiyang-li !

wgtmac (approved):
+1

ffacs (approved):
LGTM

Issue #2037, opened 2024-09-23
Incorrect positions might be recorded after ORC-1711


nickitat:
Hi 👋🏼 
It seems we stepped on a bug after [updating](https://github.com/ClickHouse/orc/tree/b0c0571d122b8f3281119443d027e217aebf85e3) ORC submodule on 360c0613ed1b8c0106ee7a52dec995af04e85806.

In the corresponding PR we start seeing [test failure](https://s3.amazonaws.com/clickhouse-test-reports/69473/63d1cd1bd2a8f07bcb34989af51d1c1c06fad82b/stateless_tests__debug__[2_2].html). 
The failing test: `select count(), sum(number) from file('02892.orc') where indexHint(sometimes_null is NULL);`.
The exception: `orc::ParseError, e.what() = Bad seek to (chunkHeader=0, posInChunk=65536) in zstd(ORCInputStream from 1401 for 12)`

While looking at this test I noticed that we indeed read wrong positions here:
https://github.com/apache/orc/blob/f919fab762f189cacb96f19f435da70a6496da82/c%2B%2B/src/Reader.cc#L457-L462
And then I found that we actually also write incorrect values here:
https://github.com/apache/orc/blob/f919fab762f189cacb96f19f435da70a6496da82/c%2B%2B/src/ByteRLE.cc#L190-L194
And then I found that the difference between previously written values and the values we're getting after the update is only that in the positions when we previously had zeroes we now have 65536. And this number is equal to MemoryBlockSize in the [PR](https://github.com/ClickHouse/orc/commit/e8402b233fd671f8fe42ec7eff24ee4fea51256b#diff-ce1a0e200c60df38aa82c7b9d6eae495972fcf8e463037c418cebd1fd6d20187R281). It happens when all three `unusedBufferSize, bufferLength, bufferPosition` were equal to 0.

I'm not familiar with the code, so all I did is just implemented a small crutch to work around that issue: https://github.com/ClickHouse/orc/commit/e77893cd1212f0abdc0ad616f0b4932fbca4f692 and at least it fixed our tests.
Now I'd like to ask the actual experts in this code to have a look and implement a proper fix.

Thanks!

nickitat:
cc @luffy-zh @ffacs @wgtmac

luffy-zh:
Thanks for the report, I'll take a look.

nickitat:
i checked on our tests and everything looks fine now - big thanks @luffy-zh

PR #2038, opened 2024-09-27
ORC-1711: [C++] Fix recordPosition with unexpected value

luffy-zh:
Make sure the recorded position in blockCompressStream is correct.

Fix the issue discussed [here](https://github.com/apache/orc/issues/2037).

UTs in testwriter.cc can cover this patch.

No.

### DIFF
--- a/c++/src/io/OutputStream.hh
+++ b/c++/src/io/OutputStream.hh
@@ -75,10 +75,6 @@ namespace orc {
       return false;
     }
 
-    virtual bool isBlockCompressed() const {
-      return false;
-    }
-
     virtual void finishStream();
   };
   DIAGNOSTIC_POP
--- a/c++/src/io/OutputStream.hh
+++ b/c++/src/io/OutputStream.hh
@@ -74,6 +74,11 @@ namespace orc {
     virtual bool isCompressed() const {
       return false;
     }
+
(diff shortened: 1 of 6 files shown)


dongjoon-hyun:
Thank you for making a PR, @luffy-zh .

cc @wgtmac

c++/src/io/OutputStream.hh
```diff
@@ -74,6 +74,11 @@ namespace orc {
     virtual bool isCompressed() const {
       return false;
     }
+
+    virtual bool isBlockCompressed() const {
```
wgtmac, line 78: I would be against adding this to the `BufferedOutputStream`. We already have `isCompressed()` above, we can use static_cast to get a `CompressionStream` and call this from there. WDYT?

c++/src/io/OutputStream.hh
```diff
@@ -74,6 +74,7 @@ namespace orc {
     virtual bool isCompressed() const {
       return false;
     }
+
```
wgtmac, line 77:
```suggestion
```

ffacs (commented):
Nit: keep the comments. LGTM. Thank you @luffy-zh

wgtmac (approved):
+1

Thanks @luffy-zh and @ffacs!

PR #2048, opened 2024-10-09
ORC-262: [C++] Support async I/O prefetch

taiyang-li:
Support async io prefetch for orc c++ lib. Close https://issues.apache.org/jira/browse/ORC-262 

Changes: 
- Added new interface `InputStream::readAsync`(default unimplemented). It reads io asynchronously within the specified range.
- Added IO Cache implementation `ReadRangeCache` to cache async io results. This borrows from a similar design of Parquet Reader in https://github.com/apache/arrow
- Added interface `Reader::preBuffer` to trigger io prefetch. In the specific implementation of `ReaderImpl::preBuffer`, the io ranges will be calculated according to the selected stripe and columns, and then these ranges will be merged and sorted, and `ReadRangeCache::cache` will be called to trigger the asynchronous io in the background, waiting for the use of the upper layer
- Added the interface `Reader::releaseBuffer`, which is used to release all cached io ranges before an offset

Async io prefetch could hide io latency during reading orc files, which improves performance of scan operators in ClickHouse.

Pass the CIs.

No.


c++/src/io/Cache.hh
```diff
@@ -0,0 +1,230 @@
...
+
+      // Remove zero-sized ranges
+      auto end = std::remove_if(ranges.begin(), ranges.end(),
+                                [](const ReadRange& range) { return range.length == 0; });
+      // Sort in position order
+      std::sort(ranges.begin(), end,
+                [](const ReadRange& a, const ReadRange& b) { return a.offset < b.offset; });
+
+      // Remove ranges that overlap 100%
+      end = std::unique(ranges.begin(), end, [](const ReadRange& left, const ReadRange& right) {
```
ffacs, line 78: The binary predicate does not establish an equivalence relation, the behavior is undefined. https://en.cppreference.com/w/cpp/algorithm/unique

taiyang-li: fixed.

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,230 @@
...
+  struct CacheOptions {
+    /// The maximum distance in bytes between two consecutive
+    /// ranges; beyond this value, ranges are not combined
+    uint64_t holeSizeLimit = 8192;
+
+    /// The maximum size in bytes of a combined range; if
+    /// combining two consecutive ranges would produce a range of a
+    /// size greater than this, they are not combined
+    uint64_t rangeSizeLimit = 32 * 1024 * 1024;
+  };
```
zhanglistar, line 43: Do we have memory limit for cache?

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,230 @@
...
+  ///
+  /// To use:
+  /// 1. Cache() the ranges you expect to read in the future. Ideally, these ranges have
+  ///    the exact offset and length that will later be read. The cache will combine those
+  ///    ranges according to parameters (see constructor).
+  ///
+  ///    By default, the cache will also start fetching the combined ranges in parallel in
+  ///    the background, unless CacheOptions.lazy is set.
+  ///
+  /// 2. Call WaitFor() to be notified when the given ranges have been read. If
```
zhanglistar, line 182: Where is `WaitFor`?

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,230 @@
...
+#include <cassert>
+#include <cstdint>
+#include <future>
+#include <utility>
+#include <vector>
+
+namespace orc {
+  class InputStream;
+
+  struct CacheOptions {
```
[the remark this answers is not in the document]
taiyang-li: `lazy` means only read io asynchronously when it is actually needed, but we need to prefetch io in advance. Currently we don't need this function.

c++/src/io/Cache.hh
taiyang-li: Useless. I'll delete it.

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,230 @@
...
+    explicit ReadRangeCache(InputStream* stream, CacheOptions options, MemoryPool* memoryPool)
+        : stream_(stream), options_(std::move(options)), memoryPool_(memoryPool) {}
+
+    ~ReadRangeCache() = default;
+
+    /// Cache the given ranges in the background.
+    ///
+    /// The caller must ensure that the ranges do not overlap with each other,
+    /// nor with previously cached ranges.  Otherwise, behaviour will be undefined.
+    void cache(std::vector<ReadRange> ranges);
```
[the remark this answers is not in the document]
taiyang-li: ranges are moved into function.

c++/src/io/Cache.hh
taiyang-li: I'll do it.

c++/src/Reader.cc
```diff
@@ -1474,6 +1476,77 @@ namespace orc {
...
+      return;
+    }
+
+    orc::RowReaderOptions row_reader_options;
+    row_reader_options.includeTypes(includeTypes);
+    ColumnSelector column_selector(contents_.get());
+    std::vector<bool> selected_columns;
+    column_selector.updateSelected(selected_columns, row_reader_options);
+
+    std::vector<ReadRange> ranges;
```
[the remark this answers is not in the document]
taiyang-li: Single include type may correspond to serveral ranges for complex type or dict type. By calling `ranges.reserve(includeTypes.size());`, we give `ranges` an initial capacity to avoid too much realloc.

taiyang-li:
> `Reader::preBuffer` prefetch stripes as a unit which might be too large. For those users who don't want to prefetch entire file one-shot, they have to know the structure of the file. Do you think it is a good idea to make prefetch transparent to users and let the orc reader prefetch data(eg. 1MB for each column at a time) when it's proper. What's more, we could make enable async IO a option and expose a cache interface for users so they can implement their eviction policy.

I agree with you. It is totally decided by users to choose whether to prefetch the whole orc file or single stripe or single column in single/multiple stripes. `Reader::preBuffer` already supported all those options.

taiyang-li:
@ffacs  @wgtmac  any more comments ? Thanks!

wgtmac:
Sorry that I'm a little bit overwhelmed these days. Will take a look when I get the chance.

BTW, @luffy-zh is implementing exposing RowIndex positions: https://github.com/apache/orc/pull/2005. Perhaps there is an opportunity to further prefetch io together with predicate pushdown.

taiyang-li:
@wgtmac  That's a great work. We could do more improvements on IO latency hiding after it is merged.

c++/src/MemoryPool.cc
```diff
@@ -86,10 +86,6 @@ namespace orc {
       for (uint64_t i = currentSize_; i > newSize; --i) {
         (buf_ + i - 1)->~T();
       }
-    } else if (newSize > currentSize_) {
-      for (uint64_t i = currentSize_; i < newSize; ++i) {
-        new (buf_ + i) T();
```
wgtmac, line 91: I don't think we can remove these lines.

c++/src/MemoryPool.cc
```diff
@@ -134,9 +130,6 @@ namespace orc {
   template <>
   void DataBuffer<char>::resize(uint64_t newSize) {
     reserve(newSize);
-    if (newSize > currentSize_) {
```
wgtmac, line 137: Lines below seem to be unrelated changes. Could you please revert them? I remember there was an effort to remove these but caused issues to the downstream.

c++/src/Reader.hh
```diff
@@ -19,6 +19,7 @@
 #ifndef ORC_READER_IMPL_HH
 #define ORC_READER_IMPL_HH
 
+#include "io/Cache.hh"
```
wgtmac, line 22: Move it below line 28.

c++/src/StripeStream.hh
```diff
@@ -19,6 +19,7 @@
 #ifndef ORC_STRIPE_STREAM_HH
 #define ORC_STRIPE_STREAM_HH
 
+#include "io/Cache.hh"
```
wgtmac, line 22: What about using forward declaration of `ReadRangeCache` instead of adding the header file?

c++/src/StripeStream.hh
```diff
@@ -45,6 +46,7 @@ namespace orc {
     InputStream& input_;
     const Timezone& writerTimezone_;
     const Timezone& readerTimezone_;
+    std::shared_ptr<ReadRangeCache> cachedSource;
```
wgtmac, line 49:
```suggestion
    std::shared_ptr<ReadRangeCache> readCache_;
```

We don't need to copy the name `source` from Arrow.

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,208 @@
...
+      return !(left == right);
+    }
+
+    bool contains(const ReadRange& other) const {
+      return (offset <= other.offset && offset + length >= other.offset + other.length);
+    }
+  };
+
+  struct ReadRangeCombiner {
+    std::vector<ReadRange> coalesce(std::vector<ReadRange> ranges) const {
```
wgtmac, line 65: Move the implementation detail to the cc file?

c++/include/orc/Reader.hh
```diff
@@ -605,6 +612,26 @@ namespace orc {
...
+     * Get the footer of the ORC file.
+     */
+    virtual const proto::Footer* getFooter() const = 0;
+
+    /**
+     * Get the schema of the ORC file.
+     */
+    virtual const proto::Metadata* getMetadata() const = 0;
+
+    virtual void preBuffer(const std::vector<int>& stripes, const std::list<uint64_t>& includeTypes,
```
wgtmac, line 631: Is it possible to call these functions internally instead of user intervention? Reader has the knowledge of selected stripes (from offset/length reader options) and selected columns.

wgtmac (changes requested):
I have just finished the initial review. Thanks @taiyang-li! Please see my inline comments. My main concern is the usability that it requires user to call `preBuffer` instead of automatically prefetching required data.

c++/src/io/Cache.hh
taiyang-li: Ok I'll do it.

c++/src/MemoryPool.cc
taiyang-li: There is performance benefits if we remove it. In fact, similar optimization exists in ClickHouse.  
If you find any code relying on it, please let me know. At least removing it doesn't influence the CI.

c++/src/MemoryPool.cc
taiyang-li: Do you think it makes sence if we add a build option like `MEMSET_AFTER_RESIZE` with false as default value

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,208 @@
...
+        assert(left.offset < right.offset);
+        assert(!left.contains(right));
+      }
+#endif
+
+      std::vector<ReadRange> coalesced;
+
+      auto itr = ranges.begin();
+      // Ensure ranges is not empty.
+      assert(itr <= ranges.end());
```
ffacs, line 102: Remove this

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,208 @@
...
+      // Start of the current coalesced range and end (exclusive) of previous range.
+      // Both are initialized with the start of first range which is a placeholder value.
+      uint64_t coalesced_start = itr->offset;
+      uint64_t prev_range_end = coalesced_start;
+
+      for (; itr < ranges.end(); ++itr) {
```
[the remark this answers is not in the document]
ffacs: What's more, could you follow the naming style?

c++/include/orc/Reader.hh
taiyang-li: Consider this case : there are 6 stripes in an ORC file,   and the user wants to read the first/third/fifth stripe in it(). He may create a RowReader for each stripe to get `ColumnVectorBatch` from it and convert it to other data columnar representations. To hide more io latency, we should prefetch the next stripe while reading/processing current stripe. For example: 
- prefetch the third stripe before reading the first stripe
- prefetch the fifth stripe before reading the third stripe 

Actually above case happens in ClickHouse. What I want to illustrate is that only user knows best about which is the next stripe to prefetch and when to prefetch it. That's why user has to invoke `preBuffer` explicitly. Besides, parquet also has similar designs.

ffacs: > Consider this case : there are 6 stripes in an ORC file, and the user wants to read the first/third/fifth stripe in it(). He may create a RowReader for each stripe to get `ColumnVectorBatch` from it and convert it to other data columnar representations. To hide more io latency, we should prefetch the next stripe while reading/processing current stripe. For example:
> 
> * prefetch the third stripe before reading the first stripe
> * prefetch the fifth stripe before reading the third stripe
> 
> Actually above case happens in ClickHouse. What I want to illustrate is that only user knows best about which is the next stripe to prefetch and when to prefetch it. That's why user has to invoke `preBuffer` explicitly. Besides, parquet also has similar designs.

Do you means there are some users who want to read certain stripes of a file, and the requirement of stripes would change on processing of data, so they want to call prefetch on their own?

taiyang-li: @ffacs it is a typical use case in spark. ORC files may be very large, and spark will split it into several file splits, each file split contains several stripes, we can even discard some candidate stripes by stripe statistics after that.

taiyang-li: @wgtmac I thought about it. But I'm afraid it is not enough to hide enough io because the gap between the creation of stripe reader and reading from it may be very small, which means we will meet a lot of cache misses during reading from current stripe.

taiyang-li: I think the `preBuffer` doesn't bother

wgtmac: SGTM

c++/include/orc/OrcFile.hh
```diff
@@ -58,6 +73,17 @@ namespace orc {
      */
     virtual void read(void* buf, uint64_t length, uint64_t offset) = 0;
 
+    /**
+     * Read data asynchronously.
+     * @param offset the position in the stream to read from.
+     * @param length the number of bytes to read.
+     * @return a future that will be set to the buffer when the read is complete.
+     */
+    virtual std::future<BufferPtr> readAsync(uint64_t /*offset*/, uint64_t /*length*/,
```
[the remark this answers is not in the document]
taiyang-li: readAsync is an asynchronous process, which means `DataBuffer` is constructed and filled asynchronously in the future. If we construct `DataBuffer` in the main thread in which `readAsync` is invoked, the main thread may be blocked by page faults. It is not friendly to performance.

c++/include/orc/OrcFile.hh
```diff
@@ -36,6 +37,20 @@ namespace orc {
    */
   class InputStream {
    public:
+    using Buffer = DataBuffer<char>;
+    using BufferPtr = std::shared_ptr<Buffer>;
+
+    struct BufferSlice {
```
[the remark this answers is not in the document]
taiyang-li: It is used in `ReadRangeCache`. That is why I expose it as public class.

c++/include/orc/Reader.hh
```diff
@@ -605,6 +612,26 @@ namespace orc {
...
+
+    /**
+     * Get the input stream for the ORC file.
+     */
+    virtual InputStream* getStream() const = 0;
+
+    /**
+     * Get the footer of the ORC file.
+     */
+    virtual const proto::Footer* getFooter() const = 0;
```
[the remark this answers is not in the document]
taiyang-li: I'll try.

c++/src/StripeStream.hh
taiyang-li: Good idea

c++/src/io/Cache.hh
taiyang-li: But assert check the conditions in debug mode, keeping it doesn't influence release version

c++/src/io/Cache.hh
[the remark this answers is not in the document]
taiyang-li: done.

c++/include/orc/Reader.hh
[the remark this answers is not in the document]
taiyang-li: @wgtmac I found it is needed by https://github.com/bigo-sg/ClickHouse/blob/6e9d19fdd05bf951323c7b89c7d8c9089a604f0b/src/Processors/Formats/Impl/NativeORCBlockInputFormat.cpp#L990-L991 to get candidate stripes to prefetch.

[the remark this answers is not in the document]
wgtmac: The principle is that we cannot expose any protobuf structure directly. The generated protobuf header is not installed to the include directory. If you need access to certain metadata, please call (or add) getXXX function to the ORC reader.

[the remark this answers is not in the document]
taiyang-li: removed.

c++/src/Reader.cc
```diff
@@ -1474,6 +1476,77 @@ namespace orc {
...
+              << ": streamOffset=" << offset << ", streamLength=" << stream.length()
+              << ", stripeOffset=" << stripe_info.offset()
+              << ", stripeIndexLength=" << stripe_info.index_length()
+              << ", stripeDataLength=" << stripe_info.data_length();
+          throw ParseError(msg.str());
+        }
+
+        if (stream.has_kind() && selected_columns[stream.column()]) {
+          const auto& kind = stream.kind();
+          if (kind == proto::Stream_Kind_DATA || kind == proto::Stream_Kind_DICTIONARY_DATA ||
```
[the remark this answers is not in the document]
taiyang-li: image 

Above codes show where `getStream` is called in column reader. Only those kind of streams are used in column reader. It is designed on purpose to avoid useless prefetch.

c++/src/MemoryPool.cc
taiyang-li: reverted

c++/src/MemoryPool.cc
taiyang-li: > Do you think it makes sence if we add a build option like `MEMSET_AFTER_RESIZE` with false as default value.

already reverted. But I still think it is worth to try. If you think the solution is ok, I'll do it in another newe pr.

taiyang-li:
@wgtmac @ffacs I had finished the requsted changes. Hope for your further reviews, thanks!

c++/include/orc/Reader.hh
```diff
@@ -605,6 +612,21 @@ namespace orc {
      */
     virtual std::map<uint32_t, BloomFilterIndex> getBloomFilters(
         uint32_t stripeIndex, const std::set<uint32_t>& included) const = 0;
+
+    /**
+     * Get the input stream for the ORC file.
+     */
+    virtual InputStream* getStream() const = 0;
```
wgtmac, line 619: I don't think we need to add it to the base class. You can cast to `ReaderImpl` when required.

c++/include/orc/Reader.hh
```diff
@@ -605,6 +612,21 @@ namespace orc {
...
+
+    /**
+     * Get the schema of the ORC file.
+     */
+    virtual const proto::Metadata* getMetadata() const = 0;
```
wgtmac, line 624: Please do not expose proto object here.

c++/include/orc/Reader.hh
```diff
@@ -605,6 +612,21 @@ namespace orc {
...
+
+    virtual void preBuffer(const std::vector<int>& stripes, const std::list<uint64_t>& includeTypes,
```
wgtmac, line 626: Please add documentation for preBuffer and releaseBuffer as they are user-facing.

dongjoon-hyun:
Thank you all. Given the active status of this PR, I added a milestone label, `2.1.0`. Hopefully, Apache ORC 2.1 can have this.
- https://github.com/apache/orc/milestone/28 (2025-01-17).

dongjoon-hyun:
Could you resolve the conflicts, @taiyang-li ?

taiyang-li:
> Could you resolve the conflicts, @taiyang-li ?

Done.

c++/src/Reader.cc
```diff
@@ -1523,6 +1525,77 @@ namespace orc {
...
+    ColumnSelector columnSelector(contents_.get());
+    std::vector<bool> selectedColumns;
+    columnSelector.updateSelected(selectedColumns, rowReaderOptions);
+
+    std::vector<ReadRange> ranges;
+    ranges.reserve(includeTypes.size());
+    for (auto stripe : stripes) {
+      // get stripe information
+      const auto& stripeInfo = footer_->stripes(stripe);
+      uint64_t stripe_footer_start =
```
ffacs, line 1551: Please keep the naming style

c++/include/orc/Reader.hh
```diff
@@ -39,6 +39,13 @@ namespace orc {
   // classes that hold data members so we can maintain binary compatibility
   struct ReaderOptionsPrivate;
   struct RowReaderOptionsPrivate;
+  struct CacheOptions;
+  class InputStream;
+
+  namespace proto {
```
ffacs, line 45: Remove this

c++/src/Reader.cc
```diff
@@ -1523,6 +1525,77 @@ namespace orc {
...
+          stripeInfo.offset() + stripeInfo.index_length() + stripeInfo.data_length();
+      uint64_t stripe_footer_length = stripeInfo.footer_length();
```
ffacs, line 1553: ditto

c++/include/orc/OrcFile.hh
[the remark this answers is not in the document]
wgtmac: ReadRangeCache is not (and should not be) exported. Please do not declare the buffer here. We should be strict about public API.

c++/include/orc/OrcFile.hh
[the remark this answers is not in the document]
wgtmac: I don't think this is a good practice. How to allocate the buffer is an implementation detail. In the current implementation, the DataBuffer instance can be managed by ReadRangeCache. So I would prefer the following signature:
```
virtual std::future<void> read(void* buf, uint64_t length, uint64_t offset) = 0;
```

wgtmac (changes requested):
I still have some concerns about the public API. Please see my inline comments.

BTW, I believe `void preBuffer(const std::vector<int>& stripes, const std::list<uint64_t>& includeTypes)` is a little bit coarse. We need to think about how to work together with selective read (e.g. when predicate pushdown is able to filter most rows).

c++/include/orc/OrcFile.hh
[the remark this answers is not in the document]
taiyang-li: How to allocate buffer is only related to MemoryPool implementation. If we use the new interface, the cost of allocating will occupy the main thread instead of backgroud thread. It is not good to performance.

c++/include/orc/Reader.hh
```diff
@@ -39,6 +39,8 @@ namespace orc {
   // classes that hold data members so we can maintain binary compatibility
   struct ReaderOptionsPrivate;
   struct RowReaderOptionsPrivate;
+  struct CacheOptions;
```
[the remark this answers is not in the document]
taiyang-li: `ReadRangeCache` is already an internal implementation. I'll put `CacheOptions` as a member of `ReaderOptions`.

c++/include/orc/OrcFile.hh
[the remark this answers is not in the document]
taiyang-li: There is an extra advantage if we use the old interface: we could easily know `readAsync` is success or not by `BufferPtr != nullptr`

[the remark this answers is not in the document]
wgtmac: Do you have the number of the performance penalty? 

We need to trade off between performance and usability. In apache/arrow, its input stream supports both styles of returning a buffer or copying into a provided buffer. If we need to return a buffer in the readAsync function, perhaps we need to add an equivalent sync function to return a buffer as well.

> There is an extra advantage if we use the old interface: we could easily know readAsync is success or not by BufferPtr != nullptr

Isn't an exception expected if the read is unsuccessful?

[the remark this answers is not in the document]
taiyang-li: I added a benchmark about it. 
``` cpp

static constexpr size_t numThreads = 16;
static constexpr size_t bufferSize = 1024 * 1024;
static std::vector<std::thread> threads1(numThreads);
static std::vector<std::thread> threads2(numThreads);

static void allocateAndWriteInBackground()
{
    for (size_t i = 0; i < numThreads; ++i)
    {
        threads1[i] = std::thread(
            []()
            {
                orc::DataBuffer<char> buffer(*orc::getDefaultPool(), bufferSize);
                for (size_t j = 0; j < bufferSize; ++j)
                    buffer[j] = static_cast<char>(j);
            });
    }

    for (auto & thread : threads1)
    {
        thread.join();
    }
}

static void BM_allocateAndWriteInBackground(benchmark::State & state)
{
    for (auto _ : state)
        allocateAndWriteInBackground();
}

static void writeInBackground()
{
    std::vector<std::shared_ptr<orc::DataBuffer<char>>> buffers(numThreads);
    for (size_t i = 0; i < numThreads; ++i)
    {
        auto buffer = std::make_shared<orc::DataBuffer<char>>(*orc::getDefaultPool(), bufferSize);
        buffers[i] = buffer;
        threads2[i] = std::thread(
            [&buffer]()
            {
                for (size_t j = 0; j < bufferSize; ++j)
                    (*buffer)[j] = static_cast<char>(j);
            });
    }

    for (auto & thread : threads2)
    {
        thread.join();
    }
}

static void BM_writeInBackground(benchmark::State & state)
{
    for (auto _ : state)
        writeInBackground();
}

BENCHMARK(BM_allocateAndWriteInBackground);
BENCHMARK(BM_writeInBackground);
BENCHMARK_MAIN();
```

``` bash
Running ./build_gcc/src/Columns/benchmarks/column_insert_many_from
Run on (32 X 2100 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x16)
  L1 Instruction 32 KiB (x16)
  L2 Unified 1024 KiB (x16)
  L3 Unified 11264 KiB (x2)
Load Average: 11.87, 43.43, 55.11
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
BM_allocateAndWriteInBackground    2122695 ns       631300 ns         1121
BM_writeInBackground               3096734 ns       529098 ns         1367
```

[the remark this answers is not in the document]
ffacs: Remind that user could use the buffer when it's
`virtual std::future<void> read(void* buf, uint64_t length, uint64_t offset) = 0;`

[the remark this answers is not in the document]
taiyang-li: In this pr, the buffers are all members of ReadRangeCache. The interface change doesn't influence the behavior of buffer reusing.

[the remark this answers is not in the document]
taiyang-li: Do you think providing both interfances is ok ?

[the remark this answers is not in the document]
taiyang-li: @wgtmac @ffacs I had renamed the old interface to `readAsyncInternal` and moved it under `protected`. User could not invoke it. What user could see is the new interface `readAsync`. And it has a default implementation. Users could override it if they are not satisfied with it.

c++/src/Reader.hh
```diff
@@ -352,10 +367,6 @@ namespace orc {
       return contents_->blockSize;
     }
 
-    const proto::Footer* getFooter() const {
```
wgtmac, line 355: Please do not delete this. It is not a public header.

c++/include/orc/Reader.hh
```diff
@@ -39,6 +39,8 @@ namespace orc {
...
+  class InputStream;
```
wgtmac, line 43: Please remove this line.

c++/include/orc/Reader.hh
wgtmac, line 42: `CacheOptions` is defined in src/Cache.hh, which is not installed so that downstream users do not have access to it. I'd suggest to choose one of these options:
1. Move the definition of CacheOptions to this file.
2. Directly adding two parameters to ReaderOptions and totally remove CacheOptions.

Personally I'd prefer the 2nd option.

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,137 @@
...
+    RangeCacheEntry(const ReadRange& range, std::future<BufferPtr> future)
+        : range(range), future(std::move(future).share()) {}
+
+    friend bool operator<(const RangeCacheEntry& left, const RangeCacheEntry& right) {
+      return left.range.offset < right.range.offset;
+    }
+  };
+
+  /// A read cache designed to hide IO latencies when reading.
+  class ReadRangeCache {
```
wgtmac, line 91: nit: add some cache related metrics to `ReaderMetrics` class.

c++/include/orc/OrcFile.hh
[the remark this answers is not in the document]
wgtmac: TBH, I don't think this penalty is unacceptable. In practice, the number of issued I/Os should not be too many after being coalesced. When the data is in a hot cache (e.g. SSD or memory), this feature may not improve too much and may get worse when predicate pushdown is effective (due to reading more chunks than needed). It is super helpful when data is in the cloud object store, in which case this penalty is trivial.

c++/include/orc/OrcFile.hh
```diff
@@ -58,10 +62,38 @@ namespace orc {
...
+      return std::async(std::launch::async,
+                        [this, buf, length, offset] { this->read(buf, length, offset); });
+    }
+
     /**
      * Get the name of the stream for error messages.
      */
     virtual const std::string& getName() const = 0;
+
+   protected:
```
wgtmac, line 82: `protected` is indeed visible to downstream users. BTW, it looks weird that we need to pass a memory pool to a read function. Therefore I think `readAsyncInternal` is ill-designed. Together with my comment to your comment of the benchmark, please remove this function and stick to `readAsync`.

c++/src/io/Cache.cc
```diff
@@ -0,0 +1,162 @@
...
+    return combiner.coalesce(std::move(ranges));
+  }
+
+  void ReadRangeCache::cache(std::vector<ReadRange> ranges) {
+    ranges = coalesceReadRanges(std::move(ranges), options_.holeSizeLimit, options_.rangeSizeLimit);
+
+    std::vector<RangeCacheEntry> newEntries = makeCacheEntries(ranges);
+    // Add new entries, themselves ordered by offset
+    if (entries_.size() > 0) {
+      std::vector<RangeCacheEntry> merged(entries_.size() + newEntries.size());
```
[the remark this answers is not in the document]
taiyang-li: That is exactly how `std::merge` works. You must resize for enough space instead of reserve.

c++/include/orc/Reader.hh
taiyang-li: I think `CacheOptions` being exposed to user is not a big deal.

wgtmac: OK. A side note is that perhaps this is more like a IO coalescing options since the cache instance is not visible to end user.

taiyang-li:
@wgtmac thanks for your advice. I had already finished the requested changes.

c++/src/StripeStream.cc
```diff
@@ -19,6 +19,7 @@
 #include "StripeStream.hh"
 #include "RLE.hh"
 #include "Reader.hh"
+#include "io/Cache.hh"
```
wgtmac, line 22: Please sort alphabetically.

c++/src/StripeStream.cc
```diff
@@ -37,7 +38,8 @@ namespace orc {
         stripeStart_(stripeStart),
         input_(input),
         writerTimezone_(writerTimezone),
-        readerTimezone_(readerTimezone) {
+        readerTimezone_(readerTimezone),
+        readCache_(reader.getReadCache()) {
```
wgtmac, line 42: You might directly call `RowReaderImpl.getFileContents()` to get readCache as suggested above.

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,133 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <orc/MemoryPool.hh>
+#include <orc/OrcFile.hh>
```
wgtmac, line 22:
```suggestion
#include "orc/MemoryPool.hh"
#include "orc/OrcFile.hh"
```

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,133 @@
...
+
+#include <algorithm>
+#include <cassert>
+#include <cstdint>
+#include <future>
+#include <utility>
+#include <vector>
+
+namespace orc {
+  class InputStream;
```
wgtmac, line 32:
```suggestion
```

c++/test/TestReader.cc
```diff
@@ -105,6 +105,15 @@ namespace orc {
                                                         nextSkippedRows));
   }
 
+  void prefetchAllStripes(Reader* reader) {
+    auto num_stripes = reader->getNumberOfStripes();
+    std::vector<int> stripes;
+    for (size_t i = 0; i < num_stripes; ++i) {
+      stripes.push_back(i);
+    }
+    reader->preBuffer(stripes, {0});
```
wgtmac, line 114: Please add test case where preBuffer is called for multiple times and with different stripe/column, etc.

wgtmac (changes requested):
Thanks for the quick update! Now the public interfaces look good.

I haven't carefully reviewed `Cache.hh/cc` yet. Hopefully I can make it by the end of this week.

cc @ffacs @luffy-zh

c++/src/io/Cache.cc
```diff
@@ -0,0 +1,172 @@
...
+                                     [](const RangeCacheEntry& entry, const ReadRange& range) {
+                                       return entry.range.offset + entry.range.length <
+                                              range.offset + range.length;
+                                     });
+
+    BufferSlice result{};
+    bool hit_cache = false;
+    if (it != entries_.end() && it->range.contains(range)) {
+      hit_cache = it->future.valid();
+      it->future.get();
```
luffy-zh, line 139: If `it->future.valid()` returns false, we might encounter an exception here.

wgtmac, line 139: Should we catch and rethrow an orc::Exception?
Should we use timeout here to fallback to direct read?

c++/src/Reader.hh
```diff
@@ -260,6 +268,13 @@ namespace orc {
     // footer
     proto::Footer* footer_;
     uint64_t numberOfStripes_;
+
+    // cached io ranges. only valid when preBuffer is invoked.
+    std::shared_ptr<ReadRangeCache> readCache_;
+
+    // mutex to protect readCache_ from concurrent access
+    std::mutex readCacheMutex_;
```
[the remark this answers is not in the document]
taiyang-li: I think this is the proper place of mutex. Because the mutex could not only protect the interfaces of `ReadRangeCache`, but also the creation of `ReadRangeCache`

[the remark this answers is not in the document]
wgtmac: Make sense. I have noticed this since last review but forgot to reply to my own comment.

c++/src/io/Cache.cc
taiyang-li: Yes. It is as expected, just like the behavior when reading orc file without prefetch.

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,132 @@
...
+
+    // The result may be get multiple times, so we use shared_future instead of std::future
+    BufferPtr buffer;
+    std::shared_future<void> future;
+
+    RangeCacheEntry() = default;
+    RangeCacheEntry(const ReadRange& range, BufferPtr buffer, std::future<void> future)
+        : range(range), buffer(std::move(buffer)), future(std::move(future).share()) {}
+
+    friend bool operator<(const RangeCacheEntry& left, const RangeCacheEntry& right) {
```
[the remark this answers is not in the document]
taiyang-li: Without friend, we have to move it out of RangeCacheEntry code block, which is not tight.

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,132 @@
...
+  struct RangeCacheEntry {
+    using BufferPtr = InputStream::BufferPtr;
+
+    ReadRange range;
+
+    // The result may be get multiple times, so we use shared_future instead of std::future
+    BufferPtr buffer;
+    std::shared_future<void> future;
+
+    RangeCacheEntry() = default;
```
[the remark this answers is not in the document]
taiyang-li: It shouldn't.  Because default constructor is needed in `ReadRangeCache::cache`

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,132 @@
...
+    const uint64_t holeSizeLimit;
+    const uint64_t rangeSizeLimit;
+
+    std::vector<ReadRange> coalesce(std::vector<ReadRange> ranges) const;
+  };
+
+  std::vector<ReadRange> coalesceReadRanges(std::vector<ReadRange> ranges, uint64_t holeSizeLimit,
+                                            uint64_t rangeSizeLimit);
+  struct RangeCacheEntry {
+    using BufferPtr = InputStream::BufferPtr;
```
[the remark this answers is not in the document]
taiyang-li: I'm afraid it is a bit ugly with too many level of template.

c++/src/io/Cache.cc
```diff
@@ -0,0 +1,172 @@
...
+    auto itr = ranges.begin();
+    // Ensure ranges is not empty.
+    assert(itr <= ranges.end());
+
+    // Start of the current coalesced range and end (exclusive) of previous range.
+    // Both are initialized with the start of first range which is a placeholder value.
+    uint64_t coalescedStart = itr->offset;
+    uint64_t coalescedEnd = coalescedStart + itr->length;
+
+    for (++itr; itr < ranges.end(); ++itr) {
```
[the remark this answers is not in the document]
taiyang-li: I'm not sure we can keep the same semantics if we merge the 3 loops. Later loops rely on previous loops, and sometimes multiple elements are processed at one time. Maybe you are right, I just personally think it is not worth putting efforts in a performance insensitive place.

c++/src/io/Cache.cc
taiyang-li: Maybe keeping the behavior consistency with `InputStream::read` is better? If `InputStream::read` throws an exception, it is transmitted directly to user.

c++/include/orc/Reader.hh
```diff
@@ -624,6 +647,24 @@ namespace orc {
...
         uint32_t stripeIndex, const std::set<uint32_t>& included = {}) const = 0;
+
+    /**
+     * Trigger IO prefetch and cache the prefetched contents asynchronously.
+     * It is thread safe. Users should make sure requested stripes and columns
+     * are not overlapped, otherwise the overlapping part will be prefetched multiple time,
+     * which doesn't affect correctness but waste IO and memory resources.
+     * @param stripes the stripes to prefetch
+     * @param includeTypes the types to prefetch
+     * @param options the cache options for prefetched contents
```
wgtmac, line 658:
```suggestion
```

c++/include/orc/Reader.hh
```diff
@@ -624,6 +647,24 @@ namespace orc {
...
+     */
+    virtual void preBuffer(const std::vector<int>& stripes,
```
wgtmac, line 660:
```suggestion
    virtual void preBuffer(const std::vector<uint32_t>& stripes,
```

Please use explicit integer type.

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,129 @@
...
+
+  using Buffer = DataBuffer<char>;
+  using BufferPtr = std::shared_ptr<Buffer>;
+
+  struct RangeCacheEntry {
+    ReadRange range;
+
+    // The result may be get multiple times, so we use shared_future instead of std::future
+    BufferPtr buffer;
+    std::shared_future<void> future;
```
wgtmac, line 71:
```suggestion
    ReadRange range;
    BufferPtr buffer;
    std::shared_future<void> future;  // use shared_future in case of multiple get calls
```

c++/src/io/Cache.hh
```diff
@@ -0,0 +1,129 @@
...
+  struct BufferSlice {
+    BufferSlice() : buffer(nullptr), offset(0), length(0) {}
+
+    BufferSlice(BufferPtr buffer, uint64_t offset, uint64_t length)
+        : buffer(std::move(buffer)), offset(offset), length(length) {}
+
+    BufferPtr buffer;
+    uint64_t offset;
+    uint64_t length;
+  };
```
wgtmac, line 91:
```suggestion
  struct BufferSlice {
    BufferPtr buffer = nullptr;
    uint64_t offset = 0;
    uint64_t length = 0;
  };
```

c++/test/TestReader.cc
```diff
@@ -218,7 +239,9 @@ namespace orc {
     auto inStream = std::make_unique<MemoryInputStream>(memStream.getData(), memStream.getLength());
     ReaderOptions readerOptions;
     readerOptions.setMemoryPool(*pool);
-    return createReader(std::move(inStream), readerOptions);
+    auto reader = createReader(std::move(inStream), readerOptions);
+    prefetchStripesAndColumnsRandomly(reader.get(), *type);
```
wgtmac, line 243: Could you write new test cases for prefetching instead of mixing the existing cases? Perhaps it requires some refactoring work to share the common code. In addition, I'd prefer clear parameters instead of (pseudo) random ones.

wgtmac (commented):
Thanks for the changes! It generally looks good. I think the test cases need to be polished.

c++/test/TestReader.cc
taiyang-li: @wgtmac Thanks for your advice. I'll try.

taiyang-li:
> Thanks for the changes! It generally looks good. I think the test cases need to be polished.

Thanks for detailed reviews. I had improved the test cases.

wgtmac (approved):
LGTM. Thanks!

ffacs (approved):
LGTM. Thank you  @wgtmac @taiyang-li ~

dongjoon-hyun:
Merged to main for Apache ORC 2.1.0 on January 2025.

I added you to the Apache ORC contributor group and assigned ORC-252 to you, @taiyang-li .

Also, I updated ORC-1767 JIRA issue by assigning to you.

Thank you and welcome to the Apache ORC community again!

taiyang-li:
Thank you very much @dongjoon-hyun, I am very happy to join the Apache ORC Contributor Group. Apache Gluten relies heavily on this library, and I'm looking forward to contributing more in the future.

Issue #2079, opened 2024-12-09
Incompatibility with ORC files written in version 0.12 due to missing hasNull field in C++ Reader


suxiaogang223:
# Issue Description:
>From Hive 1.1.0 onwards, the column statistics will also record if there are any null values within the row group by setting the hasNull flag. The hasNull flag is used by ORC’s predicate pushdown to better answer ‘IS NULL’ queries.

We encountered an issue with the C++ implementation of the ORC reader when handling ORC files written with version 0.12. Specifically, files written in this version do not include the hasNull field in the column statistics metadata. While the Java implementation of the ORC reader handles this gracefully by defaulting hasNull to true when the field is absent, the C++ implementation does not handle this scenario correctly.
**This issue prevents predicates like IS NULL from being pushed down to the ORC reader!!! As a result, all rows in the file are filtered out, leading to incorrect query results :(**
# Steps to Reproduce:
1.	Use an ORC file written with ORC version 0.12 (without hasNull in its column statistics).
2.	Attempt to read the file using the C++ ORC reader.
## Expected Behavior:
The C++ ORC reader should default the hasNull field to true when it is absent, ensuring compatibility with older file versions.
## Observed Behavior:
The C++ ORC reader default the hasNull field to false, resulting in incorrect metadata interpretation.

# Comparison with Java Implementation:
The Java implementation includes the following logic:
```java
if (stats.hasHasNull()) {
    hasNull = stats.getHasNull();
} else {
    hasNull = true;
}
```
In contrast, the C++ implementation directly uses the has_null value without any fallback logic:
```c++
ColumnStatisticsImpl::ColumnStatisticsImpl(const proto::ColumnStatistics& pb) {
    stats_.setNumberOfValues(pb.number_of_values());
    stats_.setHasNull(pb.has_null());
}
```
# Suggested Fix:
Introduce fallback logic in the C++ reader to set hasNull to true when the field is missing, similar to the Java implementation.

ffacs:
HI @suxiaogang223 , there is a patch https://github.com/apache/orc/pull/2055 to disable PPD when `pb.has_hasnull() == false`, but `ColumnStatistics::hasNull` still return false when the field in pb is missing. It Seems your patch would make it return a right value, would you like to make a PR? CC @wgtmac

suxiaogang223:
[...]

PR #2055, opened 2024-10-25
ORC-2054: [C++] fix return wrong result if lack of hasnull

shuai-xu:
### What changes were proposed in this pull request?
This pr fix the bug that if the column statistics in a orc file is not fully written, and lack of hasnull field, user may get a wrong result using c++ to read it.
For example, a file struct<string col1, string col2>, has 10 lines, col1 all has value, col2 all is null. the column 1's stat written by trino may be 
numberOfValues: 10
stringStatistics {
  minimum: "10"
  maximum: "100"
  sum: 565
}. col2's stat is  numberOfValues: 0. They all have no hasnull field. When we want to get where col2 is null, we will get nothing.

### Why are the changes needed?
User may get a wrong result with this bug.

### How was this patch tested?
Add unit tests.

### Was this patch authored or co-authored using generative AI tooling?
No

PR #2082, opened 2024-12-10
fix has_null forward compatibility

suxiaogang223:
cose issue: #2079 
relate pr: #2055
Introduce fallback logic in the C++ reader to set hasNull to true when the field is missing, similar to the Java implementation.
The Java implementation includes the following logic:
```java
if (stats.hasHasNull()) {
    hasNull = stats.getHasNull();
} else {
    hasNull = true;
}
```
In contrast, the C++ implementation directly uses the has_null value without any fallback logic:
```c++
ColumnStatisticsImpl::ColumnStatisticsImpl(const proto::ColumnStatistics& pb) {
    stats_.setNumberOfValues(pb.number_of_values());
    stats_.setHasNull(pb.has_null());
}
```
We encountered an issue with the C++ implementation of the ORC reader when handling ORC files written with version 0.12. Specifically, files written in this version do not include the hasNull field in the column statistics metadata. While the Java implementation of the ORC reader handles this gracefully by defaulting hasNull to true when the field is absent, the C++ implementation does not handle this scenario correctly.
**This issue prevents predicates like IS NULL from being pushed down to the ORC reader!!! As a result, all rows in the file are filtered out, leading to incorrect query results :(**
I have tested this using [Doris](https://github.com/apache/doris) external pipeline: 
https://github.com/apache/doris/pull/45104
https://github.com/apache/doris-thirdparty/pull/259

No

### DIFF
--- a/c++/src/sargs/PredicateLeaf.cc
+++ b/c++/src/sargs/PredicateLeaf.cc
@@ -390,10 +390,15 @@ namespace orc {
 
   DIAGNOSTIC_POP
 
+  static bool col_has_null_forward_compatible(const proto::ColumnStatistics& stats) {
+    // for foward compatibility, if has_null is not set, assume that the column has nulls
+    return stats.has_has_null() ? stats.has_null() : true;
+  }
+
   static TruthValue evaluateBoolPredicate(const PredicateLeaf::Operator op,
                                           const std::vector<Literal>& literals,
                                           const proto::ColumnStatistics& stats) {
-    bool hasNull = stats.has_null();
+    bool hasNull = col_has_null_forward_compatible(stats);
     if (!stats.has_bucket_statistics() || stats.bucket_statistics().count_size() == 0) {
       // does not have bool stats
       return hasNull ? TruthValue::YES_NO_NULL : TruthValue::YES_NO;
@@ -512,8 +517,9 @@ namespace orc {
         if (colStats.has_int_statistics() && colStats.int_statistics().has_minimum() &&
             colStats.int_statistics().has_maximum()) {
           const auto& stats = colStats.int_statistics();
-          result = evaluatePredicateRange(operator_, literal2Long(literals_), stats.minimum(),
-                                          stats.maximum(), colStats.has_null());
+          result =
+              evaluatePredicateRange(operator_, literal2Long(literals_), stats.minimum(),
+                                     stats.maximum(), col_has_null_forward_compatible(colStats));
         }
         break;
       }
@@ -522,10 +528,12 @@ namespace orc {
             colStats.double_statistics().has_maximum()) {
           const auto& stats = colStats.double_statistics();
           if (!std::isfinite(stats.sum())) {
-            result = colStats.has_null() ? TruthValue::YES_NO_NULL : TruthValue::YES_NO;
+            result = col_has_null_forward_compatible(colStats) ? TruthValue::YES_NO_NULL
+                                                               : TruthValue::YES_NO;
           } else {
-            result = evaluatePredicateRange(operator_, literal2Double(literals_), stats.minimum(),
-                                            stats.maximum(), colStats.has_null());
+            result =
+                evaluatePredicateRange(operator_, literal2Double(literals_), stats.minimum(),
+                                       stats.maximum(), col_has_null_forward_compatible(colStats));
           }
         }
         break;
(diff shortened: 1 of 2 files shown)


dongjoon-hyun:
Thank you for making a PR, @suxiaogang223 .

suxiaogang223:
I have fix some orc c++ uts because the orc file orc_index_int_string.orc is written by old orc version which don't write the has_null statistics.

c++/src/sargs/PredicateLeaf.cc
```diff
@@ -390,10 +390,15 @@ namespace orc {
 
   DIAGNOSTIC_POP
 
+  static bool col_has_null_forward_compatible(const proto::ColumnStatistics& stats) {
```
ffacs, line 393: Please rename the function in camel case

ffacs (changes requested):

c++/src/sargs/PredicateLeaf.cc
```diff
@@ -390,10 +390,15 @@ namespace orc {
 
   DIAGNOSTIC_POP
 
+  static bool colHasNullForwardCompatible(const proto::ColumnStatistics& stats) {
```
wgtmac, line 393: Why calling it `forward compatible`? It seems to be a bug from a legacy writer IMHO.

wgtmac, line 393:
```suggestion
  static bool mayHaveNull(const proto::ColumnStatistics& stats) {
```

c++/src/sargs/PredicateLeaf.cc
```diff
@@ -701,24 +707,22 @@ namespace orc {
       }
     }
 
-    // files written by trino may lack of hasnull field.
-    if (!colStats.has_has_null()) return TruthValue::YES_NO_NULL;
```
wgtmac, line 705: Is it better to keep this line so lines below do not need to change?

c++/test/TestPredicateLeaf.cc
```diff
@@ -672,7 +672,7 @@ namespace orc {
   TEST(TestPredicateLeaf, testLackOfSataistics) {
     PredicateLeaf pred(PredicateLeaf::Operator::IS_NULL, PredicateDataType::STRING, 1, {});
     EXPECT_EQ(TruthValue::YES_NO, evaluate(pred, createStringStats("c", "d", true)));
-    EXPECT_EQ(TruthValue::YES_NO_NULL, evaluate(pred, createIncompleteNullStats()));
+    EXPECT_EQ(TruthValue::YES, evaluate(pred, createIncompleteNullStats()));
```
wgtmac, line 675: IIRC, this is wrong because `not(is_null(x))` returns `TruthValue::NO_NULL` after this change. For malformed stats, we should return `YES_NO_NULL` since we cannot do anything.

wgtmac:
Thanks for fixing this! Could you create a JIRA issue? We use JIRA instead of GitHub issue for tracking purpose.

suxiaogang223:
> Thanks for fixing this! Could you create a JIRA issue? We use JIRA instead of GitHub issue for tracking purpose.

How to create jira?  it seems that your jira system cannot be used :(

wgtmac:
You may request a JIRA account via: https://selfserve.apache.org/jira-account.html

It was open for registration. Unfortunately it has been disabled due to spam :(

dongjoon-hyun:
Thank you for fling JIRA and update the PR title, @suxiaogang223 .

BTW, does it affect all ORC versions like 1.9.x and 1.8.x? The JIRA issue seems to be reported on Apache ORC 2.0.x only.

dongjoon-hyun:
For now, I set the milestone, 2.0.4, based on `Affected Versions` of ORC-1813.

c++/src/sargs/PredicateLeaf.cc
suxiaogang223: It seems better

ffacs (approved):
LGTM. Thank you @suxiaogang223 @wgtmac @dongjoon-hyun ~

wgtmac (approved):
+1

dongjoon-hyun (approved):

dongjoon-hyun:
Thank you all, @suxiaogang223 , @wgtmac , @ffacs !

Merged to `main`.

Could you make a backporting PR to `branch-2.0`? There was a conflict.

suxiaogang223:
> Thank you all, @suxiaogang223 , @wgtmac , @ffacs !
> 
> Merged to `main`.
> 
> Could you make a backporting PR to `branch-2.0`? There was a conflict.

There is a conflict fixed pr : https://github.com/apache/orc/pull/2086

dongjoon-hyun:
For the record, the following landed to branch-2.0.
- https://github.com/apache/orc/pull/2086

PR #2086, opened 2024-12-16
ORC-1813: [C++] Fix has_null forward compatibility

suxiaogang223:
close: #2079
relate pr: #2055
Introduce fallback logic in the C++ reader to set hasNull to true when the field is missing, similar to the Java implementation. The Java implementation includes the following logic:
```java
if (stats.hasHasNull()) {
    hasNull = stats.getHasNull();
} else {
    hasNull = true;
}
```
In contrast, the C++ implementation directly uses the has_null value without any fallback logic:
```c++
ColumnStatisticsImpl::ColumnStatisticsImpl(const proto::ColumnStatistics& pb) {
    stats_.setNumberOfValues(pb.number_of_values());
    stats_.setHasNull(pb.has_null());
}
```
We encountered an issue with the C++ implementation of the ORC reader when handling ORC files written with version 0.12. Specifically, files written in this version do not include the hasNull field in the column statistics metadata. While the Java implementation of the ORC reader handles this gracefully by defaulting hasNull to true when the field is absent, the C++ implementation does not handle this scenario correctly. **This issue prevents predicates like IS NULL from being pushed down to the ORC reader!!! As a result, all rows in the file are filtered out, leading to incorrect query results :(** I have tested this using [Doris](https://github.com/apache/doris) external pipeline: https://github.com/apache/doris/pull/45104
https://github.com/apache/doris-thirdparty/pull/259 No

Closes #2082 from suxiaogang223/fix_has_null.

Authored-by: Socrates <suxiaogang223@icloud.com>


wgtmac (approved):

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
Thank you for this backporting, @suxiaogang223 and @wgtmac .
Merged to branch-2.0.

PR #2094, opened 2024-12-24
ORC-1810: [C++] Add environment variable ORC_FORMAT_URL

luffy-zh:

Add an CMake option ORC_FORMAT_URL to indicate where to download the orc-format_ep

Handle the issue discussed https://github.com/apache/arrow/pull/43417

Test it locally

NO

### DIFF
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -33,6 +33,7 @@ option(ORC_PREFER_STATIC_LZ4      "Prefer static lz4 library, if available"     
 option(ORC_PREFER_STATIC_ZSTD     "Prefer static zstd library, if available"     ON)
 option(ORC_PREFER_STATIC_ZLIB     "Prefer static zlib library, if available"     ON)
 option(ORC_PREFER_STATIC_GMOCK    "Prefer static gmock library, if available"    ON)
+
 # zstd requires us to add the threads
 FIND_PACKAGE(Threads REQUIRED)
 
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -33,7 +33,8 @@ option(ORC_PREFER_STATIC_LZ4      "Prefer static lz4 library, if available"     
 option(ORC_PREFER_STATIC_ZSTD     "Prefer static zstd library, if available"     ON)
 option(ORC_PREFER_STATIC_ZLIB     "Prefer static zlib library, if available"     ON)
 option(ORC_PREFER_STATIC_GMOCK    "Prefer static gmock library, if available"    ON)
-
+set(DEFAULT_ORC_FORMAT_URL "https://www.apache.org/dyn/closer.lua/orc/orc-format-${ORC_FORMAT_VERSION}/orc-format-${ORC_FORMAT_VERSION}.tar.gz?action=download")
+set(ORC_FORMAT_URL ${DEFAULT_ORC_FORMAT_URL} CACHE STRING "URL from which to download ORC format library" FORCE)
 # zstd requires us to add the threads
 FIND_PACKAGE(Threads REQUIRED)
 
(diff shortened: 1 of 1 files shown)


cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -136,8 +137,9 @@ endfunction()
 
 # ----------------------------------------------------------------------
 # ORC Format
+message(STATUS "Using ORC_FORMAT_URL: ${ORC_FORMAT_URL}")
 ExternalProject_Add (orc-format_ep
-  URL "https://www.apache.org/dyn/closer.lua/orc/orc-format-${ORC_FORMAT_VERSION}/orc-format-${ORC_FORMAT_VERSION}.tar.gz?action=download"
+  URL ${ORC_FORMAT_URL}
   URL_HASH SHA256=739fae5ff94b1f812b413077280361045bf92e510ef04b34a610e23a945d8cd5
```
wgtmac, line 143: User provided URL may not match the `URL_HASH` below. We should probably avoid passing URL_HASH to ExternalProject_Add

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -33,7 +33,8 @@ option(ORC_PREFER_STATIC_LZ4      "Prefer static lz4 library, if available"
 option(ORC_PREFER_STATIC_ZSTD     "Prefer static zstd library, if available"     ON)
 option(ORC_PREFER_STATIC_ZLIB     "Prefer static zlib library, if available"     ON)
 option(ORC_PREFER_STATIC_GMOCK    "Prefer static gmock library, if available"    ON)
-
+set(DEFAULT_ORC_FORMAT_URL "https://www.apache.org/dyn/closer.lua/orc/orc-format-${ORC_FORMAT_VERSION}/orc-format-${ORC_FORMAT_VERSION}.tar.gz?action=download")
```
wgtmac, line 36: What about moving it to below line 139?

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -33,7 +33,8 @@ option(ORC_PREFER_STATIC_LZ4      "Prefer static lz4 library, if available"
...
+set(ORC_FORMAT_URL ${DEFAULT_ORC_FORMAT_URL} CACHE STRING "URL from which to download ORC format library" FORCE)
```
wgtmac, line 37: What about using environment variable?

```
if(DEFINED ENV{ORC_FORMAT_URL})
  set(ORC_FORMAT_SOURCE_URL "$ENV{ORC_FORMAT_URL}")
else()
  set(ORC_FORMAT_SOURCE_URL
    "https://www.apache.org/dyn/closer.lua/orc/orc-format-${ORC_FORMAT_VERSION}/orc-format-${ORC_FORMAT_VERSION}.tar.gz?action=download"
  )
endif()
```

dongjoon-hyun (changes requested):
Thank you, @luffy-zh .

- I'm also +1 for the above @wgtmac 's comment to use env variable.
- However, I'm -1 for allowing a different `URL_HASH` code because it could expose a vulnerability.

wgtmac (approved):
+1 Thanks @luffy-zh!

cc @kou

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -33,7 +33,6 @@ option(ORC_PREFER_STATIC_LZ4      "Prefer static lz4 library, if available"
 option(ORC_PREFER_STATIC_ZSTD     "Prefer static zstd library, if available"     ON)
 option(ORC_PREFER_STATIC_ZLIB     "Prefer static zlib library, if available"     ON)
 option(ORC_PREFER_STATIC_GMOCK    "Prefer static gmock library, if available"    ON)
-
```
kou, line 36: We may want to revert this needless change.

kou (approved):
+1

Thanks!

FYI: Apache Arrow accepts not only source URL but also checksum. It's for accepting newer/older versions than pre-defined versions at https://github.com/apache/arrow/blob/main/cpp/thirdparty/versions.txt . Apache Arrow uses `ExternalProject_Add()`/`FetchContent_Declare()` for optional external libraries. But Apache ORC uses `ExternalProject_Add()` for mandatory orc-format dependency. So situations in Apache Arrow and Apache ORC may be different.

dongjoon-hyun (approved):
+1, LGTM. Thank you, @luffy-zh and all.

Happy New Year.

Merged to main for Apache ORC 2.1.

PR #2108, opened 2025-01-13
ORC-1833: [C++] Fix CMake script to be used inside another project

wgtmac:

The change to add support for exporting CMake config and target has introduced some minor issues when the ORC C++ library is used inside a larger CMake project. Mainly there are three issues:

- We should prepend (not append) our CMake modules so we always use our own modules when there are naming conflict.
- Do not use CMAKE_SOURCE_DIR and CMAKE_BINARY_DIR because they are tied to the root project and it is no longer ORC project when we are inside another project.
- Add orc_ prefix to our CMake functions to avoid potential conflict.

See: https://github.com/apache/arrow/pull/45226

We had a problem with upgrading Apache ORC to 2.1.0 in the Apache Arrow. See: https://github.com/apache/arrow/pull/45226

- Pass all CIs.
- Manually integrated it with Apache Arrow. 

No.


CMakeLists.txt
```diff
@@ -85,6 +85,10 @@ option(ORC_PACKAGE_KIND
     "Arbitrary string that identifies the kind of package"
     "")
 
+option(ORC_ENABLE_CLANG_TOOLS
```
ffacs, line 88: Maybe we should check `CMAKE_CXX_COMPILER_ID MATCHES "Clang"` when the option is ON

wgtmac: I think they are unrelated. Sometimes I want to use gcc to build but Clang is available on my machine to do the checking.

ffacs: > I think they are unrelated. Sometimes I want to use gcc to build but Clang is available on my machine to do the checking.

Fine, but there are some compile options used by GCC that clang did not understand.

ffacs (approved):
LGTM.

CMakeLists.txt
wgtmac: Actually I think we can disable it by default because we have switched to use `cpp-linter-action` already. But this PR is required to backport to 2.1.1 so I decided to keep it to make the CI happy on branch-2.1

wgtmac:
cc @kou

c++/src/CMakeLists.txt
```diff
@@ -218,8 +218,8 @@ target_include_directories (orc
   INTERFACE
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
   PUBLIC
-    $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/c++/include>
-    $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/c++/include>
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/c++/include>
+    $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/c++/include>
```
kou, line 222: This is not related to this PR but... In general, build directory -> source directory order is better.

kou (approved):
+1

wgtmac:
Thanks for reviewing it! @ffacs @kou

dongjoon-hyun (commented):
+1, LGTM. Thank you, @wgtmac and @ffacs .

cc @williamhyun

PR #2121, opened 2025-01-21
ORC-1846: [C++] Fix imported libraries in the Conan build

wgtmac:

Fix all imported libraries of thirdparty dependencies to build in Conan.

Upgrading the C++ ORC 2.1.0 in Conan will fail due to the CMake refactoring. See https://github.com/conan-io/conan-center-index/pull/26426

See CIs from Conan: https://github.com/conan-io/conan-center-index/pull/26426/checks?check_run_id=35916226382

No.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,7 @@ option(ORC_PACKAGE_KIND
 
 option(ORC_ENABLE_CLANG_TOOLS
     "Enable Clang tools"
-    ON)
+    OFF)
 
 # Make sure that a build type is selected
 if (NOT CMAKE_BUILD_TYPE)
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -156,7 +156,8 @@ ExternalProject_Add (orc-format_ep
 # Snappy
 if (ORC_PACKAGE_KIND STREQUAL "conan")
   find_package (Snappy REQUIRED CONFIG)
-  orc_add_resolved_library (orc_snappy ${Snappy_LIBRARIES} ${Snappy_INCLUDE_DIR})
+  add_library (orc_snappy INTERFACE)
+  target_link_libraries(orc_snappy INTERFACE Snappy::snappy)
   list (APPEND ORC_SYSTEM_DEPENDENCIES Snappy)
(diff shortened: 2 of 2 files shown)


ffacs (approved):
+1

wgtmac:
Thanks for the review @ffacs !

CMakeLists.txt
```diff
@@ -87,7 +87,7 @@ option(ORC_PACKAGE_KIND
 
 option(ORC_ENABLE_CLANG_TOOLS
     "Enable Clang tools"
-    ON)
+    OFF)
```
dongjoon-hyun, line 90: Hi, @wgtmac .

May I ask why we need to disable this for all build?
I guess `ORC_PACKAGE_KIND` has the information of `conan`.

Can we narrow down our change into the specific `conan` build?

dongjoon-hyun:
Thank you for making a PR and backporting it. 

I left a comment in the above to remove any side-effects on `non-Conan` build. Please let me know your opinion, @wgtmac .

CMakeLists.txt
wgtmac: `ORC_ENABLE_CLANG_TOOLS` was required by the previous GitHub workflow for clang-tidy and clang-format enforcement. Now it is replaced by the `cpp-linter-action` and is no longer needed. At the same time, it defines custom targets like `check-clang-tidy` and `check-format` which may have name conflict with other projects (like Apache Arrow). In many build environments (like Conan or Github CIs) usually these tools are missing and result in annoying warning messages. So I think it makes sense to disable it by default.

dongjoon-hyun: It sounds valid but independent from this PR. Maybe, could you handle them seperately next time please?

dongjoon-hyun: BTW, do you think we need Apache ORC 2.1.1 for downstreams to deliver this? I'm wondering if this is the last one to help them.

wgtmac: I'll keep it in mind next time :)

> BTW, do you think we need Apache ORC 2.1.1 for downstreams to deliver this? I'm wondering if this is the last one to help them.

I don't think we need it yet and we can still keep the release cadence for now. Let's wait for feedback from the Conan community. But releasing 2.1.1 is helpful anyway.

PR #2185, opened 2025-04-12
ORC-1875: Support `ubuntu-24.04-arm` in GitHub Action CIs

dongjoon-hyun:

This PR aims to support `ubuntu-24.04-arm` in GitHub Action CIs.

`Linux` arm64 architecture becomes public preview as a new GitHub Action runner. We had better utilize this resource in our test infra.
- https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Pass the CIs.

No.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -67,6 +67,7 @@ jobs:
         os:
           - ubuntu-22.04
           - ubuntu-24.04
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Merged to main/2.1.

PR #2223, opened 2025-05-10
ORC-1898: When column is all null, NULL_SAFE_EQUALS pushdown doesn't get evaluated correctly

jayhan94:

When all values in column `col_0` are `NULL`s within a row group, and we attempt to apply the predicate pushdown `col_0 <=> 'xxx'`, the `evaluatePredicateProto` function returns `TruthValue.NULL`. In this case, we can directly determine the result based on the literal value: if the literal is `NULL`, return `TruthValue.YES`, otherwise, return `TruthValue.NO`.

See [SPARK-52032](https://issues.apache.org/jira/projects/SPARK/issues/SPARK-52032).
When we pushdown the NULL_SAFE_EQUALS predicate, all values of the column are `NULL`. The `evaluatePredicateProto` returns `TruthValue.NULL`, whose `isNeeded` returns false so that the whole row group is skipped by `SargApplier.pickRowGroups`, which actually is incorrect.

There already exists unit test -- `TestOrcTimezonePPD.testTimestampAllNulls`

Co-authored using generative AI tooling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ dependency-reduced-pom.xml
 *.ipr
 *.iws
 .idea
-.vscode
 .DS_Store
 .java-version
 java/bench/data
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ dependency-reduced-pom.xml
 *.ipr
 *.iws
 .idea
+.vscode
 .DS_Store
 .java-version
 java/bench/data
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -9,6 +9,7 @@ dependency-reduced-pom.xml
 *.ipr
 *.iws
 .idea
+.vscode
```
dongjoon-hyun, line 12: Please remove this because this is orthogonal topic. You can make another PR for this.

dongjoon-hyun:
Thank you for making a PR, @jayhan94 .

.gitignore
jayhan94: I've removed it.

dongjoon-hyun (approved):
+1, LGTM.

Thank you, @jayhan94 and @wgtmac .

Sorry for being later. I was a little busy until Today.

dongjoon-hyun:
Thank you for your first contribution @jayhan94 . I added you to the Apache ORC contributor group and assigned ORC-1898 to you. Welcome to the Apache ORC community.

dongjoon-hyun:
This is merged to all live release branches for Apache ORC 2.2/2.1/2.0/1.9/1.8.

PR #2240, opened 2025-05-24
Fix CMake modules for finding external libraries

lahwaacz:

- allow finding header files in default paths - this is consistent with the logic for finding the library files
- replace undefined variables `${SNAPPY_LIB_NAME}` and `${LZ4_LIB_NAME}` with the actual names

The CMake modules are broken. Building anything that depends on orc using cmake (e.g. Apache Arrow) does not work.

Better than the original files :smirk: 

No


wgtmac:
Thanks for creating this PR!

> The CMake modules are broken. Building anything that depends on orc using cmake (e.g. Apache Arrow) does not work.

Could you elaborate this? I was thinking to remove these obsolete `FindXXX.cmake` files to use config mode of find_package so we don't have to maintain them.

lahwaacz:
> Could you elaborate this?

Try to build the latest orc version and then arrow using the following cmake options (notice especially `-DARROW_DEPENDENCY_SOURCE=SYSTEM` and `-DARROW_ORC=ON`)

https://gitlab.archlinux.org/archlinux/packaging/packages/arrow/-/blob/main/PKGBUILD?ref_type=heads#L80-120

> I was thinking to remove these obsolete `FindXXX.cmake` files to use config mode of find_package so we don't have to maintain them.

The "config mode" is still looking for some `.cmake` file provided by the library you are finding. Many packages do that, including protobuf, snappy, and zstd, but some don't (e.g. lz4, zlib). But cmake also has its standard `FindZLIB` module so you don't have to reimplement your own.

wgtmac:
I think lz4 and zlib have exported their packages to be used by config mode:
- https://github.com/lz4/lz4/blob/dev/build/cmake/lz4Config.cmake.in
- https://github.com/madler/zlib/blob/develop/CMakeLists.txt#L240

lahwaacz:
Apparently this is not relevant after https://github.com/apache/orc/pull/2416

PR #2261, opened 2025-06-09
ORC-1913: Fix `TestColumnStatistics` to set `testFilePath` with absolute path

dongjoon-hyun:

This PR aims to fix `TestColumnStatistics` to set `testFilePath` with absolute path.

The test case is fragile because it depends on the assumption that new FileSystem will be identical.

For example, it fails when we set `fs.file.impl.disable.cache=false`.

```
$ git diff
...
     conf = new Configuration();
+    conf.set("fs.file.impl.disable.cache", "true");
     fs = FileSystem.getLocal(conf);

$ mvn package --pl core -Dtest=TestColumnStatistics
...
[ERROR] Errors:
[ERROR]   TestColumnStatistics.testDecimalMinMaxStatistics:695 » FileNotFound File TestOrcFile.testDecimalMinMaxStatistics.orc does not exist
```

Pass the CIs with the updated code.

No.

### DIFF
--- a/java/core/src/test/org/apache/orc/TestColumnStatistics.java
+++ b/java/core/src/test/org/apache/orc/TestColumnStatistics.java
@@ -753,9 +753,9 @@ public class TestColumnStatistics {
   @BeforeEach
   public void openFileSystem(TestInfo testInfo) throws Exception {
     conf = new Configuration();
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Merged to main/2.1.

PR #2292, opened 2025-06-20
ORC-1932: Use `setIfUnset` for `fs.defaultFS` and `fs.file.impl.disable.cache`

dongjoon-hyun:

This PR aims to use `setIfUnset` instead of `set` for `fs.defaultFS` and `fs.file.impl.disable.cache` in `TestConf` and `tools` module.

These are newly added to Apache ORC 2.2, and `setIfUnset` is better because it respects the existing setting.

Pass the CIs (including Java 25-ea CI)

No.


dongjoon-hyun:
Merged to main.

PR #2293, opened 2025-06-20
ORC-1933: Change `org.jetbrains:annotations` dependency to the `provided` scope

dongjoon-hyun:

This PR aims to lower `org.jetbrains:annotations` dependency from `compile` scope to `provided` scope for Apache ORC 2.2.0.

From now, we can update it to the latest version because this doesn't affect the downstream. So, this PR also
- Remove the pinning the dependency.
- Upgrade it to the latest version

To simplify Apache ORC dependency because this is not required at the runtime.

Pass the CIs.

No.


dongjoon-hyun:
Merged to main for Apache ORC 2.2.0.

PR #2298, opened 2025-06-23
ORC-1937: Remove generating release build in Meson config by default

WillAyd:
This removes setting the Meson configuration to remove the buildtype of release by default. 

While coming from CMake it may seem like a good idea to build release builds by default, the setting is either ignored or problematic when orc is used as a subproject in other projects, especially on Windows where a main project building a library with debug symbols may fail if a subproject tries to force release symbols.

Compiled locally

No

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -252,6 +252,6 @@ jobs:
         pip install meson
     - name: Test
       run: |
-        meson setup build
+        meson setup build -Dbuildtype=release
         meson compile -C build
         meson test -C build
--- a/README.md
+++ b/README.md
@@ -126,10 +126,10 @@ meson compile -C build
 meson test -C build
 ```
 
-By default, Meson will build release libraries. If you would instead like debug libraries, you can use the ``-Dbuildtype=debug`` argument to setup. You must either remove the existing build directory before changing that setting, or alternatively pass the ``--reconfigure`` flag:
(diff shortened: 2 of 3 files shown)


dongjoon-hyun (changes requested):
Well, I'm a little negative.

Does Mesos has no corresponding feature like `RELWITHDEBINFO` of CMake?

WillAyd:
Yes it does - that would be the debugoptimized build type 

https://mesonbuild.com/Builtin-options.html#core-options

dongjoon-hyun:
Could you provide some reference what is the standard (or recommended way) in Meson? `debug` or `debugoptimized` ?

In addition, I'm wondering if Mesos eco-system is enforcing `debug` for all libraries or not? To me, it's a little weird that `release` is not enabled by default in source release.

WillAyd:
By default meson will produce debug libraries. If you want something else, it is encouraged for the user to specify that from the command line when configuring their project, rather than having a (sub-)project enforce a particular build type. There is some upstream discussion about this in https://github.com/mesonbuild/wrapdb/issues/2208

dongjoon-hyun:
The upstream discussion looks like a new one (5 days ago) instead of the stabilized one. Let's have more time to think about this since we still have enough time for Apache ORC 2.2 release.

dongjoon-hyun (approved):
+1, LGTM. I slightly revisited the PR title according to the AS-IS status.

PR #2300, opened 2025-06-24
ORC-1939: set TimestampColumnVector isUTC flag in TimestampFromXXXTreeReader

vrozov:
`TimestampColumnVector` `isUTC` flag should be set based on the reader `useUtc` flag similar to how `TimestampTreeReader` sets the flag based on `context.getUseUTCTimestamp()`.

Correctly reflects whether `TimestampColumnVector` represents time in local or UTC time zones.

Added new assertions to existing tests

No

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
+++ b/java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
@@ -1533,6 +1533,7 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         // Allocate column vector for file; cast column vector for reader.
         longColVector = new LongColumnVector(batchSize);
         timestampColVector = (TimestampColumnVector) previousVector;
+        timestampColVector.setIsUTC(useUtc);
       } else {
         longColVector.ensureSize(batchSize, false);
       }
@@ -1597,6 +1598,7 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         // Allocate column vector for file; cast column vector for reader.
         doubleColVector = new DoubleColumnVector(batchSize);
         timestampColVector = (TimestampColumnVector) previousVector;
+        timestampColVector.setIsUTC(useUtc);
       } else {
         doubleColVector.ensureSize(batchSize, false);
       }
@@ -1661,6 +1663,7 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
         // Allocate column vector for file; cast column vector for reader.
         decimalColVector = new DecimalColumnVector(batchSize, precision, scale);
         timestampColVector = (TimestampColumnVector) previousVector;
+        timestampColVector.setIsUTC(useUtc);
       } else {
         decimalColVector.ensureSize(batchSize, false);
       }
@@ -1676,6 +1679,7 @@ public class ConvertTreeReaderFactory extends TreeReaderFactory {
   public static class TimestampFromStringGroupTreeReader extends ConvertTreeReader {
     private BytesColumnVector bytesColVector;
     private TimestampColumnVector timestampColVector;
+    private final boolean useUtc;
     private final DateTimeFormatter formatter;
     private final boolean useProlepticGregorian;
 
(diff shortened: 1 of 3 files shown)


vrozov:
@dongjoon-hyun Please review

java/core/src/java/org/apache/orc/impl/ConvertTreeReaderFactory.java
```diff
@@ -1676,13 +1679,15 @@ public void nextVector(ColumnVector previousVector,
...
     private TimestampColumnVector timestampColVector;
+    private final boolean useUtc;
     private final DateTimeFormatter formatter;
     private final boolean useProlepticGregorian;
 
     TimestampFromStringGroupTreeReader(int columnId, TypeDescription fileType,
                                        Context context, boolean isInstant)
         throws IOException {
       super(columnId, getStringGroupTreeReader(columnId, fileType, context), context);
+      useUtc = isInstant || context.getUseUTCTimestamp();
```
dongjoon-hyun, line 1690: Do we have this kind of logic in C++ implementation, @wgtmac ?

vrozov: @dongjoon-hyun @wgtmac does it apply to C++? I think it is hive-storage-api specific.

vrozov: @dongjoon-hyun @wgtmac ?

wgtmac: Could you help verify this? @ffacs

ffacs: No, C++ does not have the option `UseUTCTimestamp`; it uses the provided timezone or defaults to GMT if none is specified.

vrozov: @dongjoon-hyun ^^^

dongjoon-hyun (commented):
Sorry for the delaying. I need more time to think about this and validate this with other systems. I'll get back to you in a week.

java/core/src/test/org/apache/orc/impl/TestSchemaEvolution.java
```diff
@@ -2420,50 +2423,62 @@ public void testEvolutionToTimestamp() throws Exception {
           assertEquals(expected1.replace(".1 ", " "),
               timestampToString(l1.time[current], l1.nanos[current], READER_ZONE),
               msg);
+          assertFalse(l1.isUTC(), EXPECT_LOCAL.formatted("l1"));
```
dongjoon-hyun, line 2426: For my understanding, is it correct, `isUTC` == `EXPECT_LOCAL`? It's a little counter-intuitive to me.

java/core/src/test/org/apache/orc/impl/TestSchemaEvolution.java
```diff
@@ -2420,50 +2423,62 @@ public void testEvolutionToTimestamp() throws Exception {
...
 
           assertEquals(expected2.replace(".1 ", " "),
               timestampToString(l2.time[current], l2.nanos[current], WRITER_ZONE),
               msg);
+          assertTrue(l2.isUTC(), EXPECT_UTC.formatted("l2"));
```
dongjoon-hyun, line 2431: `isUTC` == `EXPECT_UTC` is too.

dongjoon-hyun (commented):
Could you add a new minimal test case outside `TestSchemaEvolution`, @vrozov ? It would be very helpful for the review.

java/core/src/test/org/apache/orc/impl/TestSchemaEvolution.java
vrozov: No, it asserts that `isUTC` is `false`, so `!isUTC == EXPECT_LOCAL`.

vrozov:
> Could you add a new minimal test case outside TestSchemaEvolution

Sorry, I don't follow. It will duplicate `TestSchemaEvolution`. Note that the fix only affects schema evolution use cases.

dongjoon-hyun (approved):
+1, LGTM. Let's merge this to main and test more.

dongjoon-hyun:
Since schema evolution is a tricky area, I'll keep this in main for a while.

dongjoon-hyun:
BTW, thank you, @vrozov , @wgtmac , @ffacs .

dongjoon-hyun:
I backported this to branch-2.2 for Apache ORC 2.2.0.

cc @williamhyun

PR #2274, opened 2025-06-17
ORC-1923: Remove `Windows 2019` GitHub Action job

dongjoon-hyun:
### What changes were proposed in this pull request?

This PR aims to remove `Windows 2019` GitHub Action job.

Note that ORC-1924 is created in order to recover Windows test coverage later.
- #2273

### Why are the changes needed?

`Windows 2019` is deprecated already and will be removed soon.
- https://github.com/actions/runner-images/issues/12045

### How was this patch tested?

Manual review because this is a removal of CI.

### Was this patch authored or co-authored using generative AI tooling?

No.

PR #2301, opened 2025-06-25
ORC-1940: Add threads dependency to orc lib in Meson

WillAyd:

This explicitly adds the threads dependency to the orc library when built in the Meson configuration

In the current CI, no failure is reported because the threads dependency is transitively provided by the protobuf dependency, when protobuf is included from source. However, some older protobuf libraries do not correctly specify in their pkg-config files that threads is a transitive dependency, so if built on a system where an older protobuf version exists orc will fail to build. This is exactly the case in the Apache Arrow CI, which you can see https://github.com/apache/arrow/actions/runs/15878114506/job/44770826425?pr=46906#step:6:1669

Compiled and tested locally

No

### DIFF
--- a/c++/src/meson.build
+++ b/c++/src/meson.build
@@ -173,6 +173,8 @@ source_files += files(
 
 incdir = include_directories('../include')
 orc_format_proto_dep = dependency('orc_format_proto')
+# zstd requires us to add the threads
+threads_dep = dependency('threads')
 
 orc_lib = library(
     'orc',
@@ -184,6 +186,7 @@ orc_lib = library(
         snappy_dep,
         lz4_dep,
         zstd_dep,
+        threads_dep,
     ],
     include_directories: incdir,
     install: true,


c++/src/meson.build
```diff
@@ -173,6 +173,8 @@ source_files += files(
 
 incdir = include_directories('../include')
 orc_format_proto_dep = dependency('orc_format_proto')
+# zstd requires us to add the threads
+threads_dep = dependency('threads')
```
wgtmac, line 177: Does it work on Windows? Or does Meson support Windows?

dongjoon-hyun: Good question. 

BTW, FYI, we lost `Windows` test coverage recently due to the EOL of `Windows 2019`.
- #2274
- https://lists.apache.org/thread/jtromvktcgb37vgw1po3plfxw3ws76hq

WillAyd: Yes this works across platforms

WillAyd: Tangentially we can look to add Meson CI for Windows. Typically the hardest part for Windows is getting the export macros correct, but *if* that's already taken care of it should be a trivial amount of effort

wgtmac (approved):

c++/src/meson.build
wgtmac: @dongjoon-hyun Thanks for the info! It seems that we need to bump version of protobuf to make the Windows 2025 CI happy.

@WillAyd It may or may not work out of box. But I think we can postpone this until required to do.

dongjoon-hyun (approved):
+1, LGTM. Thank you, @WillAyd and @wgtmac .

c++/src/meson.build
dongjoon-hyun: BTW, do you mean this which requires ORC 3.0.0 or something else, @wgtmac ?
- #1857

wgtmac: Yes, I think so.

PR #2304, opened 2025-06-30
ORC-1942: Fix `PhysicalFsWriter` not to change `ZstdCodec`'s default option

dongjoon-hyun:

This PR aims to fix `PhysicalFsWriter` to change `tempOptions` directly.

In the following code path, `tempOptions` is supposed to be updated and used. However, `codec.getDefaultOptions() instanceof ZstdCodec.ZstdOptions options` code is currently updating the return value of `codec.getDefaultOptions()` via a variable `options`.

https://github.com/apache/orc/blob/d3843bc043bea97bdd81a0f8e1fab3329efc7308/java/core/src/java/org/apache/orc/impl/PhysicalFsWriter.java#L118-L127

Technically, `ZstdCodec.getDefaultOptions()` returns the final static variable. This AS-IS code is trying to change this static object which leads unintended side-effects. We should avoid this code pattern.

https://github.com/apache/orc/blob/d3843bc043bea97bdd81a0f8e1fab3329efc7308/java/core/src/java/org/apache/orc/impl/ZstdCodec.java#L150-L156

Pass the CIs with a newly added test case.

No.

### DIFF
--- a/java/core/src/java/org/apache/orc/impl/ZstdCodec.java
+++ b/java/core/src/java/org/apache/orc/impl/ZstdCodec.java
@@ -152,7 +152,7 @@ public class ZstdCodec implements CompressionCodec, DirectDecompressionCodec {
 
   @Override
   public Options getDefaultOptions() {
-    return DEFAULT_OPTIONS;
+    return DEFAULT_OPTIONS.copy();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


java/core/src/java/org/apache/orc/impl/ZstdCodec.java
```diff
@@ -152,7 +152,7 @@ public int hashCode() {
 
   @Override
   public Options getDefaultOptions() {
-    return DEFAULT_OPTIONS;
+    return DEFAULT_OPTIONS.copy();
```
dongjoon-hyun, line 155: In addition, this PR makes this return a copied one to make it sure.

dongjoon-hyun:
cc @cxzl25

dongjoon-hyun:
Merged to main/2.1/2.0.

cxzl25 (commented):
Late LGTM. Thank you @dongjoon-hyun

dongjoon-hyun:
Thank you for reviewing, @cxzl25 !

PR #2316, opened 2025-07-06
[C++] add missing include for protobuf types

### LINKED ISSUE #2315: Missing protobuf header / compatibility with protobuf v6+
While [migrating](https://github.com/conda-forge/orc-feedstock/pull/90) orc for newer protobuf (from 5.29 --> 6.31 from the POV of the C++ version, as counted in protobuf's very particular versioning scheme), I'm running into

```
$SRC_DIR/c++/src/io/InputStream.cc:115:21: error: 'int64' in namespace 'google::protobuf' does not name a type
  115 |   google::protobuf::int64 SeekableArrayInputStream::ByteCount() const {
      |                     ^~~~~
```

That type still [exists](https://github.com/protocolbuffers/protobuf/blob/74211c0dfc2777318ab53c2cd2c317a2ef9012de/src/google/protobuf/stubs/port.h#L83), but the required header isn't being included.

h-vetinari:
Fixes #2315

### DIFF
--- a/c++/src/io/InputStream.hh
+++ b/c++/src/io/InputStream.hh
@@ -23,6 +23,8 @@
 #include "orc/OrcFile.hh"
 #include "wrap/zero-copy-stream-wrapper.h"
 
+#include "google/protobuf/stubs/port.h"
+
 #include <fstream>
 #include <iostream>
 #include <list>
--- a/c++/src/io/OutputStream.hh
+++ b/c++/src/io/OutputStream.hh
@@ -24,6 +24,8 @@
 #include "orc/OrcFile.hh"
 #include "wrap/zero-copy-stream-wrapper.h"
 
+#include "google/protobuf/stubs/port.h"
+
 namespace orc {
 
   /**


ffacs:
LGTM

wgtmac:
Thanks for fixing this! Could you create a JIRA ticket for this? The PR title should use JIRA id not the Github issue id.

h-vetinari:
> Could you create a JIRA ticket for this? The PR title should use JIRA id not the Github issue id.

Why do you have duplicate issue tracking systems? I opened #2315 already. You can just copy & paste that to JIRA if you want, but IMO you should either switch off issues on this repos or decommission JIRA. Asking contributors to do redundant work like that will just mean people contribute less (because who wants to waste their time like that).

dongjoon-hyun (changes requested):
To @h-vetinari , I agree with your argument. However, to simply put, it's because your PR title is wrong.

https://issues.apache.org/jira/browse/ORC-2315

dongjoon-hyun:
I will remove the PR title here to prevent accidental merging. We can add a legit one later.

dongjoon-hyun:
This is superseded by [ORC-1973: Use `int64_t` instead of `google::protobuf::int64` for Protobuf v22+](https://issues.apache.org/jira/browse/ORC-1973)

PR #2317, opened 2025-07-07
ORC-1946: [C++] Fix potential overflow issue in left shift operation identified by UBSAN

luffy-zh:

Remove some temporary workarounds in this [PR](https://github.com/apache/orc/pull/2117).

Fix potential overflow issue in left shift operation.

Test it locally.

No.

### DIFF
--- a/c++/src/RLE.cc
+++ b/c++/src/RLE.cc
@@ -77,9 +77,9 @@ namespace orc {
     add<int16_t>(data, numValues, notNull);
   }
 
-  NO_SANITIZE_ATTR
   void RleEncoder::writeVslong(int64_t val) {
-    writeVulong((val << 1) ^ (val >> 63));
+    uint64_t uval = static_cast<uint64_t>(val);
+    writeVulong((uval << 1) ^ (static_cast<int64_t>(uval) >> 63));
   }
 
   void RleEncoder::writeVulong(int64_t val) {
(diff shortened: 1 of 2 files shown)


c++/src/RLE.cc
```diff
@@ -77,9 +77,9 @@ namespace orc {
     add<int16_t>(data, numValues, notNull);
   }
 
-  NO_SANITIZE_ATTR
   void RleEncoder::writeVslong(int64_t val) {
-    writeVulong((val << 1) ^ (val >> 63));
+    uint64_t uval = static_cast<uint64_t>(val);
+    writeVulong((uval << 1) ^ (static_cast<int64_t>(uval) >> 63));
```
wgtmac: Why not `writeVulong((uval << 1) ^ (uval >> 63))` ?

luffy-zh: Uval is of type uint64_t, and shifting it directly with >> 63 would perform a logical shift (zero-fill).
However, we need a sign extension (i.e., fill with copies of the sign bit), which only happens with signed types.

wgtmac: Ah, I think I had a typo. I mean `writeVulong((uval << 1) ^ (val >> 63))`. Then we can simply write these two lines as `writeVulong((static_cast<uint64_t>(val) << 1) ^ (val >> 63));`

dongjoon-hyun:
Merged to main/2.2 for Apache ORC 2.2.0 (the end of July).

Thank you, @luffy-zh and @wgtmac .

cc @williamhyun , too

PR #2320, opened 2025-07-08
ORC-1949: Fix `BoundingBox#isXEmpty/isYEmpty/isZEmpty/isMEmpty` returning incorrect results for invalid dimension range

usberkeley:
Modify the logic for determining emptiness in `BoundingBox#isXEmpty/isYEmpty/isZEmpty/isMEmpty`

For example, in the x dimension, `xMin` > `xMax` should be considered empty.

`TestBoundingBox` covers all possibilities for determining emptiness in a certain dimension.

No

### DIFF
--- a/java/core/src/java/org/apache/orc/geospatial/BoundingBox.java
+++ b/java/core/src/java/org/apache/orc/geospatial/BoundingBox.java
@@ -201,7 +201,7 @@ public class BoundingBox {
    * @return true if the X dimension is empty, false otherwise.
    */
   public boolean isXEmpty() {
-    return Double.isInfinite(xMin) && Double.isInfinite(xMax);
+    return xMin > xMax;
   }
 
   /**
@@ -210,7 +210,7 @@ public class BoundingBox {
    * @return true if the Y dimension is empty, false otherwise.
    */
   public boolean isYEmpty() {
-    return Double.isInfinite(yMin) && Double.isInfinite(yMax);
+    return yMin > yMax;
   }
 
   /**
@@ -219,7 +219,7 @@ public class BoundingBox {
    * @return true if the Z dimension is empty, false otherwise.
    */
   public boolean isZEmpty() {
-    return Double.isInfinite(zMin) && Double.isInfinite(zMax);
+    return zMin > zMax;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


dongjoon-hyun:
cc @ffacs and @wgtmac

java/core/src/java/org/apache/orc/geospatial/BoundingBox.java
```diff
@@ -201,7 +201,7 @@ public boolean isXYEmpty() {
    * @return true if the X dimension is empty, false otherwise.
    */
   public boolean isXEmpty() {
-    return Double.isInfinite(xMin) && Double.isInfinite(xMax);
+    return xMin > xMax;
```
wgtmac, line 204: I don't think this is correct. Empty means the bbox is still in its initial state (a.k.a. haven't set any value). Please correct me if I was wrong. @paleolimbot @jiayuasu

paleolimbot: Yes, I also don't think this is correct. Where `xMin > xMax`, this is defined in the spec as the "wraparound case", meaning the union of the intervals `[-Inf, xMin]` and `[xMax, Inf]`. I think in C++ I did this check as `Double.isInfinite(xMin - xMax)`.

wgtmac: Thanks for the quick response!

paleolimbot: Whoops, there was a typo there. It's the union of `[-Inf, xMax]` and `[xMin, Inf]` 🤦

usberkeley: @wgtmac 
@paleolimbot 
Thank you for your answers. I got it. ❤️

dongjoon-hyun:
Could you address the above comment, @usberkeley ?

usberkeley:
> Could you address the above comment, @usberkeley ?

Sure, I'm reviewing the relevant code. Once I understand everyone's comments, if I realize I misunderstood, I'll close the PR later

dongjoon-hyun:
Feel free to reopen this when the PR is ready. Let me close it for now per the above discussion.

PR #2324, opened 2025-07-11
ORC-1952: [C++] Fix the issue where the value of headerThirdByte exceeds the valid byte range

luffy-zh:

Ensure DIRECT encoding is employed when the input value exceeds the valid byte range.

In Patched Base Encoding, the value of headerThirdByte exceeds the valid byte range.

Add the RleV2_value_limit_test to the RleTest suite.

No.

### DIFF
--- a/c++/src/RLEv2.hh
+++ b/c++/src/RLEv2.hh
@@ -123,7 +123,7 @@ namespace orc {
     int64_t* zigzagLiterals_;
     int64_t* baseRedLiterals_;
     int64_t* adjDeltas_;
-    static const int64_t BASE_VALUE_LIMIT = int64_t(1) << 56;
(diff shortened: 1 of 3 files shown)


wgtmac:
Thanks for fixing this! Please create a JIRA issue for this.

cc @ffacs

c++/src/RLEv2.hh
```diff
@@ -123,6 +123,7 @@ namespace orc {
     int64_t* zigzagLiterals_;
     int64_t* baseRedLiterals_;
     int64_t* adjDeltas_;
+    static const int64_t BASE_VALUE_LIMIT = int64_t(1) << 56;
```
ffacs:
```suggestion
    static constexpr int64_t BASE_VALUE_LIMIT = int64_t(1) << 56;
```

dongjoon-hyun:
Merged to main/2.2/2.1.

PR #2321, opened 2025-07-08
ORC-1950: [C++] Replace std::unorder_map with google dense_hash_map in SortedStringDictionary and remove reorder to improve write performance of dict-encoding columns

taiyang-li:
### What changes were proposed in this pull request?

Replace std::unorder_map with google dense_hash_map in SortedStringDictionary and remove reorder to improve write performance of dict-encoding columns 

### Why are the changes needed?

For performance improvement.s

### How was this patch tested?

Pass the CIs.

I tested it in clickhouse. Here are the results

``` sql
Query: select concat('gluten ', cast(rand()%1000 as String)) from numbers(10000000) into outfile 'dict.orc' truncate;

-- Enable dictionary encoding before optimization
set max_threads =1 ;
set output_format_orc_dictionary_key_size_threshold = 1;

10000000 rows in set. Elapsed: 0.760 sec. Processed 10.00 million rows, 80.00 MB (13.16 million rows/s., 105.31 MB/s.)
10000000 rows in set. Elapsed: 0.796 sec. Processed 10.00 million rows, 80.00 MB (12.56 million rows/s., 100.51 MB/s.)
10000000 rows in set. Elapsed: 0.803 sec. Processed 10.00 million rows, 80.00 MB (12.45 million rows/s., 99.63 MB/s.)


-- Enable dictionary encoding after optimization
set max_threads =1 ;
set output_format_orc_dictionary_key_size_threshold = 1;

10000000 rows in set. Elapsed: 0.645 sec. Processed 10.00 million rows, 80.00 MB (15.50 million rows/s., 124.00 MB/s.)
10000000 rows in set. Elapsed: 0.622 sec. Processed 10.00 million rows, 80.00 MB (16.08 million rows/s., 128.64 MB/s.)
10000000 rows in set. Elapsed: 0.631 sec. Processed 10.00 million rows, 80.00 MB (15.85 million rows/s., 126.83 MB/s.)  



-- Disable dictionary encoding
set max_threads =1 ;
set output_format_orc_dictionary_key_size_threshold = 0;

10000000 rows in set. Elapsed: 0.707 sec. Processed 10.00 million rows, 80.00 MB (14.14 million rows/s., 113.15 MB/s.)
10000000 rows in set. Elapsed: 0.671 sec. Processed 10.00 million rows, 80.00 MB (14.90 million rows/s., 119.17 MB/s.)
[...]

PR #2337, opened 2025-07-22
ORC-1950: [C++] Make sure dictionary is sorted before flushed into ORC file to follow ORC specs

taiyang-li:

Make sure dictionary is sorted before flushed into ORC file to follow ORC specs. The [issue](https://github.com/apache/orc/pull/2321#discussion_r2219569602) was brought by https://github.com/apache/orc/pull/2336.

### DIFF
--- a/c++/src/ColumnWriter.cc
+++ b/c++/src/ColumnWriter.cc
@@ -939,6 +939,14 @@ namespace orc {
       std::unique_ptr<std::string> data;
     };
 
+    struct DictEntryWithIndex {
+      DictEntryWithIndex(const char* str, size_t len, size_t index)
+          : entry(str, len), index(index) {}
+
+      DictEntry entry;
+      size_t index;
+    };
+
     SortedStringDictionary() : totalLength_(0) {
       /// Need to set empty key otherwise dense_hash_map will not work correctly
       keyToIndex_.set_empty_key(std::string_view{});
@@ -950,8 +958,11 @@ namespace orc {
     // write dictionary data & length to output buffer
     void flush(AppendOnlyBufferedStream* dataStream, RleEncoder* lengthEncoder) const;
 
+    // reorder input index buffer from insertion order to dictionary order
+    void reorder(std::vector<int64_t>& idxBuffer) const;
+
     // get dict entries in insertion order
-    const std::vector<DictEntry>& getEntriesInInsertionOrder() const;
+    void getEntriesInInsertionOrder(std::vector<const DictEntry*>&) const;
 
     // return count of entries
     size_t size() const;
@@ -962,8 +973,13 @@ namespace orc {
     void clear();
 
    private:
+    struct LessThan {
+      bool operator()(const DictEntryWithIndex& l, const DictEntryWithIndex& r) {
+        return l.entry.data < r.entry.data;  // use std::string's operator<
+      }
+    };
     // store dictionary entries in insertion order
-    mutable std::vector<DictEntry> flatDict_;
+    mutable std::vector<DictEntryWithIndex> flatDict_;
 
     // map from string to its insertion order index
     google::dense_hash_map<std::string_view, size_t> keyToIndex_;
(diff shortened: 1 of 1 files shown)


c++/src/ColumnWriter.cc
```diff
@@ -962,8 +973,13 @@ namespace orc {
     void clear();
 
    private:
+    struct LessThan {
+      bool operator()(const DictEntryWithIndex& l, const DictEntryWithIndex& r) {
+        return l.entry.data < r.entry.data;  // use std::string's operator<
```
wgtmac: Originally I used `memcmp` to avoid wrong results if there is any `\0` in the string. Does `std::string's operator<` resolve this issue?

c++/src/ColumnWriter.cc
```diff
@@ -939,6 +939,14 @@ namespace orc {
...
 
+    struct DictEntryWithIndex {
+      DictEntryWithIndex(const char* str, size_t len, size_t index)
+          : entry(str, len), index(index) {}
+
+      DictEntry entry;
+      size_t index;
+    };
+
     SortedStringDictionary() : totalLength_(0) {
```
wgtmac: Can we add `src/Dictionary.hh` and `src/Dictionary.cc` to make it abstract? We can use macro and pimpl idiom to avoid virtual functions. The reason is that adding a new dependency may enforce downstream projects (e.g. Apache Arrow) to do the same thing to manage an extra dependency with marginal benefit. We can add a new CMake option `ORC_ENABLE_SPARSEHASH` with `OFF` by default. In our CI settings, it can be turned on.

c++/src/ColumnWriter.cc
taiyang-li: Yes. `std::string` marks its ending with `length` field instead of `\0`.  
image

c++/src/ColumnWriter.cc
taiyang-li: If `ORC_ENABLE_SPARSEHASH` is OFF, what should be used inside SortedStringDictionary ?

wgtmac: std::unordered_map ?

taiyang-li: done.

c++/src/CMakeLists.txt
```diff
@@ -189,7 +189,8 @@ set(SOURCE_FILES
   Timezone.cc
   TypeImpl.cc
   Vector.cc
-  Writer.cc)
+  Writer.cc
+  Dictionary.cc)
```
wgtmac: Please sort them alphabetically.

c++/src/meson.build
```diff
@@ -170,6 +170,7 @@ source_files += files(
     'TypeImpl.cc',
     'Vector.cc',
     'Writer.cc',
+    'Dictionary.cc',
```
wgtmac: ditto

c++/src/meson.build
```diff
@@ -180,6 +181,9 @@ threads_dep = dependency('threads')
 orc_lib = library(
     'orc',
     sources: source_files,
+    cpp_args: [
+        '-DBUILD_SPARSEHASH'
```
wgtmac, line 185: @WillAyd Could you help review this? If we depend on the sparsehash library in the meson build, what should we do on the Apache Arrow side?

wgtmac (commented):
Generally LGTM. Thanks!

ffacs (commented):
@taiyang-li Thank you for the fix, could you please add a test to make sure dictionary is ordered?

taiyang-li:
> @taiyang-li Thank you for the fix, could you please add a test to make sure dictionary is ordered?

Of course.

taiyang-li:
> > @taiyang-li Thank you for the fix, could you please add a test to make sure dictionary is ordered?

done

wgtmac (approved):

williamhyun (approved):
+1 LGTM, will merge this now.

dongjoon-hyun:
Thank you, @williamhyun .

PR #2345, opened 2025-07-31
Fix Windows build

georgthegreat:
See
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/byteswap-uint64-byteswap-ulong-byteswap-ushort?view=msvc-170

PR #2346, opened 2025-08-03
ORC-1967: [C++] Fix case for byteswap functions

nehaljwani:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/byteswap-uint64-byteswap-ulong-byteswap-ushort

```
%SRC_DIR%\c++\src\Geospatial.cc(78): error C3861: '_byteSwap_ulong': identifier not found
%SRC_DIR%\c++\src\Geospatial.cc(81): error C3861: '_byteSwap_uint64': identifier not found
```

`s/byteSwap/byteswap`

To fix a compilation error on Windows

Conda Forge CI

No

### DIFF
--- a/c++/src/Geospatial.cc
+++ b/c++/src/Geospatial.cc
@@ -66,8 +66,8 @@ namespace orc::geospatial {
 
 #if defined(_MSC_VER)
 #include <intrin.h>  // IWYU pragma: keep
-#define ORC_BYTE_SWAP64 _byteSwap_uint64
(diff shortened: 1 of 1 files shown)


wgtmac:
Thanks for your report and fix! This seems to be a duplicate of https://github.com/apache/orc/pull/2345 which comes earlier. So I'm closing this in favor of that. Looking forward to your future contribution!

PR #2352, opened 2025-08-06
ORC-1924: Add Windows 2025 GitHub Action job

wgtmac:


dongjoon-hyun:
Thank you for working on this.

wgtmac:
It seems that the CIs are all green now. But we still have some issues. To break them down:

**1. Failed to build zlib.**

- I switched to install it via vcpkg and set ZLIB_HOME to make it work.

**2. Failed to build protobuf.**

- I was too lazy so I just bumped it to 3.21.12 and it succeeded.
- I think we should use vcpkg installed thirdparty libraries in the Windows CI as it is not easy to reproduce those failures.

**3. Failed to start cpp-linter-action.**

- We just need to bump it to cpp-linter/cpp-linter-action@f91c446a32ae3eb9f98fef8c9ed4c7cb613a4f8a

**4. Failed test cases due to missing TZDB**

- `tzdata` has moved the legacy TZDB files to `tzdata-right` so I changed the download location in the cmake_modules/ThirdpartyToolchain.cmake.
- Also the env has been changed to set(TZDATA_DIR ${SOURCE_DIR}/share/zoneinfo/right)

**5. Bad test case isolation of TEST(TestTimezone, testTzdbFromCondaEnv)** 

- It seems that TEST(TestTimezone, testTzdbFromCondaEnv) can globally change the TZDIR state and mess other test cases. I have disabled it for now.

**6. Inconsistent test result of TEST(TestTimezone, testZoneCache)**

- I just commented out some cases in it to make it pass. I don't know why.

wgtmac:
Unfortunately I don't have time to fix all of these. Perhaps @luffy-zh can take it over and create separate PRs to address them one by one?

cc @dongjoon-hyun @ffacs

dongjoon-hyun:
It seems that @luffy-zh mistakenly created a new JIRA issue when he starts to take over this.

- #2363

dongjoon-hyun:
To @luffy-zh , all ASF community recommend to reuse the existing JIRA issues for the same bug and work items.

wgtmac:
Close this in favor of https://github.com/apache/orc/pull/2363

PR #2353, opened 2025-08-08
ORC-1969: [C++] Support async I/O prefetch of next stripe

wgtmac:

Add a reader option to support prefetch next stripe automatically.

The C++ reader has added two functions (preBuffer and releaseBuffer) to support manual async I/O prefetch. This is not easy to use since it requires manual control of I/O operations.

Added test cases to verify that async prefetch is triggered and worked as expected.

No.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -173,7 +173,7 @@ jobs:
           mkdir build && cd build
           cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DBUILD_JAVA=OFF
           cmake --build .
-      - uses: cpp-linter/cpp-linter-action@v2.13.3
+      - uses: cpp-linter/cpp-linter-action@f91c446a32ae3eb9f98fef8c9ed4c7cb613a4f8a
         id: linter
         continue-on-error: true
         env:
--- a/c++/src/Options.hh
+++ b/c++/src/Options.hh
@@ -25,6 +25,7 @@
 
 #include "io/Cache.hh"
 
+#include <iostream>
 #include <limits>
 
 namespace orc {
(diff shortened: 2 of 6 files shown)


wgtmac:
cc @taiyang-li @luffy-zh

c++/src/Options.hh
```diff
@@ -25,6 +25,7 @@
 
 #include "io/Cache.hh"
 
+#include <iostream>
```
ffacs, line 28: Maybe we can remove this?

wgtmac: No, linter will complain it because std::cerr is used in this file.

.github/workflows/build_and_test.yml
```diff
@@ -173,7 +173,7 @@ jobs:
           mkdir build && cd build
           cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DBUILD_JAVA=OFF
           cmake --build .
-      - uses: cpp-linter/cpp-linter-action@v2.13.3
+      - uses: cpp-linter/cpp-linter-action@f91c446a32ae3eb9f98fef8c9ed4c7cb613a4f8a
```
dongjoon-hyun, line 176: Just a question. Is this required for this code change, @wgtmac ?

c++/src/Reader.cc
```diff
@@ -1099,6 +1137,16 @@ namespace orc {
     } while (sargsApplier_ && currentStripe_ < lastStripe_);
 
     if (currentStripe_ < lastStripe_) {
+      if (enableAsyncPrefetch_) {
+        // FIXME: this is very coarse since I/O ranges of all selected columns are about to
```
dongjoon-hyun, line 1141: nit. Please use IDed TODO like `TODO(ORC-XXX)` style instead of `FIXME`. It's because IDed TODOs give us a community-wide visibility than the hidden comments.

dongjoon-hyun:
+1 for the code. I just left two nit comments. 😄

.github/workflows/build_and_test.yml
wgtmac: Yes, otherwise the cpp-linter ci was broken.

PR #2406, opened 2025-09-18
ORC-2000: [C++] Add support to prefetch small stripes

wgtmac:

- Support reader to prefetch and coalesce small stripes.
- Add a reader option to limit number of stripes to prefetch.

Sometimes ORC files may have a lot of small stripes which hurt performance a lot.

Added a test case to verify that the prefetch works and so as the limit.

No.

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -1171,6 +1171,8 @@ namespace orc {
 
     if (currentStripe_ < lastStripe_) {
       if (enableAsyncPrefetch_) {
+        contents_->evictCache(currentStripeInfo_.offset());
+
         if (fullyCachedStripes_.find(currentStripe_) != fullyCachedStripes_.cend()) {
           // Current stripe has been fully cached, do nothing.
         } else if (isSmallStripe(currentStripeInfo_, contents_->cacheOptions.rangeSizeLimit)) {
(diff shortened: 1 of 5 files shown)


c++/src/Reader.cc
```diff
@@ -1070,6 +1095,14 @@ namespace orc {
     }
   }
 
+  uint64_t getStripeSize(const proto::StripeInformation& stripeInfo) {
```
luffy-zh, line 1098: Could we perhaps use StripeInformation::getLength() instead?

c++/src/Reader.cc
```diff
@@ -1138,13 +1171,37 @@ namespace orc {
...
+          // Cache footer of next stripe to avoid blocking I/O.
+          if (currentStripe_ + 1 < lastStripe_) {
+            const auto& nextStripe = footer_->stripes(static_cast<int>(currentStripe_ + 1));
+            contents_->cacheRanges(std::vector<ReadRange>{ReadRange{
+                nextStripe.offset() + nextStripe.index_length() + nextStripe.data_length(),
+                nextStripe.footer_length()}});
+          }
+        }
+
         contents_->evictCache(currentStripeInfo_.offset());
```
ffacs: Maybe evictCache before prefetching would be better?

wgtmac: Good point!

c++/src/Reader.cc
wgtmac: We do not have access to it since it is still a proto object.

dongjoon-hyun (approved):
+1, LGTM.

PR #2409, opened 2025-09-20
ORC-2002: [C++] Improve prefetching stripe

wgtmac:

- Fix prefetching when small stripe limit is 0 to prefetch current stripe data and next stripe footer.
- Prefetch next small stripe if current stripe is currently the last small stripe that cached fully.
- Prefetch and coalesce ranges of index. 
- Refactor code to try read from cache and then fall back to original stream.

To make stripe prefetching more effective.

Added test case for small stripe limit 0. Other fixes are verified manually due to not easy to reflect in the test.

No

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -27,7 +27,6 @@
 #include "wrap/coded-stream-wrapper.h"
 
 #include <algorithm>
-#include <iostream>
 #include <iterator>
 #include <memory>
 #include <set>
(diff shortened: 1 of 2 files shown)


c++/src/Reader.cc
```diff
@@ -1173,21 +1189,33 @@ namespace orc {
...
 
-        if (fullyCachedStripes_.find(currentStripe_) != fullyCachedStripes_.cend()) {
-          // Current stripe has been fully cached, do nothing.
-        } else if (isSmallStripe(currentStripeInfo_, contents_->cacheOptions.rangeSizeLimit)) {
+        auto extractSmallStripeRanges = [this](uint64_t startStripe) {
           std::vector<ReadRange> ranges;
-          uint64_t maxStripe =
-              std::min(lastStripe_, currentStripe_ + smallStripeLookAheadLimit_ + 1);
-          for (uint64_t stripe = currentStripe_; stripe < maxStripe; stripe++) {
+          uint64_t maxStripe = std::min(lastStripe_, startStripe + smallStripeLookAheadLimit_ + 1);
```
luffy-zh, line 1194: It seems there's no need to +1？

wgtmac: If we don't add one, then the loop will do nothing if `smallStripeLookAheadLimit_` is 0.

wgtmac:
Merged this. Thanks @luffy-zh and @ffacs for the review!

Issue #2408, opened 2025-09-19
Orc unable to find dependencies, even when they exist


WillAyd:
To reproduce the issue you can create a conda environment with orc and a third party dependency, like Snappy:

```shell
$ conda create -n foo orc snappy
$ conda activate foo
```

With the minimal CMake project below, no errors occur during setup:

```cmake
cmake_minimum_required(VERSION 3.2)
project(foo)

find_package(Snappy REQUIRED)
```

However, if you add a call to find orc, you will then get an error that Snappy could not be found:

```cmake
cmake_minimum_required(VERSION 3.2)
project(foo)

find_package(Snappy REQUIRED)
find_package(orc REQUIRED)
```

```shell
$ cmake -S . -B build
-- SNAPPY_HOME: 
CMake Error at /home/will/miniforge3/envs/foo/lib/cmake/orc/FindSnappy.cmake:80 (message):
  Could not find Snappy in system search paths.
Call Stack (most recent call first):
  /usr/share/cmake-3.31/Modules/CMakeFindDependencyMacro.cmake:76 (find_package)
  /home/will/miniforge3/envs/foo/lib/cmake/orc/orcConfig.cmake:72 (find_dependency)
  CMakeLists.txt:5 (find_package)
```
This issue was first discovered downstream in Apache Arrow (see https://github.com/apache/arrow/pull/46906#issuecomment-3286997441) and appears to have started with Orc 2.1.0

WillAyd:
This may be the same or similar to @h-vetinari issue in https://github.com/apache/orc/issues/2147

FWIW, the orc distribution provides custom CMake Find* files at ${CONDA_PREFIX}/lib/cmake/orc/:

```shell
$ ls ${CONDA_PREFIX}/lib/cmake/orc/Find*
/home/will/miniforge3/envs/foo/lib/cmake/orc/FindLZ4.cmake
/home/will/miniforge3/envs/foo/lib/cmake/orc/FindProtobuf.cmake
/home/will/miniforge3/envs/foo/lib/cmake/orc/FindSnappy.cmake
/home/will/miniforge3/envs/foo/lib/cmake/orc/FindZLIB.cmake
/home/will/miniforge3/envs/foo/lib/cmake/orc/FindZSTD.cmake
```

If you remove or rename FindSnappy.cmake, then it appears that CMake can correctly resolve the location for it

wgtmac:
[...]

Issue #2147, opened 2025-03-07
Depend on CMake and/or pkgconfig package data by default


h-vetinari:
Orc should recognize CMake / pkgconfig metadata for dependencies, ideally by default. In https://github.com/conda-forge/arrow-cpp-feedstock/pull/1697, we had to add
```batch
      -DLZ4_HOME="%LIBRARY_PREFIX%" ^
      -DLZ4_INCLUDE_DIR="%LIBRARY_INC%" ^
      -DLZ4_LIBRARY="%LIBRARY_LIB%\lz4.lib" ^
      -DZSTD_HOME="%LIBRARY_PREFIX%" ^
      -DZSTD_INCLUDE_DIR="%LIBRARY_INC%" ^
      -DZSTD_LIBRARY="%LIBRARY_LIB%\libzstd.lib" ^
```
otherwise orc would fail to find its dependencies. That should not be necessary, as conda-forge has complete metadata for those packages, and so orc should search _first_ try to look in `CMAKE_MODULE_PATH`, `PKG_CONFIG_PATH` (and ideally just `$PREFIX` / `CMAKE_PREFIX_PATH`) before falling back to testing `ZSTD_*` / `LZ4_*` / ... variables (and eventually failing if the dependencies really cannot be found anywhere).

From my POV think this is healthy _default_ behaviour in general, but if you prefer, you can also put this behind some choice of `-DORC_PACKAGE_KIND=`. Based on the suggestion of @wgtmac, I tried this (shortly before the release of 2.1.1) in https://github.com/conda-forge/orc-feedstock/pull/86, and the existing `"conan"` / `"vcpkg"` options do not work for this purpose.

h-vetinari:
Ping @wgtmac

wgtmac:
Sorry that I may not have bandwidth to work on it at this moment

Issue #2016, opened 2024-08-30
Allow using external lib instead of trying to build bundled duplicates


barracuda156:
Firstly, this simply fails:
```
--->  Extracting orc-2.0.2.tar.gz
Executing:  cd "/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_apache-orc/apache-orc/work" && /usr/bin/gzip -dc '/opt/local/var/macports/distfiles/apache-orc/orc-2.0.2.tar.gz' | /usr/bin/gnutar --no-same-owner -xf - 
--->  Configuring apache-orc
        (using ccache)
[...]

PR #2416, opened 2025-09-26
ORC-2013: [C++] Bump to CMake 3.25+ to use FetchContent

wgtmac:

- Rename builtin FindXXX.cmake to FindXXXAlt.cmake to not interfere with cannonical package name. These are now only used when `XXX_HOME` or its friends are set. This includes:
  - FindGTest.cmake -> FindGTestAlt.cmake
  - FindLZ4.cmake -> FindLZ4Alt.cmake
  - FindProtobuf.cmake -> FindProtobufAlt.cmake
  - FindSnappy.cmake -> FindSnappyAlt.cmake
  - FindZLIB.cmake -> FindZLIBAlt.cmake
  - FindZSTD.cmake -> FindZSTDAlt.cmake
- Bump CMake minimum requirement to 3.25 to leverage `FetchContent` with `FIND_PACKAGE_ARGS`. For 3rd party dependencies now, we try to find system installed one via `CONFIG` mode or fallback to vendor it. This includes:
  - GTest
  - LZ4
  - Protobuf
  - Snappy
  - ZSTD
  - ZLIB: note that we use an unreleased version (1.4.0) to work around its misuse of build and install interface
  - Sparsehash: always vendor it with workaround to its CMake misuse

- https://github.com/apache/orc/issues/2408
- https://github.com/apache/orc/issues/2147
- https://github.com/apache/orc/issues/2016
- https://github.com/apache/arrow/pull/46906

- Pass all CIs
- Test it manually

No

### DIFF
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -207,17 +207,17 @@ else ()
       set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-deprecated-declarations")
     endif()
 
-    fetchcontent_declare(protobuf
+    fetchcontent_declare(Protobuf
       URL "https://github.com/google/protobuf/archive/v${PROTOBUF_VERSION}.tar.gz"
       SOURCE_SUBDIR "cmake"
       FIND_PACKAGE_ARGS
       NAMES Protobuf
       CONFIG
       )
-    fetchcontent_makeavailable(protobuf)
+    fetchcontent_makeavailable(Protobuf)
 
     if(protobuf_SOURCE_DIR)
-      message(STATUS "Using vendored protobuf")
+      message(STATUS "Using vendored Protobuf")
 
       add_library(protobuf::libprotobuf ALIAS libprotobuf)
       add_executable(protobuf::protoc ALIAS protoc)
@@ -238,7 +238,7 @@ else ()
 
       list(APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:orc::libprotobuf>")
     else()
-      message(STATUS "Using system protobuf")
+      message(STATUS "Using system Protobuf")
       list(APPEND ORC_SYSTEM_DEPENDENCIES Protobuf)
       list(APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:protobuf::libprotobuf>")
     endif()
@@ -292,7 +292,7 @@ else ()
     fetchcontent_makeavailable(Snappy)
 
     if(snappy_SOURCE_DIR)
-      message(STATUS "Using vendored snappy")
+      message(STATUS "Using vendored Snappy")
       if(NOT TARGET Snappy::snappy)
         add_library(Snappy::snappy INTERFACE IMPORTED)
         target_link_libraries(Snappy::snappy INTERFACE snappy)
@@ -319,7 +319,7 @@ else ()
 
       list(APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:orc::snappy>")
     else()
-      message(STATUS "Using system snappy")
+      message(STATUS "Using system Snappy")
       list(APPEND ORC_SYSTEM_DEPENDENCIES Snappy)
       list(APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:Snappy::snappy>")
     endif()
@@ -368,7 +368,7 @@ else ()
     set(ZLIB_INSTALL OFF)
     set(ZLIB_PREFIX OFF)
 
-    fetchcontent_declare(zlib
+    fetchcontent_declare(ZLIB
       # URL "https://zlib.net/fossils/zlib-${ZLIB_VERSION}.tar.gz"
       # See https://github.com/madler/zlib/issues/937
       GIT_REPOSITORY "https://github.com/madler/zlib.git"
@@ -377,10 +377,10 @@ else ()
       NAMES ZLIB
       CONFIG
       )
-    fetchcontent_makeavailable(zlib)
+    fetchcontent_makeavailable(ZLIB)
 
     if(zlib_SOURCE_DIR)
-      message(STATUS "Using vendored zlib")
+      message(STATUS "Using vendored ZLIB")
       if(NOT TARGET ZLIB::ZLIB)
         add_library(ZLIB::ZLIB ALIAS zlibstatic)
       endif()
@@ -402,11 +402,11 @@ else ()
       set(ZLIB_LIBRARIES "zlibstatic")
       set(ZLIB_INCLUDE_DIRS "${zlib_SOURCE_DIR}")
     else()
-      message(STATUS "Using system zlib")
+      message(STATUS "Using system ZLIB")
       # FindZLIB guarantees that ZLIB::ZLIB target exists if found
       # See https://cmake.org/cmake/help/latest/module/FindZLIB.html#imported-targets
       if(NOT TARGET ZLIB::ZLIB)
-        message(FATAL_ERROR "Using system zlib, but ZLIB::ZLIB not found")
+        message(FATAL_ERROR "Using system ZLIB, but ZLIB::ZLIB not found")
       endif()
       list(APPEND ORC_SYSTEM_DEPENDENCIES ZLIB)
       list(APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:ZLIB::ZLIB>")
@@ -547,7 +547,7 @@ else ()
     fetchcontent_makeavailable(lz4)
 
     if(lz4_SOURCE_DIR)
-      message(STATUS "Using vendored LZ4")
+      message(STATUS "Using vendored lz4")
       if(NOT TARGET LZ4::lz4_static)
         add_library(LZ4::lz4_static ALIAS lz4_static)
       endif()
@@ -567,7 +567,7 @@ else ()
 
       list(APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:orc::lz4_static>")
     else()
-      message(STATUS "Using system LZ4")
+      message(STATUS "Using system lz4")
       list(APPEND ORC_SYSTEM_DEPENDENCIES lz4)
       list(APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:$<TARGET_NAME_IF_EXISTS:LZ4::lz4_shared>>")
       list(APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:$<TARGET_NAME_IF_EXISTS:LZ4::lz4_static>>")
@@ -650,15 +650,15 @@ if (BUILD_CPP_TESTS)
   else ()
     block()
       prepare_fetchcontent()
-      fetchcontent_declare(googletest
+      fetchcontent_declare(GTest
         URL "https://github.com/google/googletest/archive/refs/tags/v${GTEST_VERSION}.tar.gz"
         FIND_PACKAGE_ARGS
         NAMES GTest
         CONFIG
         )
-      fetchcontent_makeavailable(googletest)
+      fetchcontent_makeavailable(GTest)
 
-      if(googletest_SOURCE_DIR)
+      if(gtest_SOURCE_DIR)
         message(STATUS "Using vendored GTest")
         if(NOT TARGET GTest::gtest)
           add_library(GTest::gtest ALIAS gtest)


wgtmac:
This PR is ready for review.

For now `XXX_HOME` variables are still supported and take higher priority than `FetchContent` call, so Apache Arrow C++ should not be affected. Is it possible for Apache Arrow to leverage this improvement? @kou @raulcd

Can conda leverage improvement from this PR, or is there any additional change that is required to add? @h-vetinari @WillAyd

cmake_modules/FindGTestAlt.cmake
```diff
@@ -52,14 +52,14 @@ find_library (GTEST_STATIC_LIB NAMES ${CMAKE_STATIC_LIBRARY_PREFIX}gtest${CMAKE_
   PATH_SUFFIXES "lib")
 
 if (GTEST_INCLUDE_DIR AND GMOCK_LIBRARY)
-  set (GTEST_FOUND TRUE)
+  set (GTESTAlt_FOUND TRUE)
```
kou: How about using `GTestAlt_FOUND` because this files uses `FindGTestAlt.cmake` not `FindGTESTAlt.cmake`?

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -155,6 +183,111 @@ ExternalProject_Add (orc-format_ep
...
+    orc_add_resolved_library (orc_protoc ${PROTOC_STATIC_LIB} ${PROTOBUF_INCLUDE_DIR})
+  else ()
+    orc_add_resolved_library (orc_protoc ${PROTOC_LIBRARY} ${PROTOBUF_INCLUDE_DIR})
+  endif ()
+
+  list (APPEND ORC_SYSTEM_DEPENDENCIES ProtobufAlt)
+  list (APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:protobuf::libprotobuf>")
+  orc_provide_find_module (ProtobufAlt)
+else ()
+  prepare_fetchcontent()
```
kou: If we call this in top-level, changed variables in this context are visible after this clause.

How about using `function()` or `block()`?

```cmake
function(build_protobuf)
  prepare_fetchcontent()
  ...
endfunction()
build_protobuf()
```

https://cmake.org/cmake/help/latest/command/block.html

```cmake
block()
  prepare_fetchcontent()
  ...
endblock()
```

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -155,6 +183,111 @@ ExternalProject_Add (orc-format_ep
...
+    SOURCE_SUBDIR "cmake"
+    FIND_PACKAGE_ARGS
+    NAMES Protobuf
+    CONFIG
+    )
+  fetchcontent_makeavailable(protobuf)
+
+  # Restore original compiler flags after protobuf configuration
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS_BACKUP}")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS_BACKUP}")
```
kou: We can remove this by creating new variables scope by `function()`/`block()`.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -155,6 +183,111 @@ ExternalProject_Add (orc-format_ep
...
+
+  if(protobuf_SOURCE_DIR)
+    message(STATUS "Using vendored protobuf")
+
+    if(NOT TARGET protobuf::libprotobuf)
+      add_library(protobuf::libprotobuf ALIAS libprotobuf)
+    endif()
+    if(NOT TARGET protobuf::protoc)
+      add_executable(protobuf::protoc ALIAS protoc)
+    endif()
```
kou: It seems that we don't need these `if`s:

```suggestion
    add_library(protobuf::libprotobuf ALIAS libprotobuf)
    add_executable(protobuf::protoc ALIAS protoc)
```

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -170,41 +303,68 @@ elseif (ORC_PACKAGE_KIND STREQUAL "vcpkg")
...
     orc_add_resolved_library (orc_snappy ${SNAPPY_STATIC_LIB} ${SNAPPY_INCLUDE_DIR})
   else ()
     orc_add_resolved_library (orc_snappy ${SNAPPY_LIBRARY} ${SNAPPY_INCLUDE_DIR})
   endif ()
-  list (APPEND ORC_SYSTEM_DEPENDENCIES Snappy)
+  list (APPEND ORC_SYSTEM_DEPENDENCIES SnappyAlt)
   list (APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:Snappy::snappy>")
-  orc_provide_find_module (Snappy)
+  orc_provide_find_module (SnappyAlt)
 else ()
```
kou, line 278: We may be able to remove `SnappyAlt` entirely by setting `Snappy_ROOT`:

```suggestion
else ()
  if (NOT "${SNAPPY_HOME}" STREQUAL "" AND "${Snappy_ROOT}" STREQUAL "")
    set(Snappy_ROOT "${SNAPPY_HOME}")
  endif()
```

See also: https://cmake.org/cmake/help/latest/variable/PackageName_ROOT.html

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -223,48 +383,74 @@ elseif (ORC_PACKAGE_KIND STREQUAL "vcpkg")
   list (APPEND ORC_SYSTEM_DEPENDENCIES ZLIB)
   list (APPEND ORC_INSTALL_INTERFACE_TARGETS "$<INSTALL_INTERFACE:ZLIB::ZLIB>")
 elseif (NOT "${ZLIB_HOME}" STREQUAL "")
-  find_package (ZLIB REQUIRED)
+  find_package (ZLIBAlt REQUIRED)
```
kou, line 351: We may be able to remove `FindZLIBAlt.cmake`. CMake's `FindZLIB` provides `ZLIB_ROOT` hint: https://cmake.org/cmake/help/latest/module/FindZLIB.html#hints

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -16,6 +16,7 @@
 # under the License.
 
 INCLUDE(ExternalProject)
+INCLUDE(FetchContent)
 
 set(ORC_VENDOR_DEPENDENCIES)
```
kou: It seems that we can remove this.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -170,41 +303,68 @@ elseif (ORC_PACKAGE_KIND STREQUAL "vcpkg")
...
+      set_target_properties(snappy POSITION_INDEPENDENT_CODE ON)
+    endif()
+
+    if(INSTALL_VENDORED_LIBS)
+      set_target_properties(snappy PROPERTIES OUTPUT_NAME "orc_vendored_snappy")
+
+      install(FILES ${snappy_SOURCE_DIR}/snappy-c.h DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+      install(FILES ${snappy_SOURCE_DIR}/snappy-sinksource.h DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+      install(FILES ${snappy_SOURCE_DIR}/snappy.h DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+      install(FILES ${snappy_BINARY_DIR}/snappy-stubs-public.h DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
```
kou: We may not need to install header files.

ORC doesn't use Snappy in public headers, right?

wgtmac: We need to install it, otherwise downstream project may fail like below:

```
CMake Error at CMakeLists.txt:9 (target_link_libraries):
  Cannot find source file:

    /tmp/orc/include/snappy-c.h
```

The reason is that Snappy adds these 4 header files as install interface: https://github.com/google/snappy/blob/1.2.2/CMakeLists.txt#L266

cmake_modules/ThirdpartyToolchain.cmake
wgtmac: Good suggestion! I've switched to use blocks.

cmake_modules/ThirdpartyToolchain.cmake
wgtmac: Doesn't it break arrow-cpp (where `SNAPPY_LIBRARY ` and `SNAPPY_INCLUDE_DIR` are explicitly provided)?

Update: I've tried to apply this change with or without `CONFIG` in the FIND_PACKAGE_ARGS. In both cases, system snappy is used instead of arrow vendored one.

```
-- Apache ORC: Found the Protobuf headers: /Users/gangwu/Projects/arrow/cpp/orc_build/protobuf_ep-install/include
-- Apache ORC: Found the Protobuf library: arrow::protobuf::libprotobuf
-- Apache ORC: Found the Protoc library: arrow::protobuf::libprotoc
-- Apache ORC: Found the Protoc executable: arrow::protobuf::protoc
-- Apache ORC: Found the Protobuf static library: /Users/gangwu/Projects/arrow/cpp/orc_build/protobuf_ep-install/lib/libprotobuf.a
-- Apache ORC: Found the Protoc static library: /Users/gangwu/Projects/arrow/cpp/orc_build/protobuf_ep-install/lib/libprotoc.a
-- Apache ORC: Providing CMake module for FindProtobufAlt as part of CMake package
-- Apache ORC: Using system snappy                     <==================
-- Apache ORC: ZLIB_HOME: /Users/gangwu/Projects/arrow/cpp/orc_build/zlib_ep/src/zlib_ep-install
-- Apache ORC: Found the ZLIB header: /Users/gangwu/Projects/arrow/cpp/orc_build/zlib_ep/src/zlib_ep-install/include/zlib.h
-- Apache ORC: Found the ZLIB library: ZLIB::ZLIB
-- Apache ORC: Found the ZLIB static library: /Users/gangwu/Projects/arrow/cpp/orc_build/zlib_ep/src/zlib_ep-install/lib/libz.a
-- Apache ORC: Providing CMake module for FindZLIBAlt as part of CMake package
```

wgtmac: I'm now using `CONFIG` mode in the `FIND_PACKAGE_ARGS` so it is not compatible if we remove `XXXAlt`.

wgtmac:
The two failed CIs are due to outdated CMake versions from amazonlinux23 and debian11. Should we remove them? @dongjoon-hyun

dongjoon-hyun:
> The two failed CIs are due to outdated CMake versions from amazonlinux23 and debian11. Should we remove them? @dongjoon-hyun

Do you mean to remove `amazonlinux23` and `debian11`, @wgtmac ? Can we upgrade `CMake` on those Docker images? Then, I can help you by publishing new images.

dongjoon-hyun:
For `Debian 11`, we can remove it because we added `Debian 12` and `Debian 13`. However, `Amazonlinux23` is the latest one. I believe we need to support it.

dongjoon-hyun:
Here is the `Debian 11` removal PR.
- https://github.com/apache/orc/pull/2434

dongjoon-hyun:
For AmazonLinux2023, I confirmed that it has `CMake 3.22.2`. Let me update it locally first.
```
$ docker run -it --rm apache/orc-dev:amazonlinux23 cmake --version
cmake version 3.22.2
```

dongjoon-hyun:
Unfortunately, `CMake` version is still old even after refreshing the base image.

```
-FROM amazonlinux:2023
+FROM amazonlinux:2023.9.20250929.0
```

```
$ docker run -it --rm 4facd3ecc339 cmake --version
cmake version 3.22.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

kou:
FYI: Apache Arrow installs newer CMake manually on Amazon Linux 2023: https://github.com/apache/arrow/blob/33d1f32136f950520dcca278b77acff430dc8dca/dev/tasks/linux-packages/apache-arrow/yum/amazon-linux-2023/Dockerfile#L64-L68

dongjoon-hyun:
Good. I'll borrow that code. Thank you, @kou .

dongjoon-hyun:
Here is the PR for `amazonlinux:2023`
- https://github.com/apache/orc/pull/2435

dongjoon-hyun:
New `amazonlinux23` image is published and I re-triggered CI on this PR.

```
$ docker run -it --rm apache/orc-dev:amazonlinux23 cmake --version
cmake version 3.26.0

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

WillAyd:
Thanks for doing this @wgtmac . I'm not sure if there's a way for me to test this out locally but if so I'm happy to try it. Otherwise looking forward to this making its way through conda

wgtmac:
@WillAyd I created https://github.com/conda-forge/orc-feedstock/pull/94 to test this. Let's see what's going on. (cc @h-vetinari)

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -54,50 +54,22 @@ string(TOUPPER ${CMAKE_BUILD_TYPE} UPPERCASE_BUILD_TYPE)
 
 if (DEFINED ENV{SNAPPY_HOME})
   set (SNAPPY_HOME "$ENV{SNAPPY_HOME}")
-elseif (Snappy_ROOT)
```
wgtmac, line 57: @kou I have deleted lines like these so setting `XXX_ROOT` will not set `XXX_HOME` any more and it will skip `FindXXXAlt.cmake` by directly jumping to the `FetchContent` branch.

kou: I'm OK with removing them as an Apache Arrow developer. But controlling wheter `FindXXXAlt.cmake` or `FetchContent` by only this may not be enough... I'll comment it on the https://github.com/apache/orc/pull/2416/files#r2387031578 thread.

cmake_modules/ThirdpartyToolchain.cmake
kou: Ah, you're right. Sorry.

How about the following?

```cmake
if (ORC_PACKAGE_KIND STREQUAL "conan")
  # ...
elseif (ORC_PACKAGE_KIND STREQUAL "vcpkg")
  # ...
elseif  (TARGET Snappy::snappy)
  # Use existing target if possible: This is for FetchContent use case like Apache Arrow does
  add_library(orc_snappy ALIAS Snappy::snappy)
elseif (TARGET Snappy::snappy-static)
  # Use existing target if possible: This is for FetchContent use case like Apache Arrow does
  add_library(orc_snappy ALIAS Snappy::snappy-static)
elseif (ORC_PACKAGE_KIND_SNAPPY STREQUAL "system")
  find_package(Snappy REQUIRED)
  # ....
else ()
  if (ORC_PACKAGE_KIND_SNAPPY STREQUAL "bundled")
    set(Snappy_FOUND FALSE)
  else ()
    find_package(Snappy)
  endif ()
  if (Snappy_FOUND)
    add_library(orc_snappy ALIAS Snappy::snappy)
  else ()
    # Use FetchContent
    # ...
  endif ()
endif ()
```

Apache Arrow will use the following part for bundled Apache ORC:

```cmake
elseif  (TARGET Snappy::snappy)
  # Use existing target if possible: This is for FetchContent use case like Apache Arrow does
  add_library(orc_snappy ALIAS Snappy::snappy)
elseif (TARGET Snappy::snappy-static)
  # Use existing target if possible: This is for FetchContent use case like Apache Arrow does
  add_library(orc_snappy ALIAS Snappy::snappy-static)
```

If this works, we can remove `FindSnappyAlt.cmake` and `SNAPPY_HOME` entirely.
```

wgtmac: Thanks for the suggestion! I will explore this approach.

Let me merge this PR first since it is large enough and backward compatible. BTW, I do not want to introduce variables like `ORC_PACKAGE_KIND_SNAPPY` to make it more complicated. IMHO, Apache ORC C++ is a low-level library and should use provided dependencies as much as possible.

wgtmac:
Merged this. Thanks everyone!

Issue #2225, opened 2025-05-20
Patch to handle missing /etc/localtime


rickysarraf:
On Arch Linux, the documentation clearly states that /etc/localtime is not
required, and in CI it's not present. Patch the timezone handling to
assume UTC when /etc/localtime is not there.

This is hopefully a sane fallback for systems where `/etc/localtime` may not be available.

```patch
--- a/c++/src/Timezone.cc             2025-03-12 12:09:01.267309399 +0000
+++ b/c++/src/Timezone.cc             2025-03-12 12:12:12.303295722 +0000
@@ -28,6 +28,12 @@
 #include <map>
 #include <sstream>
 
+#ifndef _MSC_VER
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+#endif
+
 namespace orc {
 
   // default location of the timezone files
@@ -694,6 +700,14 @@
 #ifdef _MSC_VER
     return getTimezoneByName("UTC");
 #else
+    // The absence of LOCAL_TIMEZONE means UTC conventionally
+    {
+      struct stat _ignored;
+      if (stat(LOCAL_TIMEZONE, &_ignored) == -1) {
+        return getTimezoneByName("UTC");
+      }
+    }
+
     return getTimezoneByFilename(LOCAL_TIMEZONE);
 #endif
   }

```

dongjoon-hyun:
Thank you for reporting, @rickysarraf .

cc @wgtmac

wgtmac:
I'm not sure if such fallback makes sense because it may slightly produce data shift for timestamp values written by a different host. I also see that the fallback exists on Windows.

@ffacs WDYT?

ffacs:
> I'm not sure if such fallback makes sense because it may slightly produce data shift for timestamp values written by a different host. I also see that the fallback exists on Windows.
> 
> [@ffacs](https://github.com/ffacs) WDYT?

I think the fallback makes sense. Because when using the local timezone, reading on different machines may produce different results.

wgtmac:
The problem may come from the writer which blindly uses UTC to serialize timestamp values. In terms of fallback, I think it is better to make the fallback value configurable and fail if unset.

dongjoon-hyun:
[...]

PR #2432, opened 2025-10-01
Fall back to UTC when /etc/localtime does not exist

lahwaacz:
Fixes https://github.com/apache/orc/issues/2225


dongjoon-hyun:
Thank you again, @lahwaacz and @wgtmac .

Merged to main.

PR #2436, opened 2025-10-11
ORC-2022: [C++] Add support to use dictionary for IN expression

wgtmac:

Leverage dictionary to evaluate IN expression.

For IN expression, min/max stats are not that useful. It would be good to support evaluating dictionary if it is not too large.

Added test case to verify it takes effect.

No.

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -19,7 +19,9 @@
 #include "Reader.hh"
 #include "Adaptor.hh"
 #include "BloomFilter.hh"
+#include "DictionaryLoader.hh"
 #include "Options.hh"
+#include "RLE.hh"
 #include "Statistics.hh"
 #include "StripeStream.hh"
 #include "Utils.hh"
@@ -32,6 +34,7 @@
 #include <set>
 #include <sstream>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 namespace orc {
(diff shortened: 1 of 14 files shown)


c++/src/sargs/SargsApplier.cc
```diff
@@ -185,4 +202,90 @@ namespace orc {
...
+    for (size_t i = 0; i < dictSize; ++i) {
+      int64_t start = offsets[i];
+      int64_t length = offsets[i + 1] - start;
+      std::string_view dictEntry(blob + start, static_cast<size_t>(length));
+
+      // Check if this dictionary entry matches any literal in the IN list
+      for (const auto& literal : literals) {
+        if (dictEntry == literal.getStringView()) {
+          // Found a match - stripe might contain matching rows
+          return TruthValue::YES_NO_NULL;
```
ffacs, line 232: We can return TruthValue::YES when dictionary is a subset of literals

c++/src/Reader.cc
```diff
@@ -1119,6 +1123,73 @@ namespace orc {
...
+    }
+
+    const Timezone& writerTimezone = stripeFooter.has_writer_timezone()
+                                         ? getTimezoneByName(stripeFooter.writer_timezone())
+                                         : localTimezone;
+    StripeStreamsImpl stripeStreams(reader, currentStripe, stripeInfo, stripeFooter,
+                                    stripeInfo.offset(), *reader.getFileContents().stream,
+                                    writerTimezone, readerTimezone);
+
+    auto dictionaries =
```
ffacs, line 1183: Dictionaries would be read twice if the stripe is needed,  we could add a stripe wise dictionary cache for column reader.

wgtmac:
Close it as I don't have time to move it forward.

PR #2445, opened 2025-10-16
ORC-2028: [C++] Fix ReadRangeCache uses freed memory

HuaHuaY:

If a `Buffer` is used in coroutine, it will be released after the coroutine finishes.

If `Buffer` is release before the coroutine finishes, coroutine will try to write data to freed memory and cause panic.

Tested in private repo by ASan.

No.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,3 @@ java/bench/data
 subprojects/*
 !subprojects/packagefiles
 !subprojects/*.wrap
-.vscode


.gitignore
```diff
@@ -17,3 +17,4 @@ java/bench/data
 subprojects/*
 !subprojects/packagefiles
 !subprojects/*.wrap
+.vscode
```
dongjoon-hyun: Thank you, but please revert this change, @HuaHuaY . We don't want to have a hidden change under the title, `ORC-2028: [C++] Fix ReadRangeCache uses freed memory`. I'd like to recommend to open another PR for this.

HuaHuaY: Fixed

wgtmac:
Merged it. Thanks @HuaHuaY and @dongjoon-hyun!

PR #2446, opened 2025-10-16
ORC-2029: [C++] support Float fast read by memcpy in DoubleColumnReader

HuaHuaY:

1. Simplfy `DoubleColumnReader::readXXX()`.
2. Support memcpy Float array in `DoubleColumnReader::next`.

Optimize `DoubleColumnReader::next` speed.

Ran existing tests.

No.

### DIFF
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -472,8 +472,7 @@ namespace orc {
       // Number of values in the buffer that we can copy directly.
       // Only viable when the machine is little-endian.
       uint64_t bufferNum = 0;
-      if constexpr (isLittleEndian &&
-                    (columnKind == TypeKind::DOUBLE ? true : std::is_same_v<ValueType, float>)) {
+      if constexpr (isLittleEndian && std::is_same_v<ValueType, ItemType>) {
         bufferNum =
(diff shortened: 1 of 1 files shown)


HuaHuaY:
It seems that there are some undefined behaviors after removing template. I will check them tomorrow.

HuaHuaY:
> It seems that there are some undefined behaviors after removing template. I will check them tomorrow.

My bad. Wrote something wrong. I think it is fixed. I tested in my local environment.

c++/src/ColumnReader.cc
```diff
@@ -375,7 +376,7 @@ namespace orc {
 
    private:
     std::unique_ptr<SeekableInputStream> inputStream_;
-    const uint64_t bytesPerValue_ = (columnKind == FLOAT) ? 4 : 8;
+    const int64_t bytesPerValue_ = (columnKind == DOUBLE) ? 8 : 4;
```
luffy-zh: Could we please replace the hard-coded values with sizeof(double) and sizeof(float)?

HuaHuaY: I adjusted the code.

wgtmac:
Thanks @HuaHuaY @dongjoon-hyun and @luffy-zh!

PR #2453, opened 2025-10-27
ORC-2022: [C++] Add support to use dictionary for IN expression 

luffy-zh:

Use column dictionaries to evaluate IN predicates.

Optimize IN predicate pruning: consult column dictionary (when reasonably sized) instead of relying on less effective min/max statistics.

Unit tests in TestPredicatePushdown.cc verify this change.

NO.

### DIFF
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -18,9 +18,7 @@
 
 #include "ColumnReader.hh"
 
-#include <math.h>
-
-#include <iostream>
+#include <cmath>
 
 #include "Adaptor.hh"
 #include "ByteRLE.hh"


c++/src/sargs/SargsApplier.cc
```diff
@@ -185,4 +202,97 @@ namespace orc {
...
+    for (size_t i = 0; i < dictSize; ++i) {
+      int64_t start = offsets[i];
+      int64_t length = offsets[i + 1] - start;
+      std::string_view dictEntry(blob + start, static_cast<size_t>(length));
+
+      // Check if this dictionary entry matches any literal in the IN list
+      for (const auto& literalView : literalViews) {
+        if (dictEntry == literalView) {
+          // Found a match - stripe might contain matching rows
+          return TruthValue::YES;
```
ffacs: We can only return TruthValue::YES when **all** rows satisfied the predication.

c++/src/sargs/SargsApplier.cc
```diff
@@ -185,4 +202,97 @@ namespace orc {
...
+        }
+      }
+    }
+
+    // No dictionary entry matches any literal in the IN list - skip stripe
+    return TruthValue::NO;
```
ffacs: ditto，We can only return TruthValue::NO when **no** rows satisfied the predication.

c++/src/ColumnReader.cc
```diff
@@ -529,43 +524,34 @@ namespace orc {
     void nextEncoded(ColumnVectorBatch& rowBatch, uint64_t numValues, char* notNull) override;
 
     void seekToRowGroup(std::unordered_map<uint64_t, PositionProvider>& positions) override;
+
+    // Method to set dictionary if we have a pre-loaded one
+    void setDictionary(std::shared_ptr<StringDictionary> dictionary) {
```
ffacs: Where is the function used?

c++/src/ColumnReader.cc
```diff
@@ -1717,8 +1703,16 @@ namespace orc {
...
         switch (static_cast<int64_t>(stripe.getEncoding(type.getColumnId()).kind())) {
           case proto::ColumnEncoding_Kind_DICTIONARY:
-          case proto::ColumnEncoding_Kind_DICTIONARY_V2:
-            return std::make_unique<StringDictionaryColumnReader>(type, stripe);
+          case proto::ColumnEncoding_Kind_DICTIONARY_V2: {
+            // Check if we have a pre-loaded dictionary we can use
+            auto dictionary = stripe.getSharedDictionary(type.getColumnId());
+            if (dictionary) {
+              return std::unique_ptr<ColumnReader>(
+                  new StringDictionaryColumnReader(type, stripe, dictionary));
```
ffacs: std::make_unique would be better.

c++/src/sargs/SargsApplier.cc
```diff
@@ -185,4 +202,97 @@ namespace orc {
...
+    for (size_t i = 0; i < dictSize; ++i) {
+      int64_t start = offsets[i];
+      int64_t length = offsets[i + 1] - start;
+      std::string_view dictEntry(blob + start, static_cast<size_t>(length));
+
+      // Check if this dictionary entry matches any literal in the IN list
+      for (const auto& literalView : literalViews) {
+        if (dictEntry == literalView) {
+          // Found a match - stripe might contain matching rows
+          return TruthValue::YES_NO_NULL;
```
ffacs: How about calculating the size of the intersect of literals and the dictionary, by which we return TruthValue more precisely?

c++/test/TestPredicatePushdown.cc
```diff
@@ -548,4 +548,122 @@ namespace orc {
...
+
+  // Create a test file with dictionary-encoded string columns for testing
+  // dictionary-based predicate pushdown
+  void createDictionaryTestFile(MemoryOutputStream& memStream) {
+    MemoryPool* pool = getDefaultPool();
+    auto type = std::unique_ptr<Type>(
+        Type::buildTypeFromString("struct<id:bigint,category:string,status:string>"));
+    WriterOptions options;
+    options
+        .setStripeSize(64 * 1024)  // Small stripe size to create multiple stripes
```
ffacs: Let's add a assert to make sure there are more than one stripes, since the pattern of data is to easy for encoding.

c++/src/sargs/SargsApplier.cc
```diff
@@ -227,6 +227,9 @@ namespace orc {
     const char* blob = dictionary.dictionaryBlob.data();
     size_t dictSize = dictionary.dictionaryOffset.size() - 1;
 
+    // Use a set to store matching dictionary entries
+    std::set<size_t> matchingDictEntries;
```
ffacs: size_t matchingDictEntries;

ffacs (approved):
LGTM. cc @wgtmac

c++/src/ColumnReader.cc
```diff
@@ -16,15 +16,20 @@
  * limitations under the License.
  */
 
-#include "orc/Int128.hh"
+#include "ColumnReader.hh"
+
+#include <math.h>
+
+#include <iostream>
```
wgtmac:
```suggestion
#include <cmath>
```

wgtmac:
I just committed this. Thanks @luffy-zh and @ffacs!

PR #2461, opened 2025-12-11
ORC-2036: [C++] Optimize SortedStringDictionary performance

HuaHuaY:
1. There are lots of `std::unique_ptr<std::string>` in `SortedStringDictionary`. This PR uses `google::protobuf::Arena` to store these continuously.
2. Add a new method `SortedStringDictionary::reserve` to avoid frequent expansion.
3. Delay the construction of `std::vector<DictEntryWithIndex> flatDict_` in `flush` method to reducing vector expansion.

It optimizes the performance of `SortedStringDictionary`.

Ran existing tests. Tested performance in my private environment.

No.

### DIFF
--- a/c++/src/Dictionary.cc
+++ b/c++/src/Dictionary.cc
@@ -119,5 +119,6 @@ namespace orc {
   void SortedStringDictionary::clear() {
     totalLength_ = 0;
     keyToIndex_.clear();
+    arena_->Reset();
   }
 }  // namespace orc


HuaHuaY:
There is another thing that can be optimized. `SortedStringDictionary` frequently performs hash operations on `std::string_view` in insert and expansion operations.
```cpp
size_t SortedStringDictionary::insert(const char* str, size_t len) {
  size_t index = keyToIndex_.size();
  auto it = keyToIndex_.find(std::string_view{str, len}); // <-- hash once
  if (it != keyToIndex_.end()) {
    return it->second;
  } else {
    auto s = Arena::Create<std::string>(arena_.get(), str, len);
    keyToIndex_.emplace(std::string_view{s->data(), s->length()}, index); // <-- hash twice, and may hash more in expansion.
    totalLength_ += len;
    return index;
  }
}
```
There are at least two ways to optimize this behavior:
1. Use some maps which support `preHash`, such as latest folly's F14Map. First, call `preHash` and get a token. Second, use the token to find and insert.
2. Use a wrapper to store the hash result. This way may be better than the previous one because the expansion operation also benefits. But I think it's a little hacky.
```cpp
namespace orc {
  template <typename T>
  class HashWrapper {
   public:
    HashWrapper(T val) : val_(std::move(val)), hashVal_(std::hash<T>()(val_)) {}

    HashWrapper(T val, size_t hashVal) : val_(std::move(val)), hashVal_(hashVal) {}

    bool operator==(const HashWrapper& other) const {
      return val_ == other.val_;
    }

    size_t hash() const {
      return hashVal_;
    }

    const T& value() const {
      return val_;
    }

   private:
    T val_;
    size_t hashVal_;
  };
}  // namespace orc

namespace std {
  template <typename T>
  struct hash<orc::HashWrapper<T>> {
    size_t operator()(const orc::HashWrapper<T>& hw) const {
      return hw.hash();
    }
  };
}  // namespace std
```

.github/workflows/build_and_test.yml
```diff
@@ -227,7 +227,7 @@ jobs:
           mkdir build && cd build
           cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DBUILD_JAVA=OFF
           cmake --build .
-      - uses: cpp-linter/cpp-linter-action@f91c446a32ae3eb9f98fef8c9ed4c7cb613a4f8a
+      - uses: cpp-linter/cpp-linter-action@0f6d1b8d7e38b584cbee606eb23d850c217d54f8
```
HuaHuaY, line 230: Need to be upgraded, otherwise ci will not work well. See 
https://github.com/apache/orc/actions/runs/20126059369 
https://github.com/apache/infrastructure-actions/blob/main/actions.yml#L278

c++/src/Dictionary.hh
```diff
@@ -28,62 +28,46 @@
...
     // insert a new string into dictionary, return its insertion order
     size_t insert(const char* str, size_t len);
 
-    // write dictionary data & length to output buffer
-    void flush(AppendOnlyBufferedStream* dataStream, RleEncoder* lengthEncoder) const;
+    // reserve space for dictionary entries
+    void reserve(size_t size);
 
+    // write dictionary data & length to output buffer and
     // reorder input index buffer from insertion order to dictionary order
```
wgtmac:
```suggestion
    // reorder input index buffer from insertion order to dictionary order
    // and then write dictionary data & length streams to the output buffer
```

c++/src/ColumnWriter.cc
```diff
@@ -1038,6 +1038,8 @@ namespace orc {
 
     if (!useDictionary) {
       directLengthEncoder->add(length, numValues, notNull);
+    } else if (dictionary.capacity() < numValues) {
```
wgtmac, line 1041: This looks weird. Is it better to grow its size only when required? I think the map internally can optimize this by reserving a larger number.

HuaHuaY: I think this can solve the expansion overhead at startup at a small cost. Usually the map does not reserve a large value immediately, but gradually expands it by a multiple of the size. When the map is very large, calling reserve to allocate memory for each insertion will take more time, so a condition is used here to determine.

wgtmac (approved):

Issue #2430, opened 2025-09-30
Incorrect error message in TypeDescription.withPrecision: misleading "precision X is out of range 1 .. Y" when scale > precision


wgzhao:
Hi,  
I encountered a confusing error message when using ORC decimal types in CDH 6.2. Here is the log:

```
2025-09-30 10:24:41.701 [       job-0] ERROR Engine               - com.wgzhao.addax.core.exception.AddaxException: java.lang.IllegalArgumentException: precision 5 is out of range 1 .. 10
        at org.apache.orc.TypeDescription.withPrecision(TypeDescription.java:219)
        at com.wgzhao.addax.plugin.writer.hdfswriter.OrcWriter.buildOrcSchema(OrcWriter.java:344)
        at com.wgzhao.addax.plugin.writer.hdfswriter.OrcWriter.write(OrcWriter.java:294)
        at com.wgzhao.addax.plugin.writer.hdfswriter.HdfsWriter$Task.startWrite(HdfsWriter.java:473)
        at com.wgzhao.addax.core.taskgroup.runner.WriterRunner.run(WriterRunner.java:74)
        at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.IllegalArgumentException: precision 5 is out of range 1 .. 10
        at org.apache.orc.TypeDescription.withPrecision(TypeDescription.java:219)
        ...
```

After checking the source code, I found that in `TypeDescription.withPrecision(int precision)`, the exception message is constructed as:

```java
throw new IllegalArgumentException("precision " + precision + " is out of range 1 .. " + this.scale);
```

This is misleading, because `this.scale` is not the upper bound for precision. In my case, precision=5 and scale=10 (default), so the error message says "precision 5 is out of range 1 .. 10", which is not correct. The real problem is that scale cannot be greater than precision.

**Expected behavior:**  
The error message should clearly indicate the actual constraint, e.g.  
- "precision must be between 1 and 38, and scale must be less than or equal to precision"
- Or, if the check fails because scale > precision, the message should reflect that.

**Suggestion:**  
Update the exception message in `withPrecision` to avoid using `this.scale` as the upper bound, and clarify the real constraint.

**Environment:**  
- ORC 1.9.5
- Java 17

Thanks!

cxzl25:
@wgzhao Thanks for reporting this issue, can you create a PR to fix this issue?

PR #2462, opened 2025-12-17
ORC-2038: Improve error message in TypeDescription.withPrecision()

wgzhao:

This PR improves the error message in the `TypeDescription.withPrecision(int precision)` method. The previous error message incorrectly stated `precision {value} is out of range 1 .. {scale}` when the scale was greater than precision, which was misleading because it suggested the scale value was the upper bound for precision.

The new error message clearly states: `the scale {scale} must be less than or equal to precision {precision}`, which accurately describes the actual constraint violation.

The previous error message was confusing and misleading. When users encountered a case where scale > precision (e.g., precision=5, scale=10), the error message "precision 5 is out of range 1 .. 10" incorrectly suggested that precision must be between 1 and 10, when the real issue was that scale (10) cannot be greater than precision (5).

This improvement enhances the developer experience by providing a clear, accurate error message that directly points to the actual problem, making it easier to diagnose and fix the issue without needing to examine the source code.

The existing test suite should validate this change. The modification only affects the error message text and does not change the validation logic itself. The same conditions that trigger the exception will continue to do so, but with a clearer message.

To verify the change manually:
1. Create a TypeDescription with decimal type
2. Set scale to a value (e.g., 10)
3. Attempt to set precision to a value less than scale (e.g., 5)
4. Verify the new error message is clear and accurate

No

Try to address the issue #2430

### DIFF
--- a/java/core/src/java/org/apache/orc/TypeDescription.java
+++ b/java/core/src/java/org/apache/orc/TypeDescription.java
@@ -247,8 +247,8 @@ public class TypeDescription
           "precision " + precision + " must be between 1 and " + MAX_PRECISION
       );
     } else if (scale > precision) {
-      throw new IllegalArgumentException("the scale " + scale + " must be less than" +
-          " or equal to precision " + precision);
+      throw new IllegalArgumentException("scale " + scale +
+          " must be less than or equal to precision " + precision);
     }
     this.precision = precision;
     return this;


cxzl25 (approved):

wgzhao:
@cxzl25 @ffacs could you merge this PR ?

ffacs:
Thanks @wgzhao @cxzl25 ~

ffacs:
The jira issue is not created, let me remove this patch first.

PR #2502, opened 2026-01-29
ORC-2064: Update `oraclelinux9` to use `dnf` instead of `yum`

dongjoon-hyun:

This PR aims to update `oraclelinux9` to use `dnf` instead of `yum`.

Since `Oracle Linux 8`, `yum` is an alias of `dnf` command. We had better use `dnf` directly.

```
$ docker run -it --rm apache/orc-dev:oraclelinux9 ls -al /usr/bin/yum
lrwxrwxrwx 1 root root 5 Aug 13 03:53 /usr/bin/yum -> dnf-3
```

Pass the CIs.

No.


dongjoon-hyun:
Merged to main/2.2.

PR #2601, opened 2026-04-15
ORC-2149: Supports merging multiple ORC files with the same schema into multiple ORC files

QianyongY:
Extends the Java merge tool so that, for inputs sharing the same schema, you can still merge to one ORC file by default, or use -m / --maxSize to write multiple ORC files under an output directory as part-xxxxx.orc, batching by input file size. 

Users often need to merge many compatible ORC files without producing a single huge output file. This adds an optional mode that caps output size at whole-file boundaries while keeping the existing single-file behavior when --maxSize is not set.

Add UT

### DIFF
--- a/java/tools/src/java/org/apache/orc/tools/MergeFiles.java
+++ b/java/tools/src/java/org/apache/orc/tools/MergeFiles.java
@@ -214,9 +214,6 @@ public class MergeFiles {
           "Input path must be an existing directory with --preserveStructure: " + inputRoot);
     }
 
-    FileSystem outFs = outputRoot.getFileSystem(conf);
-    prepareOutputDir(outFs, outputRoot, overwrite, "--preserveStructure");
-
     Path qualifiedInputRoot = inFs.makeQualified(inputRoot);
     List<Path> leaves = new ArrayList<>();
     collectLeafDirs(inFs, qualifiedInputRoot, ignoreExtension, leaves);
@@ -225,6 +222,9 @@ public class MergeFiles {
       System.err.println("No leaf directories containing ORC files found under: " + inputRoot);
       System.exit(1);
     }
+
+    FileSystem outFs = outputRoot.getFileSystem(conf);
+    prepareOutputDir(outFs, outputRoot, overwrite, "--preserveStructure");
 
     long effectiveMax = maxSizeBytes > 0 ? maxSizeBytes : Long.MAX_VALUE;
     int totalInputFiles = 0;
@@ -305,13 +305,26 @@ public class MergeFiles {
     }
 
     int totalMerged = 0;
+    int partFilesWritten = 0;
     List<Path> allUnmerged = new ArrayList<>();
 
-    for (int i = 0; i < batches.size(); i++) {
-      List<Path> batch = batches.get(i);
-      Path partOutput = new Path(outputDir, String.format(PART_FILE_FORMAT, i));
+    // Advance the part index only when a file is actually written, so
+    // part-NNNNN.orc stays contiguous and the reported count matches disk.
+    for (List<Path> batch : batches) {
+      Path partOutput = new Path(outputDir, String.format(PART_FILE_FORMAT, partFilesWritten));
       List<Path> merged = OrcFile.mergeFiles(partOutput, writerOptions.clone(), batch);
+
+      if (merged.isEmpty()) {
+        // Drop any 0-stripe placeholder left behind and skip this slot.
+        if (outFs.exists(partOutput)) {
+          outFs.delete(partOutput, false);
+        }
+        allUnmerged.addAll(batch);
+        continue;
+      }
+
       totalMerged += merged.size();
+      partFilesWritten++;
 
       if (merged.size() != batch.size()) {
         Set<Path> mergedSet = new HashSet<>(merged);
@@ -323,7 +336,7 @@ public class MergeFiles {
       }
     }
 
-    return new DirMergeResult(batches.size(), totalMerged, allUnmerged);
+    return new DirMergeResult(partFilesWritten, totalMerged, allUnmerged);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


dongjoon-hyun:
Thank you for making a PR, @QianyongY .

java/tools/src/test/org/apache/orc/tools/TestMergeFiles.java
```diff
@@ -107,4 +108,85 @@ public void testMerge() throws Exception {
...
+      if (batch.size != 0) {
+        writer.addRowBatch(batch);
+      }
+      writer.close();
+    }
+
+    // Measure the size of the first source file to compute a threshold that forces a split.
+    long singleFileSize = fs.getFileStatus(new Path(sourceNames[0])).getLen();
+    // Threshold: slightly larger than one file so at most one file fits per part.
+    long maxSize = singleFileSize + 1;
```
cxzl25, line 153: Is it possible to group by the first two file sizes + 1 to test if the merge really works？

sourceNames[0] len + sourceNames[1] len +1

java/tools/src/java/org/apache/orc/tools/MergeFiles.java
```diff
@@ -56,27 +60,60 @@ public static void main(Configuration conf, String[] args) throws Exception {
...
       System.err.println("No files found.");
       System.exit(1);
     }
 
-    List<Path> mergedFiles = OrcFile.mergeFiles(
-        new Path(outputFilename), writerOptions, inputFiles);
+    List<Path> inputFiles = new ArrayList<>(inputStatuses.size());
+    for (LocatedFileStatus s : inputStatuses) {
+      inputFiles.add(s.getPath());
+    }
```
dongjoon-hyun: Shall we sort `inputFiles` here by path to be more deterministic?

java/tools/src/test/org/apache/orc/tools/TestMergeFiles.java
```diff
@@ -107,4 +108,85 @@ public void testMerge() throws Exception {
...
     }
   }
+
+  /**
+   * Verifies that --maxSize splits input files into multiple part files under the output
+   * directory. Three source files are created; a tight size threshold forces them to be
+   * written into at least two part files.
+   */
+  @Test
+  public void testMergeWithMaxSize() throws Exception {
```
dongjoon-hyun, line 118: Could you add a test coverage for a single giant file which is bigger than `maxSize`?

java/tools/src/java/org/apache/orc/tools/MergeFiles.java
```diff
@@ -113,6 +227,16 @@ private static Options createOptions() {
         .desc("Ignore ORC file extension")
         .build());
 
+    result.addOption(Option.builder("m")
+        .longOpt("maxSize")
```
dongjoon-hyun, line 539: What happens when `maxSize` is `0` or negative?

java/tools/src/java/org/apache/orc/tools/MergeFiles.java
```diff
@@ -100,11 +137,88 @@ public static void main(Configuration conf, String[] args) throws Exception {
...
+                                              long maxSizeBytes) throws Exception {
+    FileSystem outFs = outputDir.getFileSystem(conf);
+    if (outFs.exists(outputDir)) {
+      if (!outFs.getFileStatus(outputDir).isDirectory()) {
+        throw new IllegalArgumentException(
+            "Output path already exists and is not a directory: " + outputDir);
+      }
+      if (outFs.listStatus(outputDir).length > 0) {
+        throw new IllegalArgumentException(
+            "Output directory must be empty for multi-file merge: " + outputDir);
```
dongjoon-hyun: In this case, shall we behave as an **overwrite** for a better UX? I guess you can delete the directory recursively.

java/tools/src/java/org/apache/orc/tools/MergeFiles.java
```diff
@@ -100,11 +137,88 @@ public static void main(Configuration conf, String[] args) throws Exception {
...
+    }
+
+    // Group input files into batches where each batch's total size <= maxSizeBytes.
+    List<List<Path>> batches = new ArrayList<>();
+    List<Path> currentBatch = new ArrayList<>();
+    long currentBatchSize = 0;
+
+    for (LocatedFileStatus status : inputStatuses) {
+      long fileSize = status.getLen();
+      if (!currentBatch.isEmpty() && currentBatchSize + fileSize > maxSizeBytes) {
```
dongjoon-hyun: Actually, I understand this is practical. However, can we avoid the chance of overflow, @QianyongY ?
```java
- if (!currentBatch.isEmpty() && currentBatchSize + fileSize > maxSizeBytes) {
+ if (!currentBatch.isEmpty() && currentBatchSize > maxSizeBytes - fileSize) {
```

Apache ORC community had a bad experience about overflow like the following.

- https://github.com/apache/orc/commit/589d5cab8a27aac94aa4aa2ebc50504d4875b258

QianyongY:
Thanks @cxzl25 and @dongjoon-hyun . Latest push:
- Deterministic ordering, overflow-safe batching, `--maxSize` docs clarified.
- Non-empty output dirs now abort by default; opt in with `--overwrite` to
  delete existing contents. Shared between `--maxSize` and `--preserveStructure`.
- New `--preserveStructure` mode (partition-tree-aware merge, skips hidden entries).
- Tests added for all of the above. PTAL.

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
For the unaddressed comments, I believe they are minors which we can address later independently.

dongjoon-hyun:
Merged to main/2.3 for Apache ORC 2.3.1.

dongjoon-hyun:
Thank you, @QianyongY and @cxzl25 .

QianyongY:
Thanks @dongjoon-hyun and @cxzl25. The review items above were already fixed in the code—I had forgotten to mark some conversations as Resolved earlier, but I’ve done that now. Thanks again for your time and feedback!

PR #2609, opened 2026-04-22
ORC-2151: [C++] Move the timestamp compensation before timezone conversion to prevent incorrect DST offset lookup

lszskye:
Moving the compensation to before the timezone conversion, we ensure that getVariant() always receives the corrected wall-clock time, producing the correct DST offset and an accurate final result.

Previously, the -1 second compensation for pre-1970 timestamps was applied after the writer-to-reader timezone adjustment. This means writerTimezone.getVariant(writerTime) was called with an uncorrected writerTime that could be 1 second larger than the true value. If this 1-second discrepancy happens to cross a DST (Daylight Saving Time) boundary, the timezone lookup would return an incorrect GMT offset, causing the final timestamp result to be off by the entire DST shift (typically 1 hour) rather than just 1 second.

No

### DIFF
--- a/c++/test/TestWriter.cc
+++ b/c++/test/TestWriter.cc
@@ -839,6 +839,73 @@ namespace orc {
     testWriteTimestampWithTimezone(fileVersion, "America/Los_Angeles", "America/Los_Angeles",
                                    "2014-06-06 12:34:56", IS_DST);
   }
+
+  // Test that the ORC-306 compensation (-1s for pre-1970 timestamps with
+  // nanos > 999999) is applied BEFORE timezone conversion in the Reader.
+  //
+  // We write secs = -5756401 (1s before the 1969 PDT->PST boundary in LA)
+  // with nanos = 1000000. The C++ Writer applies +1, storing -5756400 in the
+  // ORC file (exactly at the DST boundary). On read, the Reader must apply
+  // -1 BEFORE timezone conversion so that getVariant() sees -5756401 (PDT).
+  // If compensation were applied AFTER timezone conversion, getVariant() would
+  // see -5756400 (PST), producing a gmtOffset that differs by 3600 seconds.
+  TEST_P(WriterTest, DISABLED_writeNegativeTimestampAtDSTBoundary) {
+    MemoryOutputStream memStream(DEFAULT_MEM_STREAM_SIZE);
+    MemoryPool* pool = getDefaultPool();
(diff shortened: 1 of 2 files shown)


wgtmac (approved):
Thanks for fixing this! Is it possible to add a test case?

dongjoon-hyun (approved):
+1, LGTM, too. Also, I agree with the above comment about the test coverage.

lszskye:
> Thanks for fixing this! Is it possible to add a test case?

OK, let me add a unit test for this change.

lszskye:
@wgtmac All tests have passed. Please help merge. Thanks!

c++/test/TestWriter.cc
```diff
@@ -839,6 +839,73 @@ namespace orc {
...
+  // Test that the ORC-306 compensation (-1s for pre-1970 timestamps with
+  // nanos > 999999) is applied BEFORE timezone conversion in the Reader.
+  //
+  // We write secs = -5756401 (1s before the 1969 PDT->PST boundary in LA)
+  // with nanos = 1000000. The C++ Writer applies +1, storing -5756400 in the
+  // ORC file (exactly at the DST boundary). On read, the Reader must apply
+  // -1 BEFORE timezone conversion so that getVariant() sees -5756401 (PDT).
+  // If compensation were applied AFTER timezone conversion, getVariant() would
+  // see -5756400 (PST), producing a gmtOffset that differs by 3600 seconds.
+  TEST_P(WriterTest, DISABLED_writeNegativeTimestampAtDSTBoundary) {
```
wgtmac, line 852: Why is it `DISABLED`?

lszskye: Thanks for the feedback! I added DISABLED_ to keep it consistent with the other timestamp-related tests in the same file (DISABLED_writeTimestamp, DISABLED_writeNegativeTimestamp, DISABLED_writeTimestampWithTimezone, DISABLED_writeTimestampInstant), which are all disabled due to [ORC-1976]. Also, I'm happy to remove the DISABLED_ prefix if you prefer — please let me know.

wgtmac: Yes, it would be great if these can be fixed together.

wgtmac: As discussed offline, I've merged it to move forward. Thanks @lszskye for fixing this!

PR #2615, opened 2026-05-06
ORC-2159: [C++] Fix C++ build with nodiscard return values

wgtmac:

Handle protobuf and ZeroCopyStream boolean results required by newer headers. Check serialization and text-format failures in writer/tool code, and make affected tests assert parse, next, and skip results explicitly while preserving existing reader skip behavior.

The C++ build fails with newer protobuf / compiler headers because several boolean-returning APIs are marked nodiscard and their return values are ignored. Since the C++ build treats warnings as errors, these become hard build failures.

The build fails with errors like:

  ColumnReader.cc:438:9: error: ignoring return value of function declared with 'nodiscard' attribute [-Werror,-Wunused-result]
    inputStream_->Skip(static_cast<int>(step));

  ColumnReader.cc:684:9: error: ignoring return value of function declared with 'nodiscard' attribute [-Werror,-Wunused-result]
    blobStream_->Skip(static_cast<int>(step));

  ColumnWriter.cc:215:5: error: ignoring return value of function declared with 'nodiscard' attribute [-Werror,-Wunused-result]
    rowIndex->SerializeToZeroCopyStream(indexStream.get());

Pass CIs and build are happy locally.

Generated-by: Codex Version 26.429.61741 (2429)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -144,7 +144,7 @@ jobs:
       run: |
         mkdir build
         cd build
-        cmake .. -G "Visual Studio 17 2022" \
+        cmake .. -G "Visual Studio 18 2026" \
           -DCMAKE_BUILD_TYPE=RELEASE \
           -DBUILD_TOOLS=OFF \
           -DBUILD_JAVA=OFF \
--- a/c++/src/ColumnReader.cc
+++ b/c++/src/ColumnReader.cc
@@ -435,7 +435,7 @@ namespace orc {
       const size_t cap = static_cast<size_t>(std::numeric_limits<int>::max());
       while (sizeToSkip != 0) {
         size_t step = sizeToSkip > cap ? cap : sizeToSkip;
-        inputStream_->Skip(static_cast<int>(step));
+        static_cast<void>(inputStream_->Skip(static_cast<int>(step)));
         sizeToSkip -= step;
       }
       bufferEnd_ = nullptr;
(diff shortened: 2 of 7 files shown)


c++/src/ColumnReader.cc
```diff
@@ -435,7 +435,7 @@ namespace orc {
       const size_t cap = static_cast<size_t>(std::numeric_limits<int>::max());
       while (sizeToSkip != 0) {
         size_t step = sizeToSkip > cap ? cap : sizeToSkip;
-        inputStream_->Skip(static_cast<int>(step));
+        static_cast<void>(inputStream_->Skip(static_cast<int>(step)));
```
ffacs: I'd prefer std::ignore

ffacs (approved):
Generally LGTM~

dongjoon-hyun:
The JIRA issue seems to be resolved with a wrong version.

Screenshot 2026-07-04 at 20 35 39

.github/workflows/build_and_test.yml
```diff
@@ -144,7 +144,7 @@ jobs:
       run: |
         mkdir build
         cd build
-        cmake .. -G "Visual Studio 17 2022" \
+        cmake .. -G "Visual Studio 18 2026" \
```
dongjoon-hyun, line 147: Is this changed because `GitHub Actions` removed `Visual Studio 17 2022`?

dongjoon-hyun:
For `branch-2.3`, I upgraded to `Visual Studio 18 2026` only without source code changes.
- https://github.com/apache/orc/pull/2656

For this PR (ORC-2159), I updated the fix version to `3.0.0`.

dongjoon-hyun:
For older branches, I disabled the warnings for now.
  - branch-2.2: https://github.com/apache/orc/pull/2657
  - branch-2.1: https://github.com/apache/orc/pull/2658
  - branch-2.0: https://github.com/apache/orc/pull/2659

.github/workflows/build_and_test.yml
wgtmac: Yes

PR #2616, opened 2026-05-07
ORC-2160: [C++] Expose prefetch range planning via Reader::preBufferRange and refactor preBuffer to reuse it

lucasfang:

This PR introduces a new Reader API, preBufferRange(stripes, includeTypes), which computes and returns the exact prefetch ranges (offset/length pairs), implements it in ReaderImpl, and refactors the existing preBuffer path to reuse that range-planning logic before populating cache.

The change is needed because preBuffer previously did not expose the planned ranges, but external users now have a multi-Reader scenario where different Readers prefetch different ranges of the same file and share an external prebuffer cache, so they must know the precise ranges in advance to coordinate prefetching and avoid duplicated work.

No

### DIFF
--- a/c++/include/orc/Reader.hh
+++ b/c++/include/orc/Reader.hh
@@ -32,6 +32,7 @@
 #include <memory>
 #include <set>
 #include <string>
+#include <utility>
 #include <vector>
 
 namespace orc {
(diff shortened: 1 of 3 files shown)


wgtmac (approved):
Thanks for adding this!

PR #2618, opened 2026-05-07
ORC-2161: [C++] UnionColumnReader should reject out-of-range union tags

wgtmac:

This PR adds validation for C++ union tag values before they are used as child indexes.

The change covers:
- UnionColumnReader::skip
- UnionColumnReader::nextInternal
- UnionColumnPrinter::printRow

If a malformed ORC file contains a union tag that is greater than or equal to the number of union children, the C++ reader/printer now throws ParseError instead of indexing out of bounds.

Union tags are decoded from the ORC data stream as byte values, but the valid range depends on the number of union children. Malformed input can contain a tag outside that range. The C++ reader previously trusted the tag value directly when indexing per-child state.

This patch makes malformed union tags fail cleanly.

Added C++ unit tests for a two-child union with invalid tag value 200, covering:
- next
- next with nulls
- skip
- ColumnPrinter

Generated-by: OpenAI Codex GPT-5.

### DIFF
--- a/c++/src/ColumnPrinter.cc
+++ b/c++/src/ColumnPrinter.cc
@@ -540,11 +540,17 @@ namespace orc {
     if (hasNulls && !notNull[rowId]) {
       writeString(buffer, "null");
     } else {
+      const size_t tag = static_cast<size_t>(tags_[rowId]);
+      if (tag >= fieldPrinter_.size()) {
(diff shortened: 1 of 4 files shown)


ffacs (approved):
LGTM~ Thank you for the fix @wgtmac .

wgtmac:
Thanks for the quick review, @ffacs!

PR #2622, opened 2026-05-12
ORC-2166: [C++] Guard C++ integer arithmetic against overflow

ffacs:

  This PR adds reusable checked integer arithmetic helpers for the C++ code: `addWithOverflow` and `multiplyWithOverflow`.

  The patch uses these helpers to reject overflow-prone arithmetic in reader and buffer paths, including direct string length accumulation, list/map length accumulation, dictionary offsets, float/double skip byte counts, union child counts, `DataBuffer` allocation sizes, `BlockBuffer` size/capacity calculations, and list/map vector offset capacity calculations.

  Regression tests were added for the overflow helpers and malformed string/list/map length streams.

Malformed ORC files can provide invalid or extremely large values in length streams. Negative lengths or overflowing length sums can cause unchecked integer wraparound, incorrectly sized allocations, wrapped offsets, or unsafe skip/copy behavior.

Centralizing checked arithmetic makes these cases fail cleanly before unsafe allocation or memory access. Reader-facing malformed input now fails with `ParseError`.

  Ran the full C++ test target:

 ```bash
  cmake --build build --target test-out
```
  Result:

  1/2 Test #1: orc-test  .... Passed
  2/2 Test #2: tool-test .... Passed

  100% tests passed, 0 tests failed out of 2

  Also ran targeted tests during development for the new overflow checks.

  Yes. Generated with OpenAI Codex.

### DIFF
--- a/c++/src/BlockBuffer.cc
+++ b/c++/src/BlockBuffer.cc
@@ -17,10 +17,12 @@
  */
 
 #include "BlockBuffer.hh"
+#include "Utils.hh"
 #include "orc/OrcFile.hh"
 #include "orc/Writer.hh"
 
 #include <algorithm>
+#include <stdexcept>
 
 namespace orc {
 
@@ -51,10 +53,19 @@ namespace orc {
     if (currentSize_ < currentCapacity_) {
       Block emptyBlock(blocks_[currentSize_ / blockSize_] + currentSize_ % blockSize_,
                        blockSize_ - currentSize_ % blockSize_);
-      currentSize_ = (currentSize_ / blockSize_ + 1) * blockSize_;
+      uint64_t nextBlockNumber = currentSize_ / blockSize_ + 1;
+      uint64_t nextSize = 0;
+      if (multiplyWithOverflow(nextBlockNumber, blockSize_, &nextSize)) {
+        throw std::length_error("Block buffer size overflow");
+      }
+      currentSize_ = nextSize;
       return emptyBlock;
     } else {
-      resize(currentSize_ + blockSize_);
+      uint64_t nextSize = 0;
+      if (addWithOverflow(currentSize_, blockSize_, &nextSize)) {
+        throw std::length_error("Block buffer size overflow");
+      }
+      resize(nextSize);
       return Block(blocks_.back(), blockSize_);
     }
   }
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: c++/src/ColumnReader.cc [hunk 1]
```diff
@@ -694,17 +698,32 @@ namespace orc {
   size_t StringDirectColumnReader::computeSize(const int64_t* lengths, const char* notNull,
                                                uint64_t numValues) {
     size_t totalLength = 0;
+    bool hasNegativeLength = false;
+    bool hasLengthOverflow = false;
+    auto addLength = [&](int64_t value) {
+      hasNegativeLength |= value < 0;
+      size_t length = static_cast<size_t>(value);
+      size_t nextTotalLength = totalLength + length;
+      hasLengthOverflow |= nextTotalLength < totalLength;
```
dongjoon-hyun: Like `skip()` method, shall we throw here instead of waiting for the whole loop is finished?

c++/src/MemoryPool.cc
```diff
@@ -52,6 +56,15 @@ namespace orc {
     // PASS
   }
 
+  template <class T>
+  uint64_t checkedBufferSize(uint64_t count) {
+    uint64_t bytes = 0;
+    if (multiplyWithOverflow(static_cast<uint64_t>(sizeof(T)), count, &bytes)) {
+      throw std::length_error("DataBuffer allocation size overflow");
```
wgtmac, line 63: Do we want to define a orc-specific exception that derives from `orc::Exception`? arrow-cpp catches all `orc::Exception` to avoid crashing.

c++/src/BlockBuffer.cc
```diff
@@ -70,10 +81,19 @@ namespace orc {
...
     while (currentCapacity_ < newCapacity) {
+      uint64_t nextCapacity = 0;
+      if (addWithOverflow(currentCapacity_, blockSize_, &nextCapacity)) {
+        throw std::length_error("Block buffer capacity overflow");
+      }
       char* newBlockPtr = memoryPool_.malloc(blockSize_);
       if (newBlockPtr != nullptr) {
-        blocks_.push_back(newBlockPtr);
-        currentCapacity_ += blockSize_;
+        try {
```
wgtmac, line 90: Why do we need this try-catch block?

PR #2629, opened 2026-05-25
ORC-2167: [C++] Fix integer overflow in PostScript footer length validation

luffy-zh:

 Added overflow-safe bounds checking in three locations in c++/src/Reader.cc using __builtin_add_overflow to prevent integer overflow when parsing malformed
   ORC files with extremely large length values in PostScript:

  1. readMetadata() - Use __builtin_add_overflow to safely compute total tail size (footerLength + metadataSize + postscriptLength + 1) before comparing
  against fileLength
  2. createReader() - Use __builtin_add_overflow to safely compute tailSize (1 + postscriptLength + footerSize) before bounds validation
  3. startNextStripe() - Use __builtin_add_overflow to safely sum stripe length components (offset + indexLength + dataLength + footerLength) before
  comparing against fileLength

When an ORC file's PostScript is crafted with footer_length or metadata_length set to UINT64_MAX, the bounds check using unsigned addition (e.g.,
  metadataSize + footerLength + postscriptLength_ + 1) can overflow and wrap around to a small value, bypassing validation.

  For example, given a 58-byte file with footerLength = UINT64_MAX, metadataSize = 0, and postscriptLength_ = 23:
  - The sum wraps to 23 (mod 2^64)
  - The check 58 < 23 evaluates to false, allowing invalid offset calculation
  - This can cause SIGBUS crash due to out-of-bounds memory access

  Using __builtin_add_overflow (a GCC/Clang builtin) provides a standard, efficient way to detect overflow without manual subtraction logic.

Added two unit tests in c++/test/TestReader.cc:

  1. testMalformedFooterLengthOverflow - Constructs a minimal malformed ORC file with footer_length = UINT64_MAX and verifies that createReader() throws
  ParseError
  2. testMalformedMetadataLengthOverflow - Constructs a malformed file with metadata_length = UINT64_MAX and verifies ParseError is thrown

Generated-by: Claude code

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -1253,10 +1253,8 @@ namespace orc {
           stripeTotalLength >= fileLength) {
         std::stringstream msg;
         msg << "Malformed StripeInformation at stripe index " << currentStripe_
-            << ": fileLength=" << fileLength
-            << ", StripeInfo=(offset=" << stripeOffset
-            << ", indexLength=" << indexLength
-            << ", dataLength=" << dataLength
+            << ": fileLength=" << fileLength << ", StripeInfo=(offset=" << stripeOffset
+            << ", indexLength=" << indexLength << ", dataLength=" << dataLength
             << ", footerLength=" << footerLength << ")";
         throw ParseError(msg.str());
       }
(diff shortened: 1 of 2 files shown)


c++/src/Reader.cc
```diff
@@ -884,14 +884,21 @@ namespace orc {
   void ReaderImpl::readMetadata() const {
     uint64_t metadataSize = contents_->postscript->metadata_length();
     uint64_t footerLength = contents_->postscript->footer_length();
-    if (fileLength_ < metadataSize + footerLength + postscriptLength_ + 1) {
+
+    // Check for overflow in length calculations
+    uint64_t totalTail;
+    if (__builtin_add_overflow(footerLength, metadataSize, &totalTail) ||
```
wgtmac: This does not work for Windows unfortunately

wgtmac (approved):
Could you fix the linter error?

wgtmac (approved):
Thank you, @luffy-zh!

PR #2644, opened 2026-06-16
ORC-2179: Add RISC-V Docker image and GitHub Actions support

zhanchangbao-sanechips:
### What this PR does

Adds a RISC-V (riscv64) Docker image for Ubuntu 24.04 and wires it into GitHub Actions so we can continuously build ORC on RISC-V.

### Changes

- `docker/riscv64-ubuntu24/Dockerfile`: new image based on the existing Ubuntu 24.04 Dockerfile. Supports both `cc=gcc` (default) and `cc=clang` build args.
- `docker/os-list.txt`: registered the new image.
- `.github/workflows/build_and_test.yml`: added a `riscv64` job running under QEMU on `ubuntu-latest`.

### Notes

- The job currently only builds ORC. RVV-specific tests will be added in follow-up PRs after the RVV decoder implementation is in.
- I followed the same pattern as the existing `aarch64` / `ubuntu24` Docker images to keep things consistent.
- Local verification was done with `docker buildx` on a host with QEMU user-static registered.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -63,6 +63,38 @@ jobs:
         cd docker
         ./run-one.sh local main ${{ matrix.os }}
 
+  riscv64:
+    name: "RISC-V ${{ matrix.os }}"
+    needs: [license-check]
+    runs-on: ubuntu-latest
+    timeout-minutes: 180
+    strategy:
+      fail-fast: false
+      max-parallel: 20
+      matrix:
+        os:
+          - riscv64-ubuntu24
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v6
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+      with:
+        platforms: linux/riscv64
+    - name: Set up Buildx
+      uses: docker/setup-buildx-action@v3
+    - name: Build Docker image
+      run: |
+        cd docker/${{ matrix.os }}
+        docker buildx build --platform linux/riscv64 --load \
+          --cache-from type=gha,scope=riscv64-ubuntu24 \
+          --cache-to type=gha,mode=max,scope=riscv64-ubuntu24 \
+          -t orc-riscv64 .
+    - name: Build ORC with RVV
+      run: |
+        docker run --platform linux/riscv64 --rm -v $(pwd):/root/orc orc-riscv64 /bin/bash -c \
+          "mkdir -p build && cd build && cmake -DBUILD_JAVA=OFF -DBUILD_ENABLE_RVV=ON -DCMAKE_BUILD_TYPE=Release ../orc && make -j\$(nproc) package"
+
   build:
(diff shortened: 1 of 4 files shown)


dongjoon-hyun (changes requested):
Please use `riscv64/ubuntu:26.04` because 26.04 is the latest LTS.

zhanchangbao-sanechips:
> Please use `riscv64/ubuntu:26.04` because 26.04 is the latest LTS.

Thanks for the review. Updated to use riscv64/ubuntu:26.04 as suggested.

Changes made:

Renamed docker/riscv64-ubuntu24/ to docker/riscv64-ubuntu26/
Updated FROM --platform=linux/riscv64 ubuntu:26.04
Updated docker/os-list.txt
Updated the GitHub Actions matrix and cache scope to riscv64-ubuntu26
Please take another look when you have time.

.github/workflows/build_and_test.yml
```diff
@@ -63,6 +63,38 @@ jobs:
...
+    - name: Build Docker image
+      run: |
+        cd docker/${{ matrix.os }}
+        docker buildx build --platform linux/riscv64 --load \
+          --cache-from type=gha,scope=riscv64-ubuntu26 \
+          --cache-to type=gha,mode=max,scope=riscv64-ubuntu26 \
+          -t orc-riscv64 .
+    - name: Build ORC with RVV
+      run: |
+        docker run --platform linux/riscv64 --rm -v $(pwd):/root/orc orc-riscv64 /bin/bash -c \
```
ffacs: we should make sure v extension is enable

zhanchangbao-sanechips: > we should make sure v extension is enable

I have updated the docker run command to include the environment variable QEMU_CPU=rv64,v=on,vlen=128,vext_spec=v1.0, and verified that RVV is working correctly.

The vext_spec=v1.0 explicitly specifies the vector extension version to suppress the warning: "vector version is not specified, use the default value v1.0".

PixPin_2026-06-18_11-01-21

ffacs (approved):

dongjoon-hyun (approved):

wgtmac (approved):

ffacs:
Merged. Thanks @zhanchangbao-sanechips @wgtmac @dongjoon-hyun ~

dongjoon-hyun:
While preparing Apache ORC 1.9.9 EOL release, I've checked the CIs. 

Unfortunately, this PR seems to cause the CI outage because we missed the CI didn't pass indeed at the last commit.

Screenshot 2026-07-04 at 20 23 08

dongjoon-hyun:
In the commit builders, this also fails due to the timeout like the following. We need to find another way to deliver this.
- https://github.com/apache/orc/actions/runs/28529515772/job/84575485372

Screenshot 2026-07-04 at 20 25 08

So, let me revert this to recover the CIs.

PR #2661, opened 2026-07-05
ORC-2187: Set protobuf message size limit in C++ reader

dongjoon-hyun:

This PR caps the size of protobuf messages parsed from ORC file contents in the C++ reader, matching the Java reader.

The Java reader already caps protobuf message size at 1 GB, but the C++ reader parsed with no limit.

https://github.com/apache/orc/blob/7cc86ec686b4edcdb3c4d641ae362f101d4c8015/java/core/src/java/org/apache/orc/impl/InStream.java#L45

https://github.com/apache/orc/blob/7cc86ec686b4edcdb3c4d641ae362f101d4c8015/java/core/src/java/org/apache/orc/impl/InStream.java#L922

`SetTotalBytesLimit` has the same mechanism with `InStream.setSizeLimit`.

Pass the existing C++ tests (`orc-test`).

Generated-by: Claude Fable 5

### DIFF
--- a/c++/src/Reader.cc
+++ b/c++/src/Reader.cc
@@ -531,13 +531,13 @@ namespace orc {
 
         if (pbStream.kind() == proto::Stream_Kind_ROW_INDEX) {
           proto::RowIndex rowIndex;
-          if (!rowIndex.ParseFromZeroCopyStream(inStream.get())) {
+          if (!parseProtobufFromStream(&rowIndex, inStream.get())) {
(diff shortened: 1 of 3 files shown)


dongjoon-hyun:
Merged to main/2.3 for Apache ORC 2.3.1.

dongjoon-hyun:
This is reverted from branch-2.3 due to the difference of protobuf.

For `branch-2.3`, the following is created.
- #2669

PR #2680, opened 2026-07-08
ORC-2197: Reject negative decoded dictionary entry length in Java string dictionary reader

dongjoon-hyun:

This PR aims to reject a negative decoded dictionary entry length in the Java string dictionary reader by throwing a `FileFormatException` in `StringDictionaryTreeReader.readDictionaryLengthStream`.

A corrupt file can encode a negative entry length in the LENGTH stream, making `dictionaryOffsets` non-monotonic and causing `setRef` to receive a negative length. The C++ reader already rejects this case (`DictionaryLoader.cc`), so this restores C++/Java reader parity.

Pass the CIs with a newly added test case, `TestStringDictionaryTreeReader#testNegativeDictionaryEntryLength`.

Generated-by: Claude Fable 5


dongjoon-hyun:
Merged to main/2.3.

PR #2682, opened 2026-07-08
ORC-2199: Validate union tag against the number of children in `UnionTreeReader`

dongjoon-hyun:

This PR aims to validate the union tag against the number of children in `UnionTreeReader`.

Both the `nextVector` and `skipRows` paths now check that each tag read from the `DATA` stream is within `[0, fields.length)` and throw `FileFormatException` with a clear message, consistent with the C++ reader's `getCheckedUnionTag`.

The Java `UnionTreeReader` used the raw tag byte from the `DATA` stream without validation. For a corrupt or malicious file with a tag greater than or equal to the number of union children (or a negative sign-extended byte), `skipRows` fails with `ArrayIndexOutOfBoundsException` and `nextVector` propagates the invalid tag downstream. The C++ reader already rejects this via `getCheckedUnionTag` (`ParseError`), so this closes a C++/Java parity gap.

Pass the CIs with a newly added `TestTreeReaderFactory`.

Generated-by: Claude Fable 5


dongjoon-hyun:
Merged to main

PR #2683, opened 2026-07-08
ORC-2200: Validate stripe offset/index/data/footer lengths in the Java reader

dongjoon-hyun:

This PR aims to validate stripe `offset`/`indexLength`/`dataLength`/`footerLength` in the Java reader before reading a stripe, like the C++ reader (`RowReaderImpl::startNextStripe`). `RecordReaderImpl` now throws `FileFormatException` if any value is negative, `footerLength` exceeds `Integer.MAX_VALUE`, or the sum overflows or exceeds the file length.

A crafted `StripeInformation` could cause a `NegativeArraySizeException`, a huge allocation, or a read at a bogus offset. This matches the existing C++ reader validation.

Pass the CIs with the newly added test case.

Generated-by: Claude Fable 5


dongjoon-hyun:
Merged to main

PR #2685, opened 2026-07-10
ORC-2202: Use `debian:13` instead of `13.1` in `docker/debian13/Dockerfile`

dongjoon-hyun:

This PR aims to use `debian:13` instead of `debian:13.1` in `docker/debian13/Dockerfile`.

Debian 13 has moved past the `13.1` point release, so the pinned tag no longer tracks the latest point releases. Using the `debian:13` tag keeps the image up to date automatically and is consistent with the other OS Dockerfiles (`ubuntu:24.04`, `oraclelinux:9`, `oraclelinux:10`), as well as the previous Debian Dockerfiles which used floating codename tags (`debian:bullseye`, `debian:bookworm`).

Pass the CIs.

Generated-by: Claude Fable 5

### DIFF
--- a/docker/debian13/Dockerfile
+++ b/docker/debian13/Dockerfile
@@ -17,7 +17,7 @@
 # ORC compile for Debian 13
 #
 
-FROM debian:13.1
+FROM debian:13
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Merged to main

PR #2699, opened 2026-07-20
ORC-2213: [C++] Add option to read timestamps with writer timezone

wgtmac:

Add a RowReaderOptions flag that lets timestamp readers use the writer timezone recorded in each stripe footer instead of the configured reader timezone. This preserves the stored timestamp value for callers that need to avoid ORC timestamp timezone conversion.

In old days, ORC only supports orc::TIMESTAMP type. The semantics of orc::TIMESTAMP is equivalent to TIMESTAMP_NTZ but with a complex writer and reader timezone adjustment.  So it is difficult to support both TIMESTAMP_NTZ and TIMESTAMP_LTZ types using a single orc::TIMESTAMP. In later versions, orc::TIMESTAMP_INSTANT has been added to support TIMESTAMP_LTZ type. However, users may not know the difference under the hood and still use legacy systems to write both TIMESTAMP_LTZ and TIMESTAMP_NTZ semantics to the old orc::TIMESTAMP type. When users use orc::TIMESTAMP as TIMESTAMP_LTZ values, we should set reader timezone to the writer timezone to avoid value conversion. However, users may not know the writer timezone in advance and ORC files may have different writer timezone (considering files are produced by teams in different time zones.) We need an approach to enforce reader to use writer timezone via explicit configuration.

Added a new test case.

Generated-by: Codex GPT-5.6 Sol

### DIFF
--- a/c++/include/orc/Reader.hh
+++ b/c++/include/orc/Reader.hh
@@ -354,6 +354,16 @@ namespace orc {
     const std::string& getTimezoneName() const;
 
     /**
+     * Use the writer timezone when reading timestamp values.
+     */
+    RowReaderOptions& setUseWriterTimezone(bool useWriterTimezone);
+
+    /**
+     * Get whether to use the writer timezone when reading timestamp values.
+     */
+    bool getUseWriterTimezone() const;
+
+    /**
      * Get the IdReadIntentMap map that was supplied by client.
      */
     const IdReadIntentMap getIdReadIntentMap() const;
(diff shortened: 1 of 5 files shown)


ffacs (approved):
LGTM.

dongjoon-hyun (approved):
+1, LGTM.

wgtmac:
Thank you, @ffacs and @dongjoon-hyun!

PR #2709, opened 2026-09-02
ORC-624: [C++] Avoid unaligned trailer loads in LZO decompressor

wgtmac:

This PR replaces two potentially unaligned 16-bit trailer loads with an alignment-safe little-endian reader.

It also adds a long-distance match test covering the previously untested `0x10` LZO command path. This is a focused replacement for the LZO portion of #496, which was closed without merging.

The LZO input pointer advances byte by byte and is not guaranteed to be 2-byte aligned. Dereferencing it as `uint16_t*` or `int16_t*` is undefined behavior and is reported by UBSan.

```bash
cmake --build build --target orc-test -j8
cd build/c++/test
./orc-test --gtest_filter='TestDecompression.testLzoLong:TestDecompression.testLzoLongDistanceMatch:TestDecompression.testLzoOverflow:TestDecompression.testLzoTruncatedStopCommand'
```

All four tests passed. The existing long test covers the `0x20` command path, and the new test covers the `0x10` path.

Generated-by: OpenAI Codex (GPT-5)

### DIFF
--- a/c++/src/LzoDecompressor.cc
+++ b/c++/src/LzoDecompressor.cc
@@ -32,6 +32,11 @@ namespace orc {
   static const int32_t SIZE_OF_INT = 4;
   static const int32_t SIZE_OF_LONG = 8;
 
+  static uint32_t readLittleEndianShort(const char* input) {
+    return static_cast<uint32_t>(static_cast<uint8_t>(input[0])) |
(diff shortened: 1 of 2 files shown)


ffacs (approved):
LGTM

